PR #2, opened 2016-02-14
fix up warnings and use definition to specialize _rclpy.c

wjwwood:
The big change here is that I renamed `_rclpy.c.in` to `_rclpy.c` and changed to using a preprocessor definition instead of the CMake template that was being used before. I thought this was better so that editors could more easily detect and color the file, and for me it allowed me to add a definition to my C++ checker and then it could parse and check for warnings/errors in my editor, which is obviously harder when it's a CMake template.

I also addressed the style stuff in `_rclpy.c` according to uncrusfity and cpplint, added a copyright header, and changed it so that rcl errors are propagated up to Python as `RuntimeError`'s.

@esteve I made this as a pull request so you could consider the changes before merging it into your pull request.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,15 +198,12 @@ macro(target)
     return()
   endif()
 
-  configure_file(
-    src/rclpy/_rclpy.c.in
-    src/rclpy/_rclpy${target_suffix}.c
-  )
-
   add_library(
     ${PROJECT_NAME}${target_suffix}
-    SHARED src/rclpy/_rclpy${target_suffix}.c
+    SHARED src/rclpy/_rclpy.c
   )
+  target_compile_definitions(${PROJECT_NAME}${target_suffix}
+    PRIVATE "RMW_IMPLEMENTATION_SUFFIX=${target_suffix}")
 
   set_target_properties(${PROJECT_NAME}${target_suffix} PROPERTIES
     PREFIX ""
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,15 +3,12 @@ cmake_minimum_required(VERSION 2.8.3)
 project(rclpy)
 
 find_package(ament_cmake REQUIRED)
+find_package(ament_cmake_python REQUIRED)
 find_package(rcl REQUIRED)
 find_package(rmw REQUIRED)
 find_package(rmw_implementation_cmake REQUIRED)
-find_package(ament_cmake_python REQUIRED)
-
-include_directories(include)
 
 if(NOT WIN32)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
 endif()
 
@@ -267,10 +264,10 @@ if(AMENT_ENABLE_TESTING)
   ament_lint_auto_find_test_dependencies()
 endif()
 
-  file(
-    COPY "${PROJECT_NAME}"
-    DESTINATION "${CMAKE_CURRENT_BINARY_DIR}"
-  )
+file(
+  COPY "${PROJECT_NAME}"
+  DESTINATION "${CMAKE_CURRENT_BINARY_DIR}"
+)
 
 ament_python_install_package(${PROJECT_NAME} PACKAGE_DIR "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -261,12 +261,18 @@ if(AMENT_ENABLE_TESTING)
   ament_lint_auto_find_test_dependencies()
 endif()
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -3,15 +3,12 @@ cmake_minimum_required(VERSION 2.8.3)
...
+find_package(ament_cmake_python REQUIRED)
 find_package(rcl REQUIRED)
 find_package(rmw REQUIRED)
 find_package(rmw_implementation_cmake REQUIRED)
-find_package(ament_cmake_python REQUIRED)
-
-include_directories(include)
 
 if(NOT WIN32)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
```
dirk-thomas: If this package doesn't do any C++ stuff `project(...)` should explicitly list only `C` to avoid searching for a C++ compiler.

CMakeLists.txt
```diff
@@ -267,10 +261,10 @@ if(AMENT_ENABLE_TESTING)
   ament_lint_auto_find_test_dependencies()
 endif()
 
-  file(
-    COPY "${PROJECT_NAME}"
-    DESTINATION "${CMAKE_CURRENT_BINARY_DIR}"
-  )
+file(
+  COPY "${PROJECT_NAME}"
```
dirk-thomas: What is being installed here? If it is the library the `file(TARGET` signature should be used.

wjwwood: I don't know, I was just fixing the indention. It looks like it is copying Python stuff from the source directory to the build directory.

CMakeLists.txt
wjwwood: 398ba46

CMakeLists.txt
wjwwood: I looked into it and I think what's going on is the best way to do it for now. I added a comment to the CMake code explaining what's going on (AFAIK): 20288a8

CMakeLists.txt
```diff
@@ -267,12 +261,18 @@ if(AMENT_ENABLE_TESTING)
...
-
-ament_python_install_package(${PROJECT_NAME} PACKAGE_DIR "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}")
+# Copy the Python code to the folder where the C extensions will be generated.
+# This creates a folder in ${CMAKE_CURRENT_BINARY_DIR}/rclpy which has all the
+# necessary Python code and C Python extensions for running tests.
+# This folder is also ready to be installed directly as a Python package.
+file(
+  COPY "${PROJECT_NAME}"
+  DESTINATION "${CMAKE_CURRENT_BINARY_DIR}"
+)
```
dirk-thomas: I still don't see why this should be done. Copying files like this should be avoided at all cost. Can someone argue why this is absolutely necessary?

wjwwood: > Copying files like this should be avoided at all cost.

Why?  That seems to be an absolute assertion without an obvious reason (as if it should be clear to everyone that this is always implicitly true).

> Can someone argue why this is absolutely necessary?

It is necessary because in order to import `rclpy` you need a folder which contains both the Python code (which starts in the source space) and the generated C extensions (`.so`'s) which are generated into the build space. So this copies the Python source into the build space so that there can be a folder in the build space which contains both the Python source and the C extensions. This is needed so that the tests can import `rclpy` as if it were installed, and the tests do so by importing `rclpy` from the build space.

I'd suggest you look into why it is that way by trying to change it (as I did), if you don't believe me. Maybe you can come up with an improvement.

esteve:
@wjwwood thanks for the improvements! Yeah, the explanation you added as a comment is why I copied the files from the source tree to the build tree so that a single call to `ament_python_install_package` would install both the Python files and the native extensions.

CMakeLists.txt
dirk-thomas: > > Copying files like this should be avoided at all cost.

> Why?  That seems to be an absolute assertion without an obvious reason (as if it should be clear to everyone that this is always implicitly true).

The reasons have been iterated many times. Therefore I didn't mention them. But for completeness:
* when files get copied the developer might unintentionally change the wrong file
* changes to the source files won't have any effect on future builds since the files are only copied once at configure time
* either CMake needs to be reinvoked every time (which is time consuming during the development) or the developer edits the copied files in the build space for faster iteration and then has to copy back his changes (always potentially overwriting them accidentally when reconfiguring)

> > Can someone argue why this is absolutely necessary?

> It is necessary because in order to import `rclpy` you need a folder which contains both the Python code (which starts in the source space) and the generated C extensions (`.so`'s) which are generated into the build space. So this copies the Python source into the build space so that there can be a folder in the build space which contains both the Python source and the C extensions. This is needed so that the tests can import `rclpy` as if it were installed, and the tests do so by importing `rclpy` from the build space.
> 
> I'd suggest you look into why it is that way by trying to change it (as I did), if you don't believe me. Maybe you can come up with an improvement.

See PR #3 which achieves the described goal.

PR #3, opened 2016-02-16
avoid copying Python files from source to build space

dirk-thomas:
Based on the discussion in https://github.com/ros2/rclpy/pull/2#discussion_r52928354

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,6 +252,9 @@ macro(target)
     PUBLIC
     ${PYTHON_INCLUDE_DIRS}
   )
+
+  install(TARGETS ${PROJECT_NAME}${target_suffix}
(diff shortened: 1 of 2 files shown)


esteve:
I'm far from a CMake expert, but why is copying from the source to build space a bad practice?

esteve:
Oops, I just saw your comment on https://github.com/ros2/rclpy/pull/2#discussion_r53070269

wjwwood:
I agree that having the installed code come directly from the source space is better. 

For the code in the build space used for running tests I think it is less important. The tests rely on both generated and compiled code so building before iterating on the test is already pretty common. We could have continued to use the file copy for the build space setup, and installed from the source space for a similar effect without the need for the complicated and somewhat confusing init file that mocks the source space. 

But since you've already implemented it we should probably keep it. +1

PR #7, opened 2016-02-29
wait_set API change

jacquelinekay:
Connects to ros2/rcl#27

We'll delay implementing Python services until C services are merged. This is necessary for the C services branch to compile.


wjwwood:
lgtm

PR #8, opened 2016-03-11
Change for fixed guard conditions API

jacquelinekay:
Connects to ros2/rcl#33 but does not add any fixed guard conditions.


jacquelinekay:
closing this to take a new approach.

PR #9, opened 2016-03-29
Add rosnode and rostopic functionality

vmayoral:


rclpy/rclpy/rosnode.py
```diff
@@ -0,0 +1,132 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import rclpy
+import time
+import os
+import errno
+import sys
+
+from optparse import OptionParser
```
wjwwood: We'll probably have to change this to `argparse`, we don't use `optparse` as it is obsolete (`argparse` was designed to replace it).

rclpy/rclpy/rosnode.py
```diff
@@ -0,0 +1,132 @@
...
+
+NAME='rosnode'
+
+# TODO implement
+def _rosnode_cmd_ping(argv):
+    print("NOT IMPLEMENTED\n")
+    sys.exit(0)
```
wjwwood: There is a standard Python exception called `NotImplementedError`: https://docs.python.org/3.5/library/exceptions.html#NotImplementedError

PR #10, opened 2016-04-06
Don't allow fastrtps to be selected in rclpy

dhood:
If I have both Connext and fastRTPS installed, I get the following error when trying to run either `listener_py` or `talker_py`

```
  File "/home/dhood/ros2_ws/install_isolated/rclpy_examples/bin/listener_py", line 9, in <module>
    load_entry_point('rclpy-examples==0.0.0', 'console_scripts', 'listener_py')()
  File "/home/dhood/ros2_ws/build_isolated/rclpy_examples/listener_py.py", line 30, in main
    rclpy.init(args)
  File "/home/dhood/ros2_ws/install_isolated/rclpy/lib/python3.4/site-packages/rclpy/__init__.py", line 42, in init
    rmw_implementation_tools.import_rmw_implementation()
  File "/home/dhood/ros2_ws/install_isolated/rclpy/lib/python3.4/site-packages/rclpy/impl/rmw_implementation_tools.py", line 85, in import_rmw_implementation
    __rmw_implementation_module = importlib.import_module(module_name, package='rclpy')
  File "/usr/lib/python3.4/importlib/__init__.py", line 109, in import_module
  File "<frozen importlib._bootstrap>", line 2254, in _gcd_import
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2224, in _find_and_load_unlocked
ImportError: No module named 'rclpy._rclpy__rmw_fastrtps_cpp'
```
From what I understand, fastrtps is chosen [here](https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/impl/rmw_implementation_tools.py#L79) even though it has been skipped over [here](https://github.com/ros2/rclpy/blob/master/rclpy/CMakeLists.txt#L47)

This is a pretty hack-y fix, but since this is the current default situation if following the linux install from source, I thought it would be worth patching.

### DIFF
--- a/rclpy/rclpy/impl/rmw_implementation_tools.py
+++ b/rclpy/rclpy/impl/rmw_implementation_tools.py
@@ -29,6 +29,9 @@ def reload_rmw_implementations():
     """(Re)Load the available rmw implementations by inspecting the ament index."""
     global __rmw_implementations
     __rmw_implementations = sorted(ament_index_python.get_resources('rmw_implementation').keys())
+
+    # Remove implementations that are being filtered for in the rclp CMakeLists so they cannot be selected
+    __rmw_implementations = [rmw_impl for rmw_impl in __rmw_implementations if rmw_impl not in ['rmw_connext_dynamic_cpp', 'rmw_fastrtps_cpp']]
     return __rmw_implementations
 
 


wjwwood:
Yeah, I'm not sure if we should fix it here, or update the tutorial to do something like:

```
$ RCLPY_IMPLEMENTATION=rmw_opensplice_cpp python3 `which talker_py`
$ RCLPY_IMPLEMENTATION=rmw_opensplice_cpp python3 `which listener_py`
```

dhood:
personally when I was trying out ros2 demos I just ran all of the executables I could see, so the error may still come up for some with that approach

jacquelinekay:
hmm. Connext Dynamic should work actually.

I think if these lines were removed:
https://github.com/ros2/rclpy/blob/master/rclpy/CMakeLists.txt#L43-L46

you could also remove the reference to rmw_connext_dynamic_cpp from this PR.

wjwwood:
We could certainly do with a better error message than this, but the fact that there is currently no python implementation built for FastRTPS is only a temporary problem. Also, it would be possible in the future for you to accidentally build your workspace in such a way that FastRTPS is there, but there is no Python implementation built yet (it would be tricky to do this on purpose, but possible).

At one point I had a TODO to only use a separate `ament_index` name for *python* rmw implementations that have been built, which would avoid this issue partially.

> personally when I was trying out ros2 demos I just ran all of the executables I could see, so the error may still come up for some with that approach

I agree, that's why I think an error message like:

```
<traceback>
Could not find a Python implementation for the '<selected rmw impl>' rmw implementation, these are the available ones you can select with the 'RCLPY_IMPLEMENATIONS' environment variable: <foo>, <bar>, <baz rmw impls>
```

It would similar to what you get when you select a bogus rmw implementations, e.g. try:

```
$ RCLPY_IMPLEMENTATIONS=bogus python3 `which talker_py`
```

wjwwood:
For example:

```
% RCLPY_IMPLEMENTATION=bogus python3 `which talker_py`
The rmw implementation specified in 'RCLPY_IMPLEMENTATION=bogus', is not one of the available implementations: ['rmw_fastrtps_cpp', 'rmw_opensplice_cpp']
  File "/Users/william/ros2_ws/install_isolated/rclpy_examples/bin/talker_py", line 9, in <module>
    load_entry_point('rclpy-examples', 'console_scripts', 'talker_py')()
  File "/Users/william/ros2_ws/build_isolated/rclpy_examples/talker_py.py", line 27, in main
    rclpy.init(args)
  File "/Users/william/ros2_ws/install_isolated/rclpy/lib/python3.5/site-packages/rclpy/__init__.py", line 38, in init
    raise InvalidRCLPYImplementation()
rclpy.exceptions.InvalidRCLPYImplementation: requested invalid rmw implementation
```

dhood:
I agree that we should add a better error message. 

At the moment though, it's a fault of the code and not of the user. So, if I were testing ros2 for the first time I would not expect to have to do the workaround for this myself (especially in one of the first nodes that someone would try)

wjwwood:
Ok, we're going to merge this and fix it in a more permanent way after the alpha 5 release.

wjwwood:
I opened a new issue to track this: https://github.com/ros2/rclpy/issues/11

dirk-thomas:
Fails PEP8 check: http://ci.ros2.org/job/ci_linux/1161/testReport/

PR #12, opened 2016-04-06
Improved rclpy import failed message

wjwwood:
I achieved this by adding a helpful message which prints when the import fails, but only if it goes unhandled.

I wanted to avoid printing it always incase someone is explicitly trying each of the implementations and wants to avoid the logging output. To accomplish this I added a custom `excepthook` for `rclpy` to override the built-in `sys.excepthook`. You can also add "addendum's" for uncaught exceptions, which will printed before the traceback. In the custom `excepthook`, it looks for a matching addendum, prints it if it finds one, and then passes along the unhandled exception to the original `excepthook`.

This is the observed behavior:

- In the normal case you get something like:

```
% python3 -c "import talker_py; talker_py.main()"

Failed to import the Python extension for the 'rmw_fastrtps_cpp' rmw implementation.
A different rmw implementation can be selected using the 'RCLPY_IMPLEMENTATION' env variable.
These are the available rmw implementations:
  - rmw_fastrtps_cpp
  - rmw_opensplice_cpp

  File "<string>", line 1, in <module>
  File "/Users/william/ros2_ws/build_isolated/rclpy_examples/talker_py.py", line 27, in main
    rclpy.init(args)
  File "/Users/william/ros2_ws/install_isolated/rclpy/lib/python3.5/site-packages/rclpy/__init__.py", line 47, in init
    rmw_implementation_tools.import_rmw_implementation()
  File "/Users/william/ros2_ws/install_isolated/rclpy/lib/python3.5/site-packages/rclpy/impl/rmw_implementation_tools.py", line 90, in import_rmw_implementation
    __rmw_implementation_module = importlib.import_module(module_name, package='rclpy')
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/importlib/__init__.py", line 126, in import_module
  File "<frozen importlib._bootstrap>", line 986, in _gcd_import
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 956, in _find_and_load_unlocked
ImportError: No module named 'rclpy._rclpy__rmw_fastrtps_cpp'
```

- In the case that you catch it, you get:
```
% python3 -c "
import talker_py
try:
    talker_py.main()
except ImportError:
    print('failed to import')
"
failed to import
```

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -20,8 +20,12 @@ from rclpy.exceptions import InvalidRCLPYImplementation
 from rclpy.impl import implementation_singleton
 from rclpy.impl import rmw_implementation_tools
 from rclpy.node import Node
+from rclpy.impl import excepthook
 
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+
+# install the excepthook
+excepthook.install_rclpy_excepthook()
 
 
 def init(args=None):
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -19,6 +19,7 @@ import sys
 from rclpy.exceptions import InvalidRCLPYImplementation
 from rclpy.impl import implementation_singleton
 from rclpy.impl import rmw_implementation_tools
+from rclpy.impl.rmw_implementation_tools import RCLPY_IMPLEMENTATION_ENV_NAME
 from rclpy.node import Node
 from rclpy.impl import excepthook
 
@@ -29,7 +30,7 @@ excepthook.install_rclpy_excepthook()
 
 
 def init(args=None):
-    rclpy_rmw_env = os.getenv('RCLPY_IMPLEMENTATION', None)
+    rclpy_rmw_env = os.getenv(RCLPY_IMPLEMENTATION_ENV_NAME, None)
     if rclpy_rmw_env is not None:
         available_rmw_implementations = rmw_implementation_tools.get_rmw_implementations()
         if rclpy_rmw_env not in available_rmw_implementations:
--- a/rclpy/rclpy/exceptions.py
+++ b/rclpy/rclpy/exceptions.py
@@ -13,13 +13,6 @@
 # limitations under the License.
 
 
-class NotInitializedException(Exception):
-    """Raised when the rclpy implementation is accessed before rclpy.init()."""
-
-    def __init__(self, *args):
-        Exception.__init__(self, 'rclpy.init() has not been called', *args)
-
-
 class ImplementationAlreadyImportedException(Exception):
     """Raised on select_rmw_implemenation() after import_rmw_implementation() has been called."""
 
@@ -32,3 +25,17 @@ class InvalidRCLPYImplementation(Exception):
 
     def __init__(self, *args):
         Exception.__init__(self, 'requested invalid rmw implementation', *args)
+
+
+class NotInitializedException(Exception):
+    """Raised when the rclpy implementation is accessed before rclpy.init()."""
+
+    def __init__(self, *args):
+        Exception.__init__(self, 'rclpy.init() has not been called', *args)
+
+
+class NoImplementationAvailableException(Exception):
+    """Raised when there is no rmw implementation with a Python extension available."""
+
+    def __init__(self, *args):
+        Exception.__init__(self, 'no rmw implementation with a Python extension available')
--- a/rclpy/rclpy/impl/excepthook.py
+++ b/rclpy/rclpy/impl/excepthook.py
@@ -20,7 +20,7 @@ __unhandled_exception_addendums = {}
 
 
 def rclpy_excepthook(exc_type, value, traceback):
-    """rclpy's custom except hook for unhandled exceptions.
+    """The rclpy custom except hook for unhandled exceptions.
 
     This excepthook will check for any unhandled exception addendum's,
     log them if they exist, and then call the original excepthook.
--- a/rclpy/rclpy/impl/excepthook.py
+++ b/rclpy/rclpy/impl/excepthook.py
@@ -1,0 +1,60 @@
+# Copyright 2016 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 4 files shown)


wjwwood:
I also addressed the style problems from https://github.com/ros2/rclpy/pull/10#issuecomment-206076846.

wjwwood:
Actually, scratch that. I remove the commit that cleans up #10 so I can put it in a different pr.

rclpy/rclpy/impl/rmw_implementation_tools.py
```diff
@@ -72,17 +75,48 @@ def import_rmw_implementation():
...
         logger.debug("Implicitly selecting the '{0}' rmw implementation."
                      .format(__selected_rmw_implementation))
     module_name = '._rclpy__{rmw_implementation}'.format(
         rmw_implementation=__selected_rmw_implementation,
     )
-    __rmw_implementation_module = importlib.import_module(module_name, package='rclpy')
+    try:
+        __rmw_implementation_module = importlib.import_module(module_name, package='rclpy')
+    except ImportError as exc:
+        if "No module named 'rclpy._rclpy__" in str(exc):
```
jacquelinekay: I wonder if you could do slightly better than hardcoding the module prefix here. If the module name changed from `rclpy._rclpy__rmw_....` to something else then this logic would break. Wouldn't something like `"No module named 'rclpy.{0}".format(module_name)` work?

wjwwood:
CI:

- http://ci.ros2.org/job/ci_linux/1174/
- http://ci.ros2.org/job/ci_osx/949/
- http://ci.ros2.org/job/ci_windows/1229/

wjwwood:
I fixed http://ci.ros2.org/job/ci_linux/1174/testReport/junit/projectroot.home.rosbuild.ci_scripts.ws.src.ros2.rclpy/rclpy/pep257/ in b994e52, it shouldn't need a retest.

wjwwood:
I believe all the CI is satisfactory (with my pep257 fix), looking for final reviews before squash and merge.

wjwwood:
> I believe all the CI is satisfactory (with my pep257 fix), looking for final reviews before squash and merge.

bump

jacquelinekay:
+1

rclpy/rclpy/impl/rmw_implementation_tools.py
```diff
@@ -72,17 +75,48 @@ def import_rmw_implementation():
...
     if __rmw_implementation_module is not None:
         return __rmw_implementation_module
+    available_implementations = get_rmw_implementations()
     if __selected_rmw_implementation is None:
         logger = logging.getLogger('rclpy')
-        select_rmw_implementation(get_rmw_implementations()[0])  # select the first one
+        if available_implementations:
+            select_rmw_implementation(available_implementations[0])  # select the first one
+        else:
+            raise NoImplementationAvailableException()
```
dirk-thomas: I would suggest the opposite order:

```
if not available_implementations:
    raise NoImplementationAvailableException()
select_rmw_implementation(available_implementations[0])  # select the first one
```

rclpy/rclpy/__init__.py
```diff
@@ -19,13 +19,18 @@
 from rclpy.exceptions import InvalidRCLPYImplementation
 from rclpy.impl import implementation_singleton
 from rclpy.impl import rmw_implementation_tools
+from rclpy.impl.rmw_implementation_tools import RCLPY_IMPLEMENTATION_ENV_NAME
 from rclpy.node import Node
+from rclpy.impl import excepthook
```
dirk-thomas: Is there a reason for `excepthook` being imported at the end (instead of four lines above)? If yes, I think a comment would be helpful.

rclpy/rclpy/exceptions.py
```diff
@@ -32,3 +25,17 @@ class InvalidRCLPYImplementation(Exception):
 
     def __init__(self, *args):
         Exception.__init__(self, 'requested invalid rmw implementation', *args)
+
+
+class NotInitializedException(Exception):
+    """Raised when the rclpy implementation is accessed before rclpy.init()."""
+
+    def __init__(self, *args):
+        Exception.__init__(self, 'rclpy.init() has not been called', *args)
```
dirk-thomas: Why the reorder?

rclpy/rclpy/impl/excepthook.py
```diff
@@ -0,0 +1,60 @@
...
+    """Replace the system excepthook with the custom rclpy excepthook.
+
+    Repeated calls to this function will return without any side effects.
+    """
+    global __original_excepthook
+    if __original_excepthook is not None:
+        # The rclpy excepthook has already been installed.
+        return
+    __original_excepthook = sys.excepthook
+    sys.excepthook = rclpy_excepthook
```
dirk-thomas: Would it make sense to return `True` or `False` to indicate if the function changed the hook or not?

rclpy/rclpy/exceptions.py
wjwwood: alpha of the classes

rclpy/rclpy/impl/rmw_implementation_tools.py
```diff
@@ -72,17 +75,48 @@ def import_rmw_implementation():
...
+                rmw_implementations_msg,
+            ]
+            # This message will only be printed if this exception goes unhandled.
+            add_unhandled_exception_addendum(
+                exc,
+                "\n"
+                "Failed to import the Python extension for the '{0}' rmw implementation.\n"
+                "A different rmw implementation can be selected using the '{1}' env variable.\n"
+                "These are the available rmw implementations:\n"
+                "{2}\n".format(*log_args)
```
dirk-thomas: Would named args be easier to read then positional args?

rclpy/rclpy/impl/excepthook.py
wjwwood: I don't see a need to do so, but you can pr it if you think it's important.

rclpy/rclpy/impl/rmw_implementation_tools.py
wjwwood: I don't think it makes a difference, but if you want to change it I'm fine with that.

rclpy/rclpy/__init__.py
wjwwood: No, I'll fix that up.

rclpy/rclpy/impl/rmw_implementation_tools.py
wjwwood: I guess it would be clearer, but this works for me. Since it's not a functional change, I'm not going to do it right now. Feel free to change it yourself.

rclpy/rclpy/__init__.py
wjwwood: https://github.com/ros2/rclpy/pull/19

PR #13, opened 2016-04-06
register rmw python extensions separately

wjwwood:
I added a new "ament index" entry specifically for rmw implementations with Python extensions and use that to load from. This removes the need for the blacklist in #10 and also addresses the style issues from #10.

I also removed the exceptions in the tests for these failures. However, in order to get the tests working I had to add another patch to `ament_cmake` so that I can use the ament index before installing.


wjwwood:
See: https://github.com/ament/ament_cmake/pull/65

dirk-thomas:
LGTM

PR #15, opened 2016-04-06
Reenable connext dynamic

jacquelinekay:
With the changes made in ros2/rosidl#115 this will allow us to use the dynamic implementation of Connext in rclpy.


jacquelinekay:
Seeing if Connext Dynamic and Python get along:

http://ci.ros2.org/job/ci_linux/1248/
http://ci.ros2.org/job/ci_osx/1003/
http://ci.ros2.org/job/ci_windows/1286/

jacquelinekay:
I forgot that `rclpytests` doesn't produce different test results for the different middlewares. Locally I can see that the package built a `_rclpy__rmw_connext_dynamic_cpp.cpython(etc...).so`.

jacquelinekay:
nevermind, I have a better solution coming up.

PR #19, opened 2016-04-26
fix import order

wjwwood:
Follow up to https://github.com/ros2/rclpy/pull/12


wjwwood:
I'm gonna merge this since it's trivial. I tested it locally for bizarre consequences.

PR #23, opened 2016-05-04
add tests for array

mikaelarguedas:
Add testing for primitive arrays bounded or not + tests for nested messages arrays
Note: feature working but tests failing on windows because of underlying rclpy issues that should be addressed in a separate PR.

Connects to https://github.com/ros2/rosidl/issues/114

Issue #17, opened 2016-04-14
Publisher crashes if message contains empty strings


mikaelarguedas:
If message contains either uninitialized string, an empty string, or string containing only a null terminating character, then `rclpy` publisher crashes:

```
>>> rclpy._rclpy.rclpy_publish(self.publisher_handle, msg)
RuntimeError: Failed to publish: error not set
```

PR #25, opened 2016-05-06
Fix null terminator

mikaelarguedas:
This removes the now useless string initialization in test files

Connects to https://github.com/ros2/rclpy/issues/17

Relies on https://github.com/ros2/rclpy/pull/23


wjwwood:
d86dddf6ee1a09733d6fe7997201842577ccd7d7 +1

PR #26, opened 2016-05-12
update style to pass new uncrustify version

dirk-thomas:
While the wrapped line without any extra indentation doesn't look appealing this is the only way I got it to pass the new uncrustify version.

Connect to ament/uncrustify#13

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -363,43 +363,33 @@ rclpy_shutdown(PyObject * Py_UNUSED(self), PyObject * Py_UNUSED(args))
 }
 
 static PyMethodDef rclpy_methods[] = {
-  {"rclpy_init", rclpy_init, METH_VARARGS,
-   "Initialize RCL."},
-  {"rclpy_create_node", rclpy_create_node, METH_VARARGS,
-   "Create a Node."},
-  {"rclpy_create_publisher", rclpy_create_publisher, METH_VARARGS,
-   "Create a Publisher."},
-  {"rclpy_publish", rclpy_publish, METH_VARARGS,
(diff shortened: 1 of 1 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -363,43 +363,33 @@ rclpy_shutdown(PyObject * Py_UNUSED(self), PyObject * Py_UNUSED(args))
...
-   "rclpy_ok."},
+  {"rclpy_ok", rclpy_ok, METH_NOARGS, "rclpy_ok."},
 
-  {"rclpy_shutdown", rclpy_shutdown, METH_NOARGS,
-   "rclpy_shutdown."},
+  {"rclpy_shutdown", rclpy_shutdown, METH_NOARGS, "rclpy_shutdown."},
 
   {"rclpy_get_rmw_implementation_identifier", rclpy_get_rmw_implementation_identifier,
-   METH_NOARGS, "Retrieve the identifier for the active RMW implementation."},
+  METH_NOARGS, "Retrieve the identifier for the active RMW implementation."},
```
wjwwood: This doesn't look right to me. Maybe you have to expand the block so that the opening and closing braces are on their own lines.

dirk-thomas: I agree with you. Please see my comment on the referenced issue (https://github.com/ament/uncrustify/pull/13#issue-154550137). The new version has some severe regressions and I don't expect to merge this as-is. Therefore this is still marked with the label "in progress".

wjwwood: :+1:

dirk-thomas:
Obsolete with the latest version of uncrustify from master.

PR #27, opened 2016-05-12
add testing for init and shutdown

mikaelarguedas:


rclpy/test/test_init_shutdown.py
```diff
@@ -0,0 +1,86 @@
...
+    except:
+        print('exception in {0}():'.format(func.__name__), file=sys.stderr)
+        traceback.print_exc()
+        raise
+
+
+def func_init_shutdown():
+    import rclpy
+    rclpy.init()
+    return ctypes.c_bool(rclpy.shutdown() is None)
```
dirk-thomas: Why is `ctypes.c_bool` needed here (and below)?

mikaelarguedas: It is not. removed in https://github.com/ros2/rclpy/pull/27/commits/52e6158bf252114c91234815728f10a63c2e29ad

rclpy/test/test_init_shutdown.py
```diff
@@ -0,0 +1,85 @@
...
+    except:
+        print('exception in {0}():'.format(func.__name__), file=sys.stderr)
+        traceback.print_exc()
+        raise
+
+
+def func_init_shutdown():
+    import rclpy
+    rclpy.init()
+    return rclpy.shutdown() is None
```
dirk-thomas: What is `rclpy.shutdown` supposed to return?

rclpy/test/test_init_shutdown.py
```diff
@@ -0,0 +1,85 @@
...
+        [func_double_shutdown, 'Expected Runtime error when shutting down rclpy twice']
+    ]
+
+    for fct in testlist:
+        result = pool.apply(
+            func=run_catch_report_raise,
+            args=(fct[0],)
+        )
+
+        assert result, fct[1]
```
dirk-thomas: The problem of writing the test like this is that the first failing condition will abort the test. It would be better if all tests are run all the time - despite the result of the individual tests.

mikaelarguedas: Good point, I thought that given the limited scope of testing it would be ok, but I can definitely convert each function into an independent test

rclpy/test/test_init_shutdown.py
mikaelarguedas: `rclpy_shutdown` indeed raises a RuntimeError if anything goes wrong otherwise returns None on success. So I'll just catch RuntimeErrors instead of testing return value

rclpy/test/test_init_shutdown.py
```diff
@@ -0,0 +1,143 @@
...
+def test_double_shutdown():
+    result = pool = multiprocessing.Pool(1)
+    pool.apply(
+        func=run_catch_report_raise,
+        args=(func_double_shutdown,)
+    )
+
+    assert result, 'Expected Runtime error when shutting down rclpy twice'
+    pool.close()
+    pool.join()
```
dirk-thomas: Each test function duplicates 9 lines of code. I would suggest to create a function for that and call it from all test functions.

mikaelarguedas: Yeah sorry about that, poor refactoring

tfoote:
The code lgtm, do you want to run ci on this?

mikaelarguedas:
Pending round of CI:
- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1353/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1060/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/1362/)

mikaelarguedas:
CI passed on all platforms

tfoote:
+1

rclpy/test/test_init_shutdown.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def func_double_shutdown():
+    import rclpy
+    rclpy.init()
+    rclpy.shutdown()
+    try:
+        rclpy.shutdown()
+    except RuntimeError:
+        return True
```
dhood: I know this has been merged already but perhaps we should think about making custom exceptions for these tests. It seems like it could be easy for the tests to pass even if something is not working correctly, in the case that whatever is broken is also raising a `RuntimeError`

wjwwood: I agree. I think there are lots of "RuntimeErrors" because they are bubbling up from the Python c-extension. But that should change. Can you ticket it?

dhood: ok I will open an issue. for my understanding: even using the python c-extension, we can make custom exceptions [like this](https://github.com/ros/geometry2/blob/indigo-devel/tf2_py/src/tf2_py.cpp#L497), right? (or is there a reason that's not possible in this context)

mikaelarguedas: I agree that given that everything in rclpy returns RuntimeError on failure it's not ideal for testing. Then given that these tests are mutually exclusive I can't think of a scenario where they could all pass while the init/shutdown functions are not working as expected.
Ticketed here https://github.com/ros2/rclpy/issues/31

mikaelarguedas: Yeah we can definitely create our own errors for it even if it comes from the c extension side.

dhood: Ok thanks for confirming that

Issue #24, opened 2016-05-04
de-duplicate the logic to select the default rmw implementation


wjwwood:
Carried over from https://github.com/ros2/rclpy/issues/11.

The logic exists in CMake: https://github.com/ros2/rmw/blob/739190be1ddfb7470652152148297704360287e3/rmw_implementation_cmake/cmake/get_default_rmw_implementation.cmake#L36

And here in Python: https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/impl/rmw_implementation_tools.py#L89

wjwwood:
This has been done in https://github.com/ros2/rclpy/commit/8aabbd2037a88034a2f0b85e3fe751f4515608a3 and https://github.com/ros2/rclpy/commit/9e0597807ad9b64397ba35eaa549f569978a039e

PR #28, opened 2016-05-12
refactor talker and listener to run in independent processes

mikaelarguedas:
addresses current failing rclpy tests

Waiting for CI to see if it actually fixes the problem on windows


mikaelarguedas:
I don't know if https://github.com/ros2/rclpy/pull/28/commits/d98d61e6750f2c10f23b19ba738ee7f086a94bf2 actually makes sense.

Seems to be solving the failing tests on windows:
http://ci.ros2.org/job/ci_windows/1329/
http://ci.ros2.org/job/ci_linux/1297/
http://ci.ros2.org/job/ci_osx/1028/

rclpy/test/listener.py
```diff
@@ -80,12 +81,18 @@ def listener(message_pkg, message_name, rmw_implemenatation, number_of_cycles=5)
         'Should have received a {} message from talker'.format(message_name)
 
 if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-p', '--message_pkg', default=None)
+    parser.add_argument('-m', '--message_name', default=None)
```
dirk-thomas: Does the default make any sense? What happens if these are `None`?

rclpy/test/listener.py
```diff
@@ -43,7 +44,7 @@ def listener_cb(msg, message_name, received_messages):
     received_messages.append(msg)
 
 
-def listener(message_pkg, message_name, rmw_implemenatation, number_of_cycles=5):
+def listener(message_pkg, message_name, rmw_implementation, number_of_cycles):
```
dirk-thomas: Is it intended that `rmw_implementation` is not used anywhere?

rclpy/test/listener.py
```diff
@@ -80,12 +81,18 @@ def listener(message_pkg, message_name, rmw_implemenatation, number_of_cycles=5)
         'Should have received a {} message from talker'.format(message_name)
 
 if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-p', '--message_pkg', default=None)
+    parser.add_argument('-m', '--message_name', default=None)
+    parser.add_argument('-r', '--rmw_implementation', default=None)
+    parser.add_argument('-n', '--number_of_cycles', default=5, type=int)
```
dirk-thomas: Use `ArgumentDefaultsHelpFormatter` to show the default in the help text.

mikaelarguedas: Done

rclpy/test/listener.py
mikaelarguedas: Yes it is, the test will be extended to use it in https://github.com/ros2/rclpy/pull/29

rclpy/test/listener.py
mikaelarguedas: Done

mikaelarguedas:
no more rclpy test failure on CI:
http://ci.ros2.org/job/ci_linux/1300/
http://ci.ros2.org/job/ci_windows/1335/
http://ci.ros2.org/job/ci_osx/1030/

rclpy/test/talker.py
```diff
@@ -20,34 +22,78 @@
...
         time.sleep(1)
+    rclpy.shutdown()
 
 if __name__ == '__main__':
+    parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('-p', '--message_pkg', default='std_msgs',
+                        help='name of the message package')
+    parser.add_argument('-m', '--message_name', default='String',
+                        help='name of the ROS message')
+    parser.add_argument('-r', '--rmw_implementation', default='rmw_opensplice_cpp',
```
dirk-thomas: If the system has only one rmw implementation but not OpenSplice this default is not very helpful. Should the default be selected based on the ament index instead?

rclpy/test/test_talker_listener.py
```diff
@@ -12,126 +12,83 @@
...
-        chatter_callback,
-        qos_profile_default)
+    rc = launcher.launch()
 
     spin_count = 0
-    while len(received_messages) == 0 and spin_count < 10 and launcher.is_launch_running():
-        rclpy.spin_once(node)  # This will wait 1 second or until something has been done.
+    while spin_count < nbmessages and launcher.is_launch_running():
+        time.sleep(1)
         spin_count += 1
```
dirk-thomas: `spin_count` doesn't really make sense anymore if `spin` is not called?

mikaelarguedas: I could call this variable differently indeed, it's just to have a metric to be sure no test will keep hanging forever. Would `msg_count` work? (to stay coherent with `nbmessages` used above)

rclpy/test/talker.py
mikaelarguedas: That's a good point. I'll look into how to use ament index then

rclpy/test/test_talker_listener.py
dirk-thomas: But you can't just increment `msg_count` in the loop either. That semantic would be weird too. Currently `nbmessages` is more like "max-runtime-in_seconds".

mikaelarguedas: makes sense thanks.

mikaelarguedas:
Done, thanks @dirk-thomas for your patience

rclpy/test/listener.py
```diff
@@ -0,0 +1,89 @@
...
+
+if __name__ == '__main__':
+    rmw_implementations = sorted(ament_index_python.get_resources('rmw_python_extension').keys())
+    parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('-p', '--message_pkg', default='std_msgs',
+                        help='name of the message package')
+    parser.add_argument('-m', '--message_name', default='String',
+                        help='name of the ROS message')
+    # using the same logic as the default import_rmw_implementation()
+    parser.add_argument('-r', '--rmw_implementation', default=rmw_implementations[0],
```
dirk-thomas: I guess implementing the logic to choose the default rmw impl. here again would go against the goal to only implement this in as few as possible locaitons: #24

wjwwood: You could default to `None` and let `rclpy` decide (it will use the default by default).

rclpy/test/listener.py
```diff
@@ -0,0 +1,89 @@
...
+# this is needed to allow rclpy to be imported from the build folder
+sys.path.insert(0, os.getcwd())
+
+
+def listener_cb(msg, message_name, received_messages):
+    print('received: %r' % msg)
+    received_messages.append(msg)
+
+
+def listener(message_pkg, message_name, rmw_implementation, number_of_cycles):
```
wjwwood: Also, `rmw_implementation` is not being used in this function?

rclpy/test/listener.py
wjwwood: You could, however, use `rmw_implementations` to limit the choices to the argument: https://docs.python.org/3/library/argparse.html#choices.

rclpy/test/listener.py
mikaelarguedas: Yeah sorry, this shouldn't be here and only in the following PR #29 which will actually use it

rclpy/test/listener.py
mikaelarguedas: Sounds good to me, and the talker will try to `select_rmw_implementation` only if the provided argument is not None.

mikaelarguedas: Good point, I'll do that and move all the `rmw_implementation` related stuff to #29 to avoid confusion

tfoote:
From the meeting it would be good to refactor the message definitions into a separate test_interfaces package for shared coverage of message datatype corner cases. But this can be done in a followup.

mikaelarguedas:
Closing this in favour of #30 and ros2/system_tests#121

Issue #22, opened 2016-05-03
test multiple rmw implementations for talker/listener test


jacquelinekay:
If I set the environment variable `RCLPY_IMPLEMENTATION` to a valid rmw implementation name and then run tests for rclpy, the set_rmw_implementation tests fail with output like:

```
1: Traceback (most recent call last):
1:   File "/usr/lib/python3.4/multiprocessing/pool.py", line 119, in worker
1:     result = (True, func(*args, **kwds))
1:   File "/home/jackie/code/ros2_ws/src/ros2/rclpy/rclpy/test/test_select_rmw_implementation.py", line 24, in run_catch_report_raise
1:     return func(*args, **kwargs)
1:   File "/home/jackie/code/ros2_ws/src/ros2/rclpy/rclpy/test/test_select_rmw_implementation.py", line 46, in func_import_each_available_rmw_implementation
1:     rmw_implementation, set_rmw)
1: AssertionError: expected 'rmw_connext_dynamic_cpp' but got 'rmw_connext_cpp'
1: """
1: 
1: The above exception was the direct cause of the following exception:
1: 
1: Traceback (most recent call last):
1:   File "/usr/lib/python3/dist-packages/nose/case.py", line 198, in runTest
1:     self.test(*self.arg)
1:   File "/home/jackie/code/ros2_ws/src/ros2/rclpy/rclpy/test/test_select_rmw_implementation.py", line 56, in test_import_each_available_rmw_implementation
1:     args=(func_import_each_available_rmw_implementation, rmw_implementation,)
1:   File "/usr/lib/python3.4/multiprocessing/pool.py", line 253, in apply
1:     return self.apply_async(func, args, kwds).get()
1:   File "/usr/lib/python3.4/multiprocessing/pool.py", line 599, in get
1:     raise self._value
1: AssertionError: expected 'rmw_connext_dynamic_cpp' but got 'rmw_connext_cpp'
```

The purposes of trying to was so that I could set a different rmw implementation for rclpytests to see if connext dynamic was working.

Perhaps the test should unset the variable before running. Additionally, the talker/listener test should be changed to iterate through the different rmw implementations.

mikaelarguedas:
@wjwwood I could take that off your plate if you want to

wjwwood:
Sure, it will be a few days until I get to it.

mikaelarguedas:
Consider rewriting these tests using independent files for the listener side to fix flaky/failing tests on buildfarm.

[...]

PR #29, opened 2016-05-12
Rclpy test multiple rmw implementations

mikaelarguedas:
Connects to #22 

Depends on #28 

Test passes with connext_static and opensplice but it seems that connext_dynamic crashes on nested messages. Investigating now


rclpy/test/listener.py
```diff
@@ -0,0 +1,95 @@
...
+    assert len(received_messages) > 0,\
+        'Should have received a {} message from talker'.format(message_name)
+
+if __name__ == '__main__':
+    try:
+        listener(
+            message_pkg=sys.argv[1],
+            message_name=sys.argv[2],
+            rmw_implementation=sys.argv[3],
+            number_of_cycles=sys.argv[4])
```
dirk-thomas: Please use `argparse` for those arguments. Since it is only a test they don't need a help string imo but just naming them (and having `--help`) would be beneficial.

Same below.

rclpy/test/listener.py
```diff
@@ -0,0 +1,95 @@
...
+
+    node.create_subscription(
+        msg_mod,
+        'chatter__{}__{}'.format(rmw_implementation, message_name),
+        chatter_callback,
+        qos_profile_default)
+
+    spin_count = 1
+    print('talker: beginning loop')
+    while rclpy.ok() and spin_count < int(number_of_cycles) and len(received_messages) == 0:
```
dirk-thomas: `number_of_cycles` should always be `int`, once you use `argparse`.

rclpy/test/listener.py
```diff
@@ -0,0 +1,95 @@
...
+            msg.transform.rotation.x,
+            msg.transform.rotation.y,
+            msg.transform.rotation.z,
+            msg.transform.rotation.w))
+    elif message_name == 'Imu':
+        print('received: ({})'.format(msg.angular_velocity_covariance))
+    elif message_name == 'LaserScan':
+        print('received: ({})'.format(msg.intensities))
+    elif message_name == 'PointCloud2':
+        print('received: ({})'.format([f.name for f in msg.fields]))
```
dirk-thomas: Each message should be able to print itself and make this conditionals obsolete:

    print('received: %r' % msg)

See ros2/rosidl#134 for the necessary `__repr__` method for Python message objects.

rclpy/test/listener.py
mikaelarguedas: ok I added it to the parent PR commits https://github.com/ros2/rclpy/pull/28/commits/a8be540c38cad8dae1391d9bd34a00395a25b098 and will rebase this one

mikaelarguedas:
Closing this since the tests will not be stored here anymore. A follow up PR adding rmw_implementations testing will be opened in https://www.github.com/ros2/system_tests

PR #32, opened 2016-06-08
Disable on Android

esteve:
Disable on Android

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -2,6 +2,11 @@ cmake_minimum_required(VERSION 2.8.3)
 
 project(rclpy C)
 
+if(ANDROID)
+  message(STATUS "rclpy does not target Android, skipping...")
+  return()
+endif()
+
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
 find_package(rcl REQUIRED)


tfoote:
+1

wjwwood:
There was some concerns raised about where this returns from. Since it doesn't even call `ament_package()` it won't install anything, including a `package.xml`. And so it leaves us, at least me, wondering what happens to packages which depend on `rclpy`? Maybe this style of "graceful failing" intended, but it's not clear how this package (and in general other packages) should be skipped when building for Android. Maybe it would useful to have some more insight into how you're doing the Android build.

esteve:
@wjwwood I disable `rclpy_examples` as well (which is the only one depending on `rclpy`), by either invoking `--skip-packages rclpy_examples` or adding an `AMENT_IGNORE` file, ideally there should be a way to skip `ament_python` packages without an explicit `ament` incantation, but I haven't looked into it. Any ideas?

wjwwood:
The `AMENT_IGNORE` files or just avoiding downloading the packages in question would be my suggestion. The reason I ask, is I'm wondering why you don't use `AMENT_IGNORE` in this package too rather than the suggested change here. Or put another way, why is this the only package which is disabled this, why not disable the other packages like rclpy_examples this way too?

My personal preference would be to disable any packages with `AMENT_IGNORE` files or remove them from the source space, rather than explicitly detecting and disabling all packages which don't work on Android. Simply because it keeps the `CMakeLists.txt` boilerplate in all of those packages smaller.

However, I'm willing to go with whatever is most convenient for you since you're working on Android, but we should be consistent either way.

Issue #38, opened 2016-09-27
qos support


geoffviola:
It seems that the qos settings are not available in rclpy: https://github.com/ros2/rclpy/blob/master/rclpy/src/rclpy/_rclpy.c#L63. Is this true? When might such a feature be included?

wjwwood:
No, it's not supported yet. The Python interface is under active development and we hope to add the remaining features (like QoS) and functions (like `count_publishers` and friends) soon.

@mikaelarguedas fyi

PR #39, opened 2016-09-30
Addin QoS support

mikaelarguedas:
connects to #38


mikaelarguedas:
I don't really like the idea to have to go back and forth between c and python to get `rmw_qos_profile_t` but that's the only way I found to avoid creating an extension for the `QoSProfile` class. @wjwwood any advice about a better way to do this ?

rclpy/rclpy/qos.py
```diff
@@ -12,8 +12,114 @@
...
+    def depth(self):
+        """Get int 'depth'."""
+        return self._depth
+
+    @depth.setter
+    def depth(self, value):
+        assert isinstance(value, int)
+        self._depth = value
+
+    def get_QoS_Profile(self):
```
wjwwood: nitpick: I think it would be better to use all lowercase in this method name.

rclpy/rclpy/qos.py
```diff
@@ -12,8 +12,114 @@
...
+    RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT = 0
+    RMW_QOS_POLICY_TRANSIENT_LOCAL_DURABILITY = 1
+    RMW_QOS_POLICY_VOLATILE_DURABILITY = 2
+
+
+qos_profile_default = QoSProfile(
+    history=QoSHistoryPolicy.RMW_QOS_POLICY_HISTORY_SYSTEM_DEFAULT,
+    depth=10,
+    reliability=QoSReliabilityPolicy.RMW_QOS_POLICY_RELIABILITY_SYSTEM_DEFAULT
+    # durability=QoSDurabilityPolicy.RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT
```
wjwwood: Why is this commented? If it should remain in the code, but commented out I'd suggest a TODO.

rclpy/rclpy/qos.py
wjwwood: You might want to call this something like `get_c_qos_profile`, which follows from the ctypes conventions:

https://docs.python.org/3/library/ctypes.html#ctypes-fundamental-data-types-2

I don't think this is a bad way of making this work. It's nice that you can access the C version of the structure from Python if needed, though I don't expect anyone will need to do this.

rclpy/rclpy/qos.py
```diff
@@ -12,8 +12,114 @@
...
+    RMW_QOS_POLICY_BEST_EFFORT = 2
+
+
+class QoSDurabilityPolicy(IntEnum):
+    RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT = 0
+    RMW_QOS_POLICY_TRANSIENT_LOCAL_DURABILITY = 1
+    RMW_QOS_POLICY_VOLATILE_DURABILITY = 2
+
+
+qos_profile_default = QoSProfile(
```
wjwwood: Also, wouldn't it be better to get these from the C implementation? That way the defaults aren't duplicated here?

rclpy/rclpy/qos.py
mikaelarguedas: fixed function name in https://github.com/ros2/rclpy/commit/93503d0be408db4d4906023bd0f062c5ce05a45a

rclpy/rclpy/qos.py
mikaelarguedas: removed when adressing https://github.com/ros2/rclpy/pull/39#pullrequestreview-2386728

rclpy/rclpy/qos.py
mikaelarguedas: done now importing qos profile defines in rmw

mikaelarguedas:
This PR now creates a new python module for rmw_agnostic rclpy functions. This allow to import and use these functions before `rclpy.init()` is called.
All comments have been addressed and it is ready for a new round of review

rclpy/rclpy/node.py
```diff
@@ -27,14 +27,16 @@ def __init__(self, handle):
     def create_publisher(self, msg_type, topic, qos_profile):
```
dhood: would it be appropriate to set the profile to the default profile here, so that the [talker example](https://github.com/ros2/examples/blob/1df40225eb0c0e6e8d1a201093fe90b0a247c898/rclpy_examples/talker_py.py#L33) can simplified further? or is it complicated because of default vs system default

rclpy/CMakeLists.txt
```diff
@@ -89,6 +89,32 @@ endmacro()
...
+)
+
+target_include_directories("${PROJECT_NAME}_implementation_rmw_agnostic"
+  PUBLIC
+  ${rmw_INCLUDE_DIRS}
+  ${PythonExtra_INCLUDE_DIRS}
+)
+
+install(TARGETS "${PROJECT_NAME}_implementation_rmw_agnostic"
+  DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}")
```
mikaelarguedas: Good point, I copy pasted the install rule from line 86 without thinking further. I'll update both

rclpy/rclpy/qos.py
```diff
@@ -12,8 +12,108 @@
...
+            QoSDurabilityPolicy.RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT)
+
+    @property
+    def history(self):
+        """Get QoSHistoryPolicy 'history'."""
+        return self._history
+
+    @history.setter
+    def history(self, value):
+        assert isinstance(value, QoSHistoryPolicy) or isinstance(value, int)
```
mikaelarguedas: Good point

rclpy/rclpy/qos.py
```diff
@@ -12,8 +12,108 @@
...
+        self.reliability = kwargs.get(
+            'reliability',
+            QoSReliabilityPolicy.RMW_QOS_POLICY_RELIABILITY_SYSTEM_DEFAULT)
+        self.durability = kwargs.get(
+            'durability',
+            QoSDurabilityPolicy.RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT)
+
+    @property
+    def history(self):
+        """Get QoSHistoryPolicy 'history'."""
```
mikaelarguedas: I assumed that properties and setters were self explanatory, anyone else think it's necessary? If so I'll open another PR to update that in all the python classes of the codebase that don't comply to this convention

dhood: the suggestion was more to make it clearer that `QoSHistoryProfile` in this comment is referring to the type (when I got to [this line](https://github.com/ros2/rclpy/pull/39/files/2ca4e21a729efaafce76ae43dd7ced7d04bd7293#diff-8110981f890a0a14e8a04fb4d65d5832R75) it wasn't immediately obvious)

rclpy/rclpy/node.py
mikaelarguedas: Confirmed that this parameter is optional in rclcpp API so I'll make it optional here as you suggested. Thanks for pointing it out.
I don't know if we'll simplify the examples though, underlying question is "do we want to show that the API allows you to pass qos profiles from the highest level ?"

rclpy/CMakeLists.txt
mikaelarguedas: fixed in https://github.com/ros2/rclpy/pull/39/commits/9f2a4512998175558d4a743a1346a307553680a9

rclpy/rclpy/qos.py
mikaelarguedas: fixed in https://github.com/ros2/rclpy/pull/39/commits/32ee2cf9a08f411e1bfc6b6f9bb84999fefdcb03

rclpy/rclpy/node.py
mikaelarguedas: fixed in https://github.com/ros2/rclpy/pull/39/commits/7d679518e3caa4413130276b222b1d76c5adf475

rclpy/rclpy/qos.py
mikaelarguedas: done https://github.com/ros2/rclpy/pull/39/commits/490d1d491df939c1cda1745c93eae7c2dda967f9

rclpy/rclpy/node.py
dhood: yeah, good point - thanks for making it possible so we can decide either way

wjwwood:
I think this is good to go, with CI.

mikaelarguedas:
CI:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1769)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1330/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1685/)

Now that we comply with D213 pep257 failed, no other test failure appeared.
style fixed in https://github.com/ros2/rclpy/pull/39/commits/b2f37fffc08d1f8d3a9ae7a4f116480e8b9da5e0

mikaelarguedas:
@wjwwood what is the final decision regarding this? should we hold this PR until after ROSCon ?

wjwwood:
If it's good to go, I'd say merge it.

mikaelarguedas:
better co convert from the right type... https://github.com/ros2/rclpy/commit/4c80fdc5af886e3647fc8724d88e768ef9e74131

PR #44, opened 2016-10-24
fix logic for python executable for Windows debug

mikaelarguedas:
fixes test failures on windows debug (http://ci.ros2.org/view/nightly/job/nightly_win_deb/258/#showFailuresLink)

Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1726/)


rclpy/CMakeLists.txt
```diff
@@ -14,8 +14,10 @@ find_package(PythonExtra MODULE)
 set(_PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE}")
 if(NOT WIN32)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
-elseif(CMAKE_BUILD_TYPE STREQUAL "Debug")
-  set(PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE_DEBUG}")
+else(NOT WIN32)
+  if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+    set(PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE_DEBUG}")
+  endif()
```
dirk-thomas: This patch is not changing the logic of the CMake code as far as I understand it. The Python executable variable gets updated when being on Windows (`else NOT WIN32`) and using debbug build type (`CMAKE_BUILD_TYPE STREQUAL "Debug"`) - before and after the patch. Can you please elaborate why this change should make a difference?

mikaelarguedas: Sorry for the noise, this should still be in progress, I'm still investigating why the python executable used is not the one expected a test time

mikaelarguedas:
not the right fix need investigation someday

mikaelarguedas:
will be replaced by @dirk-thomas PRs (provide link)

dirk-thomas:
Not directly but addressed by ros2/rclpy#82, ros2/ci#72 and ament/ament_tools#147.

PR #47, opened 2016-10-31
[spin_once] check waitset content to trigger subscription callbacks

mikaelarguedas:


rclpy/rclpy/__init__.py
```diff
@@ -67,11 +67,12 @@ def spin_once(node):
 
     _rclpy.rclpy_wait(wait_set)
 
-    # TODO(wjwwood): properly implement this by checking the contents of the wait_set.
-    for subscription in node.subscriptions:
-        msg = _rclpy.rclpy_take(subscription.subscription_handle, subscription.msg_type)
+    sub_ready_list = _rclpy.rclpy_get_ready_subscriptions(wait_set)
+    for sub_idx in sub_ready_list:
+        sub_ready = node.subscriptions[sub_idx]
```
wjwwood: From a multi threading standpoint this looks dangerous. What if `node.subscriptions` changes while the `rclpy_wait` function is blocking? I think something more sophisticated is needed to corrollate a subscription in the waitset from one in the node (other than a list index). Maybe you could use the underlying rcl_subscription_t pointer or the Python object's `id()` or a GUID of some sort.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -287,6 +288,27 @@ rclpy_wait_set_add_subscription(PyObject * Py_UNUSED(self), PyObject * args)
...
 
 static PyObject *
+rclpy_get_ready_subscriptions(PyObject * Py_UNUSED(self), PyObject * args)
+{
+  PyObject * pywait_set;
+  if (!PyArg_ParseTuple(args, "O", &pywait_set)) {
+    return NULL;
+  }
+
+  rcl_wait_set_t * wait_set = (rcl_wait_set_t *)PyCapsule_GetPointer(pywait_set, NULL);
```
wjwwood: What happens if `PyCapsule_GetPointer` returns null?

mikaelarguedas:
Updated this to use the subscription pointers instead of indexes. Not ideal given that these pointers could have been reallocated by the time we use them but I don't see a good way to access GUIDs from that high in the stack.
To test this with 3 subscriptions, you can checkout the branch `rclpy_spin_once` in both `rclpy` and `rclpy_examples` and run `talker_py` and `listener_py`

rclpy/rclpy/__init__.py
```diff
@@ -67,11 +67,15 @@ def spin_once(node):
...
-    # TODO(wjwwood): properly implement this by checking the contents of the wait_set.
-    for subscription in node.subscriptions:
-        msg = _rclpy.rclpy_take(subscription.subscription_handle, subscription.msg_type)
-        if msg:
-            subscription.callback(msg)
+    sub_ready_list = _rclpy.rclpy_get_ready_subscriptions(wait_set)
+
+    for sub_ptr in sub_ready_list:
+        for sub in node.subscriptions:
+            if sub_ptr == sub.subscription_pointer:
```
dirk-thomas: The two loops plus condition could be compressed into the following which expresses more clearly the intention:

```
for sub in [s for s in node.subscriptions if s.subscription_pointer in sub_ready_list]:
```

dirk-thomas: I don't think this should break anymore?

mikaelarguedas:
warning on windows: https://github.com/ros2/rclpy/commit/fa9f2864386b8e10155217f43015d8302f7a0543

PR #48, opened 2016-11-04
Add optional timeout to spin_once

mikaelarguedas:


rclpy/rclpy/__init__.py
```diff
@@ -56,16 +58,20 @@ def create_node(node_name):
...
 
     _rclpy.rclpy_wait_set_init(wait_set, len(node.subscriptions), 0, 0)
 
     _rclpy.rclpy_wait_set_clear_subscriptions(wait_set)
     for subscription in node.subscriptions:
         _rclpy.rclpy_wait_set_add_subscription(wait_set, subscription.subscription_handle)
+    if timeout_sec is None:
+        timeout = -1
+    else:
+        timeout = int(timeout_sec * S_TO_NS)
```
dirk-thomas: `int()` only needs to be called on the parameter `timeout_sec` (not on the know factor / constant).

mikaelarguedas: https://github.com/ros2/rclpy/pull/48/commits/807cf8fe49c4aa6a532c96383b141f1c9f9a85b2, thanks

dirk-thomas:
Should the parameter be checked for reasonable value ranges?

Beside that comment the patch lgtm.

mikaelarguedas:
I though relying on the checks performed by rcl in the [wait function](https://github.com/ros2/rcl/blob/master/rcl/src/rcl/wait.c#L523-L553), do you think we should expose a lower range to the users in the client libraries ?

dirk-thomas:
Sure, we can rely on the underlying function which accepts any integer. The Python API will just accept `None` as well as negative numbers for the same purpose. I guess once the functions has a docblock that should be fine.

rclpy/rclpy/__init__.py
```diff
@@ -56,16 +58,20 @@ def create_node(node_name):
...
 
     _rclpy.rclpy_wait_set_init(wait_set, len(node.subscriptions), 0, 0)
 
     _rclpy.rclpy_wait_set_clear_subscriptions(wait_set)
     for subscription in node.subscriptions:
         _rclpy.rclpy_wait_set_add_subscription(wait_set, subscription.subscription_handle)
+    if timeout_sec is None:
+        timeout = -1
+    else:
+        timeout = int(timeout_sec) * S_TO_NS
```
dirk-thomas: This logic doesn't work when passing floats e.g. `0.5`.

mikaelarguedas: I knew there was a reason the cast was perform on the entire expression, I just forgot by the time I saw your comment ... :confounded: 
reverting https://github.com/ros2/rclpy/pull/48/commits/807cf8fe49c4aa6a532c96383b141f1c9f9a85b2

dirk-thomas: :wink:

mikaelarguedas: Sorry about that: https://github.com/ros2/rclpy/pull/48/commits/528047986793cf4604119641505dcf80d76d4704

dirk-thomas: The I would suggest to do the following:

```int(float(timeout_sec) * S_TO_NS)```

**Comment by mikaelarguedas** [comment]
Agreed, rclpy is not documented at all ATM.
I'll open a pull request to add documentation for all functions (Python and C) before iterating on services.

File: rclpy/rclpy/__init__.py (hunk 2 above)
Comment (mikaelarguedas) [inline rclpy/rclpy/__init__.py]: done https://github.com/ros2/rclpy/pull/48/commits/59412f508349bb082663e7066f67db852dab72e4

Comment (dirk-thomas) [inline rclpy/rclpy/__init__.py]: Then I would suggest to do the following:

```int(float(timeout_sec) * S_TO_NS)```

Try calling the current code with 'bad' as the argument for the `timeout_sec`... :upside_down_face:

File: rclpy/rclpy/__init__.py [hunk 3]
```diff
@@ -58,18 +58,19 @@ def create_node(node_name):
...
 
     _rclpy.rclpy_wait_set_init(wait_set, len(node.subscriptions), 0, 0)
 
     _rclpy.rclpy_wait_set_clear_subscriptions(wait_set)
     for subscription in node.subscriptions:
         _rclpy.rclpy_wait_set_add_subscription(wait_set, subscription.subscription_handle)
-    if timeout_sec is None:
-        timeout = -1
-    else:
+    if isinstance(timeout_sec, float) or isinstance(timeout_sec, int):
```
dirk-thomas: This is not really necessary since the `float` call will already raise in cases where it can't be converted. Without this you can even pass `"1.23"`.

dirk-thomas:
:+1:

PR #49, opened 2016-11-08
document rclpy

mikaelarguedas:


rclpy/src/rclpy/_rclpy.c
```diff
@@ -60,6 +67,15 @@ rclpy_create_node(PyObject * Py_UNUSED(self), PyObject * args)
   return pynode;
 }
 
+/// Create a publisher
+/*
+ * \param[in] pynode Capsule pointing to the node to add the publisher to
+ * \param[in] pymsg_type ROS 2 type of the message this publisher will send
```
wjwwood: I'd recommend wording this one as "type support structure for the message type associated with the publisher". The doc should clue developers using this API about what they need to pass here (type as well as purpose). It's less import, in my opinion, to convey what this method will do with the parameter, e.g. in this case mentioning that the publisher will use this to send data of this type. If that information is important, I'd recommend specifying it in the body of the doc string rather than on the parameter description.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -319,6 +384,15 @@ rclpy_get_ready_subscriptions(PyObject * Py_UNUSED(self), PyObject * args)
   return sub_ready_list;
 }
 
+/// Wait until timeout is reached or event happened
+/*
+ * \param[in] pywait_set Capsule pointing to the waitset structure
+ * \param[in] timeout optional, time to wait in nanoseconds
+ *                negative value means wait forever
+ *                zero means no wait
+ *                positive value: time to wait in nanoseconds
```
wjwwood: This looks pretty awkward. I'd suggest explaining the different values in the body of the doc string and just leave this to something like "optional time to wait before waking up".

rclpy/src/rclpy/_rclpy.c
mikaelarguedas: Yeah I didn't know how to phrase it, that's a tricky one in the sense that the user passes a message module to the function (e.g. String) but we expect this module to already have the typesupport imported in it.
But they don't pass the typesupport structure itself given that this structure is extracted from the class in https://github.com/ros2/rclpy/blob/f36710dec992e8e1d0edb38477c0edf2e5086392/rclpy/src/rclpy/_rclpy.c#L103 
I'll try to rephrase this to make it clearer

rclpy/rclpy/qos.py
```diff
@@ -106,18 +106,36 @@ def get_c_qos_profile(self):
 
 
 class QoSHistoryPolicy(IntEnum):
+    """
+    Class defining enums for QoS History settings.
```
dirk-thomas: I would avoid starting with `Class` (similar as what the linters check for avoiding to start with `This ...`). And singular for enum (rather then enums). Suggesting:

> Enum for QoS History settings.

The plural / singular applies to all comments.

mikaelarguedas: Updated in https://github.com/ros2/rclpy/pull/49/commits/fd1e9b6597cefae1518723223a3a26e0b7683279

PR #53, opened 2016-11-16
use macro to limit code duplication

mikaelarguedas:
refactor mentionned in https://github.com/ros2/rclpy/pull/51#discussion_r88342583

Don't know where the right place to put the new macro would be ? a macros.h file ? on top of this file ?


dirk-thomas:
Since it is only used in one place I would suggest as close to that location as possible.

mikaelarguedas:
Passing CI:
http://ci.ros2.org/job/ci_linux/1921/
http://ci.ros2.org/job/ci_osx/1473/
http://ci.ros2.org/job/ci_windows/1874
linting fixed in https://github.com/ros2/rclpy/pull/53/commits/1edab98954035f8d6486c746f2df2037e4c51c8f

PR #54, opened 2016-11-17
Add timer support

mikaelarguedas:


rclpy/rclpy/timer.py
```diff
@@ -0,0 +1,44 @@
...
+        self._timer_period_ns = val
+        return val
+
+    @timer_period_ns.setter
+    def timer_period_ns(self, value):
+        print('changing timer.timer_period_ns')
+        val = int(value)
+        _rclpy.rclpy_change_timer_period(self.timer_handle, val)
+        self._timer_period_ns = val
+        print('done changing timer.timer_period_ns')
```
dirk-thomas: Should these print statements be removed?

mikaelarguedas:
It seems that we miss timing when running tests at higher frequency (kilohertz). I don't think it's tied to the implementation of this PR. I'll try to revive [this stale branch](https://github.com/ros2/rcl/tree/timer_waitset) that aimed to improve timer checking in wait_set at the rcl level

rclpy/rclpy/node.py
```diff
@@ -126,6 +137,15 @@ def destroy_client(self, client):
                 return True
         return False
 
+    def destroy_timer(self, timer):
+        for tmr in self.timers:
+            if tmr.timer_handle == timer.timer_handle:
+                _rclpy.rclpy_destroy_entity(
+                    'timer', tmr.timer_handle)
```
dirk-thomas: Should the return value of these C-calls be checked? (That would apply to many other calls as well.)

rclpy/test/test_timer.py
```diff
@@ -0,0 +1,91 @@
...
+        timestamps.append(time.time())
+
+    import rclpy
+
+    rclpy.init()
+    node = rclpy.create_node('talker')
+    timestamps = []
+    nb_iterations = 0
+
+    test_callback = functools.partial(timer_cb, timestamps=timestamps)
```
dirk-thomas: What is the purpose of `functools` here? Wouldn't a lambda be less complicated?

rclpy/test/test_timer.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def timer_cb(timestamps):
+        timestamps.append(time.time())
+
+    import rclpy
+
+    rclpy.init()
+    node = rclpy.create_node('talker')
+    timestamps = []
+    nb_iterations = 0
```
dirk-thomas: I would move this variable further down (as close to where it is actually being used).

rclpy/rclpy/node.py
mikaelarguedas: All the C functions are raising `RuntimeError` if anything goes wrong at the `rcl` level, and if the argument parsing fails the C-python API raises the [appropriate exception](https://docs.python.org/3/c-api/arg.html#c.PyArg_ParseTuple) for you. So the return value is not really meaningful in that context.

But I agree that error handling and asserts in this library need to be cleaned up/homogenized and it will be done in a follow-up PR.

rclpy/test/test_timer.py
mikaelarguedas: done in https://github.com/ros2/rclpy/pull/54/commits/57550a0e4ce3b691b8c9645fb3ab8a06bf0ee11f

mikaelarguedas:
Launching CI and documenting the C functions before merging:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1928)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1487/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1878/)

rclpy/rclpy/timer.py
```diff
@@ -35,6 +36,26 @@ def timer_period_ns(self, value):
...
+        return val
+
+    @is_canceled.setter
+    def is_canceled(self, value):
+        raise NotImplementedError('canceled is set by calling the `timer_cancel` function')
+
+    def cancel(self):
+        _rclpy.rclpy_cancel_timer(self.timer_handle)
+        self._is_canceled = True
+        return
```
dirk-thomas: Unnecessary line.

Same below.

rclpy/rclpy/timer.py
```diff
@@ -35,6 +36,26 @@ def timer_period_ns(self, value):
...
 
+    @property
+    def is_canceled(self):
+        val = _rclpy.rclpy_is_timer_canceled(self.timer_handle)
+        self._is_canceled = True
+        return val
+
+    @is_canceled.setter
+    def is_canceled(self, value):
+        raise NotImplementedError('canceled is set by calling the `timer_cancel` function')
```
dirk-thomas: Could we just leave out this method? Why define it with "not implemented"?

rclpy/rclpy/timer.py
```diff
@@ -22,6 +22,7 @@ def __init__(self, timer_handle, timer_pointer, callback, timer_period_ns):
         [self.timer_handle, self.timer_pointer] = _rclpy.rclpy_create_timer(timer_period_ns)
         self.timer_period_ns = timer_period_ns
         self.callback = callback
+        self._is_canceled = False
```
dirk-thomas: What is this member being used for? All locations seem to always call the C API anyway.

rclpy/rclpy/timer.py
wjwwood: My guess is for the custom (and more informative) error message.

dirk-thomas: Based on the name `is_canceled` I would never expect it to be a setter...

mikaelarguedas: I went back and forth because I wanted hide some if the logic inside Python but it doesn't seem to add much value so I went back to just bind directly all functions and provide a Python method for each of the C functions: see https://github.com/ros2/rclpy/pull/54/commits/25c1aabb2c58a82e4a785b6a4d09eb07629cc201

mikaelarguedas:
Commented out the higher frequency tests and kept only the tests running at 10Hz (same frequency used for [rclcpp tests](https://github.com/ros2/system_tests/blob/cfc04f728de1335a89ce2fc9330e5ad85632af7a/test_rclcpp/test/test_timer.cpp#L47))

Passing CI:
http://ci.ros2.org/job/ci_linux/1934/
http://ci.ros2.org/job/ci_osx/1492/
http://ci.ros2.org/job/ci_windows/1883/

rclpy/test/test_timer.py
```diff
@@ -133,7 +133,8 @@ def func_launch(function, args, message):
 def test_timer_zero_callbacks10hertz():
     func_launch(func_zero_callback, ['0.1'], 'received callbacks when not expected')
 
-
+# TODO(mikaelarguedas) reenable these once timer have consistent behaviour
+# on every platform at high frequency
```
dirk-thomas: You could keep the tests commented in and raise a `SkipTest` exception in their first line for now. That would make them visible in the test results. But no need to do that...

mikaelarguedas: addressed in https://github.com/ros2/rclpy/pull/54/commits/426f23fc986520a0541177e7114bc07d9431ce43

mikaelarguedas:
everything went fine with high speed tests skipped on windows.
http://ci.ros2.org/job/ci_windows/lastSuccessfulBuild/testReport/
http://ci.ros2.org/job/ci_osx/lastSuccessfulBuild/testReport/
http://ci.ros2.org/job/ci_linux/lastSuccessfulBuild/testReport/
Merging this

PR #55, opened 2016-11-21
Error handling

mikaelarguedas:
raise errors on argument mismatch rather than asserts. Also gives more descriptive error messages when asserts fail


mikaelarguedas:
thanks for the fixups

PR #56, opened 2016-12-07
reset response every call

mikaelarguedas:


mikaelarguedas:
Sorry lacking context again.
When doing [busy wait on client response](https://github.com/ros2/examples/commit/b33d3007fd1e6c6622555d7ee7bbcde0b8f00500#diff-303851010c0f5bc1343f9b48ac04b860R38) I was checking the content of the response field. Hence had to reset this before every request sent. So I decided to push that down one layer to make user code simpler

PR #57, opened 2016-12-07
restor spin timeout to 1sec

mikaelarguedas:
Right now the absence of guard_condition support prevent python script to be interrupted when waiting in c (rcl_wait)
This sets the default value to 1sec.
see https://github.com/ros2/examples/pull/141#issuecomment-263388756

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -82,7 +82,7 @@ def spin_once(node, timeout_sec=None):
             )
 
     if timeout_sec is None:
-        timeout = -1
+        timeout = S_TO_NS
     else:
         timeout = int(float(timeout_sec) * S_TO_NS)
 
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -82,7 +82,7 @@ def spin_once(node, timeout_sec=None):
             )
 
     if timeout_sec is None:
-        timeout = S_TO_NS
+        timeout = 1 * S_TO_NS
     else:
         timeout = int(float(timeout_sec) * S_TO_NS)
 


mikaelarguedas:
Thanks for pointing out there is some context missing here.

In the original implementation the timeout was hardcoded to 1 sec and not exposed to the users.
Since rclpy now handles waitset we decided to make the default behavior match the design decision that is "spin and friends should wait forever and exit only if something happens".

Putting the timeout to 1 sec by default in this PR is a temporary fix that should be reverted as soon as we have guard conditions / the ability to exit `rcl_wait` on keyboard interrupts.

The motivation for changing it here rather that in every example is that we expect user to use timers if they want to exit spin periodically rather than using a timeout. Because timeout is an upper bound and not thus doesn't ensure periodicity and also because we would prefer users letting control of the main thread to the ROS 2 machinery.
So setting it in every example may send the wrong message about the good practices in writing rclpy nodes.

dhood:
Thanks, I didn't realise that this was a temporary thing. 

If the only reason we would add timeouts to the examples is 1) where we should actually be using a timer and 2) to get around guard conditions being pending, then I think adding a nonzero default timeout is a good workaround

rclpy/rclpy/__init__.py
```diff
@@ -82,7 +82,7 @@ def spin_once(node, timeout_sec=None):
             )
 
     if timeout_sec is None:
-        timeout = -1
+        timeout = S_TO_NS
```
wjwwood: I know it's the same thing, but I think it is more readable to do `1 * S_TO_NS`.

mikaelarguedas: sounds good: https://github.com/ros2/rclpy/pull/57/commits/944f69633ee22773d4ce80f23d398c79b78ef064

mikaelarguedas:
I'm putting this back to ready hoping that I'll have time to get guard_conditions in before the release, will merge this if getting them in is not possible

mikaelarguedas:
closed in favor of #59

PR #59, opened 2016-12-08
Add sigint guard condition

mikaelarguedas:
This add support for SIGINT signals to be processed from C code and raise the interrupt in Python.
This solves the current issue of Python scripts not killable if they're spinning


dhood:
I think this broke the windows jobs:
http://ci.ros2.org/job/ci_windows/1989/consoleFull
search for `error C2065: 'SIGINT'`

mikaelarguedas:
oh how weird I thought I took that approach from MSDN website. Reverting this for investigation and allow other CI to be ran. Sorry about that

PR #63, opened 2016-12-12
Prevent null handle

mikaelarguedas:
This PR:
* prevents user from setting/modifying the handle of a python node
 * adds a bunch of tests for destruction of various node entities

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -156,30 +156,29 @@ class Node:
 
     def destroy_node(self):
         ret = True
-        if self.handle is not None:
-            for sub in self.subscriptions:
-                ret = ret and _rclpy.rclpy_destroy_node_entity(
-                    'subscription', sub.subscription_handle, self.handle)
-                self.subscriptions.remove(sub)
-            for pub in self.publishers:
-                ret = ret and _rclpy.rclpy_destroy_node_entity(
-                    'publisher', pub.publisher_handle, self.handle)
-                self.publishers.remove(pub)
-            for cli in self.clients:
-                ret = ret and _rclpy.rclpy_destroy_node_entity(
-                    'client', cli.client_handle, self.handle)
-                self.clients.remove(cli)
-            for srv in self.services:
-                ret = ret and _rclpy.rclpy_destroy_node_entity(
-                    'service', srv.service_handle, self.handle)
-                self.services.remove(srv)
-            for tmr in self.timers:
-                ret = ret and _rclpy.rclpy_destroy_entity('timer', tmr.timer_handle)
-                self.timers.remove(tmr)
-            ret = ret and _rclpy.rclpy_destroy_entity('node', self.handle)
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/node.py
```diff
@@ -147,23 +155,35 @@ def destroy_timer(self, timer):
...
-        for srv in self.services:
-            _rclpy.rclpy_destroy_node_entity('service', srv.service_handle, self.handle)
-        for tmr in self.timers:
-            _rclpy.rclpy_destroy_entity('timer', tmr.timer_handle)
-        _rclpy.rclpy_destroy_entity('node', self.handle)
-        self.handle = None
+        ret = True
+        if self.handle is not None:
+            for sub in self.subscriptions:
+                ret = ret and _rclpy.rclpy_destroy_node_entity(
```
dirk-thomas: This can be written shorter as:

```
ret &= ...
```

rclpy/rclpy/node.py
```diff
@@ -147,23 +155,35 @@ def destroy_timer(self, timer):
...
-        for cli in self.clients:
-            _rclpy.rclpy_destroy_node_entity('client', cli.client_handle, self.handle)
-        for srv in self.services:
-            _rclpy.rclpy_destroy_node_entity('service', srv.service_handle, self.handle)
-        for tmr in self.timers:
-            _rclpy.rclpy_destroy_entity('timer', tmr.timer_handle)
-        _rclpy.rclpy_destroy_entity('node', self.handle)
-        self.handle = None
+        ret = True
+        if self.handle is not None:
```
dirk-thomas: To avoid indenting the whole block I would suggest:

```
if self.handle is None:
    return True
```

rclpy/test/test_destruction.py
```diff
@@ -0,0 +1,184 @@
...
+        raise
+
+
+def func_destroy_node():
+    import rclpy
+    try:
+        rclpy.init()
+        node = rclpy.create_node('test_node1')
+        node.destroy_node()
+        ret = True
```
dirk-thomas: The lines which are not supposed to raise `RuntimeError` should not be in the `try` block but in `else`.

mikaelarguedas:
CI:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/2060)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1593/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/2076/)

rclpy/rclpy/node.py
mikaelarguedas: thanks! I didn't know that &= worked also in Python

rclpy/test/test_destruction.py
mikaelarguedas: done https://github.com/ros2/rclpy/pull/63/commits/ae732af804c989081f4a8555cddbb782ea9816dc

PR #66, opened 2017-01-05
use rmw implementation

dirk-thomas:
Connect to ros2/rcl#106

The patch explicitly selects `rmw_implementation`. As a follow up a lot of the logic to handle rmw implementations as well as the `RCLPY_IMPLEMENTATION` environment variable can be removed from `rclpy`.


rclpy/src/rclpy/_rclpy.c
```diff
@@ -834,7 +830,7 @@ rclpy_destroy_node_entity(PyObject * Py_UNUSED(self), PyObject * args)
   }
 
   rcl_node_t * node = (rcl_node_t *)PyCapsule_GetPointer(pynode, NULL);
-  rcl_ret_t ret;
+  rcl_ret_t ret = RCL_RET_ERROR;
```
wjwwood: This initial value is never used, why did you add it throughout the file? My static linter complains about this ("value assigned but never used").

rclpy/rclpy/impl/rmw_implementation_tools.py
```diff
@@ -85,23 +83,8 @@ def import_rmw_implementation():
...
-        elif available_implementations:
-            select_rmw_implementation(available_implementations[0])  # select the first one
-        else:
-            raise NoImplementationAvailableException()
-        logger.debug("Implicitly selecting the '{0}' rmw implementation."
-                     .format(__selected_rmw_implementation))
-    module_name = '._rclpy__{rmw_implementation}'.format(
-        rmw_implementation=__selected_rmw_implementation,
-    )
+    __selected_rmw_implementation = 'rmw_implementation'
```
wjwwood: I guess this is temporary? (pr is still "in progress") You will need to do something with the `select_rmw_implementation()` function. I don't know if you can preserve the functionality here that has to do with selecting the rmw implementation after launching Python but before running `rclpy.init` without also requiring the meta rmw implementation you just made. It might be better to continue to build a Python module for each rmw implementation (since it doesn't really affect user code anyways).

dirk-thomas: The PRs are "in progress" because they haven't been CI tested yet and that can only happen after all packages have been converted to not require rmw specific libraries of rcl / rclcpp. Other then that the patches are in the state they will be in for review.

We don't need to build separate Python modules anymore. You will be able to select the rmw implementation the same way as for C++ by setting the environment variable `RMW_IMPLEMENTATION`. There is certainly more cleanup possible since all this rmw impl. selection logic in Python can be removed. But I will very likely leave that to a follow up PR since the complexity of the current set of PRs is already pretty high.

rclpy/src/rclpy/_rclpy.c
wjwwood: I think I'm using `cppcheck` (under the hood), but it might be using my include path to expand the macros and see that it is actually a return statement.

A compromise would be to set it in the else case, which would be correct in both checks.

rclpy/rclpy/impl/rmw_implementation_tools.py
wjwwood: > You will be able to select the rmw implementation the same way as for C++ by setting the environment variable `RMW_IMPLEMENTATION`.

That ignores the current use case which allows you to query the available ones, and then decide in Python code which one to use. I think that's actually pretty useful, that's why I suggested you look at whether or not the env variable mechanism can be used to preserve that functionality or that we just continue to build one for each. You didn't address this point at all and simply stated "We don't need to build separate Python modules anymore." which ignores the issue.

> There is certainly more cleanup possible since all this rmw impl. selection logic in Python can be removed. But I will very likely leave that to a follow up PR since the complexity of the current set of PRs is already pretty high.

-1. Sorry, I don't support merging this while creating dead code and leaving tests disabled.

In my opinion this pr needs to either justify why the existing functionality is no longer required and remove the dead code and tests related to the removed feature or fix the existing code so the feature continues to work with the new mechanism.

dirk-thomas: Regarding the "dead code" and disabled tests. The idea as mentioned above is to remove those in a follow up PR. For sanity I wanted to separate these steps since the current PRs are already very invasive. The whole typesupport redesign has already been split over 8 independent changeset (which have all been reviewed and merged separately). I though that splitting changes up into more bite size pieces is desired (for the sanity of the programmer making the patches as well as the sanity of the reviewer). I will update this PR then to remove the "dead code" and disabled test as part of this PR since you prefer that.

dirk-thomas: Imo we don't need two separate ways to select rmw implementation. The `RMW_IMPLEMENTATION` environment variable works across all languages. In C++ you don't have the option to choose the rmw implementation from within the running executable. Therefore I don't think this is a feature we need in Python - especially since it is not possible currently to use rmw A and then switch to rmw B within the same Python script.

I don't know if a running Python script can look at the environment variable, change the value and that change being reflected in the loaded library since I haven't tried it. But imo it is a not a use case worth supporting.

We should then make a decision on the question about selecting the rmw impl. within a running Python script in order to continue on this.

rclpy/src/rclpy/_rclpy.c
dirk-thomas: I have removed the commit from this PR and will handle my local linter somehow. So nothing needs to change here.

rclpy/rclpy/impl/rmw_implementation_tools.py
wjwwood: > I though[t] that splitting changes up into more bite size pieces is desired (for the sanity of the programmer making the patches as well as the sanity of the reviewer). 

I'm fine with breaking things up, but I would not split the work up as you have here. You could have left the one Python module per rmw impl design in place and only changed this repository to work with the new system, e.g. like change the default to `rmw_implementation` rather than "`rmw_fastrtps_cpp` or the first" as it is now. You could also have done what you've done here plus remove old code and tests, but what's here is, from my perspective, in between and not a good stopping place. That's all I'm saying.

> I will update this PR then to remove the "dead code" and disabled test as part of this PR since you prefer that.

I do prefer that, but others might disagree with me and approve it as-is.

> The `RMW_IMPLEMENTATION` environment variable works across all languages.

I'm not disputing that.

> In C++ you don't have the option to choose the rmw implementation from within the running executable.

Because it is not simple to do so technically, but that's not a good reason to not do it in Python, imo. There's lots of things you can easily do in Python that aren't practical in C++.

> Therefore I don't think this is a feature we need in Python - especially since it is not possible currently to use rmw A and then switch to rmw B within the same Python script.

I'm -0 on removing this feature. I think it would be potentially useful for tests (by using multiprocessing as some of our tests already do), despite the fact that you cannot use different ones from the same script at different times (without multiprocessing). But I understand not wanting to spend more time on fixing it for the new rmw mechanism.

> We should then make a decision on the question about selecting the rmw impl. within a running Python script in order to continue on this.

I'm slightly in favor of keeping it, mostly because I think it would be easy enough to fix no matter which route we take (leave the one Python module per rmw impl or try setting the env variable before importing the module with the rmw impl). I agree others should chime in on whether or not they want to preserve the feature.

If you wanted to work around this issue and merge these pr's sooner, simply change the default rmw implementation Python uses to be `rmw_implemenation`. I don't think anything else has to change. That's the least intrusive route in my opinion. Whether or not we do away with `RCLPY_IMPLEMENTATION` and avoid building multiple Python modules, one for each rmw impl, is a separate decision from my perspective.

dirk-thomas: Imo we have been building separate modules only for the reason that we were forced to do so. I don't think it was for having it as a feature. Also removing building N variations of the Python module is a benefit inmo.

I have removed the obsolete parts in 9e0597807ad9b64397ba35eaa549f569978a039e.

Anyway since you vote with `-0` I guess we have to wait for more votes on this then before moving forward. @ros2/team Please cast your vote here.

rclpy/src/rclpy/_rclpy.c
dirk-thomas: It looks like the current linter doesn't like the even unchanged code anymore. I initialized the value in the `else` block in c15271beb25344bab31edca1ad97fcb9c3118217

rclpy/rclpy/impl/rmw_implementation_tools.py
dirk-thomas: Btw we can't "just keep the rmw specific Python modules" since they were depending on the availability of rmw specific `rclcpp` libraries. (And I don't think we should keep those just for this case.)

wjwwood: You're right, though I suppose you mean `rcl` not `rclcpp`. You'll have to either remove the code and tests for the Python selection of rmw impl (as is proposed in this pr) or fix those functions/features to use the new mechanism, assuming that is possible.

I hadn't thought about it before, but that also means that you cannot build rmw implementation specific user executables at all anymore. Before I just assumed we were removing that logic because we could, not because we had too.

Doesn't it also mean that all of the `get_X_information` macros (where X is `rcl`, `rclcpp`, etc.) are no longer useful and should be removed? If that is the case, then that is another example of dead code that should be removed.

dirk-thomas: > You're right, though I suppose you mean `rcl` not `rclcpp`. You'll have to either remove the code and tests for the Python selection of rmw impl (as is proposed in this pr) or fix those functions/features to use the new mechanism, assuming that is possible.

As mentioned above I have already removed these in this PR in 9e0597807ad9b64397ba35eaa549f569978a039e.

> Doesn't it also mean that all of the `get_X_information` macros (where X is `rcl`, `rclcpp`, etc.) are no longer useful and should be removed? If that is the case, then that is another example of dead code that should be removed.

Yes, various CMake function can be removed. All if this is being tracked in the parent ticket: ros2/ros2#302.

Again I don't see the point of squeezing as much changes as possible into this PR set. Especially since the previous rounds of PRs have moved very fast. It takes a lot of time to get these changes into the state to be "green". Imo this discussion is slowing the progress more down then if this would have been merged. Then the remaining tasks could have already been started / done by now.

I can remove the CMake functions in the corresponding PR if that makes a difference for you.

dirk-thomas: See https://github.com/ros2/rcl/pull/106/commits/6d1183e547851b737bd138e479f8e8a5be17a51a and https://github.com/ros2/rclcpp/pull/302/commits/78dfef75b4687e06885780c878895e6c54b0baed

wjwwood: > As mentioned above I have already removed these in this PR in 9e05978.

I know I said that 😛:

> > remove the code and tests for the Python selection of rmw impl (as is proposed in this pr)

My point was that if someone else reviews and agrees with me then you'll need to do the other thing.

> Again I don't see the point of squeezing as much changes as possible into this PR set. 

It's not about squeezing as much as possible in, it is about making sure the master branch stays in the best state possible even between incremental pr's. When someone asks me to review a pr, I look for anything that smells bad like: failing test, failing style, logical inconsistency, dead code, broken features. I think it is acceptable to table any of those things by marking them with a TODO or tracking them in a ticket, but only if there is no plan to fix them in the immediate future. If you're just going to merge this pr and then do those things right afterwards, then I don't see much benefit to splitting them up. The only other reason to merge short of the goal like this is to prevent getting out of sync with others, but since you're the only one working in this part of the code right now, that's not an issue in my opinion. So from my perspective you're neither at risk of conflicts nor are you going to defer doing these things for very long and so I think you should do them in these current pr's.

I can't see a good reason to do it in two steps. In contrast, you may discover something important when removing these functions (like that they're still being used somewhere that you missed) which would be important to find and fix before putting it to master.

I appreciate that you're trying to keep the changes small and reviewable, but I just think they should also be cohesive and that splitting the pr's short of removing now vestigial functions just doesn't make sense.

> Imo this discussion is slowing the progress more down then if this would have been merged. Then the remaining tasks could have already been started / done by now.

Imo if you just removed them and restarted CI when I mentioned it, then we'd also be done by now. The size of the pr doesn't affect the time it takes to run CI or anything. In fact the time it would take you to make new branches and pr's for the additional changes might make it take longer.

> I can remove the CMake functions in the corresponding PR if that makes a difference for you.

I think it is worthwhile to do that at the same time as these others changes, yes. I don't know what the rest of the @ros2/team is doing today (hint, hint -- nudge, nudge 😄), but they could override me and that would be fine.

dirk-thomas: Please see my previous comment. That already referenced the requested changes. And the CI build links on the main ticket have been update with it.

dirk-thomas:
I added a bullet to the typesupport redesign ticket.

Does the approval only cover this PR or all other related PRs referenced from the parent ticket (ros2/rcl#106)?

wjwwood:
Just this one, I'll review the others now.

esteve:
I see @wjwwood already approved this PR, so I don't know if this is still open for discussion, but I'd rather not see this functionality go away (cleaned up yes, but not removed). There are certain cases where it's more convenient to pick the RMW implementation from Python, instead of relying on an environment variable.

For example, if you're developing a Django application that uses `rclpy` (yeah, a bit crazy, but it may be the case that you want to build a web frontend for ROS2), it's far easier to deploy if the RMW implementation is specified in the Django settings (`DJANGO_SETTINGS_MODULE`) and later loaded in Python, it's also more explicit, which is always better than implicit. Of course most people will just use whatever default RMW implementation is available, but in any case, I don't see any strong reason for removing this.

dirk-thomas:
We can't keep the current functionality as is - simple because it would require to continue to build separate libraries of `rcl` / `rclcpp` for each rmw implementation (which we want to get rid of). Currently `rmw_implementation` chooses only based on an environment variable which rmw implemenation to load. I mentioned in https://github.com/ros2/rcl/pull/106#issuecomment-271106788 that a select function could be added to override that selection process. Maybe that would be useful for the described use case too.

esteve:
> @dirk-thomas We can't keep the current functionality as is

Exactly, that's what I meant with:

> I'd rather not see this functionality go away (cleaned up yes, but not removed)

so as long as there's something else functionally equivalent (i.e. being able to select the RMW implementation explicitly without relying on an environment variable), that's fine, doesn't matter how it's implemented.

PR #71, opened 2017-03-27
wrap rcl_get_names in rclpy

mikaelarguedas:


rclpy/rclpy/names_and_types.py
```diff
@@ -29,9 +29,8 @@ def __init__(self, **kwargs):
         self.topic_names = kwargs.get('topic_names', list())
         self.type_names = kwargs.get('type_names', list())
 
-    def __repr__(self):
-        s = 'number of topics = ' + repr(self._topic_count) + '\n'
-        s += '[topic_name] : [ROS type name]\n'
+    def __str__(self):
```
dirk-thomas: While not being changed in this PR I wouldn't expect the string to have a trailing newline. If the user prints this string it would result in an empty trailing line (since `print` adds its own newline at the end).

mikaelarguedas: sounds good I'll fix that. any comment on the format itself (displaying `TOPIC_NAME : TOPIC_TYPE`)?

dirk-thomas: In ROS 1 the command line tools often print it as `TOPIC_NAME [TOPIC_TYPE]`.

mikaelarguedas: gotcha, I'll do the same here then

rclpy/rclpy/names_and_types.py
```diff
@@ -29,12 +29,14 @@ def __init__(self, **kwargs):
...
+        s = ''
         for i in range(self._topic_count):
-            s += '{} : {}\n'.format(
+            s += '{} [{}]'.format(
                 self._topic_names[i], self._type_names[i])
+            if i == self._topic_count - 1:
+                break
+            s += '\n'
```
dirk-thomas: Instead of adding the newline based on the index a common approach is this:

```
lines = []
for i in range(self._topic_count):
    lines.append('{} [{}]'.format(self._topic_names[i], self._type_names[i]))
return '\n'.join(lines)
```

dirk-thomas:
Looking at the data structure `_topic_count` seems redundant to me. It is already available as `len(self._topic_names)` and `len(self._type_names)`. (Again unrelated to this PR - just noticed it when reading the code.)

mikaelarguedas:
Yeah I agree, I basically wrapped the c-structure that has that field but there is no need to reflect it in the python object.

rclpy/rclpy/names_and_types.py
mikaelarguedas: so we are trading a trailing new line for a leading new line ? Isn't the goal "no new line" ?

dirk-thomas: `'\n'.join(lines)` only adds newlines in between lines.

mikaelarguedas: That's what I though too initially, but had a different result when testing. I must have messed up something somewhere

mikaelarguedas:
merging this along https://github.com/ros2/cli_tools/pull/9
If there is any follow-up comment please post them here and I'll address them

PR #72, opened 2017-03-29
updates to support node namespace and new rcl API

wjwwood:
Connects to ros2/rmw#95

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -36,7 +36,7 @@ def init(args=None):
 
 
 def create_node(node_name, namespace=None):
-    node_handle = _rclpy.rclpy_create_node(node_name, namespace or "")
+    node_handle = _rclpy.rclpy_create_node(node_name, namespace or '')
     return Node(node_handle)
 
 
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/__init__.py
```diff
@@ -35,8 +35,8 @@ def init(args=None):
     return _rclpy.rclpy_init(args if args is not None else sys.argv)
 
 
-def create_node(node_name):
-    node_handle = _rclpy.rclpy_create_node(node_name)
+def create_node(node_name, namespace=None):
```
dirk-thomas: Please make the keyword argument explicit:

```
def create_node(node_name, *, namespace=None):
```

wjwwood: 913698ee392c73114d3a3409f94f162962b314ec

Note that the other functions here are not "keyword-only arguments".

dirk-thomas: We should either update them in this PR or in a separate one.

wjwwood: Since this change is not strictly required, I'd rather not do it in this pr. You're welcome to do a follow up.

dirk-thomas: A namespace can be empty. A topic name can't be. Since both types have separate validation criteria I would suggest to create separate validation functions for them. These implementation could share common functionality though if that makes sense.

dirk-thomas: Makes sense: #73

wjwwood:
> I think a separate validation function makes sense.

In rclpy?

PR #75, opened 2017-04-11
use string_array_t from c_utilities package

Karsten1987:
connects to ros2/rmw#100

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -14,6 +14,7 @@
 
 #include <Python.h>
 
+#include <c_utilities/types.h>
 #include <rcl/error_handling.h>
 #include <rcl/graph.h>
 #include <rcl/node.h>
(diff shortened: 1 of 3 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -1418,7 +1419,7 @@ rclpy_get_node_names(PyObject * Py_UNUSED(self), PyObject * args)
       pynode_names, idx, PyUnicode_FromString(node_names.data[idx]));
   }
 
-  ret = rcl_destroy_node_names(&node_names);
+  ret = utilities_string_array_fini(&node_names);
```
wjwwood: Comparing to the wrong return code macro again, should be `UTILITIES_RET_OK`.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1418,7 +1419,7 @@ rclpy_get_node_names(PyObject * Py_UNUSED(self), PyObject * args)
...
   if (ret != RCL_RET_OK) {
     PyErr_Format(PyExc_RuntimeError,
       "Failed to destroy node_names: %s", rcl_get_error_string_safe());
```
wjwwood: Same comment as for `rclcpp`, is the rcl error message actually set here?

PR #77, opened 2017-04-19
skip D401 on rclpy_excepthook

wjwwood:
I couldn't find a way to make the docstring imperative and still make sense. So I added a noqa as suggested by pydocstyle when this one gives a false positive. I'm 100% happy with the solution, but I'm going to merge it as soon as I get a green CI back since it is broken on master atm due to the new pydocstyle version.

If someone has a better solution please open up a new pr for it.


wjwwood:
[Build Status](http://ci.ros2.org/job/ci_linux/2484/)

PR #79, opened 2017-04-20
fixes memory leak in rclpy_publish

yamokosk:
I believe this PR fixes the memory leak referred to in #74. Here are my results with valgrind, pre- and post-fix:

```
Without fix (Float64MultiArray test):
==23766== LEAK SUMMARY:
==23766==    definitely lost: 584 bytes in 17 blocks
==23766==    indirectly lost: 2,400,702 bytes in 22 blocks
==23766==      possibly lost: 8,888 bytes in 14 blocks
==23766==    still reachable: 3,967,098 bytes in 3,247 blocks
==23766==         suppressed: 0 bytes in 0 blocks

With fix (Float64MultiArray test):
==18138== LEAK SUMMARY:
==18138==    definitely lost: 323 bytes in 13 blocks
==18138==    indirectly lost: 320 bytes in 7 blocks
==18138==      possibly lost: 8,200 bytes in 13 blocks
==18138==    still reachable: 3,949,000 bytes in 3,240 blocks
==18138==         suppressed: 0 bytes in 0 blocks


Without fix (String):
==29155== LEAK SUMMARY:
==29155==    definitely lost: 500,298 bytes in 9 blocks
==29155==    indirectly lost: 750,323 bytes in 10 blocks
==29155==      possibly lost: 6,680 bytes in 11 blocks
==29155==    still reachable: 1,714,333 bytes in 3,213 blocks
==29155==         suppressed: 0 bytes in 0 blocks

With fix (String):
==19726== LEAK SUMMARY:
==19726==    definitely lost: 500,314 bytes in 7 blocks
==19726==    indirectly lost: 600 bytes in 13 blocks
==19726==      possibly lost: 12,144 bytes in 18 blocks
==19726==    still reachable: 1,731,327 bytes in 3,220 blocks
==19726==         suppressed: 0 bytes in 0 blocks
```
Unfortunately, there is something funny still going on with ```String```. Valgrind seems to think its something with Python3:

```
==22910== 500,098 bytes in 2 blocks are definitely lost in loss record 925 of 926
==22910==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==22910==    by 0x5D3AD6: PyObject_Malloc (in /home/yamokosk/ros2_ws/rclpy/bin/python3)
==22910==    by 0x5440EF: PyUnicode_New (in /home/yamokosk/ros2_ws/rclpy/bin/python3)
[...]

PR #80, opened 2017-04-21
Fixes memory leaks in python subscribers (and probably also python services)

yamokosk:
This is a follow on from #79. It pairs every ```convert_from_py``` call with a ```destroy_ros_message```. Running valgrind on a python subscriber (same test setup used in #79) yielded something like:

```
Before fix
==7972== LEAK SUMMARY:
==7972==    definitely lost: 709,984 bytes in 12,104 blocks
==7972==    indirectly lost: 1,232,986 bytes in 28,242 blocks
==7972==      possibly lost: 217,896 bytes in 348 blocks
==7972==    still reachable: 11,638,082 bytes in 28,306 blocks
==7972==         suppressed: 0 bytes in 0 blocks
==7972== Reachable blocks (those to which a pointer was found) are not shown.
```

After adding ```destroy_ros_message``` to ```rclpy_take``` (and others):

```
After fix
==23836== LEAK SUMMARY:
==23836==    definitely lost: 1,240 bytes in 23 blocks
==23836==    indirectly lost: 2,920 bytes in 71 blocks
==23836==      possibly lost: 11,408 bytes in 17 blocks
==23836==    still reachable: 1,563,802 bytes in 3,279 blocks
==23836==         suppressed: 0 bytes in 0 blocks
==23836== Reachable blocks (those to which a pointer was found) are not shown.
```

I did not explicitly test that a memory leak exists in the service functions. But by code inspection, there should be. The same fix was applied there.

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -713,9 +713,19 @@ rclpy_send_request(PyObject * Py_UNUSED(self), PyObject * args)
   assert(convert_from_py != NULL &&
     "unable to retrieve convert_from_py function, type_support mustn't have been imported");
 
+  PyObject * pydestroy_ros_message = PyObject_GetAttrString(pymetaclass, "_DESTROY_ROS_MESSAGE");
+
+  typedef void * (* destroy_ros_message_signature)(void *);
+  destroy_ros_message_signature destroy_ros_message =
+    (destroy_ros_message_signature)PyCapsule_GetPointer(pydestroy_ros_message, NULL);
+
+  assert(destroy_ros_message != NULL &&
+    "unable to retrieve destroy_ros_message function, type_support mustn't have been imported");
+
   void * raw_ros_request = convert_from_py(pyrequest);
   int64_t sequence_number;
   rcl_ret_t ret = rcl_send_request(client, raw_ros_request, &sequence_number);
+  destroy_ros_message(raw_ros_request);
   if (ret != RCL_RET_OK) {
     PyErr_Format(PyExc_RuntimeError,
       "Failed to send request: %s", rcl_get_error_string_safe());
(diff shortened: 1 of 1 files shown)


wjwwood:
Thanks for the pr, we'll try to get to asap.

yamokosk:
No worries. I was considering pulling this PR back as I think a more complete fix would also address the `PyMem_Malloc` losses that are also occuring. I made a comment on this on another PR (#81).

mikaelarguedas:
@mikaelarguedas to review this

rclpy/src/rclpy/_rclpy.c
```diff
@@ -713,9 +713,19 @@ rclpy_send_request(PyObject * Py_UNUSED(self), PyObject * args)
...
     "unable to retrieve convert_from_py function, type_support mustn't have been imported");
 
+  PyObject * pydestroy_ros_message = PyObject_GetAttrString(pymetaclass, "_DESTROY_ROS_MESSAGE");
+
+  typedef void * (* destroy_ros_message_signature)(void *);
+  destroy_ros_message_signature destroy_ros_message =
+    (destroy_ros_message_signature)PyCapsule_GetPointer(pydestroy_ros_message, NULL);
+
+  assert(destroy_ros_message != NULL &&
+    "unable to retrieve destroy_ros_message function, type_support mustn't have been imported");
```
dirk-thomas: This check won't be present in release builds. Since the pointer is going to be used below this should always be checked.

Same below (as well as for existing code, e.g. `convert_from_py`).

mikaelarguedas: opened a separate issue for this because this is not tied to this PR (c.f. https://github.com/ros2/rclpy/issues/83)

dirk-thomas: For the existing code I agree but I would advocate to not add more code with the same flawed style.

mikaelarguedas:
The code change looks good to me. This has the same problem of leaking nested messages as reported in https://github.com/ros2/rosidl/issues/216 but is a good incremental improvement.

I rebased the branch and ran CI on it:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2585/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/176/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2053/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2711/)

rclpy/src/rclpy/_rclpy.c
mikaelarguedas: We'll have to go through all "equivalent" asserts and replace them anyway so I'm not advocating strongly for fixing this in this PR, but I see your point. If you think this is a required change for this PR to get in, please request a change here with an advice on the type of error that should be thrown.

mikaelarguedas:
I'm going to merge this as is. We'll do a follow-up PR to cleanup the asserts.
Thanks @yamokosk for the contribution!

Issue #74, opened 2017-04-07
Increasing memory usage when only using spin_once


ArkturusMengsk:
If the timeout in spin_once is set to zero (see snippet below) the program will continuously allocate new memory. If any other value or the default infinite timeout is chosen, everything works fine.

```
rclpy.init(args)
node = rclpy.create_node('test_node')
while rclpy.ok():
    rclpy.spin_once(node, 0)
```

IanTheEngineer:
I spent some time investigating this leak today, figuring it would be a good chance to dip my toe into `ros2`. It does seem there is a memory leak here, and as far as I can tell there is something funny happening beneath the [`rmw_wait()` function](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h#L312-L321) whenever you pass in `timeout_sec=0.0` into `rclpy.spin_once()`, as you did in your example code, @ArkturusMengsk. As a workaround, it seems you can make the timeout anything greater than `0.0`, and the memory will at least leak slowly. 

The middleware seems to exit without releasing the [zero-initialized wait_set memory](https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/__init__.py#L44). This issue is exacerbated by your while loop continually spinning without a pause.

**Instrumented Script:**
```
#! /usr/bin/env python3

import rclpy
import time
import tracemalloc

rclpy.init()
node = rclpy.create_node('test_spin_once_mem')
period = 0.0 # 0.1, 0.01, 0.001
count = 0
begin_time = time.time()
tracemalloc.start()
while rclpy.ok() and time.time() - begin_time < 2.0:
  rclpy.spin_once(node, timeout_sec=period)
  count = count + 1
  #time.sleep(0.01)                # Uncomment for debugging loop speed
snapshot = tracemalloc.take_snapshot()
top_stats = snapshot.statistics('lineno')

print("[Loop Count: {} ]".format(count))
for stat in top_stats[:10]:
    print(stat)
rclpy.shutdown()

```
**Output with `#time.sleep(0.01)` commented in the loop:**
```
$ ./mem_test_spin_once.py 
[Loop Count: 205663 ]
/home/imcmahon/dev/install_isolated/rclpy/lib/python3.5/site-packages/rclpy/__init__.py:44: size=17.3 MiB, count=205663, average=88 B
[...]

PR #81, opened 2017-04-26
Bugfix/rclpy mem leak

srsidd:
In reference to #74 .

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import gc
 import sys
 
 from rclpy.constants import S_TO_NS
@@ -109,6 +110,7 @@ def spin_once(node, *, timeout_sec=None):
         if request:
             response = srv.callback(request, srv.srv_type.Response())
             srv.send_response(response, header)
+    gc.collect()
 
 
 def ok():
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -713,9 +713,19 @@ rclpy_send_request(PyObject * Py_UNUSED(self), PyObject * args)
   assert(convert_from_py != NULL &&
     "unable to retrieve convert_from_py function, type_support mustn't have been imported");
 
+  PyObject * pydestroy_ros_message = PyObject_GetAttrString(pymetaclass, "_DESTROY_ROS_MESSAGE");
+
+  typedef void * (* destroy_ros_message_signature)(void *);
+  destroy_ros_message_signature destroy_ros_message =
+    (destroy_ros_message_signature)PyCapsule_GetPointer(pydestroy_ros_message, NULL);
+
+  assert(destroy_ros_message != NULL &&
+    "unable to retrieve destroy_ros_message function, type_support mustn't have been imported");
+
   void * raw_ros_request = convert_from_py(pyrequest);
   int64_t sequence_number;
   rcl_ret_t ret = rcl_send_request(client, raw_ros_request, &sequence_number);
+  destroy_ros_message(raw_ros_request);
   if (ret != RCL_RET_OK) {
     PyErr_Format(PyExc_RuntimeError,
       "Failed to send request: %s", rcl_get_error_string_safe());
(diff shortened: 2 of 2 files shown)


rclpy/rclpy/__init__.py
```diff
@@ -109,6 +110,7 @@ def spin_once(node, *, timeout_sec=None):
         if request:
             response = srv.callback(request, srv.srv_type.Response())
             srv.send_response(response, header)
+    gc.collect()
```
dirk-thomas: I don't think that we should explicitly call `gc.collect` - especially not in a function like `spinOnce` which is called in a potentially very high frequency.

wjwwood: Yeah, sorry if I miscommunicated, but we don't want to do this.

From the other thread I was just saying that we should definitely have the destructor call the fini function in C, but we should let the gc run when it feels like it. Even though that might throw valgrind off.

The rest of the patch can stay though I think (still haven't reviewed it yet).

srsidd: I tried doing that, but it did not seem to have any effect. Is there another way of achieving the same effect?

dirk-thomas: Maybe the tests can either trigger the GC before finishing or monkey patch rclpy to perform this on shutdown?

yamokosk:
Here's my suggestion for fixing this:

```cpp
/// waitset destructor
void
rclpy_destroy_wait_set(PyObject* pywait_set)
{
  rcl_wait_set_t * wait_set = (rcl_wait_set_t *)PyCapsule_GetPointer(pywait_set, NULL);
  rcl_ret_t ret = rcl_wait_set_fini(wait_set);
  if (ret != RCL_RET_OK) {
      "Failed to destroy wait set: %s", rcl_get_error_string_safe());
  }
  PyMem_Free(wait_set);
  return;
}

/// Return a Capsule pointing to a zero initialized rcl_wait_set_t structure
static PyObject *
rclpy_get_zero_initialized_wait_set(PyObject * Py_UNUSED(self), PyObject * Py_UNUSED(args))
{
  rcl_wait_set_t * wait_set = (rcl_wait_set_t *)PyMem_Malloc(sizeof(rcl_wait_set_t));
  *wait_set = rcl_get_zero_initialized_wait_set();

  PyObject * pywait_set = PyCapsule_New(wait_set, NULL, rclpy_destroy_wait_set);

  return pywait_set;
}
```

In testing on my machine, this eliminated the leak that grew linearly with the number of calls to spin_once (and confirmed with valgrind that it was coming from wait_set). The problem is that this is not a universal fix for all the other `PyMem_Malloc` calls in `_rclpy.c`. It addresses a lot of them but for things like `publisher` that take two args for its `_fini` function, this pattern won't work.

```cpp
rcl_ret_t
rcl_publisher_fini(rcl_publisher_t * publisher, rcl_node_t * node)
```

srsidd:
Closing pull request since this is not the right way of doing this. And there is an issue tracking this at #74 .

BrannonKing:
Reincorporating the above rclpy_destroy_wait_set on a current build does not eliminate the memory leak.

mikaelarguedas:
thanks @BrannonKing for following up on this

Yeah I think the main leak is elsewhere, especially given than the bigger the messages the bigger the leak (would go along this https://github.com/ros2/rclpy/issues/74#issuecomment-320333788 about PyMem_Malloc)
I'm planning on chasing it down once the refactor of rclpy to [use executors](https://github.com/ros2/rclpy/pull/104) has been merged.

For reference: can you give more detail about the code you're running and if you also observe leaks proportional to the amount of data transmitted? 

Feel free to comment on the open ticket #74 or to open another one if the leak doesn't happen in spin_once, otherwise we may lose track of comments on an already closed ticket.
Also any contribution reducing (even partially) the current leaks is more than welcome!

BrannonKing:
I'm aware of #74 -- it is the exact bug I was looking at. I'm not sending any messages (in my test). spin_once with a small/zero timeout leaks like the Titanic. It's super easy to reproduce: make a node and call spin_once repeatedly.

dirk-thomas:
@BrannonKing Can you please try the patch from #109 which should fix the memory leak described in #74 and report back.

PR #82, opened 2017-05-10
Remove agnostic lib

dirk-thomas:
Since the `rclpy` lib is now agnostic to the RMW impl. there is no need to keep some functions in a separate library. And since the loaded Python extension is always the same I removed all the fancy loading logic around it.

The second commit addresses a test failure. Instead of passing a known invalid parameter to the C function raise a `ValueError` instead.

[Build Status](http://ci.ros2.org/job/ci_windows/2705/) fixes the failing `rclpytestst` (http://ci.ros2.org/view/nightly/job/nightly_win_deb/lastCompletedBuild/testReport/(root)/projectroot/rclpytests/).


rclpy/CMakeLists.txt
```diff
@@ -57,6 +57,7 @@ if(WIN32)
 endif()
 
 target_link_libraries(${PROJECT_NAME}
+  ${rcl_LIBRARIES}
```
wjwwood: Isn't this redundant with the `ament_target_dependencies` call below?

rclpy/rclpy/node.py
```diff
@@ -160,6 +160,12 @@ def destroy_node(self):
         ret = True
         if self.handle is None:
             return ret
+
+        # ensure that the passed node contains a valid capsule
+        class_ = self.handle.__class__
+        if not class_ or class_.__name__ != 'PyCapsule':
+            raise ValueError()
```
wjwwood: Shouldn't this have a message?

rclpy/test/test_destruction.py
```diff
@@ -20,7 +20,7 @@
 def run_catch_report_raise(func, *args, **kwargs):
     try:
         return func(*args, **kwargs)
-    except:
+    except Exception:
```
wjwwood: Is this necessary?

rclpy/CMakeLists.txt
dirk-thomas: Absolutely, I moved it from the other library target and noticed the duplicate. Removed in 8076c4e56c4c142be1438c1a8f0162e9957bed4a.

rclpy/rclpy/node.py
dirk-thomas: It was only used in one bizarre test (which modified the internal handle of a node instance) so I thought it isn't a real world case. Added a message in f2b9bf5ee1b5874fc6540754dca71f41934af68d.

rclpy/test/test_destruction.py
dirk-thomas: Not necessary but recommended in [PEP 8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) and some linters will warn about it:

> When catching exceptions, mention specific exceptions whenever possible instead of using a bare `except:`

wjwwood:
Unfortunately I didn't catch this before, but the removal of the proxy object machinery has caused a regression in the check that `rclpy.init()` has been called before using other functions. Now you get a general `RuntimeError` rather than the more specific `rclpy.exceptions.NotInitializedException` when doing something like this:

```
In [1]: import rclpy

In [2]: node = rclpy.create_node('foo')
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-25-e3302d2d805e> in <module>()
----> 1 node = rclpy.create_node('foo')

/Users/william/ros2_ws/install/lib/python3.6/site-packages/rclpy/__init__.py in create_node(node_name, namespace)
     26
     27 def create_node(node_name, *, namespace=None):
---> 28     node_handle = _rclpy.rclpy_create_node(node_name, namespace or '')
     29     return Node(node_handle)
     30

RuntimeError: Failed to create node: rcl_init() has not been called, at /Users/william/ros2_ws/src/ros2/rcl/rcl/src/rcl/node.c:92
```

Please restore the proxy object stuff or check that `rclpy.init()` has been called in each of the cases and raise the right exception.

This is the code that checked on each access of `_rclpy` and raise if `init()` had not been called:

https://github.com/ros2/rclpy/pull/82/files#diff-7486716035fb1f918746b7f5b1b31b35L49

wjwwood:
I created https://github.com/ros2/rclpy/issues/84 to track the issue since this cannot be reopened.

wjwwood:
Also, after this pr, `NoImplementationAvailableException` is no longer used anywhere. I'll remove it in my upcoming pr.

PR #87, opened 2017-06-08
Fix destroy segfault

mikaelarguedas:
this reverts #80 and #79 that (by not clearing the allocated memory properly) makes the subscriber crash on message destruction for arrays of nested arrays. Leaving this in progress the time I add a regression test for it


mikaelarguedas:
Fixed in https://github.com/ros2/rosidl/pull/223 https://github.com/ros2/rmw_fastrtps/pull/126 and https://github.com/ros2/system_tests/pull/211

PR #93, opened 2017-06-15
Refactor get topic names and types

wjwwood:
Connects to ros2/ros2#361

### DIFF
--- a/rclpy/rclpy/topic_or_service_is_hidden.py
+++ b/rclpy/rclpy/topic_or_service_is_hidden.py
@@ -1,0 +1,31 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


rclpy/rclpy/topic_or_service_is_hidden.py
```diff
@@ -0,0 +1,31 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+HIDDEN_TOPIC_PREFIX = '_'
+
+
+def topic_or_service_is_hidden(name):
```
dirk-thomas: We should probably update existing code in `ros2cli` to use this new function once merged.

rclpy/rclpy/utilities.py
```diff
@@ -0,0 +1,40 @@
...
+def shutdown():
+    with g_shutdown_lock:
+        return _rclpy.rclpy_shutdown()
+
+
+def try_shutdown():
+    """Shutdown rclpy if not already shutdown."""
+    with g_shutdown_lock:
+        if _rclpy.rclpy_ok():
+            return _rclpy.rclpy_shutdown()
```
dirk-thomas: Should the function return anything in the `else` case (instead of the implicit `None`)?

wjwwood: I don't think so, the shutdown function in the C extension currently returns `None` itself. I could remove the `return` from both shutdown functions here.

dirk-thomas: Than it should be fine as-is.

PR #94, opened 2017-06-27
fix destroy logic and error handling

dirk-thomas:
The patch fixes three problems:

* The destroy logic was broken. Removing elements from a list while iterating of it results in the wrong behavior (only every second item was destroyed before).
* When an exception is being set the function must return `NULL` in order to pass it correctly from the C extension to Python.
* ~~Before using the node handle for the destroy calls in needs to be checked if it is valid.~~ **Update:** covered by the fixed error handling code already.

This can be reproduced by running these two commands with `rmw_connext_cpp`:

* `ros2 run examples_rclpy_minimal_service examples_rclpy_minimal_service`
* `ros2 run examples_rclpy_minimal_client examples_rclpy_minimal_client_async`

With the patch the problem of the node being invalid when the client is trying to be destroyed is at least reported correctly.

Updated CI links with only two commits being in this PR:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2829/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/333/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2285/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2965/)

Ready for review.

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -1362,13 +1362,13 @@ rclpy_destroy_node_entity(PyObject * Py_UNUSED(self), PyObject * args)
     ret = RCL_RET_ERROR;  // to avoid a linter warning
     PyErr_Format(PyExc_RuntimeError,
       "%s is not a known entity", entity_type);
-    Py_RETURN_FALSE;
+    return NULL;
   }
   if (ret != RCL_RET_OK) {
     PyErr_Format(PyExc_RuntimeError,
       "Failed to fini '%s': %s", entity_type, rcl_get_error_string_safe());
     rcl_reset_error();
-    Py_RETURN_FALSE;
+    return NULL;
   }
   Py_RETURN_TRUE;
 }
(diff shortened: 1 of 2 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -1345,6 +1345,20 @@ rclpy_destroy_node_entity(PyObject * Py_UNUSED(self), PyObject * args)
...
+      "rcl node is null");
+    return NULL;
+  }
+
+  if (!rcl_node_is_valid(node)) {
+    PyErr_SetString(PyExc_ValueError,
+      "rcl node is invalid");
+    rcl_reset_error();
+    return NULL;
+  }
```
wjwwood: I think both of these conditionals are redundant, as the underlying fini function for the entity should check these things. It should already be reported up to Python by the `Failed to fini '%s': %s` error below, I think.

wjwwood: Or at least they should be checked as of your pr https://github.com/ros2/rcl/pull/148.

dirk-thomas: With the error handling being fixed in the second commit of this PR these extra checks are indeed redundant. They even hide some of the error details. Therefore I remove the third commit.

dirk-thomas:
I missed that the new CI builds haven't finished yet. I will check back to make sure that is the case.

Issue #84, opened 2017-05-19
raise NotInitializedException when rclpy.init() is not called first


wjwwood:
This is a regression of https://github.com/ros2/rclpy/pull/82, see this for more details:

https://github.com/ros2/rclpy/pull/82#issuecomment-302818268

allenh1:
Is this something that could be worked on in the near future?

mikaelarguedas:
:+1: yep that should be a fairly small change that can be addressed in the near future

PR #98, opened 2017-07-13
Raise NotInitializedException if rclpy has not been initialized

allenh1:
connects to #84

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -15,10 +15,11 @@
 import sys
 
 from rclpy.constants import S_TO_NS
+from rclpy.exceptions import NotInitializedException
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 from rclpy.node import Node
 from rclpy.utilities import get_rmw_implementation_identifier  # noqa
-from rclpy.utilities import ok  # noqa
+from rclpy.utilities import ok
 from rclpy.utilities import shutdown  # noqa
 from rclpy.utilities import try_shutdown  # noqa
 from rclpy.validate_namespace import validate_namespace
@@ -32,6 +33,8 @@ def init(*, args=None):
 def create_node(node_name, *, namespace=None):
     namespace = namespace or ''
     failed = False
+    if not ok():
+        raise NotInitializedException('cannot create node')
     try:
         node_handle = _rclpy.rclpy_create_node(node_name, namespace)
     except ValueError:
--- a/rclpy/test/test_init_shutdown.py
+++ b/rclpy/test/test_init_shutdown.py
@@ -83,6 +83,16 @@ def func_double_shutdown():
     return False
 
 
+def func_create_node_without_init():
+    import rclpy
+    from rclpy.exceptions import NotInitializedException
+    try:
+        rclpy.create_node('foo')
+    except NotInitializedException:
+        return True
+    return False
+
+
 def func_launch(function, message):
     pool = multiprocessing.Pool(1)
     result = pool.apply(
@@ -113,3 +123,7 @@ def test_double_init():
 
 def test_double_shutdown():
     func_launch(func_double_shutdown, 'Expected Runtime error when shutting down rclpy twice')
+
+
+def test_create_node_without_init():
+    func_launch(func_create_node_without_init, 'Expected NotInitializedException')


rclpy/src/rclpy/_rclpy.c
```diff
@@ -77,6 +77,17 @@ rclpy_get_sigint_guard_condition(PyObject * Py_UNUSED(self), PyObject * Py_UNUSE
...
 
+/// Check if rclpy has been initialized.
+static PyObject *
+rclpy_is_initialized(PyObject * Py_UNUSED(self), PyObject * Py_UNUSED(args))
+{
+  bool ret = rcl_ok();
+  if (!ret) {
+    Py_RETURN_FALSE;
+  }
+  Py_RETURN_TRUE;
```
dirk-thomas: How about simplifying this logic to:

```
if (rcl_ok()) {
  Py_RETURN_TRUE;
}
Py_RETURN_FALSE
```

rclpy/rclpy/__init__.py
```diff
@@ -32,6 +32,8 @@ def init(*, args=None):
 def create_node(node_name, *, namespace=None):
     namespace = namespace or ''
     failed = False
+    if not _rclpy.rclpy_is_initialized():
+        raise NotInitializedException('rclpy has not been initialized')
```
dirk-thomas: Since the exception type `rclpy.NotInitializedException` already encapsulates the semantic I would not pass an additional string message.

rclpy/rclpy/__init__.py
```diff
@@ -32,6 +32,8 @@ def init(*, args=None):
 def create_node(node_name, *, namespace=None):
     namespace = namespace or ''
     failed = False
+    if not _rclpy.rclpy_is_initialized():
```
dirk-thomas: Can't `_rclpy` actually be used in this case since it hasn't been initialized yet? Also if that would be possible the existing function `rclpy_ok` could be used (no need to introduce `rclpy_is_initialized`).

rclpy/src/rclpy/_rclpy.c
allenh1: definitely ok with that

allenh1:
Just a quick general question: how are the tests being called here?

dirk-thomas:
> Just a quick general question: how are the tests being called here?

With `nosetests`.

allenh1:
I Meant to ask where I should put the expected results?

dirk-thomas:
> I meant to ask where I should put the expected results?

What mean "expected results" here? The tests either pass or fail. In your case you probably expect an exception. If that happens (as expected) the test should pass. If the exception doesn't happen than the test should fail.

allenh1:
Is this the desired behavior?

```
allenh1@machina:~/ros2_ws$ python3
Python 3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import rclpy
>>> node = rclpy.create_node('foo')
  File "<stdin>", line 1, in <module>
  File "/home/allenh1/ros2_ws/install/lib/python3.5/site-packages/rclpy/__init__.py", line 37, in create_node
    raise NotInitializedException('cannot create node')
rclpy.NotInitializedException: cannot create node
```

```
allenh1@machina:~/ros2_ws$ python3
Python 3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import rclpy
>>> rclpy.init()
>>> node = rclpy.create_node('foo')
>>> vars(node)
{'clients': [], 'publishers': [], 'subscriptions': [], 'services': [], 'timers': [], '_handle': <capsule object NULL at 0x7f10382057b0>}
>>> quit()
```

dirk-thomas:
> Is this the desired behavior?

Looks pretty close. The exception message should come from the existing class though: https://github.com/ros2/rclpy/blob/6b217daf62f4be6b679d3b8a4946aaafff5dc2be/rclpy/rclpy/exceptions.py#L20

allenh1:
> Looks pretty close. The exception message should come from the existing class though: https://github.com/ros2/rclpy/blob/6b217daf62f4be6b679d3b8a4946aaafff5dc2be/rclpy/rclpy/exceptions.py#L20

Good catch, forgot to remove the redefined class.

Are there any other functions that should raise this exception?

mikaelarguedas:
does this function needs to exist in the c land ? I think that a pure python approach relying on `rclpy.ok()` should be enough isnt it ? (I think that's the same remark as https://github.com/ros2/rclpy/pull/98#discussion_r127295174)

Could this test be simply added to `test_init_shutdown`, or do we need a new file + class for this one test ?

It would be cleaner to use a `try, except, else` structure rather than a `self.assertEqual(0, 1)` in the try block

allenh1:
@mikaelarguedas @dirk-thomas Yes, `ok()` does the trick. Replaced the C function.

> Could this test be simply added to test_init_shutdown, or do we need a new file + class for this one test?

@mikaelarguedas I'm not sure yet. I'll give that a try. I'd definitely prefer to use an existing class.

rclpy/rclpy/__init__.py
```diff
@@ -32,6 +33,8 @@ def init(*, args=None):
 def create_node(node_name, *, namespace=None):
     namespace = namespace or ''
     failed = False
+    if not ok():
```
dirk-thomas: Since `ok()` is being used now the `# noqa` comment on the import line is obsolete.

rclpy/test/test_init_shutdown.py
```diff
@@ -83,6 +83,17 @@ def func_double_shutdown():
     return False
 
 
+def func_create_node_without_init():
+    import rclpy
+    from rclpy.exceptions import NotInitializedException
+    try:
+        node = rclpy.create_node('foo')
+        assert node  # silence flake
```
dirk-thomas: You can simply not assign the return value of `create_node` to a variable instead.

allenh1:
Ok, addressed final remarks by @dirk-thomas.

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2889/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/376/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2328/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3013/)

PR #104, opened 2017-08-24
Executors and Callback Groups

sloretz:
This adds Executors and CallbackGroups to rclpy. This is part of feature parity with rclcpp.

Executors and CallbackGroups are added in the fourth commit on this branch. The first 3 commits are prerequisites:
1. Release the GIL so other python threads can run during a call to `rcl_wait()`
2. Add functions to rclpy for guard conditions
3. Change `Node.__init__` arguments so it can be created by inheriting from it

See examples in ros2/examples#182 and ros2/demos#169
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3213/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/533/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2568/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3280/)

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -14,11 +14,11 @@
 
 import sys
 
-from rclpy.constants import S_TO_NS
+from rclpy.executors import SingleThreadedExecutor as _SingleThreadedExecutor
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 from rclpy.node import Node
 from rclpy.utilities import get_rmw_implementation_identifier  # noqa
-from rclpy.utilities import ok  # noqa
+from rclpy.utilities import ok
 from rclpy.utilities import shutdown  # noqa
 from rclpy.utilities import try_shutdown  # noqa
 
@@ -32,74 +32,19 @@ def create_node(node_name, *, namespace=None):
 
 
 def spin_once(node, *, timeout_sec=None):
-    wait_set = _rclpy.rclpy_get_zero_initialized_wait_set()
+    executor = _SingleThreadedExecutor()
+    try:
+        executor.add_node(node)
+        executor.spin_once(timeout_sec=timeout_sec)
+    finally:
+        executor.shutdown()
 
-    _rclpy.rclpy_wait_set_init(
-        wait_set,
-        len(node.subscriptions),
-        1,
-        len(node.timers),
-        len(node.clients),
-        len(node.services))
 
-    [sigint_gc, sigint_gc_handle] = _rclpy.rclpy_get_sigint_guard_condition()
-    entities = {
-        'subscription': (node.subscriptions, 'subscription_handle'),
-        'client': (node.clients, 'client_handle'),
-        'service': (node.services, 'service_handle'),
-        'timer': (node.timers, 'timer_handle'),
-    }
-    for entity, (handles, handle_name) in entities.items():
-        _rclpy.rclpy_wait_set_clear_entities(entity, wait_set)
-        for h in handles:
-            _rclpy.rclpy_wait_set_add_entity(
-                entity, wait_set, h.__getattribute__(handle_name)
-            )
-    _rclpy.rclpy_wait_set_clear_entities('guard_condition', wait_set)
-    _rclpy.rclpy_wait_set_add_entity('guard_condition', wait_set, sigint_gc)
-
-    if timeout_sec is None:
-        timeout = -1
-    else:
-        timeout = int(float(timeout_sec) * S_TO_NS)
-
-    _rclpy.rclpy_wait(wait_set, timeout)
-
-    guard_condition_ready_list = _rclpy.rclpy_get_ready_entities('guard_condition', wait_set)
-    if sigint_gc_handle in guard_condition_ready_list:
-        raise KeyboardInterrupt
-
-    timer_ready_list = _rclpy.rclpy_get_ready_entities('timer', wait_set)
-    for tmr in [t for t in node.timers if t.timer_pointer in timer_ready_list]:
-        if _rclpy.rclpy_is_timer_ready(tmr.timer_handle):
-            _rclpy.rclpy_call_timer(tmr.timer_handle)
-            tmr.callback()
-
-    sub_ready_list = _rclpy.rclpy_get_ready_entities('subscription', wait_set)
-    for sub in [s for s in node.subscriptions if s.subscription_pointer in sub_ready_list]:
-        msg = _rclpy.rclpy_take(sub.subscription_handle, sub.msg_type)
-        if msg:
-            sub.callback(msg)
-
-    client_ready_list = _rclpy.rclpy_get_ready_entities('client', wait_set)
-    for client in [c for c in node.clients if c.client_pointer in client_ready_list]:
-        response = _rclpy.rclpy_take_response(
-            client.client_handle, client.srv_type.Response, client.sequence_number)
-        if response:
-            # clients spawn their own thread to wait for a response in the wait_for_future function
-            # users can either use this mechanism or monitor the content of
-            # client.response themselves to check if a response have been received
-            client.response = response
-
-    service_ready_list = _rclpy.rclpy_get_ready_entities('service', wait_set)
-    for srv in [s for s in node.services if s.service_pointer in service_ready_list]:
-        request_and_header = _rclpy.rclpy_take_request(srv.service_handle, srv.srv_type.Request)
-        if request_and_header is None:
-            continue
-        [request, header] = request_and_header
-        if request:
-            response = srv.callback(request, srv.srv_type.Response())
-            srv.send_response(response, header)
-
-    _rclpy.rclpy_destroy_entity('guard_condition', sigint_gc)
-    _rclpy.rclpy_destroy_wait_set(wait_set)
+def spin(node):
+    executor = _SingleThreadedExecutor()
+    try:
+        executor.add_node(node)
+        while ok():
+            executor.spin_once()
+    finally:
+        executor.shutdown()
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -15,15 +15,12 @@
 import sys
 
 from rclpy.constants import S_TO_NS
-from rclpy.exceptions import NotInitializedException
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 from rclpy.node import Node
 from rclpy.utilities import get_rmw_implementation_identifier  # noqa
-from rclpy.utilities import ok
+from rclpy.utilities import ok  # noqa
 from rclpy.utilities import shutdown  # noqa
 from rclpy.utilities import try_shutdown  # noqa
-from rclpy.validate_namespace import validate_namespace
-from rclpy.validate_node_name import validate_node_name
 
 
 def init(*, args=None):
@@ -31,24 +28,7 @@ def init(*, args=None):
 
 
 def create_node(node_name, *, namespace=None):
-    namespace = namespace or ''
-    failed = False
-    if not ok():
-        raise NotInitializedException('cannot create node')
-    try:
-        node_handle = _rclpy.rclpy_create_node(node_name, namespace)
-    except ValueError:
-        failed = True
-    if failed:
-        # these will raise more specific errors if the name or namespace is bad
-        validate_node_name(node_name)
-        # emulate what rcl_node_init() does to accept '' and relative namespaces
-        if not namespace:
-            namespace = '/'
-        if not namespace.startswith('/'):
-            namespace = '/' + namespace
-        validate_namespace(namespace)
-    return Node(node_handle)
+    return Node(node_name, namespace=namespace)
 
 
 def spin_once(node, *, timeout_sec=None):
--- a/rclpy/rclpy/callback_groups.py
+++ b/rclpy/rclpy/callback_groups.py
@@ -1,0 +1,112 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from threading import Lock
+import weakref
+
+
+class CallbackGroup:
+    """Control when callbacks are allowed to be executed."""
+
+    def __init__(self):
+        super().__init__()
+        self.entities = set()
+
+    def add_entity(self, entity):
+        """
+        Add an entity to the callback group.
+
+        :param entity: a subscription, timer, client, or service instance
+        :rtype: None
+        """
+        self.entities.add(weakref.ref(entity))
+
+    def has_entity(self, entity):
+        """
+        Determine if an entity has been added to this group.
+
+        :param entity: a subscription, timer, client, or service instance
+        :rtype: bool
+        """
+        return weakref.ref(entity) in self.entities
+
+    def can_execute(self, entity):
+        """
+        Return true if an entity can be executed.
+
+        :param entity: a subscription, timer, client, or service instance
+        :rtype: bool
+        """
+        raise NotImplementedError
+
+    def beginning_execution(self, entity):
+        """
+        Get permission from the callback from the group to begin executing an entity.
+
+        Return true if the callback can be executed, false otherwise. If this returns True then
+        :func:`CallbackGroup.ending_execution` must be called after the callback has been executed.
+
+        :param entity: a subscription, timer, client, or service instance
+        :rtype: bool
+        """
+        raise NotImplementedError
+
+    def ending_execution(self, entity):
+        """
+        Notify group that a callback has finished executing.
+
+        :param entity: a subscription, timer, client, or service instance
+        :rtype: None
+        """
+        raise NotImplementedError
+
+
+class ReentrantCallbackGroup(CallbackGroup):
+    """Allow callbacks to be executed in parallel without restriction."""
+
+    def can_execute(self, entity):
+        return True
+
+    def beginning_execution(self, entity):
+        return True
+
+    def ending_execution(self, entity):
+        pass
+
+
+class MutuallyExclusiveCallbackGroup(CallbackGroup):
+    """Allow only one callback to be executing at a time."""
+
+    def __init__(self):
+        super().__init__()
+        self._active_entity = None
+        self._lock = Lock()
+
+    def can_execute(self, entity):
+        with self._lock:
+            assert weakref.ref(entity) in self.entities
+            return self._active_entity is None
+
+    def beginning_execution(self, entity):
+        with self._lock:
+            assert weakref.ref(entity) in self.entities
+            if self._active_entity is None:
+                self._active_entity = entity
+                return True
+        return False
+
+    def ending_execution(self, entity):
+        with self._lock:
+            assert self._active_entity == entity
+            self._active_entity = None
--- a/rclpy/rclpy/callback_groups.py
+++ b/rclpy/rclpy/callback_groups.py
@@ -48,7 +48,7 @@ class CallbackGroup:
         :param entity: a subscription, timer, client, or service instance
         :rtype: bool
         """
-        raise NotImplementedError
+        raise NotImplementedError()
 
     def beginning_execution(self, entity):
         """
@@ -60,7 +60,7 @@ class CallbackGroup:
         :param entity: a subscription, timer, client, or service instance
         :rtype: bool
         """
-        raise NotImplementedError
+        raise NotImplementedError()
 
     def ending_execution(self, entity):
         """
@@ -69,7 +69,7 @@ class CallbackGroup:
         :param entity: a subscription, timer, client, or service instance
         :rtype: None
         """
-        raise NotImplementedError
+        raise NotImplementedError()
 
 
 class ReentrantCallbackGroup(CallbackGroup):
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -54,7 +54,7 @@ class ResponseThread(threading.Thread):
 class Client:
     def __init__(
             self, node_handle, client_handle, client_pointer,
-            srv_type, srv_name, qos_profile):
+            srv_type, srv_name, qos_profile, callback_group):
         self.node_handle = node_handle
         self.client_handle = client_handle
         self.client_pointer = client_pointer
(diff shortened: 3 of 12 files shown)


sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3078/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/465/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2471/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3157/)

rclpy/rclpy/executors.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        _rclpy.rclpy_destroy_guard_condition(self._guard_condition)
+
+    def add_node(self, node):
+        """
+        Add a node whose callbacks should be managed by this executor.
+
+        return true if the node was added
+        """
+        self._nodes.add(node)
```
wjwwood: What happens if you add a node while spinning in another thread? Don't you need to interrupt any running spins after adding the node? Also don't you need to make this function thread-safe?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1666,7 +1752,13 @@ rclpy_wait(PyObject * Py_UNUSED(self), PyObject * args)
 #endif  // _WIN32
   previous_handler = signal(SIGINT, catch_function);
   rcl_wait_set_t * wait_set = (rcl_wait_set_t *)PyCapsule_GetPointer(pywait_set, NULL);
-  rcl_ret_t ret = rcl_wait(wait_set, timeout);
+  rcl_ret_t ret;
+
+  // Could be a long wait, release the GIL
+  Py_BEGIN_ALLOW_THREADS;
+  ret = rcl_wait(wait_set, timeout);
+  Py_END_ALLOW_THREADS;
```
wjwwood: 👍, cool

rclpy/rclpy/executors.py
```diff
@@ -0,0 +1,290 @@
...
+        self._guard_condition_handle = gc_handle
+
+    def shutdown(self, timeout_sec=None):
+        """
+        Stop executing callbacks and wait for their completion.
+
+        timeout_sec - seconds to wait for callbacks to complete. Blocks forever if None. Returns
+                  immediately if <= 0
+        """
+        _rclpy.rclpy_destroy_guard_condition(self._guard_condition)
```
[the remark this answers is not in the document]
sloretz: It might be more useful if it put a stop to the spinning and waited for all executing handlers to complete.

rclpy/rclpy/executors.py
sloretz: I didn't consider that. Yes to both questions, it should be made thread safe and trigger a rebuild of the wait set.

rclpy/rclpy/callback_groups.py
```diff
@@ -0,0 +1,107 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from threading import Lock
+import weakref
+
+
+class CallbackGroup:
```
dirk-thomas: Every class should subclass from `object` (if not from any other type).

rclpy/rclpy/callback_groups.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.entities = set()
+
+    def add_entity(self, entity):
+        """
+        Add an entity to the callback group.
+
+        entity - a subscription, timer, client, or service instance
```
dirk-thomas: Some example with a docblock for parameters: https://github.com/ament/ament_package/blob/30ec27ee5bec133bef51abd1f82b2c38c8d07cac/ament_package/__init__.py#L72-L75

rclpy/rclpy/client.py
```diff
@@ -50,7 +50,7 @@ def run(self):
 class Client:
     def __init__(
             self, node_handle, client_handle, client_pointer,
-            srv_type, srv_name, qos_profile):
+            srv_type, srv_name, qos_profile, callback_group):
```
dirk-thomas: Maybe move the default for the callback group from the `create_client` function here?

Same for other entities.

rclpy/rclpy/executors.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        Yield callbacks that are ready to be performed.
+
+        timeout_sec - seconds to wait for a callback. Blocks forever if None. Don't wait if <= 0
+        nodes - a subset of the added nodes to wait on. Wait on all nodes if None.
+
+        It's possible for this to return before the timeout without yielding any callbacks. In this
+        case the executor should call this method again.
+
+        Yeilds 3-tuple (callable, entity, node)
```
dirk-thomas: Spelling: Yields

rclpy/rclpy/executors.py
```diff
@@ -0,0 +1,290 @@
...
+    """Runs callbacks one at a time in the thread which calls spin()."""
+
+    def __init__(self):
+        super().__init__()
+
+    def spin_once(self, timeout_sec=None):
+        for handler, entity, node in self.wait_for_ready_callbacks(
+                timeout_sec=timeout_sec):
+            handler()
+            break
```
dirk-thomas: Since the `for` loop always break this could just get the `next` from the generator.

Same below.

rclpy/rclpy/callback_groups.py
```diff
@@ -0,0 +1,107 @@
...
+
+from threading import Lock
+import weakref
+
+
+class CallbackGroup:
+    """Control when callbacks are allowed to be executed."""
+
+    def __init__(self):
+        super().__init__()
```
dirk-thomas: `super` should be called with two arguments: the class type and `self`: `super(CallbackGroup, self).__init__()`

rclpy/rclpy/callback_groups.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        return weakref.ref(entity) in self.entities
+
+    def can_take(self, entity):
+        """
+        Return true if an entity can be executed.
+
+        entity - a subscription, timer, client, or service instance
+        """
+        raise NotImplementedError("Derived classes must implement this")
```
dirk-thomas: I would say the exception type already describes the reason / intention. I would skip the custom message - but that is probably pretty subjective.

Anyway the string should be in single quotes.

rclpy/test/test_callback_group.py
```diff
@@ -0,0 +1,100 @@
...
+from rclpy.callback_groups import ReentrantCallbackGroup
+from std_msgs.msg import String
+
+
+class TestCallbackGroups(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        rclpy.init()
+        cls.node = rclpy.create_node('my_node', namespace='/my_ns')
```
dirk-thomas: The node should be named after the test (`test_callback_group`) that makes it less likely to collide (in case we ever enforce uniqueness) and makes it easier to debug what is currently running.

rclpy/rclpy/callback_groups.py
dirk-thomas: Actually [PEP 3135](https://www.python.org/dev/peps/pep-3135/) seems to advocate to call it without arguments in Python 3. I remember that at least in the past with Python 2 that was "not a good idea". Our existing code passes the class name and `self`. Maybe we should update our code base to this new style?

rclpy/rclpy/callback_groups.py
sloretz: Even in python 3? [This page](https://docs.python.org/3.5/tutorial/classes.html#multiple-inheritance) says all classes inherit from object in python 3.5.

rclpy/rclpy/callback_groups.py
sloretz: Looks like the PEP guarantees the old style will still work. To be honest I don't fully understand the advantages of the new style besides less typing. I can make a jira ticket to look into updating the existing code.

rclpy/rclpy/callback_groups.py
sloretz: Sphinx style?

dirk-thomas: :+1:

rclpy/rclpy/callback_groups.py
mikaelarguedas: I think you're right @sloretz we had the same interrogation before when developing rclpy and concluded that inheriting from object was not necessary, that's why non of the classes here inherit from object.

dirk-thomas: We should follow that style in the existing code then.

mikaelarguedas: https://github.com/ros2/rclpy/pull/106 for rcllpy. I'll open PRs where necessary

rclpy/rclpy/callback_groups.py
```diff
@@ -0,0 +1,107 @@
...
+        super().__init__()
+        self._active_entity = None
+        self._lock = Lock()
+
+    def can_take(self, entity):
+        with self._lock:
+            assert weakref.ref(entity) in self.entities
+            return self._active_entity is None
+
+    def take(self, entity):
```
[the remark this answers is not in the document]
sloretz: I agree that `take`/`give_back` doesn't really describe what's happening here, but acquire/release is closer. It's almost like the executor is borrowing the entity with `take` being "please may I borrow this" and `give_back` being "thanks, here it is". I'll update it to `can_acquire`/`acquire_for_taking`/`release_for_taking` unless something more representative comes up.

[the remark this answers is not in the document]
sloretz: @wjwwood How about `can_execute`, `begin_execution`/`end_execution`?

rclpy/rclpy/__init__.py
```diff
@@ -14,109 +14,37 @@
...
-    for sub in [s for s in node.subscriptions if s.subscription_pointer in sub_ready_list]:
-        msg = _rclpy.rclpy_take(sub.subscription_handle, sub.msg_type)
-        if msg:
-            sub.callback(msg)
+    try:
+        executor = _SingleThreadedExecutor()
+        executor.add_node(node)
+        executor.spin_once(timeout_sec=timeout_sec)
+    finally:
+        executor.shutdown()
```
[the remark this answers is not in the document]
sloretz: Fixed d760b98b7c5688c8e6fb35af77fb8bcd854bd70c

rclpy/rclpy/__init__.py
```diff
@@ -14,109 +14,37 @@
...
-            response = srv.callback(request, srv.srv_type.Response())
-            srv.send_response(response, header)
+def spin(node):
+    try:
+        executor = _SingleThreadedExecutor()
+        executor.add_node(node)
+        while ok():
+            executor.spin_once()
+    finally:
+        executor.shutdown()
```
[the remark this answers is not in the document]
sloretz: Fixed d760b98b7c5688c8e6fb35af77fb8bcd854bd70c

rclpy/rclpy/callback_groups.py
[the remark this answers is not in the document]
sloretz: `can_execute`, `begin_execution`, `end_execution` in b1f882ec13bc182be2a23fdf98c1296492a69756. I'm not convinced those are the right names, but I think it's an improvement. The base `Executor` still has internal methods that begin with `_take` where it matches `rcl`.

rclpy/rclpy/executors.py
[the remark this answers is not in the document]
sloretz: `shutdown` waits for all work to complete in 1c74d65b2dd7aae68710dc5d9ac32623139f280e. Neither of the derived executors override it still. Maybe `MultiThreadedExecutor` could override it and `join` the worker threads?

rclpy/rclpy/executors.py
sloretz: Fixed in 89ed2c4fcad9f2a94aa23562476e6d340bacde43. The commit also eliminates a case where `wait_for_ready_callbacks` could return before the timeout was reached without yielding any work. Without fixing that case `add_node` would cause the next `spin_once` to return without doing anything because the wait list guard condition was triggered. This broke the executor tests.

rclpy/rclpy/callback_groups.py
sloretz: Fixed in 7b4684c17d211a387a79ebd64608906b4e352f91

rclpy/rclpy/callback_groups.py
sloretz: Message removed in fa60f5ab0be6b350a7e63b4e696bf7d5aca854e5

rclpy/rclpy/executors.py
sloretz: Fixed in 7b4684c17d211a387a79ebd64608906b4e352f91

rclpy/rclpy/executors.py
sloretz: Fixed in 073146983d5be98509feaef4841887dc0d3359a5

rclpy/test/test_callback_group.py
sloretz: Fixed in 372d3880dac68ebc1c8796b886044b78b66a1ac7

rclpy/rclpy/client.py
sloretz: I think the node needs to own the default callback group so that callbacks can be mutually exclusive by default.

rclpy/rclpy/callback_groups.py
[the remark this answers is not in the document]
wjwwood: @sloretz Those are better I think, though I would have said `can_execute_from` (meaning can execute something from this callback group), `begin_execution` or `begin_execution_of_entity` (meaning begin execution of an entity from this callback group), and `end_execution` or `end_execution_of_entity` (meaning the same as the last one), as you're not executing the callback group itself but rather something from within it. That being said, the ones I just listed with a `_from` suffix are also a bit confusing without some context. 

I'm fine with what you've got there now, with the small issue that it sounds like you're executing the callback group when you are not.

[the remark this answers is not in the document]
sloretz: @wjwwood Yeah, I have the same concern about the callback group is doing the execution. I think it could be improved by adding "ing" `beginning_execution`/`ending_execution`.

rclpy/rclpy/executors.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        self._is_shutdown = True
+        # Wait for all work to complete
+        if timeout_sec is None or timeout_sec >= 0:
+            with self._work_condition:
+                self._work_condition.wait_for(lambda: self._num_work_executing == 0, timeout_sec)
+        # Clean up stuff that won't be used anymore
+        with self._nodes_lock:
+            self._nodes = set()
+        _rclpy.rclpy_destroy_guard_condition(self._guard_condition)
```
dirk-thomas: Assuming a user wants to call `shutdown` with a timeout of a few seconds. After the function call returns the user doesn't know if the shutdown was actually "successful / clean". So I would imagine the function should return a `bool` to indicate that.

In the case the shutdown wasn't "successful / clean" I wouldn't expect that the nodes and condition are destroyed. What if the user wants to print a message to the console "waited for shutdown, but still ongoing..." and then wait for another amount of time?

sloretz: Fixed in 34c3ed257b2501ce550f3482624bf5f8a94b31e1

rclpy/rclpy/callback_groups.py
```diff
@@ -0,0 +1,112 @@
...
+        return weakref.ref(entity) in self.entities
+
+    def can_execute(self, entity):
+        """
+        Return true if an entity can be executed.
+
+        :param entity: a subscription, timer, client, or service instance
+        :rtype: bool
+        """
+        raise NotImplementedError
```
dirk-thomas: `NotImplementedError` is an exception class and should therefore be instantiated: `raise NotImplementedError()`.

Same below.

rclpy/rclpy/executors.py
```diff
@@ -0,0 +1,380 @@
...
+                    num_work += 1
+
+                arg = take_from_wait_list(entity)
+
+                # Signal that this has been 'taken' and can be added back to the wait list
+                entity._executor_event = False
+                _rclpy.rclpy_trigger_guard_condition(gc)
+
+                call_callback(entity, arg)
+            finally:
```
dirk-thomas: What kind of exceptions are you trying to catch here? If the lines 190-193 can't raise that they should be before the `try`. If they do, the code might raise before incrementing `num_work` and only decrement the counter in the `finally` block.

rclpy/rclpy/node.py
```diff
@@ -30,13 +33,35 @@
...
+        namespace = namespace or ''
+        failed = False
+        if not ok():
+            raise NotInitializedException('cannot create node')
+        try:
+            node_handle = _rclpy.rclpy_create_node(node_name, namespace)
+            self._handle = node_handle
+        except ValueError:
+            failed = True
+        if failed:
```
dirk-thomas: The variable `failed` seems to be unnecessary. The error handling code could just be in the `except` block.

rclpy/rclpy/node.py
```diff
@@ -30,13 +33,35 @@
...
+            # these will raise more specific errors if the name or namespace is bad
+            validate_node_name(node_name)
+            # emulate what rcl_node_init() does to accept '' and relative namespaces
+            if not namespace:
+                namespace = '/'
+            if not namespace.startswith('/'):
+                namespace = '/' + namespace
+            validate_namespace(namespace)
+            # Should not get to this point
+            raise RuntimeError("rclpy_create_node failed for unknown reason")
```
dirk-thomas: This should be single quotes. I would assume if you re-run CI builds they should flag that (as of ament/ament_lint#84).

rclpy/src/rclpy/_rclpy.c
```diff
@@ -77,6 +77,92 @@ rclpy_get_sigint_guard_condition(PyObject * Py_UNUSED(self), PyObject * Py_UNUSE
...
+static PyObject *
+rclpy_destroy_guard_condition(PyObject * Py_UNUSED(self), PyObject * args)
+{
+  PyObject * pygc;
+
+  if (!PyArg_ParseTuple(args, "O", &pygc)) {
+    Py_RETURN_FALSE;
+  }
+
+  rcl_guard_condition_t * gc = (rcl_guard_condition_t *)PyCapsule_GetPointer(pygc, NULL);
```
dirk-thomas: This needs to be modified in a similar way as in #109 / #110 / #111. First it needs to check if the capsule is `Py_None` and after calling `fini` it needs to `PyMem_Free` the pointer.

mikaelarguedas: This should be part of `rclpy_destroy_entity` by just adding another condition to the if else and leverage the PyMem_Free introduced in #110 (PR to come)

rclpy/rclpy/node.py
sloretz: Fixed in f2c23341c419c0cc652252e11d042770a652aae4

rclpy/rclpy/node.py
sloretz: Removed in 3f66415d4788c4c0d6c06e4529c17b13640e9948

rclpy/rclpy/callback_groups.py
sloretz: Exceptions are instantiated in c74a9947f761c9e8aca703e464d185e1ebc50f19

Side note, it should be equivalent because of some sneaky behavior. [PEP 3109](https://www.python.org/dev/peps/pep-3109/) says when a subclass of BaseException is raised it gets instantiated by calling it with no arguments.

rclpy/rclpy/executors.py
sloretz: Good point. The try/finally block was being used to make sure work got decremented if a user's callback raised. I've moved the work incrementing/decrementing to a helper class using __enter__/__exit__ so that the try/finally block could be  narrowed to just the user's callback.

c008cddaef8676af3c1f02b9263d05180864b5f5
223ab33016d7b3764ad87344517a546cc9e4ed25

rclpy/test/test_executor.py
```diff
@@ -68,6 +69,15 @@ def test_add_node_to_executor(self):
         executor.add_node(self.node)
         self.assertIn(self.node, executor.get_nodes())
 
+    def test_executor_spin_non_blocking(self):
+        self.assertIsNotNone(self.node.handle)
+        executor = SingleThreadedExecutor()
+        executor.add_node(self.node)
+        start = time.time()
```
mikaelarguedas: for consistency I'd use steady time (`time.monotonic`) here given that this is what the rcl timers are using

sloretz: Done in ac6250fcd304f3e7d5f283126bd7ac3322543238

rclpy/rclpy/node.py
```diff
@@ -30,13 +33,32 @@
...
         self.subscriptions = []
         self.timers = []
+        self._default_callback_group = MutuallyExclusiveCallbackGroup()
+
+        namespace = namespace or ''
+        if not ok():
+            raise NotInitializedException('cannot create node')
+        try:
+            node_handle = _rclpy.rclpy_create_node(node_name, namespace)
+            self._handle = node_handle
```
dirk-thomas: The temporary variable `node_handle` is needed. The result can directly be assigned to `self._handle`.

rclpy/src/rclpy/_rclpy.c
dirk-thomas: This function still needs to check that the `PyObject` is not `Py_None` before suing it.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -77,6 +77,66 @@ rclpy_get_sigint_guard_condition(PyObject * Py_UNUSED(self), PyObject * Py_UNUSE
...
+/// Create a general purpose guard condition
+/**
+ * A successful call will return a list with two elements:
+ *
+ * - a Capsule with the pointer of the created rcl_guard_condition_t * structure
+ * - an integer representing the memory address of the rcl_guard_condition_t
+ *
+ * \return a list with the capsule and memory location, or
+ * \return NULL on failure
+ */
```
dirk-thomas: The doc block should mention what function should be called to destroy the created enitty.

sloretz:
PR squashed/merged. Will address issues pointed out by @dirk-thomas in a PR to be merged after r2b3.

rclpy/rclpy/node.py
sloretz: 8c4b1ddff392a444dd20491a48da22f71ee611ad

rclpy/src/rclpy/_rclpy.c
sloretz: ec4bf8e5da4d1eaca06a856eda2a14b6bf70f0d6

sloretz:
@dirk-thomas I can't reply to your comment about checking for Py_None, but It's fixed here f6ff01e5fd23e24f4c3772d258d0c394f04c8583

PR #105, opened 2017-08-24
Remove duplicate call to rclpy_create_timer()

sloretz:
This removes an extra call to `rclpy_create_timer()`  and two unused arguments in `WallTimer.__init__()`. A timer handle was created inside of `Node.create_timer()` and given to `WallTimer.__init__()` which ignored the handle and created its own.


mikaelarguedas:
Looks good to me, we discussed it offline a few day ago. I had it as part of a bigger timer cleanup PR connected to https://github.com/ros2/rcl/pull/160 but we can merge this now and I'll remove it from the other one

PR #106, opened 2017-08-25
implicitly inherit from object

mikaelarguedas:
In python 3 classes [implicitely inherit from object](https://docs.python.org/3.5/tutorial/classes.html#multiple-inheritance), so no need to express inheritence except if the class inherits from something else.
Addresses https://github.com/ros2/rclpy/pull/104#discussion_r135370317

### DIFF
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -17,7 +17,7 @@ from enum import IntEnum
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
 
-class QoSProfile(object):
+class QoSProfile:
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/qos.py
```diff
@@ -17,7 +17,7 @@
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
 
-class QoSProfile(object):
+class QoSProfile():
```
dirk-thomas: Should the parenthesis be removed too?

mikaelarguedas: yes they should :+1:

PR #108, opened 2017-08-31
Don't add canceled timers to waitset

mikaelarguedas:
follow up of https://github.com/ros2/rcl/pull/160 and https://github.com/ros2/rcl/pull/163
connects to ros2/rcl#163

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -93,8 +93,9 @@ def spin_once(node, *, timeout_sec=None):
 
     timer_ready_list = _rclpy.rclpy_get_ready_entities('timer', wait_set)
     for tmr in [t for t in active_timer_list if t.timer_pointer in timer_ready_list]:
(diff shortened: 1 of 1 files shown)


wjwwood:
@mikaelarguedas I was just looking into https://github.com/ros2/rclcpp/issues/319 and I think that it might be the same problem but in C++.

Perhaps we should check the given timers in rcl_wait and ignore the canceled ones? I still don't think it is "correct" to pass canceled timers to wait, but at least this will prevent similar code in rclpy and rclcpp, plus the state of the timer's cancel is stored in rcl as well (rather than in the c++/py code).

wjwwood:
Oh wait, we aren't considering passing canceled timers anyways it seems...

https://github.com/ros2/rcl/blob/master/rcl/src/rcl/wait.c#L545-L552

Maybe there is something broken in that logic. I'll look into it.

wjwwood:
Closing because it is not needed after https://github.com/ros2/rcl/pull/163.

PR #109, opened 2017-09-05
add function to destroy waitset and call it in spin_once

### LINKED ISSUE #74: Increasing memory usage when only using spin_once

dirk-thomas:
Addresses the memory leak described in #74. The script still uses a large amount of memory though.

This PR only fixes the leak of the waitset. A similar patch is required for every entity allocated with `PyMem_Malloc` (node, publisher, subscriptions, service, client, timer, guard condition, header (twice), qos_profile).

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -1588,6 +1588,44 @@ rclpy_wait_set_add_entity(PyObject * Py_UNUSED(self), PyObject * args)
   Py_RETURN_NONE;
 }
 
+/// Destroy the waitset structure
+/**
+ * \param[in] pywait_set Capsule pointing to the waitset structure
+ * \return NULL
+ */
+static PyObject *
+rclpy_destroy_wait_set(PyObject * Py_UNUSED(self), PyObject * args)
+{
+  PyObject * pywait_set;
+
+  if (!PyArg_ParseTuple(args, "O", &pywait_set)) {
+    return NULL;
+  }
+  void * p = PyCapsule_GetPointer(pywait_set, NULL);
+  if (Py_None == p) {
+    PyErr_Format(PyExc_RuntimeError, "wait set is None");
+    return NULL;
+  }
+  rcl_wait_set_t * wait_set = (rcl_wait_set_t *)p;
+
+  rcl_ret_t ret = rcl_wait_set_fini(wait_set);
+  if (ret != RCL_RET_OK) {
+    PyErr_Format(PyExc_RuntimeError,
+      "Failed to fini wait set: %s", rcl_get_error_string_safe());
+    rcl_reset_error();
+    return NULL;
+  }
+
+  PyMem_Free(wait_set);
+
+  if (PyCapsule_SetPointer(pywait_set, Py_None)) {
+    // exception set by PyCapsule_SetPointer
+    return NULL;
+  }
+
+  Py_RETURN_NONE;
+}
+
 #define GET_LIST_READY_ENTITIES(ENTITY_TYPE) \
   size_t idx; \
   size_t idx_max; \
@@ -2290,6 +2328,9 @@ static PyMethodDef rclpy_methods[] = {
 
   {"rclpy_wait_set_add_entity", rclpy_wait_set_add_entity, METH_VARARGS,
    "rclpy_wait_set_add_entity."},
+
+  {"rclpy_destroy_wait_set", rclpy_destroy_wait_set, METH_VARARGS,
+   "rclpy_destroy_wait_set."},
 
   {"rclpy_get_ready_entities", rclpy_get_ready_entities, METH_VARARGS,
    "List non null subscriptions in waitset."},
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -120,3 +120,5 @@ def spin_once(node, *, timeout_sec=None):
         if request:
             response = srv.callback(request, srv.srv_type.Response())
             srv.send_response(response, header)
+
+    _rclpy.rclpy_destroy_wait_set(wait_set)


mikaelarguedas:
While that will allow to destroy the wait set, I don't think that will fix the main memory leak seen in spin_once (to be fair I didnt see any difference in memory allocation with or without this patch).
After a quick look at it I think that the guard_condition is the main reason of the leak and by removing it from `spin_once` I don't see any leak whatsoever.
Can you try it with and without the following patch and confirm that the leak stops only when removing the guard_condition from the equation ?
```
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -57,12 +57,12 @@ def spin_once(node, *, timeout_sec=None):
     _rclpy.rclpy_wait_set_init(
         wait_set,
         len(node.subscriptions),
-        1,
+        0,  # 1,
         len(node.timers),
         len(node.clients),
         len(node.services))
 
-    [sigint_gc, sigint_gc_handle] = _rclpy.rclpy_get_sigint_guard_condition()
+    # [sigint_gc, sigint_gc_handle] = _rclpy.rclpy_get_sigint_guard_condition()
     entities = {
         'subscription': (node.subscriptions, 'subscription_handle'),
         'client': (node.clients, 'client_handle'),
@@ -75,8 +75,8 @@ def spin_once(node, *, timeout_sec=None):
             _rclpy.rclpy_wait_set_add_entity(
                 entity, wait_set, h.__getattribute__(handle_name)
             )
-    _rclpy.rclpy_wait_set_clear_entities('guard_condition', wait_set)
-    _rclpy.rclpy_wait_set_add_entity('guard_condition', wait_set, sigint_gc)
+   #  _rclpy.rclpy_wait_set_clear_entities('guard_condition', wait_set)
+   #  _rclpy.rclpy_wait_set_add_entity('guard_condition', wait_set, sigint_gc)
 
     if timeout_sec is None:
         timeout = -1
@@ -85,9 +85,9 @@ def spin_once(node, *, timeout_sec=None):
 
     _rclpy.rclpy_wait(wait_set, timeout)
 
-    guard_condition_ready_list = _rclpy.rclpy_get_ready_entities('guard_condition', wait_set)
-    if sigint_gc_handle in guard_condition_ready_list:
-        raise KeyboardInterrupt
+    # guard_condition_ready_list = _rclpy.rclpy_get_ready_entities('guard_condition', wait_set)
+    # if sigint_gc_handle in guard_condition_ready_list:
+    #     raise KeyboardInterrupt
 
     timer_ready_list = _rclpy.rclpy_get_ready_entities('timer', wait_set)
     for tmr in [t for t in node.timers if t.timer_pointer in timer_ready_list]:
```

dirk-thomas:
> While that will allow to destroy the wait set, I don't think that will fix the main memory leak seen in spin_once (to be fair I didnt see any difference in memory allocation with or without this patch).

This patch does remove the repeated leak mention in #74. I used https://github.com/ros2/rclpy/issues/74#issuecomment-327333712 to confirm this. Please try it with the same example without and with this patch. I would be surprised if your results are different.

mikaelarguedas:
After quick check with @dirk-thomas, this patch does fix the memory leak reported by valgrind. The memory usage still goes up a lot when using a very small timeout but is (according to valgrind) not leaked by the process.

I can confirm that the memory accumulation stops growing when removing the guard_condition so there is definitely something to do there but can be addressed in a separate PR.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1588,6 +1588,48 @@ rclpy_wait_set_add_entity(PyObject * Py_UNUSED(self), PyObject * args)
...
+  }
+
+  PyMem_Free(wait_set);
+
+  if (PyCapsule_SetPointer(pywait_set, Py_None)) {
+    // exception set by PyCapsule_SetPointer
+    return NULL;
+  }
+
+  Py_RETURN_NONE;
```
mikaelarguedas: This could return true on success and false on failure instead of always returning none (thouhg it looks like every failure raises an exception so I'm not sure we would ever return false)

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1588,6 +1588,48 @@ rclpy_wait_set_add_entity(PyObject * Py_UNUSED(self), PyObject * args)
...
+    return NULL;
+  }
+  void * p = PyCapsule_GetPointer(pywait_set, NULL);
+  if (p == Py_None) {
+    PyErr_Format(PyExc_RuntimeError, "wait set is None");
+    return NULL;
+  }
+  rcl_wait_set_t * wait_set = (rcl_wait_set_t *)p;
+  if (!wait_set) {
+    PyErr_Format(PyExc_RuntimeError, "wait set is NULL");
```
mikaelarguedas: Isnt that check redundant? This should already be checked and reported by `rcl_wait_set_fini` right ?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1588,6 +1588,48 @@ rclpy_wait_set_add_entity(PyObject * Py_UNUSED(self), PyObject * args)
...
+rclpy_destroy_wait_set(PyObject * Py_UNUSED(self), PyObject * args)
+{
+  PyObject * pywait_set;
+
+  if (!PyArg_ParseTuple(args, "O", &pywait_set)) {
+    return NULL;
+  }
+  void * p = PyCapsule_GetPointer(pywait_set, NULL);
+  if (p == Py_None) {
+    PyErr_Format(PyExc_RuntimeError, "wait set is None");
```
mikaelarguedas: Same here that should result in the same error in `rcl_wait_set_fini`, not sure it needs to be checked in every layer

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1588,6 +1588,48 @@ rclpy_wait_set_add_entity(PyObject * Py_UNUSED(self), PyObject * args)
...
+    return NULL;
+  }
+  rcl_wait_set_t * wait_set = (rcl_wait_set_t *)p;
```
mikaelarguedas: if we remove the redundant check we should be able to just cast the return of `PyCapsule_GetPointer` to a `rcl_wait_set_t *`

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1588,6 +1588,48 @@ rclpy_wait_set_add_entity(PyObject * Py_UNUSED(self), PyObject * args)
...
+static PyObject *
+rclpy_destroy_wait_set(PyObject * Py_UNUSED(self), PyObject * args)
+{
+  PyObject * pywait_set;
+
+  if (!PyArg_ParseTuple(args, "O", &pywait_set)) {
+    return NULL;
+  }
+  void * p = PyCapsule_GetPointer(pywait_set, NULL);
+  if (p == Py_None) {
```
mikaelarguedas: Nit: `if (Py_None == p) {`

dirk-thomas: Fixed in 5e156321972b2335582c2f683c7baa398a416b3f.

rclpy/src/rclpy/_rclpy.c
dirk-thomas: Checking for Python `None` here is different from checking for `NULL` inside `rcl_wait_set_fini` and is therefore necessary.

rclpy/src/rclpy/_rclpy.c
dirk-thomas: Comparing `rcl_wait_set_t * wait_set` directly with `Py_None` would result in a compiler warning.

rclpy/src/rclpy/_rclpy.c
dirk-thomas: Removed in 219c67410e3c75f4db97369f26204706fcb6b1ff.

rclpy/src/rclpy/_rclpy.c
dirk-thomas: Since the function always raises in case of problems it doesn't need to return anything. This is also the behavior of many other function in this C extension.

mikaelarguedas: :+1:

rclpy/src/rclpy/_rclpy.c
mikaelarguedas: :+1:

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3160/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/498/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2517/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3211/)

dirk-thomas:
I tried the [above patch](https://github.com/ros2/rclpy/pull/109#issuecomment-327334724) locally. I had to add at least a time to the minimal example with the `spin_once` loop since otherwise the waitset if empty which results in an exception.

While the script is still accumulating memory over time (in 30s ~110 MB) that is much slower than without the patch (in 30 ~850 MB). This is kind of expected since otherwise a new guard condition is create every cycle.

mikaelarguedas:
Thanks for trying it out, looking at a patch right now to submit a PR soon

PR #112, opened 2017-09-06
add abilitiy to destroy guard conditions and destroy the sigint guard condition in spin_once

mikaelarguedas:
This fixes partially #74

### DIFF
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -37,6 +37,9 @@ class ResponseThread(threading.Thread):
 
         guard_condition_ready_list = \
             _rclpy.rclpy_get_ready_entities('guard_condition', self.wait_set)
+
+        # destroying here to make sure we dont call shutdown before cleaning up
+        _rclpy.rclpy_destroy_entity('guard_condition', sigint_gc)
         if sigint_gc_handle in guard_condition_ready_list:
             rclpy.utilities.shutdown()
             return
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -1434,6 +1434,10 @@ rclpy_destroy_entity(PyObject * Py_UNUSED(self), PyObject * args)
   } else if (0 == strcmp(entity_type, "timer")) {
     rcl_timer_t * timer = (rcl_timer_t *)p;
     ret = rcl_timer_fini(timer);
+  } else if (0 == strcmp(entity_type, "guard_condition")) {
+    rcl_guard_condition_t * guard_condition =
+      (rcl_guard_condition_t *)PyCapsule_GetPointer(pyentity, NULL);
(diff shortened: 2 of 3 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -1434,6 +1434,10 @@ rclpy_destroy_entity(PyObject * Py_UNUSED(self), PyObject * args)
   } else if (0 == strcmp(entity_type, "timer")) {
     rcl_timer_t * timer = (rcl_timer_t *)p;
     ret = rcl_timer_fini(timer);
+  } else if (0 == strcmp(entity_type, "guard_condition")) {
+    rcl_guard_condition_t * guard_condition =
+      (rcl_guard_condition_t *)PyCapsule_GetPointer(pyentity, NULL);
```
dirk-thomas: `PyCapsule_GetPointer(pyentity, NULL)` should be replaced by `p`.

Then this will also fit in one line.

mikaelarguedas: oh I copy pasted before rebasing :S

dirk-thomas:
The remaining memory usage is still related to the `rclpy_get_sigint_guard_condition` call. So this needs some more changes.

rclpy/rclpy/client.py
```diff
@@ -37,6 +37,9 @@ def run(self):
 
         guard_condition_ready_list = \
             _rclpy.rclpy_get_ready_entities('guard_condition', self.wait_set)
+
+        # destroying here to make sure we dont call shutdown before cleaning up
+        _rclpy.rclpy_destroy_entity('guard_condition', sigint_gc)
```
dirk-thomas: Won't this invalidate the variable `sigint_gc_handle` which is used next line?

mikaelarguedas: the pointer / handle is stored in a python int and is compared to a list of already populated int so I don't think that's a problem. But maybe I'm overseeing this

dirk-thomas: I see, while the pointer is invalid the integer value is still the same and can be used for the following comparison. :+1:

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3168/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/508/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2525/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3222/)

leander2189:
Can this merge be backported to humble?

fujitatomoya:
@leander2189 

this PR was merged in 2027, that is a way before humble is introduced. that said, all current distros should have this fix. did you actually find the same problem on humble?

leander2189:
Sorry @fujitatomoya this comment was meant for MR #1308, somehow I ended up here.

Again, sorry for the disturbance.

PR #113, opened 2017-09-06
fix ref count of list items

dirk-thomas:
Before the calling code would still own the list items and therefore their ref count would never go to zero. Preallocating the list and then using `PyList_SET_ITEM` will make sure that the ref count of each item is decremented automatically when being added to the list.

This gives me a constant memory usage for an infinite `spin_once` loop as described in #74 :tada:


dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3167/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/507/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2524/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3221/)

PR #117, opened 2017-09-13
Improve type support error message

mikaelarguedas:
this refactors the type support checks and add a hint as of the reason it is failing

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -32,12 +32,15 @@ from rclpy.validate_topic_name import validate_topic_name
 
 
 def check_for_type_support(msg_type):
-    if msg_type.__class__._TYPE_SUPPORT is None:
-        msg_type.__class__.__import_type_support__()
-    if msg_type.__class__._TYPE_SUPPORT is None:
-        raise NoTypeSupportImportedException(
-            'Possible solution:'
-            'make sure that your ROS2 workspace s before your ROS1 workspace'
+    try:
+        if msg_type.__class__._TYPE_SUPPORT is None:
+            msg_type.__class__.__import_type_support__()
+        if msg_type.__class__._TYPE_SUPPORT is None:
+            raise NoTypeSupportImportedException()
+    except AttributeError:
+        raise AttributeError(
+            "unable to access '_TYPE_SUPPORT' attribute\n"
+            'Make sure that your ROS2 workspace s before your ROS1 workspace'
             ' in your PYTHONPATH.')
 
 
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -35,10 +35,12 @@ def check_for_type_support(msg_type):
     try:
         getattr(msg_type.__class__, '_TYPE_SUPPORT')
     except AttributeError as e:
-        raise AttributeError(
+        e.args = (
+            e.args[0] +
             "unable to access '_TYPE_SUPPORT' attribute\n"
-            'Make sure to source your ROS2 workspace after your ROS1 workspace'
-        )
+            'Make sure to source your ROS2 workspace after your ROS1 workspace',
+            *e.args[1:])
+        raise
     if msg_type.__class__._TYPE_SUPPORT is None:
         msg_type.__class__.__import_type_support__()
     if msg_type.__class__._TYPE_SUPPORT is None:
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -31,6 +31,16 @@ from rclpy.validate_node_name import validate_node_name
 from rclpy.validate_topic_name import validate_topic_name
 
 
+def check_for_type_support(msg_type):
+    if msg_type.__class__._TYPE_SUPPORT is None:
+        msg_type.__class__.__import_type_support__()
+    if msg_type.__class__._TYPE_SUPPORT is None:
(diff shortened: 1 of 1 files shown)


rclpy/rclpy/node.py
```diff
@@ -31,6 +31,19 @@
 from rclpy.validate_topic_name import validate_topic_name
 
 
+def check_for_type_support(msg_type):
+    try:
+        if msg_type.__class__._TYPE_SUPPORT is None:
```
dirk-thomas: Please only `try` / `except` the attribute access and not the invocation of `__import_type_support__`.

rclpy/rclpy/node.py
```diff
@@ -31,6 +31,19 @@
...
+            msg_type.__class__.__import_type_support__()
+    except AttributeError:
+        raise AttributeError(
```
dirk-thomas: By raising a new exception the stacktrace of the previous one is being lost. Instead you could update the raised exception instead.

rclpy/rclpy/node.py
```diff
@@ -31,6 +31,19 @@
...
+            "unable to access '_TYPE_SUPPORT' attribute\n"
+            'Make sure that your ROS2 workspace s before your ROS1 workspace'
```
dirk-thomas: Spelling: `s`

rclpy/rclpy/node.py
```diff
@@ -31,6 +31,19 @@
...
+            ' in your PYTHONPATH.')
```
dirk-thomas: Maybe don't argue about the order in the environment variable but the order of sourcing (ROS 2 after ROS 1)?

rclpy/rclpy/node.py
mikaelarguedas: Yeah is did it because `AttributeError` doesnt have a `msg` attribute that I could extend. I could try to just overwrite its args but looks like args have to be 1char long so it will give a pretty ugly result.

How would you recommend to do it ?

dirk-thomas: `e.args` is a `tuple` with one element which is the message string. You can simply reassign `args` to point to your custom `tuple`. Since `args` might contains more than one element being safe is better than being sorry in the future:

```
e.args = (e.args[0] + ' foo', *e.args[1:])
```

mikaelarguedas: oh yeah totally make sense. Thanks for the explanation! Fixed in https://github.com/ros2/rclpy/pull/117/commits/72db1490b4de31b963d431728fddb5fee4f39fc5

rclpy/rclpy/node.py
```diff
@@ -31,6 +31,22 @@
 from rclpy.validate_topic_name import validate_topic_name
 
 
+def check_for_type_support(msg_type):
+    try:
+        getattr(msg_type.__class__, '_TYPE_SUPPORT')
```
dirk-thomas: I would suggest to keep the syntax and use the accessed attribute (rather than ignore it):

```
ts = msg_type.__class__._TYPE_SUPPORT
```

Later this can be used:

```
if ts is None:
```

rclpy/rclpy/node.py
```diff
@@ -31,6 +31,22 @@
...
+    except AttributeError as e:
+        e.args = (
+            e.args[0] +
+            "unable to access '_TYPE_SUPPORT' attribute\n"
```
dirk-thomas: The information of this line should already be covered by `e.args[0]`?

rclpy/rclpy/node.py
```diff
@@ -31,6 +31,22 @@
...
+            'Make sure to source your ROS2 workspace after your ROS1 workspace',
```
dirk-thomas: Maybe improve the message with:

> This might be a ROS 1 message type but it should be a ROS 2 message type. Make sure to source your ROS 2 workspace after your ROS 1 workspace.

rclpy/rclpy/node.py
mikaelarguedas: How weird I had that before but switched back to `getattr` without assignment for some reason o_O

mikaelarguedas:
Resulting output:
```
$ ros2 run examples_rclpy_minimal_subscriber subscriber_old_school 
  File "/home/mikael/work/ros2/test_sprint_ws/install/lib/examples_rclpy_minimal_subscriber/subscriber_old_school", line 11, in <module>
    load_entry_point('examples-rclpy-minimal-subscriber', 'console_scripts', 'subscriber_old_school')()
  File "/home/mikael/work/ros2/test_sprint_ws/build/examples_rclpy_minimal_subscriber/subscriber_old_school.py", line 30, in main
    subscription = node.create_subscription(String, 'topic', chatter_callback)
  File "/home/mikael/work/ros2/test_sprint_ws/install/lib/python3.5/site-packages/rclpy/node.py", line 123, in create_subscription
    check_for_type_support(msg_type)
  File "/home/mikael/work/ros2/test_sprint_ws/install/lib/python3.5/site-packages/rclpy/node.py", line 36, in check_for_type_support
    ts = msg_type.__class__._TYPE_SUPPORT
AttributeError: type object 'type' has no attribute '_TYPE_SUPPORT' This might be a ROS 1 message type but it should be a ROS 2 message type. Make sure to source your ROS 2 workspace after your ROS 1 workspace.
```

PR #118, opened 2017-09-13
default to c11

mikaelarguedas:
follow up of https://github.com/ros2/ros2/issues/362 revived during review of https://github.com/ros2/rclpy/pull/102#discussion_r137089299

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -2,9 +2,9 @@ cmake_minimum_required(VERSION 3.5)
 
 project(rclpy C)
 
-# Default to C99
+# Default to C11
 if(NOT CMAKE_C_STANDARD)
-  set(CMAKE_C_STANDARD 99)
+  set(CMAKE_C_STANDARD 11)
 endif()
 if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra)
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3231/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/539/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2575/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3293/)

dirk-thomas:
The new compiler warning seems to indicate that we need C11.

mikaelarguedas:
> The new compiler warning seems to indicate that we need C11.

Indeed.
The question is: if we "require" C11 to build without warnings on MacOS, should we lock down the standard version and not allow users to override it ? Or should we keep the conditional leaving control to the users at their own risks?

dirk-thomas:
Imo we should always allow an external override. A user might have a use case to override it - either with an older or newer version. We shouldn't prevent them from doing it (or require them to modify the sources).

mikaelarguedas:
> Imo we should always allow an external override.

Fair enough https://github.com/ros2/rclpy/pull/118/commits/822298d63ab856d6132600527f850db942909e19

Run a macOS job just to confirm that warnings disappear (even if the warnings were pretty explicit...)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2576/)

PR #119, opened 2017-09-13
define docstring as in logging module

mikaelarguedas:
define doc string as in rclpy_logging module (follow up of https://github.com/ros2/rclpy/pull/102#discussion_r137093816)

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -2520,11 +2520,14 @@ static PyMethodDef rclpy_methods[] = {
   {NULL, NULL, 0, NULL}  /* sentinel */
 };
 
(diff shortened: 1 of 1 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -2520,11 +2520,14 @@ static PyMethodDef rclpy_methods[] = {
   {NULL, NULL, 0, NULL}  /* sentinel */
 };
 
+PyDoc_STRVAR(rclpy__doc__,
+  "RCLPY: module for ROS Python client library.");
```
dirk-thomas: The name `rclpy` and the fact that it is a module are already clear by the module name itself. Therefore I would suggest to boil it down to:

> ROS 2 Python client library.

PR #120, opened 2017-09-15
Follow up to executor pull request

sloretz:
Follow up to address comments #104 that didn't make it into r2b3

connects to ros2/examples#184

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -84,6 +84,8 @@ rclpy_get_sigint_guard_condition(PyObject * Py_UNUSED(self), PyObject * Py_UNUSE
  * - a Capsule with the pointer of the created rcl_guard_condition_t * structure
  * - an integer representing the memory address of the rcl_guard_condition_t
  *
+ * \remark Call rclpy_destroy_entity() to destroy a guard condition
+ * \sa rclpy_destroy_entity()
  * \return a list with the capsule and memory location, or
  * \return NULL on failure
  */
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -123,7 +123,7 @@ rclpy_trigger_guard_condition(PyObject * Py_UNUSED(self), PyObject * args)
 {
   PyObject * pygc;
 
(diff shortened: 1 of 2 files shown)


sloretz:
Builds passed and approved, squash/merging

rclpy/src/rclpy/_rclpy.c
```diff
@@ -121,7 +123,7 @@ rclpy_trigger_guard_condition(PyObject * Py_UNUSED(self), PyObject * args)
 {
   PyObject * pygc;
 
-  if (!PyArg_ParseTuple(args, "O", &pygc)) {
+  if (!PyArg_ParseTuple(args, "O", &pygc) || !PyCapsule_CheckExact(pygc)) {
```
mikaelarguedas: @sloretz I think this changed the behavior of the statement:
if PyArg_ParseTuple fails an error will be raised while if PyCapsule_CheckExact fails we return FALSE without raising. Is it expected or something we overlooked ?

sloretz: What was the behavior before? I think this returns false in both cases. The docs don't mention [`"O"` raising an exception, although `"O!"` does](https://docs.python.org/3/c-api/arg.html#other-objects).

mikaelarguedas: O! is only to check Object types, any failure to parse an argument will raise an exception if I read the [function description](https://docs.python.org/3/c-api/arg.html#c.PyArg_ParseTuple) correctly:
"Parse the parameters of a function that takes only positional parameters into local variables. Returns true on success; on failure, it returns false and raises the appropriate exception."

sloretz: Ah oops I missed that. Sounds like should be returning NULL on PyArg_ParseTuple failure and either raising an exception or returning False if the capsule check fails.

mikaelarguedas: I noticed that when preparing a PR to address https://github.com/ros2/rclpy/pull/120#pullrequestreview-63114812. I can fix this in the same PR. Would that conflict with your current work ?

sloretz: It would not conflict. I have a couple additions to this file, but no modifications.

Issue #124, opened 2017-10-10
Use PyCapsule name to check capsule pointer types


sloretz:
#### Feature description
Currently rclpy creates `PyCapsule` with NULL names. This means if a function is given a `PyCapsule` containing `rcl_client_t *` but it expects `rcl_node_t *` the C code will still attempt to use it as such. If the capsules were given a name at creation then the methods could use the name as a type check when getting the pointer.

#### Implementation considerations
https://docs.python.org/3/c-api/capsule.html
`PyCapsule_New` accepts a `const char *name` argument. It could be given a string like `"rcl_node_t *"`. 

```C
// rcl_node_t * node;
PyObject * pynode = PyCapsule_New(node, "rcl_node_t *");
```

 `PyCapsule_GetPointer` accepts a `const char *name` argument, and it sets a python exception and returns `NULL` if the name does not match (it uses `strcmp()` ).

```C
rcl_node_t * node = (rcl_node_t *)PyCapsule_GetPointer(pynode, "rcl_node_t *");
if (!node)
{
  // GetPointer already set an exception
  return NULL;
}
```

If the default exception isn't appropriate, `PyCapsule_GetPointer` is guaranteed to succeed if `PyCapsule_IsValid` returns true. This can be checked first with custom handling code to run if it returns False.

```C
// PyObject * pynode;
if (!PyCapsule_IsValid(pynode, "rcl_node_t *"))
{
  // Do something about it
}
```

mikaelarguedas:
:+1: we discussed it but never got around renaming all the capsules

mikaelarguedas:
that's totally done and working now

PR #125, opened 2017-10-10
Check exact capsules

mikaelarguedas:
The first commit makes sure that we return `NULL` each time `PyArg_ParseTuple` fails and raises an exception.
The second commit uses `PyCapsule_CheckExact` rather than comparing capsules to `Py_None`. It is an attempt to address https://github.com/ros2/rclpy/pull/120#pullrequestreview-63114812 but doesnt reduce the amount of code unfortunately because we still need to extract the pointer (to pass it's value to the entity we create) and then free it... This could be changed to a call to `PyCapsule_IsValid` and adapt the names passed to in when our [capsules get names](https://github.com/ros2/rclpy/issues/124).

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -638,17 +638,15 @@ rclpy_create_publisher(PyObject * Py_UNUSED(self), PyObject * args)
   *publisher = rcl_get_zero_initialized_publisher();
   rcl_publisher_options_t publisher_ops = rcl_publisher_get_default_options();
 
-  if (!PyCapsule_CheckExact(pyqos_profile)) {
-    PyErr_Format(PyExc_RuntimeError, "Invalid pyqos_profile capsule in rclpy_create_publisher");
-    return NULL;
-  }
-  void * p = PyCapsule_GetPointer(pyqos_profile, NULL);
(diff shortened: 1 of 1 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -636,15 +638,17 @@ rclpy_create_publisher(PyObject * Py_UNUSED(self), PyObject * args)
   *publisher = rcl_get_zero_initialized_publisher();
   rcl_publisher_options_t publisher_ops = rcl_publisher_get_default_options();
 
+  if (!PyCapsule_CheckExact(pyqos_profile)) {
+    PyErr_Format(PyExc_RuntimeError, "Invalid pyqos_profile capsule in rclpy_create_publisher");
+    return NULL;
+  }
```
dirk-thomas: Doesn't this change semantic? Before it was possible to not pass a profile.

Same below.

mikaelarguedas: Oops I thought it was a bug not a feature XD. I'll restore the previous behavior then. Note that we do also ensure a qos_profile is passed to these functions at the rclpy Python level, but I agree it's better to keep it flexible at the C-level as well

PR #129, opened 2017-10-18
Use PyCapsule names

sloretz:
Naming capsules prevents calling the wrong kind of cleanup on a pointer. This also removes the string parameter from `rclpy_destroy_entity` and `rclpy_destroy_node_entity` since it can be figured out from the capsule name.

If a non-capsule or a capsule of the wrong type is specified, the functions let `PyCapsule_GetPointer` raise.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3381/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/630/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2714/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3432/)

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -248,8 +248,7 @@ class Node:
     def destroy_guard_condition(self, guard):
         for gc in self.guards:
             if gc.guard_handle == guard.guard_handle:
-                _rclpy.rclpy_destroy_entity(
-                    'guard_condition', gc.guard_handle)
+                _rclpy.rclpy_destroy_entity(gc.guard_handle)
                 self.guards.remove(gc)
                 return True
         return False
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -268,7 +268,7 @@ class Node:
             _rclpy.rclpy_destroy_node_entity(srv.service_handle, self.handle)
         for tmr in self.timers:
             _rclpy.rclpy_destroy_entity(tmr.timer_handle)
-        for gc in list(self.guards):
+        for gc in self.guards:
             _rclpy.rclpy_destroy_entity(gc.guard_handle)
 
         self.timers = []
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -208,8 +208,7 @@ class Node:
     def destroy_publisher(self, publisher):
         for pub in self.publishers:
             if pub.publisher_handle == publisher.publisher_handle:
-                _rclpy.rclpy_destroy_node_entity(
-                    'publisher', pub.publisher_handle, self.handle)
+                _rclpy.rclpy_destroy_node_entity(pub.publisher_handle, self.handle)
                 self.publishers.remove(pub)
                 return True
         return False
@@ -217,8 +216,7 @@ class Node:
     def destroy_subscription(self, subscription):
         for sub in self.subscriptions:
             if sub.subscription_handle == subscription.subscription_handle:
-                _rclpy.rclpy_destroy_node_entity(
-                    'subscription', sub.subscription_handle, self.handle)
+                _rclpy.rclpy_destroy_node_entity(sub.subscription_handle, self.handle)
                 self.subscriptions.remove(sub)
                 return True
         return False
@@ -226,8 +224,7 @@ class Node:
     def destroy_service(self, service):
         for srv in self.services:
             if srv.service_handle == service.service_handle:
-                _rclpy.rclpy_destroy_node_entity(
-                    'service', srv.service_handle, self.handle)
+                _rclpy.rclpy_destroy_node_entity(srv.service_handle, self.handle)
                 self.services.remove(srv)
                 return True
         return False
@@ -235,8 +232,7 @@ class Node:
     def destroy_client(self, client):
         for cli in self.clients:
             if cli.client_handle == client.client_handle:
-                _rclpy.rclpy_destroy_node_entity(
-                    'client', cli.client_handle, self.handle)
+                _rclpy.rclpy_destroy_node_entity(cli.client_handle, self.handle)
                 self.clients.remove(cli)
                 return True
         return False
@@ -244,8 +240,7 @@ class Node:
     def destroy_timer(self, timer):
         for tmr in self.timers:
             if tmr.timer_handle == timer.timer_handle:
-                _rclpy.rclpy_destroy_entity(
-                    'timer', tmr.timer_handle)
+                _rclpy.rclpy_destroy_entity(tmr.timer_handle)
                 self.timers.remove(tmr)
                 return True
         return False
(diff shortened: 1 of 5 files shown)


rclpy/rclpy/node.py
```diff
@@ -244,44 +239,25 @@ def destroy_node(self):
         if self.handle is None:
             return ret
 
-        # ensure that the passed node contains a valid capsule
-        class_ = self.handle.__class__
-        if not class_ or class_.__name__ != 'PyCapsule':
-            raise ValueError('The node handle must be a PyCapsule')
+        for sub in self.subscriptions:
+            _rclpy.rclpy_destroy_node_entity(sub.subscription_handle, self.handle)
```
dirk-thomas: The return value is not being considered anymore.

sloretz: This is intentional. Currently `rclpy_destroy_node_entity` returns True on success or raises a RuntimeError on failure, so I got rid of the return value altogether.

dirk-thomas: I see. If the return value of the function doesn't have any purpose anymore I would suggest to return `None` instead (like any Python function without a return statement).

mikaelarguedas: I believe that is what this PR is doing, raising and returning NULL on failure otherwise returning None:
https://github.com/ros2/rclpy/pull/129/files#diff-80dda03110b5be606b823d6b0558b5c2R1567

dirk-thomas: Sorry for the noise. I don't know what I was looking at.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -95,28 +97,31 @@ rclpy_create_guard_condition(PyObject * Py_UNUSED(self), PyObject * Py_UNUSED(ar
...
+  PyList_SET_ITEM(pylist, 0, pygc);
   PyList_SET_ITEM(pylist, 1, PyLong_FromUnsignedLongLong((uint64_t)&gc->impl));
 
   return pylist;
 }
 
 /// Trigger a general purpose guard condition
 /**
+ * Raises ValueError if the guard condition is not a capsule of the correcttype
+ * Raises RuntimeError if the guard condition could not be triggered
```
mikaelarguedas: Several other places do raise runtime error on this type of failure, maybe we should take advantage of this PR to update them all

rclpy/src/rclpy/_rclpy.c
```diff
@@ -597,6 +613,7 @@ rclpy_expand_topic_name(PyObject * Py_UNUSED(self), PyObject * args)
  * provided as pymsg_type to send messages over the wire.
  *
  * A ValueError is raised (and NULL returned) when the topic name is invalid.
+ * Raises ValueError if the capsules are not the correct types
```
mikaelarguedas: Can you rephrase the sentence above to match this one as well?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -772,6 +799,8 @@ rclpy_create_timer(PyObject * Py_UNUSED(self), PyObject * args)
 
 /// Returns the period of the timer in nanoseconds
 /**
+ * Raise ValueError if capsule is not a timer capsule
```
mikaelarguedas: typo: Raise**s** for consistency with other docblocks added in this PR
Can you also move the `Raise RuntimeError on rcl error` here? (same for the other functions)

rclpy/src/rclpy/_rclpy.c
sloretz: I updated all the ones I found in 451a072e372eecc46e07096f8fdcb2278051bd54

rclpy/src/rclpy/_rclpy.c
sloretz: Done 451a072e372eecc46e07096f8fdcb2278051bd54

rclpy/src/rclpy/_rclpy.c
sloretz: Fixed 451a072e372eecc46e07096f8fdcb2278051bd54

rclpy/rclpy/node.py
```diff
@@ -264,49 +258,28 @@ def destroy_node(self):
...
-        destroyed = _rclpy.rclpy_destroy_entity('node', self.handle)
-        if destroyed:
-            self._handle = None
-        ret &= destroyed
+        self.timers = []
+        self.subscriptions = []
+        self.services = []
+        self.clients = []
+        self.publishers = []
+        self.guards = []
```
dirk-thomas: Nitpick: maybe use the same order as above for the loops? That allows easier comparison that the same variables are used in both.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -126,11 +135,12 @@ rclpy_trigger_guard_condition(PyObject * Py_UNUSED(self), PyObject * args)
...
   }
-  if (!PyCapsule_CheckExact(pygc)) {
-    Py_RETURN_FALSE;
-  }
 
-  rcl_guard_condition_t * gc = (rcl_guard_condition_t *)PyCapsule_GetPointer(pygc, NULL);
+  rcl_guard_condition_t * gc = (rcl_guard_condition_t *)PyCapsule_GetPointer(
+    pygc, "rcl_guard_condition_t");
+  if (!gc) {
+    return NULL;
```
dirk-thomas: What is the expectation here? It should only return `NULL` when an error is set, right? But `PyCapsule_GetPointer` shouldn't set an error. Should a specific error message be set or should this return `Py_RETURN_NONE` instead (which would be a silent failure, probably not?)?

Same below for various other pointers.

mikaelarguedas: `PyCapsule_GetPointer` will raise an exception if the Capsule is not a valid capsule with a matching name. I think the assumption here is that the only reason for `gc` to be `NULL` is if `PyCapsule_GetPointer` fails and thus an exception is already being set so we need to return `NULL`.

dirk-thomas: You are right, it [does](https://docs.python.org/3.5/c-api/capsule.html#c.PyCapsule_GetPointer) set an error when returning `NULL`. Sorry for the noise - should have checked the docs before.

rclpy/rclpy/node.py
sloretz: pub, sub, cli, srv, tmr, gc 1561483d1d16434c7c595a3bd8d026df7da2d480

mikaelarguedas:
I found a few bugs in this PR, please hold before merging

sloretz:
@dirk-thomas maybe those two places aren't covered by CI? Is it it possible no test uses a qos profile on a subscription or calls `rclpy_get_service_names_and_types`?

sloretz:
@mikaelarguedas CI in progress

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3388/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/634/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2718/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3436/)

sloretz:
CI green. squash/merging.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1332,15 +1437,14 @@ rclpy_create_service(PyObject * Py_UNUSED(self), PyObject * args)
...
         service_name, rcl_get_error_string_safe());
-      rcl_reset_error();
-      return NULL;
+    } else {
+      PyErr_Format(PyExc_RuntimeError,
+        "Failed to create service: %s", rcl_get_error_string_safe());
     }
-    PyErr_Format(PyExc_RuntimeError,
-      "Failed to create service: %s", rcl_get_error_string_safe());
+    Py_DECREF(service);
```
dirk-thomas: This line seems to be the reason for this new test failure: http://ci.ros2.org/view/nightly/job/nightly_win_deb/lastCompletedBuild/testReport/(root)/projectroot/rclpytests/

PR #130, opened 2017-10-19
Updates for global severity threshold -> default severity threshold

dhood:
connects to ros2/rcutils#57

This is the minimum PR required to keep rclpy working with the changes in https://github.com/ros2/rcutils/pull/57

It does not attempt to leverage any of the configuration of loggers (that'll be a separate PR)

### DIFF
--- a/rclpy/rclpy/logging.py
+++ b/rclpy/rclpy/logging.py
@@ -25,11 +25,11 @@ class LoggingSeverity(IntEnum):
     This enum must match the one defined in rcutils/logging.h
     """
 
-    DEBUG = 0
-    INFO = 1
-    WARN = 2
-    ERROR = 3
-    FATAL = 4
+    DEBUG = 10
+    INFO = 20
+    WARN = 30
+    ERROR = 40
+    FATAL = 50
 
 
 def get_named_logger(name):
--- a/rclpy/src/rclpy/_rclpy_logging.c
+++ b/rclpy/src/rclpy/_rclpy_logging.c
@@ -29,19 +29,19 @@ rclpy_logging_initialize(PyObject * Py_UNUSED(self), PyObject * Py_UNUSED(args))
   Py_RETURN_NONE;
 }
 
-/// Get the global severity threshold of the logging system.
+/// Get the default severity threshold of the logging system.
 /**
  * \return severity
  */
 static PyObject *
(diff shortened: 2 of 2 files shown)


rclpy/src/rclpy/_rclpy_logging.c
```diff
@@ -29,19 +29,19 @@ rclpy_logging_initialize(PyObject * Py_UNUSED(self), PyObject * Py_UNUSED(args))
   Py_RETURN_NONE;
 }
 
-/// Get the global severity threshold of the logging system.
+/// Get the default severity threshold of the logging system.
 /**
  * \return severity
  */
 static PyObject *
 rclpy_logging_get_severity_threshold(PyObject * Py_UNUSED(self), PyObject * Py_UNUSED(args))
```
dirk-thomas: Should the name of this function be changed accordingly?

rclpy/rclpy/logging.py
```diff
@@ -25,11 +25,11 @@ class LoggingSeverity(IntEnum):
...
-    DEBUG = 0
-    INFO = 1
-    WARN = 2
-    ERROR = 3
-    FATAL = 4
+    DEBUG = 10
+    INFO = 20
+    WARN = 30
+    ERROR = 40
+    FATAL = 50
```
dirk-thomas: Should we actually get the values from `rcutils/logging.h` rather than redefining them here?

dhood: yes, we should, eventually. in my mind that would be done by moving the values into the `rcutils.logging` python module so we can include them here, and empy template them into rcutils' logging.h. Since there's a lot going on in https://github.com/ros2/rcutils/pull/57 already, I think it's best to put it off to a separate step

rclpy/src/rclpy/_rclpy_logging.c
dhood: yes, there's a conceptual adaptation of the rclpy use of rcutils logging coming in another rclpy PR, but this PR is just the bare minimum to keep it compiling if https://github.com/ros2/rcutils/pull/57 gets merged (so the PRs don't block each other)

dirk-thomas: Sounds good to me.

rclpy/rclpy/logging.py
dirk-thomas: Sounds good to me.

dhood:
closed in favour of https://github.com/ros2/rclpy/pull/132

PR #132, opened 2017-10-20
Named logger config + logger hierarchy

dhood:
connects to ros2/rcutils#57

currently including commits from https://github.com/ros2/rclpy/pull/130; if this PR is ready before https://github.com/ros2/rcutils/pull/57 gets merged then it can replace #130.

This PR exposes the new notions from https://github.com/ros2/rcutils/pull/57:
1. ~there's a default severity threshold,~
1. each logger can have its own severity threshold specified,
1. there's a hierarchy of loggers: I've added a get_child() method on loggers; children inherit the severity threshold of their parent if set (the severity threshold inheritance taken care of by rcutils now that it has a notion of logger hierarchy)
1. all loggers at some point are attached to the root rclpy logger (which is where they get their default severity threshold)

Out of scope:
1. node.get_logger() or any notion of attaching loggers to nodes (waiting on decision in https://github.com/ros2/examples/pull/186)
1. any caching of whether or not loggers are enabled and updating based on notifications when severity levels change

### DIFF
--- a/rclpy/rclpy/impl/implementation_singleton.py
+++ b/rclpy/rclpy/impl/implementation_singleton.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 """
-Provide singleton access to the rclpy C module.
+Provide singleton access to the rclpy C modules.
 
 For example, you might use it like this:
 
@@ -38,3 +38,4 @@ except ImportError as e:
             " Please refer to '%s' for possible solutions" % \
             (e.path, 'https://github.com/ros2/ros2/wiki/Rclpy-Import-error-hint')
     raise
+rclpy_logging_implementation = importlib.import_module('._rclpy_logging', package='rclpy')
--- a/rclpy/rclpy/impl/implementation_singleton.py
+++ b/rclpy/rclpy/impl/implementation_singleton.py
@@ -31,6 +31,7 @@ import os
 
 try:
     rclpy_implementation = importlib.import_module('._rclpy', package='rclpy')
+    rclpy_logging_implementation = importlib.import_module('._rclpy_logging', package='rclpy')
 except ImportError as e:
     if os.path.isfile(e.path):
         e.msg += \
@@ -38,4 +39,3 @@ except ImportError as e:
             " Please refer to '%s' for possible solutions" % \
             (e.path, 'https://github.com/ros2/ros2/wiki/Rclpy-Import-error-hint')
     raise
-rclpy_logging_implementation = importlib.import_module('._rclpy_logging', package='rclpy')
--- a/rclpy/rclpy/impl/rcutils_logger.py
+++ b/rclpy/rclpy/impl/rcutils_logger.py
@@ -221,6 +221,8 @@ class RcutilsLogger:
         self.contexts = {}
 
     def get_child(self, name):
+        if not name or len(name) == 0:
+            raise ValueError('Child logger name should not be empty.')
         return RcutilsLogger(name=self.name + '.' + name)
 
     def get_severity_threshold(self):
--- a/rclpy/rclpy/impl/rcutils_logger.py
+++ b/rclpy/rclpy/impl/rcutils_logger.py
@@ -15,12 +15,11 @@
 
 from collections import namedtuple
 from collections import OrderedDict
-import importlib
 import inspect
 import os
 import time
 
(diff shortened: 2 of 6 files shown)


rclpy/rclpy/impl/implementation_singleton.py
```diff
@@ -38,3 +38,4 @@
             " Please refer to '%s' for possible solutions" % \
             (e.path, 'https://github.com/ros2/ros2/wiki/Rclpy-Import-error-hint')
     raise
+rclpy_logging_implementation = importlib.import_module('._rclpy_logging', package='rclpy')
```
dirk-thomas: Should this have the same `ImportError` check as the other library?

rclpy/rclpy/impl/rcutils_logger.py
```diff
@@ -221,15 +220,30 @@ def __init__(self, name=''):
         self.name = name
         self.contexts = {}
 
+    def get_child(self, name):
+        return RcutilsLogger(name=self.name + '.' + name)
```
dirk-thomas: Does this need to handle the case where `self.name` is an empty string?

rclpy/rclpy/impl/implementation_singleton.py
dhood: yes, I wasn't thinking of the cases where _rclpy imports fine but _rclpy_logging has issues. 2ad25a8b68520f7d97c822633bedff1e05e525f6

rclpy/rclpy/impl/rcutils_logger.py
dhood: Checks added in 52a36611bbf91c92f4a79623b55b83e59b07344c

dirk-thomas: I was aiming for `self.name` being empty which results in the new loggers name to be ".<name>".

dhood: It doesn't _need_ to but we could; the rcutils logger hierarchy will still be functional in the case of empty name tokens. I was taking the approach of allowing `''.child_name` to be a valid logger name if a user wants to do so (since they're calling `get_child` as opposed to `get_named_logger`).

If instead we remove the prefixing dot then we have returned an independent sibling instead of a child; in your opinion is that the expected behaviour for the edge case, or was your question just if it would lead to undefined behaviour?

dirk-thomas: The question is how is the topology of the loggers? Does `foo` and `bar` not share a common parent in your case? If not, why not? If yes, what would that be if not ``?

I though of the loggers as a hierarchy in the form of a tree with a single root. `` being the root of it and `"anything_without_a_dot"` are the direct children of it.

dhood: Ah, I see where the difference in understanding has come from now: in my mind there didn't need to be a single root and you could create a logger that is not attached to the root if you wanted to.

The idea of a single root matches how log4cxx and python logging module do it, so I'll switch to that. Every logger will at some point be attached to the root logger, but the root logger won't have a name. And then, yes, the dot prefix will be removed when appending a direct child to the root logger.

I'll update rcutils to support this (there isn't really a notion of a root logger in https://github.com/ros2/rcutils/pull/57, but "default" settings).

dhood: Does this better match what you had in mind? 21e5c4438ce28cc004f3572d83173597b6efbb1c
Now a call to `get_named_logger` just calls `get_child` on the root logger: all loggers are descendants of the root logger.

4c8bdbae96007e481ced2cf014cc74abcab19878 removes the notion of default configuration in favour of configuring the root logger.

dirk-thomas: :+1: looks good to me.

dhood:
This is ready for review, it brings in a the notion of logger hierarchy but doesn't act on any not-yet-made design conclusions about what those loggers will be named in userland/how loggers are associated with nodes (e.g. the name of loggers attached to nodes, etc).

The only design-like decision that is being made in this PR is that the root logger is nameless, but we just need to change the line that creates the root logger if we decide that a specific name is more appropriate, and then it will propagate to all created loggers.

dhood:
CI for this PR is in https://github.com/ros2/rcutils/pull/57#issuecomment-343257012

PR #133, opened 2017-10-25
Rclpy named capsules fixups

mikaelarguedas:
@sloretz I made a few fixups in the modified docblocks of #129 but didnt want to push directly on your branch. Feel free to cherry pick/apply what you think makes sense.


sloretz:
It all looks good to me. Thanks for the doc fixes.

In the future feel free to push to the branch directly. Just send a heads up so I don't accidentally overwrite them if I'm about to rebase onto master and force push.

PR #134, opened 2017-10-26
protect against ImportError.path being None

wjwwood:
Error looks something like:

```
  File "C:\path\to\ws\install\Lib\site-packages\rclpy\impl\implementation_singleton.py", line 35, in <module>
    if os.path.isfile(e.path):
  File "C:\Python36\lib\genericpath.py", line 30, in isfile
    st = os.stat(path)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

### DIFF
--- a/rclpy/rclpy/impl/implementation_singleton.py
+++ b/rclpy/rclpy/impl/implementation_singleton.py
@@ -32,7 +32,7 @@ import os
 try:
     rclpy_implementation = importlib.import_module('._rclpy', package='rclpy')
 except ImportError as e:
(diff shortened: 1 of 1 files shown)


dhood:
for future reference can you advise under what circumstances this error occurs?

wjwwood:
@dhood If you run the script with the non-debug version of Python but only the debug version of the c extension exists (i.e. you use `python3` rather than `python3_d` but only `rclpy/_rclpy_d.pyd` exists).

dhood:
OK good to know

Issue #123, opened 2017-10-09
Services feature parity


mikaelarguedas:
#### Feature description

Meta ticket to track improvements we want to make to services and service-related features. This is not an exhaustive list, feel free to complete it

- create a "graph_listener" equivalent to use for `wait_for_service` (related to #58)
- refactor client to store pending requests
- refactor clients to return a future for each request sent and provide an equivalent of `rclcpp::spin_until_future_complete()` at the executor level

connects to ros2/rclpy#170

sloretz:
I see  `Executor::spin_until_future_complete()`, but not `rclcpp::spin_until_future_complete()`. How is it meant to be used?

Using it in the main thread outside of an executor looks fine.

```python
future = client.call(req)
executor.spin_until_future_complete(future)
if future.get() is not None:
    print('Doing something with result %r', future.get())
```

Used in a callback done by a single threaded executor would be ok, except from the perspective of the main thread a `spin_once()` call might handle multiple callbacks. From a multi threaded executor it would cause `rcl_wait()` to be called on the same wait_set in multiple threads.

```python
# main thread
while rclpy.ok()
    # OK, though spin_once may handle more than 1 callback since the callback can spin
    singleThreadedExecutor.spin_once()
    # Not OK, this will dispatch the callback and call rcl_wait again,
    # but the callback will also call spin which will call rcl_wait on the
    # same wait set in another thread. rcl_wait() warns it is not thread safe.
    # multiThreadedExecutor.spin_once()

# ...

# Subscriber callback
    def listener_cb(self, response):
        future = self.client.call(req)
       # Executor has to be given to this object somehow
        self.executor.spin_until_future_complete(future)
        if future.get() is not None:
            print('Doing something with result %r', future.get())
```

sloretz:
Thoughts/notes regarding Future objects:

* Every Future needs a Task to set the result of the Future
    * These need to be added to the Executor so it can schedule them
        * This seems to be the reason but concurrent.futures and asyncio recommend using the executor to 
create futures
    * `client.call(req)` could returns a future and keep a list of tasks. Executor would pull those tasks off the client object
* concurrent.futures.Future is thread safe but not awaitable
* asyncio.Future is awaitable but not thread safe
* ...

PR #135, opened 2017-10-30
[WIP] Futures/coroutines for services

sloretz:
**Note:Ignore this PR. A lot of it has already been split into separate pull requests. I'll close this when there is nothing left to split.**

This is part of service feature parity.  It is not ready for review, but feedback is welcome if you can tolerate the chaotic writing below. Once #127 is merged this will need to be rebased onto master (omitting d980051).

Futures are a way to support multiple simultaneous client requests. Coroutines are a way to support futures inside of callbacks in an executor. Futures need a task running in parallel to update the result. A task can get to the executor via
 * a callback in the executor calling `await future`
    * `Future.__await__` can invoke the task, and yield if the result is not ready yet.
    * Callback itself must be a coroutine (i.e. `async def`)
* code outside giving the future to the executor directly
    * Add an equivalent of `rclcpp::Executor::spin_until_future_complete(future)`
        * In asyncio everything has a reference to the event loop so adding a task is easy, but not everything has a reference to the executor in rclpy. One possibility is a future having a reference to a task that needs to be executed to complete it. `spin_until_future_complete` could add that task to the executor.

Current state:
 * Added executor support for coroutines
    * `async def`/`await` only. Does not support generator/`yield from` coroutines (but it could)
    * executor `wait_for_ready_callbacks` yields new callbacks first, then old coroutines that yielded the last time they were run.
        * Requires an executor to run the iterator returned from `wait_for_ready_callbacks` until `StopIteration` or old coroutines may never get resumed.
        * Maybe for some cases it makes sense to interleave new tasks with old tasks? I think `wait_for_ready_callbacks` should only `yield` new tasks, and another method should yield old ones. Custom executors should be responsible for implementing their own old-vs-new task strategy.
    * Service callbacks cannot be coroutines. (not impossible; haven't figured out a clean implementation).
 * Executor yields a `Task` instance
    * `Task` class hides the difference between a normal function and a coroutine from an executor
    * `Task` done callbacks get executed in series after the original callback/coroutine finishes
        * `Task` is greedy and executes as many done callbacks as it can, pausing only if one is a coroutine that yields.
        * `asyncio.Task` done callbacks get scheduled in the event loop, but this would be a problem in rclpy because it could mean another `rcl_wait` before executing a done callback.
    * Not using asyncio.Task because it is not thread safe, and it expects a reference to an asyncio event loop
 * Does not yet have a Future class
    * `asyncio.Future` is `await`-able but not thread safe and expects an event loop reference (though partially works without it)
    * `concurrent.futures.Future` is thread safe but not `await`-able (Maybe could be subclassed for this)
    * In `asyncio.Task` is a subclass of future, so maybe a `Task` could be it's own future too?

connects to ros2/rclpy#123


sloretz:
Closing since it was split with the wait set refactor in #140 and the future/coroutine support in #166.

PR #137, opened 2017-10-31
Fix incorrect reference to service instead of pyservice

### LINKED ISSUE #136: Negative ref count in test_node.py on windows debug
`test_create_service` has been giving errors sometimes on windows debug nightlies lately (first time Oct 28):

```
Fatal Python error: C:\J\workspace\nightly_win_deb\ws\src\ros2\rclpy\rclpy\src\rclpy\_rclpy.c:1444 object at 000001F7B40882C0 has negative ref count -7870023856155525121
ValueError: Failed to create service due to invalid topic name '/get/42parameters': topic name is invalid, at C:\J\workspace\nightly_win_deb\ws\src\ros2\rcl\rcl\src\rcl\expand_topic_name.c:67

Current thread 0x00000a38 (most recent call first):
  File "C:/J/workspace/nightly_win_deb/ws/src/ros2/rclpy/rclpy\rclpy\node.py", line 177 in create_service
  File "C:\J\workspace\nightly_win_deb\ws\src\ros2\rclpy\rclpy\test\test_node.py", line 75 in test_create_service
```

http://ci.ros2.org/job/nightly_win_deb/659/testReport/junit/(root)/projectroot/rclpytests/

It seems to be unhappy with this `Py_DECREF` upon service init failure that was added in https://github.com/ros2/rclpy/pull/129:
https://github.com/ros2/rclpy/pull/129/files#diff-80dda03110b5be606b823d6b0558b5c2R1444

Looking closer it seems to just be accidentally referencing `service` instead of `pyservice` so I'll fix it up now (still opening this ticket for future reference and visibility).

dhood:
Fixes #136 

Standard CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3403/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/648/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2733/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3450/)

Windows Debug retest-until-fail 10 on rclpy: [Build Status](http://ci.ros2.org/job/ci_windows/3451/)

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -1441,7 +1441,7 @@ rclpy_create_service(PyObject * Py_UNUSED(self), PyObject * args)
       PyErr_Format(PyExc_RuntimeError,
         "Failed to create service: %s", rcl_get_error_string_safe());
     }
-    Py_DECREF(service);
+    Py_DECREF(pyservice);
     rcl_reset_error();
     return NULL;
   }


rclpy/src/rclpy/_rclpy.c
```diff
@@ -1441,7 +1441,7 @@ rclpy_create_service(PyObject * Py_UNUSED(self), PyObject * args)
       PyErr_Format(PyExc_RuntimeError,
         "Failed to create service: %s", rcl_get_error_string_safe());
     }
-    Py_DECREF(service);
+    Py_DECREF(pyservice);
```
dirk-thomas: Is this also necessary before returning in line 1430?

Does `service` need to be deallocated?

mikaelarguedas: I think this should be a follow up PR. We should check that every creation that has been moved up a few blocks is properly deallocated in case of failure

sloretz: Yes to both. For the first one the `PyCapsule_New` line could be moved after 1430.

mikaelarguedas: So do we agree this can be merged as is ?
@sloretz would you mind submitting a separate PR to address the remaining items ?

sloretz: @mikaelarguedas WIll do

sloretz: #138

PR #138, opened 2017-11-01
Free memory when things don't go as planned

sloretz:
Follow up to comments on #137 

* Reverted moving of the PyCapsule creation to eliminate some `Py_DECREF()`
* Added `PyMem_Free` calls to failure cases (sometimes moving code down to reduce the number of places where this is called)

CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3417/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/664/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2744/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3461/)

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -62,20 +62,19 @@ rclpy_get_sigint_guard_condition(PyObject * Py_UNUSED(self), PyObject * Py_UNUSE
   rcl_guard_condition_t * sigint_gc =
     (rcl_guard_condition_t *)PyMem_Malloc(sizeof(rcl_guard_condition_t));
   *sigint_gc = rcl_get_zero_initialized_guard_condition();
-  PyObject * pygc = PyCapsule_New(sigint_gc, "rcl_guard_condition_t", NULL);
   rcl_guard_condition_options_t sigint_gc_options = rcl_guard_condition_get_default_options();
 
   rcl_ret_t ret = rcl_guard_condition_init(sigint_gc, sigint_gc_options);
   if (ret != RCL_RET_OK) {
     PyErr_Format(PyExc_RuntimeError,
       "Failed to create guard_condition: %s", rcl_get_error_string_safe());
-    Py_DECREF(pygc);
     rcl_reset_error();
+    PyMem_Free(sigint_gc);
     return NULL;
   }
   g_sigint_gc_handle = sigint_gc;
   PyObject * pylist = PyList_New(2);
-  PyList_SET_ITEM(pylist, 0, pygc);
+  PyList_SET_ITEM(pylist, 0, PyCapsule_New(sigint_gc, "rcl_guard_condition_t", NULL));
   PyList_SET_ITEM(pylist, 1, PyLong_FromUnsignedLongLong((uint64_t)&sigint_gc->impl));
 
   return pylist;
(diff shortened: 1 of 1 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -62,20 +62,19 @@ rclpy_get_sigint_guard_condition(PyObject * Py_UNUSED(self), PyObject * Py_UNUSE
...
       "Failed to create guard_condition: %s", rcl_get_error_string_safe());
-    Py_DECREF(pygc);
     rcl_reset_error();
+    PyMem_Free(sigint_gc);
     return NULL;
   }
   g_sigint_gc_handle = sigint_gc;
   PyObject * pylist = PyList_New(2);
-  PyList_SET_ITEM(pylist, 0, pygc);
+  PyList_SET_ITEM(pylist, 0, PyCapsule_New(sigint_gc, "rcl_guard_condition_t", NULL));
```
dirk-thomas: The return value of `PyCapsule_New` needs to be checked.

Same below.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -62,20 +62,19 @@ rclpy_get_sigint_guard_condition(PyObject * Py_UNUSED(self), PyObject * Py_UNUSE
...
   if (ret != RCL_RET_OK) {
     PyErr_Format(PyExc_RuntimeError,
       "Failed to create guard_condition: %s", rcl_get_error_string_safe());
-    Py_DECREF(pygc);
     rcl_reset_error();
+    PyMem_Free(sigint_gc);
     return NULL;
   }
   g_sigint_gc_handle = sigint_gc;
   PyObject * pylist = PyList_New(2);
```
clalancette: FWIW, the Python [documentation](https://docs.python.org/3.6/c-api/list.html#c.PyList_New) for `PyList_New` says:

```
Return a new list of length len on success, or NULL on failure.
```

Which means we should check the return value of `PyList_New` as well (throughout this file).  I'm not sure if we want to do that as part of this PR or a different one, though.

sloretz: I'd rather do this in a few smaller PRs. I double checked all the functions in `_rclpy.c`; these are being used and can fail but are not being checked.

`PyList_New`
`PyCapsule_New` FYI @dirk-thomas 
`PyLong_FromUnsignedLongLong`
`PyMem_Malloc`
`PyUnicode_FromString` (not in web documentation but [source code says otherwise](https://github.com/python/cpython/blob/4f469c096628af730b17798d0ebfd8925bfde836/Objects/unicodeobject.c#L2101))
`PyLong_FromSize_t`
`PyObject_GetAtterString`
`PyLong_FromLongLong`
`PyCapsule_GetName`
`PyList_Append`
`PyObject_CallObject`
`PyTuple_New`
`PyTuple_SetItem`
`PyList_SetItem`
`PyImport_ImportModule`
`PyObject_SetAttrString`

clalancette: Ah, good, you were a lot more comprehensive than I was.  It's fine by me to do it in separate PRs; I'd suggest a tracking bug since there are so many of them.  I'll leave it up to you, though.

sloretz: Tracking bug #139

rclpy/src/rclpy/_rclpy.c
sloretz: I added this to #139. If it is alright with you I would like to tackle this in a separate PR.

PR #141, opened 2017-11-13
Reset the error when rcl_take_response failed.

minggangw:
[edit dhood]: connect to ros2/rclpy#142


dhood:
Thank you for the patch @minggangw. Actually `rcl_take_response` can return `RCL_RET_CLIENT_TAKE_FAILED` if there was nothing to take, and we don't consider this an error state. Because of that I will close this PR as with https://github.com/ros2/rclcpp/pull/396

Looking at this code, however, it seems that this section is missing an equivalent of https://github.com/ros2/rclcpp/pull/373/files in order to check the correct error message. If you have been seeing an error output when you didn't expect it, it may have been caused by that bug. The fix has been proposed in https://github.com/ros2/rclpy/pull/143

PR #144, opened 2017-11-13
 Relax the tolerance on throttle timing

dhood:
Too strict on ARM debug repeated: http://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/267/

Repeating rclpy tests (manually excluding the flaky `test_timer.py`) 50 times on ARM debug: [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/736/)

### DIFF
--- a/rclpy/test/test_logging.py
+++ b/rclpy/test/test_logging.py
@@ -104,8 +104,8 @@ class TestLogging(unittest.TestCase):
                 throttle_duration_sec=1,
                 throttle_time_source_type='RCUTILS_STEADY_TIME',
             ))
-            time.sleep(0.3)
-        self.assertEqual(message_was_logged, [True] + [False] * 3 + [True])
+            time.sleep(0.4)
+        self.assertEqual(message_was_logged, [True, False, False, True, False])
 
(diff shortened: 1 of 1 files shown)


rclpy/test/test_logging.py
```diff
@@ -104,8 +104,8 @@ def test_log_throttle(self):
                 throttle_duration_sec=1,
                 throttle_time_source_type='RCUTILS_STEADY_TIME',
             ))
-            time.sleep(0.3)
-        self.assertEqual(message_was_logged, [True] + [False] * 3 + [True])
+            time.sleep(0.4)
```
dirk-thomas: With the new sleep value one of the wake up cases is "by number" exactly happening at the same time as the throttle duration (5 * 0.4 = 2.0). I though the idea of the previous value was that this is not the case?

dhood: the last call will be at t=1.6s

dirk-thomas: I was thinking about if ever one more call is made it might get hairy with the timing. For now that should be good enough, I agree.

PR #146, opened 2017-11-14
Relax the tolerance on throttle + skipfirst timing

dhood:
Timing was too strict on ARM: http://ci.ros2.org/job/ci_linux-aarch64/740/testReport/junit/rclpy.src.ros2.rclpy.rclpy.test.test_logging/TestLogging/test_log_skip_first_throttle/

followup of https://github.com/ros2/rclpy/pull/144

also added comments to explain the expected output of the throttle tests

CI ARM (not repeated): [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/743/)

### DIFF
--- a/rclpy/test/test_logging.py
+++ b/rclpy/test/test_logging.py
@@ -105,7 +105,6 @@ class TestLogging(unittest.TestCase):
                 throttle_time_source_type='RCUTILS_STEADY_TIME',
             ))
             time.sleep(0.4)
-        self.assertEqual(message_was_logged, [True, False, False, True, False])
         self.assertEqual(
             message_was_logged, [
                 True,  # t=0, not throttled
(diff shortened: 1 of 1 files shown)


rclpy/test/test_logging.py
```diff
@@ -106,6 +106,14 @@ def test_log_throttle(self):
...
             time.sleep(0.4)
         self.assertEqual(message_was_logged, [True, False, False, True, False])
+        self.assertEqual(
+            message_was_logged, [
+                True,  # t=0, not throttled
+                False,  # t=0.4, throttled
+                False,  # t=0.8, throttled
+                True,  # t=1.2, not throttled
+                False  # t=1.6, throttled
+            ])
```
wjwwood: Is this check now just duplicated?

dhood: the other uses skip first as well

mikaelarguedas: I think @wjwwood meants that it looks like the line above can be removed

dhood: ah! of course. thanks: 085e860302235ee394a7776576600d9e3f607433

PR #148, opened 2017-11-15
Give nodes a logger

dhood:
Minimum required to switch demos to using node loggers: logger names are not using namespaces at the moment nor calling to rcl to get their logger name.

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -16,7 +16,6 @@ import sys
 
 from rclpy.executors import SingleThreadedExecutor as _SingleThreadedExecutor
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
-import rclpy.logging  # noqa
 from rclpy.node import Node
 from rclpy.utilities import get_rmw_implementation_identifier  # noqa
 from rclpy.utilities import ok
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -16,6 +16,7 @@ import sys
 
 from rclpy.executors import SingleThreadedExecutor as _SingleThreadedExecutor
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+import rclpy.logging  # noqa
 from rclpy.node import Node
 from rclpy.utilities import get_rmw_implementation_identifier  # noqa
 from rclpy.utilities import ok
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -20,6 +20,7 @@ from rclpy.exceptions import NoTypeSupportImportedException
 from rclpy.expand_topic_name import expand_topic_name
 from rclpy.guard_condition import GuardCondition
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+from rclpy.logging import get_named_logger
 from rclpy.publisher import Publisher
 from rclpy.qos import qos_profile_default, qos_profile_services_default
 from rclpy.service import Service
@@ -52,6 +53,8 @@ class Node:
 
     def __init__(self, node_name, *, namespace=None):
         self._handle = None
+        # TODO(dhood): get logger name from rcl, use namespace (with slashes converted)
+        self._logger = get_named_logger(node_name)
         self.publishers = []
         self.subscriptions = []
         self.clients = []
@@ -90,6 +93,9 @@ class Node:
 
     def get_namespace(self):
         return _rclpy.rclpy_get_node_namespace(self.handle)
+
+    def get_logger(self):
+        return self._logger
 
     def _validate_topic_or_service_name(self, topic_or_service_name, *, is_service=False):
         name = self.get_name()
--- a/rclpy/test/test_node.py
+++ b/rclpy/test/test_node.py
@@ -89,6 +89,12 @@ class TestNode(unittest.TestCase):
         # test that it doesn't raise
         self.node.get_node_names()
 
+    def test_node_logger(self):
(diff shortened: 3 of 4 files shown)


rclpy/rclpy/__init__.py
```diff
@@ -16,6 +16,7 @@
 
 from rclpy.executors import SingleThreadedExecutor as _SingleThreadedExecutor
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+import rclpy.logging  # noqa
```
sloretz: Why import this in `__init__,py`? Is it to initialize the root_logger before `rclpy.init()` is called? What breaks if this import doesn't happen here?

dhood: nothing breaks, it's just to save users from having to import it themselves in each script that wants to use logging. I can add a comment for it but I'm not sure which other lines in this section are doing it for the same reason

sloretz: I don't fully understand `__init__.py`, but if I comment this line out it appears to be imported regardless.

```
$ python3
Python 3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import rclpy
>>> rclpy.logging.logerr("Hello World!")
[ERROR] []: Hello World! (<module>() at /workspace/ros2_ws/<stdin>:1)
True
```

dhood: ah, it was necessary on master to use the non-node logging methods, but since node.py is importing it now, it's not needed here: a48d6b634c08751d6af302b8c4706c861a33df9a

dhood:
Please review this PR in combination with https://github.com/ros2/demos/pull/190 to check that the usage is as expected. In particular, I opted for a `get_logger` method on the node as opposed to giving users access to the logger attribute (1) to match the other accessors, (2) in case we ever need to do something more complicated to retrieve the logger for the user.

dhood:
@ros2/team what do we think about:
1. accessing the node's logger as an attribute as opposed to via an accessor
2. updating node.get_name() and node.get_namespace() to match?

We can raise when users try to set the attributes. I checked the PRs that added the other getters (https://github.com/ros2/rclpy/pull/70 and https://github.com/ros2/rclpy/pull/86) and there wasn't a discussion about why accessors have been used

wjwwood:
I don't have a problem with using properties rather than getters, but I only hesitate to avoid a difference with C++ where we cannot easily do that I think.

clalancette:
I'm for using properties, as it is more pythonic in general.  But I don't really have a strong opinion either way.

dhood:
I suppose it could be a bit counter-intuitive if it looks like an attribute but then raises when users try to set it. It might be more user-friendly to just give them the getters without setters if we don't want users to try to modify the names/logger after creation. Does that influence your opinion from a user perspective @sloretz ?

mikaelarguedas:
May be related to the discussion (havent followed it)
https://github.com/ros2/rclpy/blob/5c1b6e10b75166f14cd381a792bc4a9ffd429df0/rclpy/rclpy/node.py#L84-L86

sloretz:
@dhood On point 2 my preference isn't strong enough for me to say the existing code should be changed. I think it would be fine to leave it as `get_logger` to stay consistent with the existing code.

dhood:
OK it doesn't seem like we have a strong enough motivation to swap to attributes, while "intuitiveness", existing code and similarity with c++ slightly lean us towards using getters, so the conclusion I'm drawing is that we'll stick with `get_logger` in this PR

dhood:
CI including changes in https://github.com/ros2/demos/pull/190, will merge once comes back green
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3572/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/769/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2912/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3665/)

mikaelarguedas:
> updating node.get_name() and node.get_namespace() to match?

name and namespace are not stored in Python land (not slots or attributes of the node class) so we need getters for them as we get them from rcl.

If we want to take the same approach here we would never store the logger in python and get it from rcl everytime a user want to access it (which is totally fine IMO).
If it's stored as an attribute like in this PR there is nothing preventing users from getting and setting it as any other attribute right ?

Not saying we should change it here, I'm just wondering where we expect this to be stored in the future and if the decision is made only for being consistent with rclcpp and that it'll be inconsistent within rclpy only for the time being or if we chose to treat some attributes differently than others in rclpy for the long-term.

dhood:
> name and namespace are not stored in Python land (not slots or attributes of the node class) so we need getters for them as we get them from rcl.

They wouldn't need to be stored in Python. If we swap them to properties as with https://github.com/ros2/rclpy/blob/5c1b6e10b75166f14cd381a792bc4a9ffd429df0/rclpy/rclpy/node.py#L82 we could have the accessor call through to rcl, and then we could remove the getters.

> If we want to take the same approach here we would never store the logger in python and get it from rcl everytime a user want to access it

Python has a notion of logger objects where rcl doesn't, so in this case it is appropriate for us to store the logger object associated with the node (this is what I envision for both short-term and long-term)

> If it's stored as an attribute like in this PR there is nothing preventing users from getting and setting it as any other attribute right ?

That's correct, but attributes with underscores are not considered public

dhood:
Edit (commenting so notification gets sent): "Python has a notion of logger objects where rcl doesn't" -> "_rclpy_ has a notion of logger objects" (as of https://github.com/ros2/rclpy/pull/102)

PR #149, opened 2017-11-17
keep consistency when destroying node raises an exception

dirk-thomas:
Fixes the bad behavior described in ros2/ros2cli#62 but not the actual case of the exception.


dirk-thomas:
@sloretz FYI since you modified the logic in #129 from "destroying and removing elements one by one" to "destroy all elements and empty the list afterwards".

PR #150, opened 2017-11-17
reenable 100Hz timer test on windows

mikaelarguedas:
now that we have better time resolution on windows these tests can be restored.

Unfortunately the 1kHz timer tests still fail and have to stay disabled.

* Windows (rclpy tests ranusing --retest-until-fail 50) [Build Status](http://ci.ros2.org/job/ci_windows/3653/)

### DIFF
--- a/rclpy/test/test_timer.py
+++ b/rclpy/test/test_timer.py
@@ -144,14 +144,12 @@ def test_timer_zero_callbacks10hertz():
     func_launch(func_zero_callback, ['0.1'], 'received callbacks when not expected')
 
 
-# TODO(mikaelarguedas) reenable these once timer have consistent behaviour
-# on every platform at high frequency
 def test_timer_zero_callbacks100hertz():
-    if os.name == 'nt':
-        raise SkipTest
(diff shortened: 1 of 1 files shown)


serge-nikulin:
> Unfortunately the 1kHz timer tests still fail and have to stay disabled.

Have you enabled 1KHz switching in Windows from the default 15ms or so?
AFAIK, MS has introduced this feature for gamers and it might help you.

mikaelarguedas:
@serge-nikulin Thanks for the tip,
I don't think we changed anything explicitly. Though if the default resolution is 15.6ms I would expect our 100Hz tests to fail, at least sporadically. I'll merge this for now and post an update here when I find time to explore that lead.

serge-nikulin:
See https://msdn.microsoft.com/en-us/library/windows/desktop/dd757624(v=vs.85).aspx.  
Call ```timeBeginPeriod```  at the startup time and don't bother to call ```timeEndPeriod``` at all.  
A lot of airport security equipment around the world works in this mode for many years, non-stop, 24/365 :)

mikaelarguedas:
Good to know, thanks for the pointer

PR #151, opened 2017-11-20
Avoid unexpected free to null

gaoethan:
the `expanded_topic` is possible to be `NULL` from
`rcl_expand_topic_name()` and then the free is problematic

Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -498,7 +498,7 @@ rclpy_get_validation_error_for_node_name(PyObject * Py_UNUSED(self), PyObject * 
 static char *
 _expand_topic_name_with_exceptions(const char * topic, const char * node, const char * namespace)
 {
-  char * expanded_topic;
+  char * expanded_topic = NULL;
   rcl_allocator_t allocator = rcl_get_default_allocator();
   rcutils_allocator_t rcutils_allocator = rcutils_get_default_allocator();
   rcutils_string_map_t substitutions_map = rcutils_get_zero_initialized_string_map();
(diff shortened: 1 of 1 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -546,7 +546,9 @@ _expand_topic_name_with_exceptions(const char * topic, const char * node, const
   if (rcutils_ret != RCUTILS_RET_OK) {
     PyErr_Format(PyExc_RuntimeError, "%s", rcutils_get_error_string_safe());
     rcutils_reset_error();
-    allocator.deallocate(expanded_topic, allocator.state);
+    if (expanded_topic) {
```
tfoote: Using an if on `free` is not necessary/recommended against. I would suggest that we hold the policy that our `deallocate` functions should also have the same behavior as free. And thus not add custom logic outside the deallocator to check before.

Two related threads: 
- https://stackoverflow.com/questions/1938735/does-freeptr-where-ptr-is-null-corrupt-memory
- https://stackoverflow.com/questions/6084218/is-it-good-practice-to-free-a-null-pointer-in-c
Which link to [ISO-IEC 9899](http://www.open-std.org/JTC1/SC22/wg14/www/docs/n1124.pdf) Section 7.20.3 is relevant.

gaoethan:
@tfoote thanks for your comments, actually, I'm not sure it's a NOP on all platform. :) and It makes sense to unify the checking in the `deallocate` if it's decided to be necessary to avoid NULL free. anyway, I'm open to that and let's conclude it. thanks.

clalancette:
I just updated it to just set `expanded_topic` to NULL, in the name of defensive programming.  I'm going to run CI against this and then, assuming that passes, merge.

clalancette:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3575/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/772/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2914/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3668/)

PR #140, opened 2017-11-10
Executor bug fixes and WaitSet refactor

sloretz:
This PR is split from #127. It has performance improvements to the executor that get the 1kHz timer test to pass consistently. If you've already reviewed #127 the new changes are:

* Implemented `WaitSet` class as a cpython extension
    * Executor updated to use it
* Moved sigint handler to a library for sharing between _rclpy and _rclpy_wait_set

Changes that were already in #127

* `wait_for_callbacks` refactored readability + performance
* Fixed bug where `wait_for_callbacks` could yield the wrong node
* Removed underscores from imported names
* added utilty `timeout_sec_to_nsec`
* `SingleThreadedExecutor` and `MultiThreadedExecutor` reuse the iterator from `wait_for_callbacks` over multiple `spin_once` calls

CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3529/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/745/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2876/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3615/)

PR #152, opened 2017-11-22
waitset -> wait_set

mikaelarguedas:
connects to ros2/rmw#131

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -1668,11 +1668,11 @@ rclpy_get_zero_initialized_wait_set(PyObject * Py_UNUSED(self), PyObject * Py_UN
   return pywait_set;
 }
 
-/// Initialize a waitset
+/// Initialize a wait_set
 /**
  * Raises RuntimeError if the wait set could not be initialized
  *
- * \param[in] pywait_set Capsule pointing to the waitset structure
+ * \param[in] pywait_set Capsule pointing to the wait_set structure
  * \param[in] node_name string name of the node to be created
  * \param[in] number_of_subscriptions int
  * \param[in] number_of_guard_conditions int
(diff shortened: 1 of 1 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -1668,11 +1668,11 @@ rclpy_get_zero_initialized_wait_set(PyObject * Py_UNUSED(self), PyObject * Py_UN
   return pywait_set;
 }
 
-/// Initialize a waitset
+/// Initialize a wait set
```
sloretz: It looks like all of this code is removed in #140. Would you mind reviewing that instead? This is assuming #140 gets in before feature freeze of course.

mikaelarguedas: Given that this PR touches only documentation don't feel strongly about merging it or holding it in favor of #140, I'll go ahead and merge all the other PRs as they are affecting rmw API.

I'll try to review #140 today, but as its a >1k lines it'll take a while

sloretz: I think we should merge this PR.

How does this sound? With the number of people on vacation this week it seems unlikely the wait_for_service feature will make it in. Since most of #140 is refactoring for that, it won't need to be in the dec 8th release. Separately I'll split out the executor bug fixes from #140 to make sure those get in.

mikaelarguedas: Sounds good, I'll merge this soon, wrapping up third party rmw PRs.

> With the number of people on vacation this week it seems unlikely the wait_for_service feature will make it in.

We'll see, I think there is a lot of value in getting some version of wait_for_service in before the release. We also have several PRs blocked until we can use that feature in tests.

> Separately I'll split out the executor bug fixes from #140 to make sure those get in.

Yeah thats always a good idea to ease review. thanks!

dhood: @sloretz you would know best if this is a reasonable proposal, but it seems there was a state at which https://github.com/ros2/rclpy/pull/127 was approved, and then work continued in order to address some executor bugs. Perhaps what has been approved already can be merged and reviewers can focus on what's on top?

sloretz: @dhood I've made the state more complicated than it needs to be. The version of #127 that was approved has a couple problems:

* It adds overhead that makes the 1kHz timer test fail 100% of the time on aarch64
* A separate GraphListener wouldn't make sense once a `wait_for_service_async` is added

#135 was split from #127 to implement `wait_for_service` in a way that makes sense with `wait_for_service_async`. #140 was split from #135 to make the PR smaller. #140 has bug fixes found while reviewing #127 plus the performance improvements needed to get the 1kHz timer test to pass.

PR #154, opened 2017-11-25
Only check logger filters if severity is sufficient

dhood:
connects to ros2/rcutils#75

no test for this has been added because there isn't currently support in rclpy for the equivalent of the "expression"/"function" filters that rcutils supports

### DIFF
--- a/rclpy/rclpy/impl/rcutils_logger.py
+++ b/rclpy/rclpy/impl/rcutils_logger.py
@@ -254,12 +254,14 @@ class RcutilsLogger:
         r"""
         Log a message with the specified severity.
 
-        The message will only be logged if its severity is not less than the severity threshold
-        of the logger, and no logging filter causes the message to be skipped.
+        The message will not be logged if:
+          * the logger is not enabled for the message's severity (the message severity is less than
+            the severity threshold of the logger), or
+          * a logging filter causes the message to be skipped.
 
         .. note::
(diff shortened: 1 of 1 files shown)


rclpy/rclpy/impl/rcutils_logger.py
```diff
@@ -257,6 +257,9 @@ def log(self, message, severity, **kwargs):
         The message will only be logged if its severity is not less than the severity threshold
         of the logger, and no logging filter causes the message to be skipped.
 
+        .. note::
+           Logging filters will only be evaluated if the message severity is not less than the
```
mikaelarguedas: > Logging filters will only be evaluated if the message severity is not less than the severity threshold of the logger.

Maybe it's because I'm not native, but I had to read this twice to understand what it meant.
Maybe having the negation earlier in the sentence (logging filters will *not* be evaluated if the message severity is less than...) or using "greater" rather than "is not less than" would be clearer.

clalancette: > Maybe it's because I'm not native, but I had to read this twice to understand what it meant.
> Maybe having the negation earlier in the sentence (logging filters will not be evaluated if the message severity is less than...) or using "greater" rather than "is not less than" would be clearer.

Agreed, the wording is a little difficult.  I personally prefer using "greater" for this sentence, but moving the "not" earlier in the sentence should also make it clearer.

PR #155, opened 2017-11-27
extension should not free the passed-in QoS

BrannonKing:
We have never seen this cause a problem; it just didn't seem right to free a variable passed in by the user. There could be a problem in some convoluted situation where QoS is reused in a second publisher after a period of time.


BrannonKing:
@sloretz , I can't understand what you're asking for. You're saying that the python wrapper for rmw_qos_profile_t doesn't clean up the underlying object properly as part of its finalization? That we should use the destructor parameter on PyCapsule_New to clean this up? That sounds like a great plan to me. However, most of those extension methods call PyCapsule_New in a similar pattern where they wrap newly allocated memory without doing anything in that method to ensure said memory is freed in the future. Are they really all wrong? They all leak?

sloretz:
@BrannonKing  Yup, that's exactly what I'm asking for. I think only the qos profile `PyCapsule` can safely use the destructor parameter at the moment.

The other `PyCapsule` can't make use of the destructor right now.  Instead of using python's garbage collector the other `PyCapsule` rely on python code calling `rclpy_destroy_entity` or `rclpy_destroy_node_entity` on them. This happens in the methods `node.destroy_*()`. Those capsules do leak memory if the appropriate destroy method is not called on them.

Using the node as an example, if `rcl_shutdown` is called before `rcl_node_fini` then `rcl` will error because the instance id doesn't match. It then returns without `free`ing memory used by the node. This means `rcl_node_fini` must be called before `rcl_shutdown`. If the node `PyCapsule` called `fini` in its destructor then it would be left up to python's garbage collector. It may not get around collecting the `PyCapsule` until after a user has called `rclpy.shutdown()`.  

I can think of a way to make the rest of the PyCapsule be `free`'d by the python garbage collector, but it would mean getting rid of `rclpy.init()` and `rclpy.shutdown()`. Say the `_rclpy` module called `rcl_init()` when it is imported, and  `rcl_shutdown()` when it is unloaded. The module would need to be `incref`'d when one of the other `PyCapsule` is created and `decref`'d it when they're collected. It's a little more complicated because node entities (clients, subscribers, etc) would need to incref the node when they're created and decref it when they're collected too.

sloretz:
Hi @BrannonKing, friendly ping. Are you still interested in working on this PR?

BrannonKing:
@sloretz , I don't feel confident in my ability to accomplish all that you were asking for on this. I don't at present have sufficient understanding of Python's API to make the necessary modifications. (I recently used pybind11 for a project, but it, fortunately, kept me from needing to learn the details of Python's internals.)

sloretz:
@BrannonKing Ok, thanks for the PR! I'll close it for now.

We are also thinking about using pybind11 for rclpy. Just curious, what version of pybind11 are you using?

BrannonKing:
Yes, use pybind11! It's a sweet deal. I was just using a snapshot from their master close to their 2.2.1 release date. I had no issues with it.

PR #157, opened 2017-11-28
adapt to the NULL removal from rmw result validation string

gaoethan:
Change to align to the update of APIs `rmw_*_validation_result_string()`
Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>

Connects to ros2/rmw#130


gaoethan:
This PR relates to PR: #https://github.com/ros2/rmw/pull/130

gaoethan:
yes, I initially intend to show the specific validation message, but the error msg is returned from the final `result_list`, there is no need now, thanks for your comments.

PR #158, opened 2017-11-28
use ament cmake ros to honor BUILD_SHARED_LIBS cmake argument

mikaelarguedas:
connects to ros2/ros2#306

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3624/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/806/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2954/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3716/)

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -10,8 +10,8 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra)
 endif()
 
-find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
+find_package(ament_cmake_ros REQUIRED)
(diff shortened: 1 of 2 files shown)


rclpy/CMakeLists.txt
```diff
@@ -72,7 +72,7 @@ endfunction()
 
 add_library(
   rclpy
-  SHARED src/rclpy/_rclpy.c
+  src/rclpy/_rclpy.c
```
wjwwood: This can never be static though right? Otherwise Python would not be able to dynamically load it.

mikaelarguedas: That was my uncertainty, I wanted to test it to confirm before placing this in review.

mikaelarguedas:
Closing this as the python part of the stack cannot be static anyway

PR #159, opened 2017-11-28
Reuse ready callbacks generator

sloretz:
This adds a wrapper around `wait_for_ready_callbacks` that reuses the callback generator when multiple things are ready. It avoids unnecessary wait list construction and waiting. It also prevents one callback (like a very fast timer) from starving other callbacks (similar to ros2/rclcpp#392).

Taken from #140 and improved upon. 

CI (New run after fixing a merge conflict)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3635/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/816/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2969/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3729/)

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -71,6 +71,12 @@ class _WorkTracker:
         return True
 
 
+class TimeoutException(Exception):
+    """Signal that a timeout occurred."""
+
+    pass
+
+
 class Executor:
     """
     A base class for an executor.
@@ -94,6 +100,10 @@ class Executor:
         # True if shutdown has been called
         self._is_shutdown = False
         self._work_tracker = _WorkTracker()
+        # State for wait_for_ready_callbacks to reuse generator
+        self._cb_iter = None
+        self._last_args = None
+        self._last_kwargs = None
 
     def shutdown(self, timeout_sec=None):
         """
@@ -114,6 +124,9 @@ class Executor:
         _rclpy.rclpy_destroy_entity(self._guard_condition)
 
         self._guard_condition = None
+        self._cb_iter = None
+        self._last_args = None
+        self._last_kwargs = None
         return True
 
     def __del__(self):
@@ -257,9 +270,11 @@ class Executor:
         """
         return not entity._executor_event and entity.callback_group.can_execute(entity)
 
-    def wait_for_ready_callbacks(self, timeout_sec=None, nodes=None):
+    def _wait_for_ready_callbacks(self, timeout_sec=None, nodes=None):
         """
         Yield callbacks that are ready to be performed.
+
+        Raises :class:`TimeoutException` on timeout.
 
         :param timeout_sec: Seconds to wait. Block forever if None or negative. Don't wait if 0
         :type timeout_sec: float or None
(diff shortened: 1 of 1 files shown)


rclpy/rclpy/executors.py
```diff
@@ -384,23 +399,41 @@ def wait_for_ready_callbacks(self, timeout_sec=None, nodes=None):
...
+        Reuse generator and return callbacks that are ready to be performed.
+
+        See :func:`Executor._wait_for_ready_callbacks` for documentation
+        """
+        if self._cb_iter is None or self._last_args != args or self._last_kwargs != kwargs:
+            # Create a new generator
+            self._last_args = args
+            self._last_kwargs = kwargs
+            self._cb_iter = self._wait_for_ready_callbacks(*args, **kwargs)
+        try:
```
clalancette: Minor nit: I'd put a blank line before this try block, just to visually distinguish it from the if block above.

rclpy/rclpy/executors.py
```diff
@@ -384,23 +399,41 @@ def wait_for_ready_callbacks(self, timeout_sec=None, nodes=None):
...
+            # Ran out of work, recursively create a new generator
+            self._cb_iter = None
+            return self.wait_for_ready_callbacks(*args, **kwargs)
 
 
 class SingleThreadedExecutor(Executor):
     """Runs callbacks in the thread which calls :func:`SingleThreadedExecutor.spin`."""
 
-    def __init__(self):
-        super().__init__()
```
clalancette: Do we really want to remove this call to `Executor.__init__`?

mikaelarguedas: My understanding is that a method that is not overloaded in the child class means that the parent version of the method is used. I believe this applies to all methods included the `__init_` one

clalancette: I wasn't sure about this, so I just tried it out with the following program:

```
class Base:
	def __init__(self):
		print("Base init")

class Sub:
	def do_something(self):
		print("Sub do_something")

x = Sub()
x.do_something()
```

If you run that under python3, the output you get is just "Sub do_something"; you never see the "Base init" printed out.

mikaelarguedas: can you try the same thing with Sub inheriting from base ?
```
class Base:
	def __init__(self):
		print("Base init")

class Sub(Base):
	def do_something(self):
		print("Sub do_something")

x = Sub()
x.do_something()
```

clalancette: D'oh, sorry.  I forgot that bit.  Never mind, you are totally correct.

rclpy/rclpy/executors.py
sloretz: Added in 786b4cf4f9bdc067183eed66f50a0bc1b885b21b

clalancette:
> It also prevents one callback (like a very fast timer) from starving other callbacks (similar to ros2/rclcpp#392).

Does this new code have this property because we always process everything from the previous wait set before constructing a new one?  (I just want to make sure I understand)

sloretz:
> Does this new code have this property because we always process everything from the previous wait set before constructing a new one?

@clalancette Yup

clalancette:
The only concern I have is with the recursive nature of `wait_for_ready_callbacks`, and whether that recursion can go without bounds in some cases.  @sloretz thinks that the maximum we can go is one level deep, but he is doing a bit of research/playing around and will report back.

sloretz:
@clalancette eliminated the recursion in b7fa6d69298ca281f16eb3ebe185280992a02c9d

CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3651/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/831/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2988/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3753/)

rclpy/rclpy/executors.py
```diff
@@ -390,23 +405,45 @@ def wait_for_ready_callbacks(self, timeout_sec=None, nodes=None):
...
+
+    def wait_for_ready_callbacks(self, *args, **kwargs):
+        """
+        Reuse generator and return callbacks that are ready to be performed.
+
+        See :func:`Executor._wait_for_ready_callbacks` for documentation
+        """
+        # if an old generator is done, this variable makes the loop get a new one before returning
+        got_generator = False
+        while not got_generator:
```
clalancette: I like the loop a lot better; thanks!

PR #160, opened 2017-11-28
Added timeout_sec_to_nsec utility

sloretz:
Taken from #140 / #127. It adds a utility for converting timeouts from seconds to nanoseconds for rcl_wait. Additionally executor timeouts now block if None or negative and are only non-blocking if the timeout is zero.

CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3632/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/814/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2963/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3726/)

### DIFF
--- a/rclpy/rclpy/utilities.py
+++ b/rclpy/rclpy/utilities.py
@@ -14,6 +14,9 @@
 
 import threading
 
+from rclpy.constants import S_TO_NS
+
+
 g_shutdown_lock = threading.Lock()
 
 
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/utilities.py
```diff
@@ -44,3 +47,23 @@ def get_rmw_implementation_identifier():
...
+
+
+def timeout_sec_to_nsec(timeout_sec):
+    """
+    Convert timeout in seconds to rcl compatible timeout in nanoseconds.
+
+    Python tends to use floating point numbers in seconds for timeouts. This utility converts a
+    python-style timeout to an integer in nanoseconds that can be used by rcl_wait.
+
+    :param timeout_sec: Seconds to wait. Block forever if None or negative. Don't wait if < 1ns
```
mikaelarguedas: Nit: `Don't wait if 0 <= timeout_sec < 1ns`
maybe it's clear enough as is

rclpy/test/test_utilities.py
```diff
@@ -0,0 +1,31 @@
...
+import rclpy.utilities
+
+
+class TestUtilities(unittest.TestCase):
+
+    def test_timeout_sec_to_nsec(self):
+        self.assertGreater(0, rclpy.utilities.timeout_sec_to_nsec(None))
+        self.assertGreater(0, rclpy.utilities.timeout_sec_to_nsec(-1))
+        self.assertEqual(0, rclpy.utilities.timeout_sec_to_nsec(0))
+        self.assertEqual(int(1.5 * S_TO_NS), rclpy.utilities.timeout_sec_to_nsec(1.5))
```
mikaelarguedas: Nit: should we add a test to show that something < 1ns results in a timeout of 0?

sloretz: c82a256b40fcb2d1e2828980deef6a588271746d

rclpy/rclpy/utilities.py
sloretz: If it is okay with you, I would like to leave this as is to avoid wrapping the line.

mikaelarguedas: :+1: works for me

sloretz:
OSX job failed because the node died. Jenkins auto-retriggered it and the second build succeeded: http://ci.ros2.org/job/ci_osx/2965/. I think this can be merged.

PR #162, opened 2017-11-30
Executor avoids list comprehensions

sloretz:
These are some minor optimizations pulled out of #140.

All of them avoid unnecessary list comprehensions. The result is the time spent in the executor during `spin_once` in the 1kHz timer test using `func_number_callbacks` drops from 25.5% to 22.2% (average of 100 runs). It's not much, but it might be enough to reduce the test flakiness. I made one change at a time checking the overhead with a script, so I'm confident that each change here is an optimization.

CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3643/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/823/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2976/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3743/)

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -140,7 +140,7 @@ class Executor:
         :rtype: list
         """
         with self._nodes_lock:
-            return [node for node in self._nodes]
+            return list(self._nodes)
 
     def spin(self):
         """Execute callbacks until shutdown."""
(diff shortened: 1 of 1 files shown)


rclpy/rclpy/executors.py
```diff
@@ -284,16 +284,16 @@ def wait_for_ready_callbacks(self, timeout_sec=None, nodes=None):
...
+                timers.extend(filter(self.can_execute, node.timers))
+                clients.extend(filter(self.can_execute, node.clients))
+                services.extend(filter(self.can_execute, node.services))
+                node_guards = filter(self.can_execute, node.guards)
                 # retrigger a guard condition that was triggered but not handled
                 for gc in node_guards:
                     if gc._executor_triggered:
                         gc.trigger()
-                guards.extend(node_guards)
+                    guards.append(gc)
```
wjwwood: I would be surprised if this is faster, and it appears to be exactly the same behavior. Do you have some reference which supports why this might be considered faster?

sloretz: It's a consequence of line 291 where `node_guards` becomes a generator instead of a list.

An alternative that uses `extend` instead of `append` is:

```python
                node_guards = list(filter(self.can_execute, node.guards))
                # retrigger a guard condition that was triggered but not handled
                for gc in node_guards:
                    if gc._executor_triggered:
                        gc.trigger()
                guards.extend(node_guards)
```

But that increases the executor overhead from 22.2% to 22.6% on my machine. I'm guessing the overhead of calling `append` for each item in the list is less than the overhead of creating an intermediate list and iterating internal to `extend`.

wjwwood: I see so the removed line was implicitly `guards.extend(list(node_guards))`. 👍

PR #164, opened 2017-12-01
Remove root logger usage (rclpy.loginfo) and get_severity_threshold()

dhood:
In light of the upcoming supported release I want to remove some things that have been bothering me about the logging API.

## Remove `rclpy.loginfo()`
We decided that in rclcpp users should be "forced" to create a named logger to make a call, so that they think about the hierarchy etc. Because of this I have removed the convenience functions `rclpy.loginfo` etc on the root logger (now "private") to match.

I did not remove the root logger completely: user-created loggers still attach to the root logger because the root logger's name, though it's currently empty, will encapsulate prefix conventions should they ever be introduced.
https://github.com/ros2/rclpy/commit/10cea411b37530eee92e8b352df771e2705a02ab
https://github.com/ros2/rclpy/commit/6d67190b698dbd979b4e5037eee422fb709d8171

## Rename `get_named_logger to get_logger()`
Matches python logging module; the "named" part is redundant.
https://github.com/ros2/rclpy/commit/4531c89b6d19cc64a75cf14eb290bae9f955266b

## Remove `logger.get_severity_threshold()`
- A logger can have its severity explicitly set with `logger.set_severity_threshold()`.
- By default loggers have `UNSET` severity threshold, which causes them to inherit their parent's severity threshold: resolved via `get_effective_severity_threshold()`
- Users almost always want to call `logger.get_effective_severity_threshold()`/`logger.is_enabled_for()`. However, a few times I have accidentally called `logger.get_severity_threshold()` which usually returns `UNSET`.

=> I've removed `get_severity_threshold()` (python logging module doesn't have it) because, except for test purposes, I can't imagine why someone would want to know whether the severity of a logger was explicitly set to X or if it just resolved to X. We lose information but I users will be less likely to trip over the similarly named methods.
https://github.com/ros2/rclpy/commit/217f1c3ab2f690b38289777c0d52fd9d8475bbf1

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3654/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/834/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2991/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3760/)

### DIFF
--- a/rclpy/rclpy/impl/rcutils_logger.py
+++ b/rclpy/rclpy/impl/rcutils_logger.py
@@ -228,12 +228,6 @@ class RcutilsLogger:
             name = self.name + '.' + name
         return RcutilsLogger(name=name)
 
-    def get_severity_threshold(self):
-        from rclpy.logging import LoggingSeverity
-        severity = LoggingSeverity(
-            _rclpy_logging.rclpy_logging_get_logger_severity_threshold(self.name))
-        return severity
-
     def set_severity_threshold(self, severity):
         from rclpy.logging import LoggingSeverity
         severity = LoggingSeverity(severity)
--- a/rclpy/rclpy/logging.py
+++ b/rclpy/rclpy/logging.py
@@ -34,13 +34,13 @@ class LoggingSeverity(IntEnum):
     FATAL = 50
 
 
-root_logger = rclpy.impl.rcutils_logger.RcutilsLogger()
+_root_logger = rclpy.impl.rcutils_logger.RcutilsLogger()
 
 
 def get_named_logger(name):
     if not name:
         raise ValueError('Logger name must not be empty.')
-    return root_logger.get_child(name)
+    return _root_logger.get_child(name)
 
 
 def initialize():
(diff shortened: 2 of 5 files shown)


mikaelarguedas:
With https://github.com/ros2/rcutils/issues/77 `logger.get_effective_severity_threshold()` will become `logger.get_effective_level()` right?
(because `logger.get_effective_severity_threshold()` seems very unconvenient to write if it's the only way to get a logger level in the public API)

dhood:
correct

mikaelarguedas:
> I've removed get_severity_threshold() (python logging module doesn't have it) because, except for test purposes, I can't imagine why someone would want to know whether the severity of a logger was explicitly set to X or if it just resolved to X. 

I guess that's to match the python module API exactly, but I think it would be more user-friendly to provide just `logger.get_level()` and an optional keyword argument if we don't want it to traverse the hierarchy ?
This way we can get the information if we really want to but have a reasonable default and a shorter function name

dhood:
> I think it would be more user-friendly to provide just logger.get_level() and an optional keyword argument if we don't want it to traverse the hierarchy ?

we spoke offline and I'll leave it as more the more explicit name since just `get_level()` hides the notion of resolving severity levels and we want to keep users conscious of logger hierarchy

> will this be affected by ros2/rcutils#77?

I'll rename the functions that remain after this is merged :+1:

Issue #58, opened 2016-12-07
Add wait_for_service


mikaelarguedas:

mikaelarguedas:
@sloretz can you also open a PR against `system_tests` to update the [requester_py.py](https://github.com/ros2/system_tests/blob/master/test_communication/test/requester_py.py) file to use this new feature. I would expect this to reduce the flakiness of these tests significantly.
Thanks!

sloretz:
@ros2/team Mind giving some feedback on this idea?

TL;DR I want to eliminate the `GraphListener` using coroutines.

Currently #127 implements `wait_for_service` using a `GraphListener` class. `GraphListener` has its own thread and waits on node graph guard conditions plus timers for the `wait_for_service` timeout. This enables a user to call `client.wait_for_service()` inside of a callback without blocking forever due to starvation of a `SingleThreadedExecutor`.

```
def my_callback(self, msg):
   if self.client.wait_for_service():
      # do something with service
```

Alternatively, `GraphListener` can be eliminated using the coroutine support in #135. `wait_for_service()` would return a `Future` object. If a user wants to wait in a callback they must make it a coroutine and `await` the `Future`. The executor would wait on the node graph guard conditions, and execute callbacks like normal. The `SingleThreadedExecutor` would not be blocked by the user's callback because the future yields back to the executor if the service isn't ready.

```
async def my_callback(self, msg):
   result = await self.client.wait_for_service()
   if result:
       # do something with service
```

The trouble happens when trying to offer both options. The executor cannot wait on the node graph guard conditions because `rcl_wait` says the same entity cannot be waited on at the same time in two wait sets. It is possible to make a second method `wait_for_service_async` that returns a `Future` which uses the `GraphListener`, but it means the same pattern has to be used anywhere a `Future` is returned. `Client.call()` and `Client.call_async()` would need a `ClientListener` managing a separate thread to wait on clients.

I see a few options. I like the first option the most.

1. Offer only `wait_for_service` that returns a `Future`. Eliminate `GraphListener`. If a user wants to block in a callback they must know about python coroutines.
1. Offer both `wait_for_service` and `wait_for_service_async`. Don't Elimate `GraphListener`. Create `ClientListener` in the future.
1. Offer both `wait_for_service` and `wait_for_service_async`. Eliminate `GraphListener`. Don't try to protect the user from starvation. If they block in a callback without using a coroutine that's their problem.

dirk-thomas:
[...]

PR #166, opened 2018-01-04
Futures and coroutines

sloretz:
This adds coroutine support to the executor and is part of service feature parity #123 and `wait_for_service` without a `GraphListener` #58.

Any callback can now be a coroutine function (`async def`). If the callback is a coroutine and it becomes blocked the executor will move on and resume executing it later. The details of this are handled in the class `Task` which is a class very similar to [asyncio.Task](https://docs.python.org/3/library/asyncio-task.html#task). `asyncio` can't be used directly because it is not thread safe.

This PR also adds a `Future` class that represents the result of a stored task. It is not used in this PR, but an example of using it is in #170.

**CI** ( with 5f2cb1eb6cc451acbe0bb58b93672ebe224b45e6)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3878/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/998/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3204/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3955/)

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -120,7 +120,7 @@ class Executor:
         self._last_args = None
         self._last_kwargs = None
 
-    def create_task(self, callback, *args):
+    def create_task(self, callback, *args, **kwargs):
         """
         Add a callback or coroutine to be executed during :meth:`spin` and return a Future.
 
@@ -130,7 +130,7 @@ class Executor:
         :type callback: callable or coroutine function
         :rtype: :class:`rclpy.task.Future` instance
         """
-        task = Task(callback, *args, executor=self)
+        task = Task(callback, args, kwargs, executor=self)
         with self._tasks_lock:
             self._tasks.append((task, None, None))
         # Task inherits from Future
@@ -283,7 +283,7 @@ class Executor:
                     # callback group can get executed
                     _rclpy.rclpy_trigger_guard_condition(gc)
         task = Task(
-            handler, entity, self._guard_condition, self._is_shutdown, self._work_tracker,
+            handler, (entity, self._guard_condition, self._is_shutdown, self._work_tracker),
             executor=self)
         with self._tasks_lock:
             self._tasks.append((task, entity, node))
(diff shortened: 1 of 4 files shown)


sloretz:
Oops, looks like the description I put on this task was completely wrong and from another PR. It's fixed now.

sloretz:
@ros2/team Friendly ping. I would appreciate a review of this one. It is a prerequisite to multiple concurrent client requests (#170) and a `wait_for_service` implementation that returns asap instead of sleep/polling.

rclpy/rclpy/executors.py
```diff
@@ -105,6 +120,22 @@ def __init__(self):
         self._last_args = None
         self._last_kwargs = None
 
+    def create_task(self, callback, *args):
+        """
+        Add a callback or coroutine to be executed during :meth:`spin` and return a Future.
+
+        Arguments to this function are passed to the callback.
```
dirk-thomas: Could this also permit keyword arguments?

sloretz: Yup, Added in 5f2cb1eb6cc451acbe0bb58b93672ebe224b45e6

dirk-thomas: I didn't think about it before. It might be necessary to pass `executor` using a setter function instead to avoid a collision if the callback wants to use a keyword argument with that name.

sloretz: In 5f2cb1e the expansion of `args` and `kwargs` in `Task.__init__()` [was removed](https://github.com/ros2/rclpy/commit/5f2cb1eb6cc451acbe0bb58b93672ebe224b45e6#diff-521ddf9ec814122daa49f03bb2974340R133) to prevent that collision. The expansion on `executor.create_task()` cannot conflict because they're now passed to `Task.__init__()` as a sequence and dictionary.

dirk-thomas: Perfect!

samiamlabs:
Hi @sloretz!

Is it possible/easy to make the equivalent of `asyncio.sleep` using the coroutine support from this PR?

(We have been trying to find a robust pattern for waiting in the execute functions of action servers in rclpy. The only pattern that works for us 100% reliably so far is using a SingleThreadedExecutor and `spin_once` while waiting in the execute callback. It just seems wrong to spin in callbacks though...)

PR #170, opened 2018-01-04
Refactor client for multiple requests

sloretz:
Opening a PR for feedback. This uses the `Future` class added in pull request #166 to allow simultaneous requests from the same client (part of #123). It is implemented like [option 3 in this comment](https://github.com/ros2/rclpy/issues/58#issuecomment-344697128).

Specifically any ideas about `call`? This new version should not be called inside a callback because it would deadlock a `SingleThreadedExecutor`. It also is not very useful outside because an executor would need to be running in another thread for the client response to be taken from rcl.

CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3902/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1020/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3225/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3995/)
    * Unrelated failing test (timeout): `projectroot.test_api_srv_composition_client_first__rmw_fastrtps_cpp`
        * Also timed out on `nightly_win_deb` http://ci.ros2.org/view/nightly/job/nightly_win_deb/764
        * However, it passed in `nightly_win_rep` http://ci.ros2.org/view/nightly/job/nightly_win_rel/705

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -54,3 +54,14 @@ def spin(node):
             executor.spin_once()
     finally:
         executor.shutdown()
+
+
+def spin_until_future_complete(node, future):
+    # imported locally to avoid loading extensions on module import
+    from rclpy.executors import SingleThreadedExecutor
+    executor = SingleThreadedExecutor()
+    try:
+        executor.add_node(node)
+        executor.spin_until_future_complete(future)
+    finally:
+        executor.shutdown()
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -57,6 +57,15 @@ def spin(node):
 
 
 def spin_until_future_complete(node, future):
+    """
+    Execute work until the future is complete.
+
+    Callbacks and other work will be executed in a SingleThreadedExecutor until future.done()
+    returns True or rclpy is shutdown.
+
+    :param future: The future object to wait on.
+    :type future: rclpy.task.Future
+    """
     # imported locally to avoid loading extensions on module import
     from rclpy.executors import SingleThreadedExecutor
     executor = SingleThreadedExecutor()
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -16,39 +16,8 @@ import threading
 import time
 
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+from rclpy.task import Future
 import rclpy.utilities
-
-
-class ResponseThread(threading.Thread):
-
-    def __init__(self, client):
-        threading.Thread.__init__(self)
-        self.client = client
-        self.wait_set = _rclpy.rclpy_get_zero_initialized_wait_set()
-        _rclpy.rclpy_wait_set_init(self.wait_set, 0, 1, 0, 1, 0)
-        _rclpy.rclpy_wait_set_clear_entities('client', self.wait_set)
-        _rclpy.rclpy_wait_set_add_entity(
-            'client', self.wait_set, self.client.client_handle)
-
-    def run(self):
-        [sigint_gc, sigint_gc_handle] = _rclpy.rclpy_get_sigint_guard_condition()
-        _rclpy.rclpy_wait_set_add_entity('guard_condition', self.wait_set, sigint_gc)
-
-        _rclpy.rclpy_wait(self.wait_set, -1)
-
-        guard_condition_ready_list = \
-            _rclpy.rclpy_get_ready_entities('guard_condition', self.wait_set)
-
-        # destroying here to make sure we dont call shutdown before cleaning up
-        _rclpy.rclpy_destroy_entity(sigint_gc)
-        if sigint_gc_handle in guard_condition_ready_list:
-            rclpy.utilities.shutdown()
-            return
-        seq, response = _rclpy.rclpy_take_response(
-            self.client.client_handle,
-            self.client.srv_type.Response)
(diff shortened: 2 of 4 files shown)


mikaelarguedas:
Mentioned it to @sloretz offline but writing it here for book-keeping.
This PR removes API used by various packages so we'll need to update those to use the new API and merge these PRs all together, of the top of my head: `demo_nodes_py`, `examples_rclpy_*` and `ros2service` need corresponding PRs

rclpy/rclpy/client.py
```diff
@@ -61,15 +30,53 @@ def __init__(
...
+    def call_async(self, req):
+        """
+        Make a service request and asyncronously get the result.
+
+        :return: a Future instance that completes when the request does
+        :rtype: :class:`rclpy.task.Future` instance
+        """
+        sequence_number = _rclpy.rclpy_send_request(self.client_handle, req)
+        future = Future()
+        self._pending_requests[sequence_number] = future
```
dirk-thomas: In order to be defensive should this check / ensure that `sequence_number` is not already in the dictionary?

rclpy/rclpy/client.py
```diff
@@ -61,15 +30,53 @@ def __init__(
...
+
+        def remove_pending_request(future):
```
dirk-thomas: The argument isn't being used at the moment. Could this be a method on `self` instead and look up the to-be-deleted key based on the passed `future`?

rclpy/rclpy/executors.py
```diff
@@ -222,11 +227,15 @@ def _take_client(self, client):
...
-            # the content of client.response to check if a response has been received
-            client.response = response
+        if sequence is not None:
+            try:
+                future = client._pending_requests[sequence]
+            except IndexError:
+                # The request was cancelled
+                pass
+            else:
+                future._set_executor(self)
```
dirk-thomas: It seems weird that the executor is calling a "private" method on the future?

sloretz: Ideas? It seems weird to me too. The future needs a reference to an executor to be able to add callbacks.

`asyncio` solves the problem by expecting futures to be created via `executor.create_future()` which [passes itself as a keyword argument](https://github.com/python/cpython/blob/bec2372b7e1da5dfdbadaf242aa8e994b164cace/Lib/asyncio/base_events.py#L373-L375) to `Future.__init__()`.

Right now the client doesn't have a reference to the executor, so rclpy can't do the same thing. Instead the client holds onto the future and hopes the user will add the node to an executor.

rclpy/rclpy/client.py
sloretz: Is bf873418ea5465d1a17e9b075e50c5cca55d2a68 what you have in mind?

rclpy/rclpy/client.py
sloretz: Added check in ece3781bbd0b05cd539683a4bb769e2fdbba6135

sloretz:
CI including updates to other packages

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3944/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1062/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3274/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4043/)

rclpy/rclpy/executors.py
wjwwood: I don't have a suggestion as to how to change it, and I guess @dirk-thomas doesn't either?

So, I think it's ok to merge like this.

rclpy/rclpy/__init__.py
```diff
@@ -54,3 +54,14 @@ def spin(node):
             executor.spin_once()
     finally:
         executor.shutdown()
+
+
+def spin_until_future_complete(node, future):
```
wjwwood: Missing a doc block, but so are many existing functions. It would be nice to add, but not required I guess.

sloretz: Added in cf17f386fd9fb2eb1dd6742add50580db2412414

sloretz:
Running CI since it's been a while and cf17f38 was added.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4139/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1217/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3421/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4244/)

rclpy/rclpy/executors.py
```diff
@@ -222,11 +227,15 @@ def _take_client(self, client):
...
         sequence, response = seq_and_response
-        if sequence is not None and sequence == client.sequence_number:
-            # clients spawn their own thread to wait for a response in the
-            # wait_for_future function. Users can either use this mechanism or monitor
-            # the content of client.response to check if a response has been received
-            client.response = response
+        if sequence is not None:
+            try:
+                future = client._pending_requests[sequence]
+            except IndexError:
```
dirk-thomas: This looks wrong. `client._pending_requests` is a `dict` so if `sequence` isn't a valid key a `KeyError` would be raised.

sloretz: Agreed. It looks like there's no test coverage on canceled requests.

PR #171, opened 2018-01-08
rclpy_take_response checks sequence number

sloretz:
This fixes a bug where rclpy_take_response ignores the sequence number in a struct populated by rcl_take_response. This becomes important once the client supports multiple outstanding requests (#170).

connects to ros2/rcl#205

CI (with  e260ce6)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3883/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1002/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3209/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3972/)

### DIFF
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -44,11 +44,11 @@ class ResponseThread(threading.Thread):
         if sigint_gc_handle in guard_condition_ready_list:
             rclpy.utilities.shutdown()
             return
-        seq_and_response = _rclpy.rclpy_take_response(
+        seq, response = _rclpy.rclpy_take_response(
             self.client.client_handle,
             self.client.srv_type.Response)
-        if seq_and_response and seq_and_response[0] == self.client.sequence_number:
-            self.client.response = seq_and_response[1]
+        if seq is not None and seq == self.client.sequence_number:
+            self.client.response = response
 
 
 class Client:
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/client.py
```diff
@@ -44,12 +44,11 @@ def run(self):
         if sigint_gc_handle in guard_condition_ready_list:
             rclpy.utilities.shutdown()
             return
-        response = _rclpy.rclpy_take_response(
+        seq_and_response = _rclpy.rclpy_take_response(
```
mikaelarguedas: Could we use something lilke `[seq, response] = _rclpy.rclpy_take_response(...` for readability?

sloretz: Sure. It just means making rclpy_take_response return `(None, None)` on when there is nothing to take instead of None.

sloretz: Changed in 76a3d07d60585984df751186fd16d2c20e5d89c1

CI is running.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -2224,13 +2229,25 @@ rclpy_take_response(PyObject * Py_UNUSED(self), PyObject * args)
...
+    if (!pysequence) {
+      Py_DECREF(pytaken_response);
+      Py_DECREF(pytuple);
+      return NULL;
+    }
+    PyTuple_SET_ITEM(pytuple, 0, pysequence);
+    PyTuple_SET_ITEM(pytuple, 1, pytaken_response);
+    return pytuple;
+  }
+  Py_INCREF(Py_None);
```
mikaelarguedas: I was surprised that Py_None needed to be incref'ed but apparently that's necessary :+1:

Issue #139, opened 2017-11-01
Check return value of Python functions that can fail


sloretz:
This is a tracking ticket for checking return values in rclpy. These functions are used and can fail, but their return value is not being checked.

- [ ] `PyList_New`
- [ ] `PyCapsule_New`
- [ ] `PyLong_FromUnsignedLongLong`
- [ ] `PyMem_Malloc`
- [ ] `PyUnicode_FromString`
- [ ] `PyLong_FromSize_t`
- [ ] `PyObject_GetAtterString`
- [ ] `PyLong_FromLongLong`
- [ ] `PyCapsule_GetName`
- [ ] `PyList_Append`
- [ ] `PyObject_CallObject`
- [ ] `PyTuple_New`
- [ ] `PyTuple_SetItem`
- [ ] `PyList_SetItem`
- [ ] `PyImport_ImportModule`
- [ ] `PyObject_SetAttrString`

From discussion here
https://github.com/ros2/rclpy/pull/138#discussion_r148367820
https://github.com/ros2/rclpy/pull/138#discussion_r148363958

Labeling this as `good first issue`. This doesnt imply that a contribution for this should tackle all the functions listed above

db118:
First time contributing to open source and interested in learning more about how to fix this issue.

sloretz:
Hi @db118 ,

This issue is about checking error codes of python functions. Here is an example of error checking done correctly: `PyArg_ParseTuple(...)` can fail, and if it does the function should return `NULL` right away.

https://github.com/ros2/rclpy/blob/62012d3650de790d89c725232451c4e998b396fb/rclpy/src/rclpy/_rclpy.c#L1550-L1552

Here is an example of what should be fixed: `PyList_New()` can set an exception and return `NULL`, so `pylist` needs to be checked before calling `PyList_SET_ITEM()` with it.

https://github.com/ros2/rclpy/blob/62012d3650de790d89c725232451c4e998b396fb/rclpy/src/rclpy/_rclpy.c#L1693-L1697

A good first PR would fix one instance of a function that can fail not being checked.

db118:
@sloretz Ahhh that make sense 😄

db118:
Thank you

db118:
@sloretz 
Do you want me to switch this null error checking 
https://github.com/ros2/rclpy/blob/62012d3650de790d89c725232451c4e998b396fb/rclpy/src/rclpy/_rclpy.c#L3642-L3645

to this implementation instead
[...]

PR #172, opened 2018-01-10
Proof of concept part of cpython extension using Cython

sloretz:
This PR is meant to generate discussion on using Cython for creating parts of the cpython extension in rclpy. Cython is a superset of python that can generate C code for cpython extensions. This PR does nothing beyond successfully build (as long as `sudo apt install cython3` happens first). If the outcome of discussion is consensus in favor, I think the next step would be to replace #140 with a PR using Cython.

Advantages
* Much less human-written code (compare this with [the WaitSet in](https://github.com/ros2/rclpy/blob/34c66de41edb14b23b7d410510c289390a4035af/rclpy/src/rclpy/_rclpy_wait_set.c) #140)
* Generated code handles reference counting
* Generated code handles exceptions and checking return values from cpython functions (#139)

Disadvantages

* New dependency on cython (how to get it on windows? Vendor package?)
* Linter tests don't work on cython files
    * Some things a linter would catch are caught at build time, like trying to use an uninitialized variable
* Some burden in declaring C functions that are used (see rcl.pxd)
* cython from CMake is a little bit of a pain (it would be much easier if rclpy used a setup.py instead)

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -80,6 +80,45 @@ ament_target_dependencies(rclpy
   "rcutils"
 )
 
+# Generate code using cython
+set( CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_LIST_DIR}/cmake )
+find_package(Cython REQUIRED)
+
+function(cythonize)
+  set(options)
+  set(oneValueArgs TARGET INPUT OUTPUT)
+  set(multiValueArgs DEPENDS)
+  cmake_parse_arguments(_ARG "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+  add_custom_command( OUTPUT ${_ARG_OUTPUT}
+    COMMAND ${CYTHON_EXECUTABLE}
+    ARGS -I rclpy/ --output-file ${_ARG_OUTPUT} ${_ARG_INPUT}
+    DEPENDS ${_ARG_INPUT} ${_ARG_DEPENDS}
+    COMMENT "Cython generating code to ${_ARG_OUTPUT}"
+  )
+  add_custom_target(${_ARG_TARGET} DEPENDS ${_ARG_OUTPUT})
+endfunction()
+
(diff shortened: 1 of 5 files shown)


dirk-thomas:
I am not familiar with `Cython`. Is it correct that the result depends on the `CPython` runtime? Is that limiting us on which interpreter our code works?

I have recently read about [pybind11](https://github.com/pybind/pybind11). Maybe that would be another alternative to consider.

sloretz:
@dirk-thomas Yes, it limits us to the cpython interpreter. Cython generates C code that uses the CPython extension API. It also [supports pypy](http://cython.readthedocs.io/en/latest/src/userguide/pypy.html) because pypy implements a subset of that. The current state and pybind11 have the same interpreter limitation for the same reason.

I can look into pybind11 and make a list of pros and cons

dirk-thomas:
> The current state and pybind11 have the same interpreter limitation for the same reason.

Thanks for the clarification.

> I can look into pybind11 and make a list of pros and cons

That would be great.

clalancette:
In general, I'm in favor of using something that generates code and reduces the burden of python binding maintenance for us.  Off the top of my head, it is probably worth surveying Cython, pybind11, boost::python, and SWIG for this purpose, and seeing which fits in with our goals the best.

sloretz:
@dirk-thomas comparison to pybind11 in #173 

@clalancette I think boost::python can be skipped since pybind11 is a header only library fork of it. I think it is safe to assume pybind11 has all of the advantages without the overhead of installing boost.

mikaelarguedas:
Thanks @sloretz for putting this together.

Agreed regarding the statement about `pybind11` over `boost::python`. Some sugar from `boost::python` is not available in `pybind11` but it's very c++ specific and shouldn't impact us for C bindings.
I used `pybind11` recently in other projects and it was a good experience, actually more intuitive than `boost::python` and pretty feature complete.
I haven't played with SWIG but it could be interesting to add it to the comparison to make an educated decision.

sloretz:
I will close this for now since it seems pybind11 is most likely, but I will wait to delete the branch until that is confirmed.

PR #173, opened 2018-01-12
Proof of concept part of cpython extension using Pybind11

sloretz:
Like #172 this is a proof of concept meant to generate discussion. It implements a `WaitSet` class in c++ and uses pybind11 to create a cpython extension. To build first `pip3 install pybind11`.

Advantages over Cython

* No new language to learn: Pybind11 is pure C++, cython uses a superset of python
* c++ Linter tests work, cython files don't have a linter
* Do not need to maintain duplicate definitions for C methods ([compare with rcl.pxd](https://github.com/ros2/rclpy/pull/172/files#diff-e1b516c28082f38bf34481acbd9d7787))
* No code generation step: Pybind11 uses C++ templates that expand to python extension code at compile time. Cython has to generate C code that is then built.
* Enables rclcpp and rclpy to possibly resuse C++ code.
* pybind11 is a "lighter" dependency since it does not install any executables

Disadvantages compared to Cython

* Cython is in debian
* Cython PR #172 required slightly less code (`rcl.pxd` + `_rclpy_wait_set.px` = 208 lines, `_rcl_wait_set.cpp` = 244)
* When calling a cpython extension API directly code using pybind11 still needs to check if a python exception was raised. Cython generated does this automatically (compare call to `PyCapsule_GetPointer` here with call [using cython](https://github.com/ros2/rclpy/pull/172/files#diff-e6ba2203133d9e04c48feaabcfd02d2eR136)).
    * Not as big of a deal as it seems, being able to throw a C++ exception that gets translated to a python exception makes error handling much easier than the current state.
    * This PR uses `is_subscription_ready` instead of `ready_subscriptions` and `pysubs` is not actually the python type because it would have been a little more verbose with the checks.
* Default parameter values are separated from C++ function definition (see default value of `timeout` on wait here and when [using cython](https://github.com/ros2/rclpy/blob/8081e224f847b11ec879f17b997865a4c24925f1/rclpy/rclpy/_rclpy_wait_set.pyx#L62))

Same as Cython

* Both reduce the code needed to check for python exceptions
* Both are in homebrew
* Both limited to interpreters that implement cpython's extension API
* Both can be rolled out in incrementally: no need to rewrite all of the C code in rclpy right away.
* Both can be built using CMake or a `setup.py`
* Both are only build dependencies, the user doesn't need anything new installed at run time

Misc info

* Requires Visual Studio 2015 or higher
* Released under a BSD license
* Pybind11 is a header only library
* Pip installation of pybind11 does not come with a CMake `<project>Config.cmake`. It appears to have some nice CMake code if it is installed from source.

connects to ros2/rclpy#172

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.5)
 
-project(rclpy C)
+project(rclpy C CXX)
 
 # Default to C11
 if(NOT CMAKE_C_STANDARD)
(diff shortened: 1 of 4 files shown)


sloretz:
This is pretty stale, and just a proof of concept, so I'll close it.

PR #174, opened 2018-01-24
Get node's logger name from rcl

dhood:
connects to ros2/rcl#212

Note that the rclpy logger associated with a node won't be created until after the rcl node is created. I don't see an issue with this because we're not using the logger before that point anyway.


dhood:
waiting for review @ros2/team

PR #176, opened 2018-02-02
Make rclpy.spin*() use a persistent executor

sloretz:
This fixes a bug where coroutine callbacks that yield while being handled in `rclpy.spin_once()` never get executed again. The task is stored in the executor, but the executor is thrown away after spinning.

This PR makes the global methods reuse an executor. The test checks that a task which yields during `rclpy.spin_once()` gets resumed on a second invocation.

CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3935/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1053/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3262/)
    * Unrelated failure in `rcl.TestGetNodeNames__rmw_fastrtps_cpp.test_rcl_get_node_names`
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4033/)
    *  false positive cmake warning ros2/build_cop#79 caused by race condition due to `--parallel`

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -66,7 +66,7 @@ def spin_once(node, *, timeout_sec=None):
     if the global executor has a partially completed coroutine.
 
     :param node: A node to add to the executor to check for work.
-    :param timeout_sec: Maximum time in seconds to wait for work.
+    :param timeout_sec: Seconds to wait. Block forever if None or negative. Don't wait if 0
     :return: Always returns None regardless whether work executes or timeout expires.
     :rtype: None
     """
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -16,7 +16,7 @@ import sys
 
 from rclpy.utilities import get_rmw_implementation_identifier  # noqa
 from rclpy.utilities import ok
-from rclpy.utilities import shutdown  # noqa
+from rclpy.utilities import shutdown as _shutdown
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
@mikaelarguedas PTAL

rclpy/rclpy/__init__.py
```diff
@@ -56,6 +56,20 @@ def create_node(node_name, *, namespace=None):
...
+    Execute one item of work or wait until timeout expires.
+
+    One callback will be executed in a SingleThreadedExecutor as long as that
+    callback is ready before the timeout expires.
+
+    It is possible the work done may be for a node other than the one passed to this method
+    if the global executor has a partially completed coroutine.
+
+    :param node: A node to add to the executor to check for work.
+    :param timeout_sec: Maximum time in seconds to wait for work.
```
mikaelarguedas: Is this blocking forever if negative timeout provided? Is the default behavior to wait forever as well? (I guess it's yes in both questions, but it may be worth mentioning it in the docblock

sloretz: Yup. I copy/pasted the timeout_sec documentation from `executor.spin_once()` in 93e42b571467d241f20f7832bb46ef468640f5de

sloretz:
New commit is only a docstring change, so running CI Linux testing only rclpy to run linters

[Build Status](https://ci.ros2.org/job/ci_linux/4198/)

PR #177, opened 2018-02-02
handle node names which are null

dirk-thomas:
Necessary since each item can potentially contain a `nullptr`.

Connect to ros2/ros2#438.

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -2525,8 +2525,13 @@ rclpy_get_node_names(PyObject * Py_UNUSED(self), PyObject * args)
   PyObject * pynode_names = PyList_New(node_names.size);
   size_t idx;
   for (idx = 0; idx < node_names.size; ++idx) {
-    PyList_SetItem(
-      pynode_names, idx, PyUnicode_FromString(node_names.data[idx]));
+    if (node_names.data[idx]) {
+      PyList_SetItem(
+        pynode_names, idx, PyUnicode_FromString(node_names.data[idx]));
+    } else {
+      Py_INCREF(Py_None);
+      PyList_SetItem(pynode_names, idx, Py_None);
+    }
   }
 
   ret = rcutils_string_array_fini(&node_names);


rclpy/src/rclpy/_rclpy.c
```diff
@@ -2320,7 +2320,8 @@ rclpy_get_node_names(PyObject * Py_UNUSED(self), PyObject * args)
   size_t idx;
   for (idx = 0; idx < node_names.size; ++idx) {
     PyList_SetItem(
-      pynode_names, idx, PyUnicode_FromString(node_names.data[idx]));
+      pynode_names, idx, PyUnicode_FromString(
+        node_names.data[idx] ? node_names.data[idx] : ""));
```
sloretz: Is there a difference between a null pointer and an empty string coming from `rcl_get_node_names()`? If so, maybe this should return Py_None instead.

dirk-thomas: Absolutely. Updated the patch to use `Py_None`.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -2319,8 +2319,9 @@ rclpy_get_node_names(PyObject * Py_UNUSED(self), PyObject * args)
   PyObject * pynode_names = PyList_New(node_names.size);
   size_t idx;
   for (idx = 0; idx < node_names.size; ++idx) {
+    char * name = node_names.data[idx];
     PyList_SetItem(
-      pynode_names, idx, PyUnicode_FromString(node_names.data[idx]));
+      pynode_names, idx, name ? PyUnicode_FromString(name) : Py_None);
```
sloretz: I think `Py_INCREF(Py_None)` [is needed here](https://docs.python.org/3/c-api/none.html#c.Py_None) since `PyList_SetItem` steals a reference.

dirk-thomas: Thank you for catching this (again). Fixed in 9135c1c865dc2c96063749dfd14fd3fbd9bd07b1. Please feel free to point out the third flaw :wink:

PR #179, opened 2018-03-05
rclpy_init() passes command line arguments to rcl_init()

sloretz:
This passes command line arguments through to rcl. Previously there was only a TODO message. This PR allows python nodes to be remapped using ros2/rcl#217.

CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4016/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1111/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3342/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4107/)

connects to ros2/ros2#450

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -160,11 +160,13 @@ rclpy_init(PyObject * Py_UNUSED(self), PyObject * args)
   // Expect one argument which is a list of strings
   PyObject * pyargs;
   if (!PyArg_ParseTuple(args, "O", &pyargs)) {
+    // Exception raised
     return NULL;
   }
 
   pyargs = PySequence_List(pyargs);
   if (NULL == pyargs) {
+    // Exception raised
     return NULL;
   }
   Py_ssize_t num_args = PyList_Size(pyargs);
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -169,7 +169,12 @@ rclpy_init(PyObject * Py_UNUSED(self), PyObject * args)
     // Exception raised
     return NULL;
   }
-  Py_ssize_t num_args = PyList_Size(pyargs);
+  Py_ssize_t pysize_num_args = PyList_Size(pyargs);
+  if (pysize_num_args > INT_MAX) {
+    PyErr_Format(PyExc_OverflowError, "Too many arguments");
+    return NULL;
+  }
+  int num_args = (int)pysize_num_args;
 
   rcl_allocator_t allocator = rcl_get_default_allocator();
   char ** arg_values = allocator.allocate(sizeof(char *) * num_args, allocator.state);
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -155,14 +155,51 @@ rclpy_trigger_guard_condition(PyObject * Py_UNUSED(self), PyObject * args)
  * Raises RuntimeError if rcl could not be initialized
  */
 static PyObject *
-rclpy_init(PyObject * Py_UNUSED(self), PyObject * Py_UNUSED(args))
+rclpy_init(PyObject * Py_UNUSED(self), PyObject * args)
 {
-  // TODO(esteve): parse args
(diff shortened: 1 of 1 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -155,14 +155,51 @@ rclpy_trigger_guard_condition(PyObject * Py_UNUSED(self), PyObject * args)
...
-  // TODO(esteve): parse args
-  rcl_ret_t ret = rcl_init(0, NULL, rcl_get_default_allocator());
-  if (ret != RCL_RET_OK) {
-    PyErr_Format(PyExc_RuntimeError,
-      "Failed to init: %s", rcl_get_error_string_safe());
-    rcl_reset_error();
+  // Expect one argument which is a list of strings
+  PyObject * pyargs;
+  if (!PyArg_ParseTuple(args, "O", &pyargs)) {
+    return NULL;
```
wjwwood: Just to check, these functions setup the exceptions right? That's why only returning null here is fine?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -155,14 +155,51 @@ rclpy_trigger_guard_condition(PyObject * Py_UNUSED(self), PyObject * args)
...
+  }
+
+  pyargs = PySequence_List(pyargs);
+  if (NULL == pyargs) {
+    return NULL;
+  }
+  Py_ssize_t num_args = PyList_Size(pyargs);
```
wjwwood: I guess this cannot fail?

sloretz: It can raise SystemError and return -1 https://github.com/python/cpython/blob/745dc65b17b3936e3f9f4099f735f174d30c4e0c/Objects/listobject.c#L188 if `PyList_Check` returns false, but I don't think that's possible if `PySequence_List` above succeeds.

rclpy/src/rclpy/_rclpy.c
sloretz: Yup, [PyArg_ParseTuple](https://docs.python.org/3.5/c-api/arg.html#c.PyArg_ParseTuple) returns false and sets an exception on failure.

sloretz: Added comments about exceptions already being set in 023ccafde1fdf0026b47f19a06d978f19f946dcc

sloretz:
New CI after fixing a windows warning

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4023/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1117/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3350/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4114/)

rclpy/src/rclpy/_rclpy.c
```diff
@@ -155,14 +155,58 @@ rclpy_trigger_guard_condition(PyObject * Py_UNUSED(self), PyObject * args)
...
+  for (int i = 0; i < num_args; ++i) {
+    // Returns borrowed reference, do not decref
+    PyObject * pyarg = PyList_GetItem(pyargs, i);
+    if (NULL == pyarg) {
+      have_args = false;
+      break;
+    }
+    // Borrows a pointer, do not free arg_values[i]
+    arg_values[i] = PyUnicode_AsUTF8(pyarg);
+  }
```
mikaelarguedas: can we just skip this entire block (L180 to 197) when `num_args` is 0 and pass a `NULL` pointer to `rcl_init` instead?

sloretz: Block skipped when no args in 93c634bf20fb57def7ca3ff7517553b53d99caa3

sloretz:
CI with 93c634b
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4024/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1118/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3351/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4115/)

PR #184, opened 2018-04-17
increase timeout as if this is running with connext as the default it…

mikaelarguedas:
… takes much longer

Tested this locally and it takes ~130 seconds for me with connext.
I put 200 as I expect it to take longer on slower machines

Examples of jobs failing when ran with connext only:
https://ci.ros2.org/job/ci_osx/3482/
https://ci.ros2.org/job/ci_linux/4235/

Jobs with this change:
https://ci.ros2.org/job/ci_linux/4241

Waiting for CI to come back to place this in review


mikaelarguedas:
Build without this patch timed out on rcply tests: https://ci.ros2.org/job/ci_linux/4242/testReport/junit/(root)/projectroot/rclpytests/

With this patch it passed: https://ci.ros2.org/job/ci_linux/4243/

As this is already approved, I'm merging it without stopping by the review column

PR #191, opened 2018-06-18
call the original signal handler when receiving SIGINT during wait

wjwwood:
Fixes ros2/launch#80

Previously we just stored it and restored it after wait woke up.

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -24,6 +24,12 @@ from rclpy.timer import WallTimer
 from rclpy.utilities import ok
 from rclpy.utilities import timeout_sec_to_nsec
 
+# TODO(wjwwood): make _rclpy_wait(...) thread-safe
+# Executor.spin_once() end's up calling _rclpy_wait(...), which right now is
+# not thread-safe, no matter if differnet wait sets are used or not.
+g_wait_set_spinning_lock = Lock()
+g_wait_set_spinning = False
+
 
 class _WaitSet:
     """Make sure the wait set gets destroyed when a generator exits."""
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/executors.py
```diff
@@ -24,6 +24,12 @@
 from rclpy.utilities import ok
 from rclpy.utilities import timeout_sec_to_nsec
 
+# TODO(wjwwood): make _rclpy_wait(...) thread-safe
+# Executor.spin_once() end's up calling _rclpy_wait(...), which right now is
+# not thread-safe, no matter if differnet wait sets are used or not.
+g_wait_set_spinning_lock = Lock()
+g_wait_set_spinning = False
```
wjwwood: Note that `_rcply_wait` was not thread-safe before, but my change also does not make it thread-safe, so I thought it would be good to prevent the user from trying to use it that way.

rclpy/rclpy/executors.py
```diff
@@ -410,9 +416,6 @@ def _wait_for_ready_callbacks(self, timeout_sec=None, nodes=None):
                 clients_ready = _rclpy.rclpy_get_ready_entities('client', wait_set)
                 services_ready = _rclpy.rclpy_get_ready_entities('service', wait_set)
 
-            # Check sigint guard condition
-            if sigint_gc_handle in guards_ready:
-                raise KeyboardInterrupt
```
wjwwood: This case is no longer appropriate because the `KeyboardInterrupt` will be raised (if at all) from within `_rclpy_wait()` by the signal handler.

rclpy/rclpy/executors.py
dhood: I added https://github.com/ros2/rclpy/pull/191/commits/c1870a216c627b4c7d4436fa5a6e52e85a9ca942 to link to an example but if there are bigger issues that should be mentioned (so mentioning that specific one is misleading) we can revert it

wjwwood: Do you intend to update the issue you linked to with an example? For now I'd say just `See: <link to issue>` would be fine.

dhood: I am not sure if I have understood you, but if by example you mean rclpy code example, I was gonna link to a specific commit once this PR was merged

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4742/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1610/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3930/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4802/)

dhood:
Linter fix [Build Status](https://ci.ros2.org/job/ci_linux/4756/)

dhood:
@wjwwood can we merge this?

wjwwood:
Yes, thanks for fixing the linter error.

PR #193, opened 2018-06-20
avoid leak of sigint guard condition used in wait sets

wjwwood:
I don't think this needs to go in for Bouncy, but it's just a small bug fix that we noticed when making some other changes.

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -376,7 +376,6 @@ class Executor:
                     if gc._executor_triggered:
                         gc.trigger()
                     guards.append(gc)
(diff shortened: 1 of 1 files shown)


sloretz:
CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4955/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1775/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4114/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4971/)

PR #199, opened 2018-06-28
Fix exception handling

### LINKED ISSUE #198: exception in callback is not being reported
E.g. when the `talker` in the `demo_nodes_py` package is modified to contain a syntax error in the timer callback that is not being reported anywhere.

dirk-thomas:
Fixes #198.

Looking for feedback on this patch.

### DIFF
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -55,8 +55,6 @@ class Client:
         future.add_done_callback(unblock)
 
         event.wait()
-        if future.exception() is not None:
-            raise future.exception()
         return future.result()
 
     def remove_pending_request(self, future):
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -55,6 +55,8 @@ class Client:
         future.add_done_callback(unblock)
 
         event.wait()
+        if future.exception() is not None:
+            raise future.exception()
         return future.result()
 
     def remove_pending_request(self, future):
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -521,7 +521,7 @@ class SingleThreadedExecutor(Executor):
             pass
         else:
             handler()
-            if handler.exception():
+            if handler.exception() is not None:
                 raise handler.exception()
 
 
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -521,8 +521,8 @@ class SingleThreadedExecutor(Executor):
             pass
         else:
             handler()
-            # check result to re-raise potential exceptions
-            handler.result()
+            if handler.exception():
+                raise handler.exception()
 
 
 class MultiThreadedExecutor(Executor):
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -551,6 +551,4 @@ class MultiThreadedExecutor(Executor):
         except TimeoutException:
             pass
         else:
-            future = self._executor.submit(handler)
-            # currently the future is not been waited on?
-            # the result needs to be checked in order to re-raise potential exceptions
+            self._executor.submit(handler)
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -521,6 +521,8 @@ class SingleThreadedExecutor(Executor):
             pass
         else:
             handler()
+            # check result to re-raise potential exceptions
+            handler.result()
 
 
 class MultiThreadedExecutor(Executor):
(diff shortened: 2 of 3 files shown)


rclpy/rclpy/task.py
```diff
@@ -79,6 +81,12 @@ def result(self):
 
         :return: The result set by the task
         """
+        if self._cancelled:
+            raise CancelledError
+        if not self._done:
+            raise Error
+        if self._exception is not None:
+            raise self._exception
```
dirk-thomas: This is following the logic of the asyncio future.

rclpy/rclpy/client.py
```diff
@@ -55,8 +55,6 @@ def unblock(future):
         future.add_done_callback(unblock)
 
         event.wait()
-        if future.exception() is not None:
-            raise future.exception()
```
dirk-thomas: With the change to `result()` this is not necessary anymore since it happens implicitly.

rclpy/rclpy/executors.py
```diff
@@ -549,4 +551,6 @@ def spin_once(self, timeout_sec=None):
         except TimeoutException:
             pass
         else:
-            self._executor.submit(handler)
+            future = self._executor.submit(handler)
+            # currently the future is been waited on?
+            # the result needs to be checked in order to re-raise potential exceptions
```
dirk-thomas: What should happen here?

dirk-thomas:
We can't check for `done` in `result()` (see 8b9fe323211e0cc1505def549ff48ab7cf71e0de) since result is now being called after a `spin_once()`. A coroutines which yields won't be finished after a single spin though.

This should at least pass our tests now. Back in review.

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4836/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1690/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4015/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4881/)

rclpy/rclpy/task.py
```diff
@@ -42,6 +44,12 @@ def __init__(self, *, executor=None):
         self._executor = None
         self._set_executor(executor)
 
+    def __del__(self):
+        if self._exception is not None:
```
wjwwood: This will print even for exceptions that were retrieved, is that intended?

rclpy/rclpy/task.py
```diff
@@ -79,6 +87,13 @@ def result(self):
 
         :return: The result set by the task
         """
+        if self._cancelled:
+            raise CancelledError
+        # a coroutine might not have finished after calling spin_once
+        # if not self._done:
+        #     raise Error
```
wjwwood: Is this the right behavior? I believe that in the other Future's Python provides, calling result before being set is an error (you should check if it is done first).

If calling `result` prematurely is an issue, and we only care about whether or not there was an exception, why not just do something like this in the executors:

```python
if future.exception() is not None:
    raise future.exception()
```

dirk-thomas: > Is this the right behavior?

This is the current behavior of our coroutines / spinning them. The commented code is what Python's future class does but we can't enforce that at the moment.

> why not just do something like this in the executors

We can do that instead. We don't even have to change the `result()` method at all for now.

rclpy/rclpy/task.py
dirk-thomas: That is correct. Unfortunately I don't see a way to avoid that since there is no API calling code invoked to "take" an exception.

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4837/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1691/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4016/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4882/)

rclpy/rclpy/task.py
dirk-thomas: I added a boolean flag. When `exception()` is being called it won't be reported assuming the caller did "the right thing" already.

rclpy/rclpy/task.py
```diff
@@ -42,6 +44,12 @@ def __init__(self, *, executor=None):
         self._executor = None
         self._set_executor(executor)
 
+    def __del__(self):
+        if self._exception is not None and self._exception_fetched:
```
wjwwood: `and not self._exception_fetched`?

rclpy/rclpy/task.py
dirk-thomas: I removed all the checking from `result()` for now. We should revisit this though after the release (see follow up ticket #200).

rclpy/rclpy/task.py
dirk-thomas: Very good point. I changed the logic to match the naming of the 
variable: c63e3be660521b8473cc6e667212669da0c8044f

Issue #205, opened 2018-07-16
Executor..shutdown() should wake calls to `executor.spin...()`


sloretz:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Installation type:
  - source
- Version or commit hash:
  - 98ad24ae08f553c46c683e98350333cd0bda42cd
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

```python
import threading
import time

import rclpy
from rclpy.executors import SingleThreadedExecutor


if __name__ == '__main__':
    print('initializing rclpy')
    rclpy.init()
    node = rclpy.create_node('NodeName', namespace='/rclpy')

    print('Start executor in a new thread')
    executor = SingleThreadedExecutor()
    executor.add_node(node)
    exec_thread = threading.Thread(target=executor.spin)
    exec_thread.start()
    print('Sleep to make sure executor is blocked in rclpy_wait()')
    time.sleep(1)

    print('Shutdown executor and library')
    executor.shutdown()
    node.destroy_node()
    rclpy.shutdown()

    print('Thread cannot be joined because executor.spin() never returns')
    exec_thread.join()
    print('If this prints the issue is fixed')
```

#### Expected behavior

The thread in which spin is called should be joinable. `executor.spin()` should return when the executor is shutdown because there's no more work for it to do.

#### Actual behavior

The thread never becomes joinable.

sloretz:
fixed by ros2/rclpy#309

PR #206, opened 2018-07-16
Create entity wake executor

### LINKED ISSUE #188: node.create_*() should wake executor
When an executor spins it builds a wait list containing every entity (clients, services, publishers, etc) from the nodes added to it. If a new entity is created while the executor is spinning then it won't get added to the wait list until the executor executes unrelated work.

Creating a subscriber, client, etc should wake the executor.

sloretz:
This probably fixes #188, but I haven't figured out how to write a test that isn't blocked by #205.

Every node gets a guard condition that is triggered when an entity is created or destroyed. The guard condition triggering will wake the executor which will cause it to rebuild the wait set.

The test is broken. It requires an executor to be blocked in `rclpy_wait()` when an entity is created/destroyed. However, `rclpy_wait()` is not woken when an executor is shutdown, and the blocked executor lives forever. Every other test using an executor fails because #191 uses a global variable to prevent two executors running at the same time and it seems pytest runs all tests in the same python process.

connects to ros2/rclpy#188
blocked by ros2/rclpy#205

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from rclpy.callback_groups import MutuallyExclusiveCallbackGroup
+from rclpy.callback_groups import ReentrantCallbackGroup
 from rclpy.client import Client
 from rclpy.constants import S_TO_NS
 from rclpy.exceptions import NotInitializedException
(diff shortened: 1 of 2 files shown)


sloretz:
Replaced by #336

PR #207, opened 2018-07-17
Revert "use test_msgs instead of std_msgs"

mikaelarguedas:
Reverts ros2/rclpy#204

My bad I was holding #204 to cut the release from master and fast forward it to the `bouncy` branch before merging...
So I'm reverting #204, cutting the release and then will reapply it.


sloretz:
Since this only changes tests and a `<test_depend>`, would it be harmless to include this in the patch release?

mikaelarguedas:
it should not be impactful for users. It just doesn't "need" to be backported but should be fine to keep it as part of the patch release

mikaelarguedas:
closing this then and cutting the release with the current state from master

Issue #186, opened 2018-04-26
Implement Time Features in rclpy


tfoote:
This ticket is to track rclpy specific time impelementation requirements

 * [x] implement a Clock on top of rcl implementation: ros2/rclpy#209
 * [x] implement a TimeSource to implement the [ROS2 time design](http://design.ros2.org/articles/clock_and_time.html): ros2/rclpy#210
    * [x] use parameters for `use_sim_time` ~when~ now that parameters are available in rclpy. https://github.com/ros2/rclpy/issues/240
    * [x] clock jump callbacks: https://github.com/ros2/rclpy/pull/222
 * [x] upgrade Timer to use a Clock on top of rcl implementation
    * [x] functionality using rcl clocks: https://github.com/ros2/rclpy/pull/211
    * [x] let users specify the clock to use: update [this TODO](https://github.com/ros2/rclpy/pull/211/files#diff-6bb1cc25474f2c2bb8dea36798a89a27R24) #419
 * [ ] Implement a Rate method using the clock on top of rcl implementation

Nickolaim:
I can take this item if nobody is looking at it yet.

tfoote:
I don't think anyone is actively working on this. If you'd like to start work on the clock, timer and rate that would be feasible. The time source is blocked on the availability of parameters. If you'd like to pick it up I'd suggest creating a separate ticket for one of the smaller parts to sketch out how you're going to approach it to get feedback before you jump in and implement it.

dhood:
Here's some additional detail about the steps that would be required.

- Time class wrapping rcl types (`rcl_time_point`, `rcl_duration`) and functions (`rcl_difference_times`, `rcl_get_time_now`). Conversions to/from builtin_interfaces msg type.
- Clock class wrapping rcl's `rcl_clock_t`, exposing get_now; Node has a clock; demos using node.now() if appropriate.
- TimeSource wrapping rcl; implementation of ROS time support: A node is used to create a TimeSource of "ROS time" (subscription to /clock, use_sim_time support by some means), node's clock is managed by the TimeSource.
- Time jump types (`TimeJump`, `JumpThreshold`) and logic moved from rclcpp to rcl, wrapped in python. Jump handlers in python.
- Timer using specific Clock (rclpy Timer currently wraps an rcl timer that rcl_wait uses; would be adapting to changes in rcl's timer to use a specific clock).
- Addition of Rate in a similar manner to what's done for rclcpp.

stonier:
[...]

PR #210, opened 2018-07-20
TimeSource and support for ROS time

dhood:
Connects to ros2/rclpy#186

Requires ros2/rcl#274

Gives nodes a clock of type ROS_TIME which will reflect the time received on `/clock` topic if the Time Source has had use of ROS time activated. Since parameters aren't available in python yet, there's no support for the `use_sim_time` parameter, but the Time Source can have ROS time activated manually by calling `time_source.ros_time_is_active = True`.

Support for clock jump notifiers not targeted for this PR.

### DIFF
--- a/rclpy/rclpy/time_source.py
+++ b/rclpy/rclpy/time_source.py
@@ -39,7 +39,7 @@ class TimeSource:
     def ros_time_is_active(self, enabled):
         self._ros_time_is_active = enabled
         for clock in self._associated_clocks:
-            clock.ros_time_is_active = True
+            clock._set_ros_time_is_active(True)
 
     def _subscribe_to_clock_topic(self):
         if self._clock_sub is not None and self._node is not None:
@@ -76,7 +76,7 @@ class TimeSource:
             raise ValueError('Only clocks with type ROS_TIME can be attached.')
         if self._last_time_set is not None:
             self._set_clock(self._last_time_set, clock)
-        clock.ros_time_is_active = self.ros_time_is_active
+        clock._set_ros_time_is_active(self.ros_time_is_active)
         self._associated_clocks.append(clock)
 
     def clock_callback(self, msg):
--- a/rclpy/rclpy/time_source.py
+++ b/rclpy/rclpy/time_source.py
@@ -1,0 +1,78 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import builtin_interfaces.msg
+from rclpy.clock import Clock
+from rclpy.clock import ClockType
+from rclpy.node import Node
+from rclpy.time import Time
+
+CLOCK_TOPIC = '/clock'
+
+
+class TimeSource:
+
+    def __init__(self, *, node=None):
+        self._clock_sub = None
+        self._node = None
+        self._associated_clocks = []
+        self._last_time_set = None
+        if node is not None:
+            self.attach_node(node)
+
+    def attach_node(self, node):
+        if not isinstance(node, Node):
+            raise TypeError('Node must be of type rclpy.node.Node')
+        # Remove an existing node.
+        if self._node is not None:
+            self.detach_node()
+
+        # Create a subscription to the clock topic using the node.
+        self._clock_sub = node.create_subscription(
+            builtin_interfaces.msg.Time,
+            CLOCK_TOPIC,
+            self.clock_callback
+        )
+        self._node = node
+
+    def detach_node(self):
+        # Remove the subscription to the clock topic.
+        if self._clock_sub is not None:
+            if self._node is None:
+                print('Unable to destroy previously created clock subscription')
+            else:
+                self._node.destroy_subscription(self._clock_sub)
+        self._clock_sub = None
+        self._node = None
+
+    def attach_clock(self, clock):
+        if not isinstance(clock, Clock):
+            raise TypeError('Clock must be of type rclpy.clock.Clock')
+        if not clock.clock_type == ClockType.ROS_TIME:
+            raise ValueError('Only clocks with type ROS_TIME can be attached.')
+        if self._last_time_set is not None:
+            self._set_clock(self._last_time_set, clock)
+        self._associated_clocks.append(clock)
+
+    def clock_callback(self, msg):
+        # Cache the last message in case a new clock is attached.
+        time_from_msg = Time.from_msg(msg)
+        self._last_time_set = time_from_msg
+        for clock in self._associated_clocks:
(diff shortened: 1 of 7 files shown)


dhood:
Putting this in review

The description of this PR has been updated to reflect a recent change to the code: Following discussion in https://github.com/ros2/rclcpp/issues/516 I understand now that all clocks managed by a time source should have the same value for `ros_time_is_active`. So, I have added `ros_time_is_active` to the `TimeSource` class, and made it read-only on the `ROSClock` class (before clocks could individually have use of ROS time configured on them, now it is 'inherited' from the TimeSource's state).

The `ROSClock` class has also been added recently as a specialisation of the `Clock` class that provides ROS time-specific methods. A time source can still take in a clock like `Clock(clock_type=ROS_TIME)`, but the expected usage is that it will be passed a `ROSClock` (that is what will be stored underneath). I did this because I thought it was cleaner than having methods that only applied for ROS time in the general Clock class (e.g. [like this](https://github.com/ros2/rclpy/blob/69a98cab46ef8e0a20c1fc8950d2e99d15fc0695/rclpy/rclpy/clock.py#L70..L73)), but am open to alternatives.

Overall I tried to mirror the implementation in rclcpp, where possible. It diverges in the following ways:

- Clock callbacks have not been implemented yet
- Support for parameters isn't available in rclpy yet so the TimeSource has to have `ros_time_is_active` set manually as opposed to with the `use_sim_time` parameter
- A ROSClock specialisation does not exist in rclcpp, but does in rclpy.
- I aimed for the end-result from https://github.com/ros2/rclcpp/issues/515 where a subscription is only created for a Timesource if sim time is in use. That is not the _current_ state in rclcpp however.

rclpy/rclpy/time_source.py
```diff
@@ -0,0 +1,96 @@
...
+        self._clock_sub = None
+        self._node = None
+
+    def attach_clock(self, clock):
+        if not isinstance(clock, Clock):
+            raise TypeError('Clock must be of type rclpy.clock.Clock')
+        if not clock.clock_type == ClockType.ROS_TIME:
+            raise ValueError('Only clocks with type ROS_TIME can be attached.')
+        # "Cast" to ROSClock subclass so ROS time methods can be used.
+        clock.__class__ = ROSClock
```
sloretz: Not asking to change anything, just sharing because I think it's interesting. This led me discover that `__new__` could be used to return a `ROSClock` instance when passed a ROS clock type.

<details>
<summary>Details</summary>

```python

class Clock:

    def __new__(cls, *, clock_type="SYSTEM_CLOCK"):
        self = None
        if clock_type == "ROS_CLOCK":
            self = super().__new__(ROSClock)
        else:
            self = super().__new__(cls)
        self._clock_type = clock_type
        # create self._clock_handle here too
        return self

    # Can't have an __init__ because it would get passed clock_type argument
    # and ROSCLOCK doesn't wan't one
    # def __init__(self):
    #   pass

    def all_clocks_have_this(self):
        print("all clocks have this")


class ROSClock(Clock):

    # Can't have an __init__ because it would get passed clock_type argument
    # and ROSCLOCK doesn't wan't one
    # def __init__(self):
    #   pass

    def only_ros_clocks_have_this(self):
        print("Only ros clocks have this")

c1 = Clock()
c2 = Clock(clock_type="STEADY_CLOCK")
c3 = Clock(clock_type="ROS_CLOCK")

print(repr(c1))
print(repr(c2))
print(repr(c3))

c1.all_clocks_have_this()
c2.all_clocks_have_this()
c3.all_clocks_have_this()
c3.only_ros_clocks_have_this()
```
</details>

rclpy/test/test_time_source.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def test_time_source_not_using_sim_time(self):
+        time_source = TimeSource(node=self.node)
+        clock = ROSClock()
+        time_source.attach_clock(clock)
+
+        # When not using sim time, ROS time should look like system time
+        now = clock.now()
+        system_now = Clock(clock_type=ClockType.SYSTEM_TIME).now()
+        assert (system_now.nanoseconds - now.nanoseconds) < 1e9
```
sloretz: `1e9` seems really small. I think this test could be flaky if the OS happens to context switch between lines 73 and 74.

Also, maybe add `abs()` in case there is a future bug where `now` is a non-zero time greater than `system_now`?

rclpy/rclpy/time_source.py
dhood: I think this approach might be better! Because there's no extra step for users to get access to ROS clock methods if they have created a ROS clock with `Clock(clock_type=ClockType.ROS_TIME)`, they will always get a ROSClock back.

I don't immediately see any disadvantages of this (it's ok in this instance that the Clock class 'knows about' the ROSClock class), so I plan to incorporate it, thank you for the suggestion

rclpy/test/test_time_source.py
dhood: I was thinking 1sec is pretty generous already but it can't hurt to increase.

About the abs, I don't think it makes sense here, since we _do_ expect the value of a later call to now() be greater. I'd rather catch that bug than mask it.

sloretz: Ooops, I got this completely wrong. For some reason I thought `1e9` meant 1ns. My bad.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3246,6 +3246,136 @@ rclpy_clock_get_now(PyObject * Py_UNUSED(self), PyObject * args)
...
+ * \param[in] enabled Override state to set
+ * \return NULL on failure
+ *         None on success
+ */
+static PyObject *
+rclpy_clock_set_ros_time_override_is_enabled(PyObject * Py_UNUSED(self), PyObject * args)
+{
+  PyObject * pyclock;
+  PyObject * pyenabled;
+  if (!PyArg_ParseTuple(args, "OO", &pyclock, &pyenabled)) {
```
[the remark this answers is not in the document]
dhood: thanks! 6a9d42b019f9ab533b6515f218ec75a6d97ca2d1

rclpy/rclpy/time_source.py
dhood: Done in 9992b1f224570c9f18f5ed26156e457602a267fb + cf38ae765503bff335b624c890ca5d0a7d30e227, I think it's better to use now: happy you pointed this out!

dhood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4999/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1811/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4155/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5012/)

dhood:
Interestingly OS X (only) is failing with `ValueError: PyCapsule_GetPointer called with invalid PyCapsule object` when `_rclpy.rclpy_clock_set_ros_time_override_is_enabled(self._clock_handle, enabled)` is called. Not sure yet what's causing that: will have to look into it before this can be merged

dhood:
Back in action!
Will go ahead and merge this now the bug's been fixed in 6ac5bf98fbba8a62d8f2f59fc39bdc481112726e

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5002/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1814/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4159/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5015/)

PR #209, opened 2018-07-18
Time, Duration, Clock wrapping rcl

dhood:
connects to ros2/rclpy#186

Still a work-in-progress but input is welcome.

The scope of this PR is the first few steps of https://github.com/ros2/rclpy/issues/186#issuecomment-405412508. It excludes TimeSource/ROS time implementation to avoid the PR getting too large.

PR #211, opened 2018-07-24
create timer with clock

dirk-thomas:
Match API change from ros2/rcl#272.


rclpy/src/rclpy/_rclpy.c
```diff
@@ -2061,6 +2074,16 @@ rclpy_destroy_entity(PyObject * Py_UNUSED(self), PyObject * args)
     rcl_timer_t * timer = (rcl_timer_t *)PyCapsule_GetPointer(pyentity, "rcl_timer_t");
     ret = rcl_timer_fini(timer);
     PyMem_Free(timer);
+
+    rcl_clock_t * clock = (rcl_clock_t *)PyCapsule_GetPointer(pyentity, "rcl_clock_t");
+    rcl_ret_t ret_clock = rcl_clock_fini(clock);
+    PyMem_Free(timer);
```
wjwwood: Is this supposed to be `clock` instead?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1193,20 +1193,33 @@ rclpy_create_timer(PyObject * Py_UNUSED(self), PyObject * args)
     return NULL;
   }
 
+  rcl_clock_t * clock = (rcl_clock_t *) PyMem_Malloc(sizeof(rcl_clock_t));
```
wjwwood: I see that `timer` already gets allocated with `PyMem_Malloc`, but what I'm wondering is why we do this rather use the `rcl_allocator_t`? Maybe @sloretz or whoever did this originally has an idea. Even if we don't have to use `PyMem_Malloc`, it might be the right preference anyways. It's just that we're mixing the two, which happen to be the same right now, but are not necessarily both using `malloc` under the hood.

We control both the allocation and free of these resources, so either should work I think. We're never relying on Python to free them for us (I think).

dhood:
could you comment on the next steps for this? we have Clocks in rclpy now (I just merged https://github.com/ros2/rclpy/pull/209) and I'm wondering if we plan to let users pass a clock into the Timer in the rclpy API at a later date or not (probably this PR is just to keep it compiling with changes in connected PRs and change to the API is followup, but want to double check)

dirk-thomas:
I would suggest to get the current PRs merged. For `rcl` and `rclcpp` they add the clock argument to the timer classes. In `rclpy` the patch only updates the code which uses that function. So the current (wall) timer continues to work.

In a follow up PR we can add new API (probably to the Node class) to create timers with an arbitrary clock type. Internally it has to maintain (at least) one clock per type. Additional timers with the same clock type should be able to share the clock.

rclpy/src/rclpy/_rclpy.c
sloretz: I don't have a good reason for using this or `mallloc` via an `rcl_allocator_t`.  There is more info about [pymalloc here](https://docs.python.org/3/c-api/memory.html#pymalloc). As long as memory is free'd by the same allocator it was allocated with then I wouldn't expect problems from mixing the two.

> We control both the allocation and free of these resources, so either should work I think. We're never relying on Python to free them for us (I think).

AFAIK this is correct.

PR #212, opened 2018-07-27
Alternative way of giving timers a clock

sloretz:
I was having trouble putting feedback for ros2/rclpy#211 into words. This passes the clock into `rclpy_create_timer()` instead of it being returned in an unused variable, and makes use of the cleanup code in `_rclpy_destroy_clock()`.

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -248,7 +248,7 @@ class Node:
             if tmr.timer_handle == timer.timer_handle:
                 _rclpy.rclpy_destroy_entity(tmr.timer_handle)
                 # TODO(sloretz) Store clocks on node and destroy them separately
-                _rclpy.rclpy_destroy_entity(tmr._clock._clock_handle)
+                _rclpy.rclpy_destroy_entity(tmr.clock._clock_handle)
                 self.timers.remove(tmr)
                 return True
         return False
@@ -282,7 +282,7 @@ class Node:
             tmr = self.timers.pop()
             _rclpy.rclpy_destroy_entity(tmr.timer_handle)
             # TODO(sloretz) Store clocks on node and destroy them separately
-            _rclpy.rclpy_destroy_entity(tmr._clock._clock_handle)
+            _rclpy.rclpy_destroy_entity(tmr.clock._clock_handle)
         while self.guards:
             gc = self.guards.pop()
             _rclpy.rclpy_destroy_entity(gc.guard_handle)
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -247,6 +247,8 @@ class Node:
         for tmr in self.timers:
             if tmr.timer_handle == timer.timer_handle:
                 _rclpy.rclpy_destroy_entity(tmr.timer_handle)
+                # TODO(sloretz) Store clocks on node and destroy them separately
+                _rclpy.rclpy_destroy_entity(tmr._clock._clock_handle)
                 self.timers.remove(tmr)
                 return True
         return False
@@ -279,6 +281,8 @@ class Node:
         while self.timers:
             tmr = self.timers.pop()
             _rclpy.rclpy_destroy_entity(tmr.timer_handle)
+            # TODO(sloretz) Store clocks on node and destroy them separately
+            _rclpy.rclpy_destroy_entity(tmr._clock._clock_handle)
         while self.guards:
             gc = self.guards.pop()
             _rclpy.rclpy_destroy_entity(gc.guard_handle)
--- a/rclpy/rclpy/timer.py
+++ b/rclpy/rclpy/timer.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from rclpy.clock import Clock
+from rclpy.clock import ClockType
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
 
@@ -19,7 +21,10 @@ from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 class WallTimer:
 
     def __init__(self, callback, callback_group, timer_period_ns):
-        [self.timer_handle, self.timer_pointer, _] = _rclpy.rclpy_create_timer(timer_period_ns)
+        # TODO(sloretz) Allow passing clocks in via timer constructor
+        self._clock = Clock(clock_type=ClockType.STEADY_TIME)
+        [self.timer_handle, self.timer_pointer] = _rclpy.rclpy_create_timer(
+            timer_period_ns, self._clock._clock_handle)
         self.timer_period_ns = timer_period_ns
         self.callback = callback
         self.callback_group = callback_group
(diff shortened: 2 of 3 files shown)


rclpy/rclpy/node.py
```diff
@@ -247,6 +247,8 @@ def destroy_timer(self, timer):
         for tmr in self.timers:
             if tmr.timer_handle == timer.timer_handle:
                 _rclpy.rclpy_destroy_entity(tmr.timer_handle)
+                # TODO(sloretz) Store clocks on node and destroy them separately
+                _rclpy.rclpy_destroy_entity(tmr._clock._clock_handle)
```
sloretz: A ROSTime clock would need to be fed from a time source that is tied to the node. Clocks may need to be gotten off a `Node` instance, so it would make sense for that instance to store and destroy them later.

rclpy/rclpy/timer.py
```diff
@@ -12,14 +12,19 @@
...
 
 
 # TODO(mikaelarguedas) create a Timer or ROSTimer once we can specify custom time sources
 class WallTimer:
 
     def __init__(self, callback, callback_group, timer_period_ns):
-        [self.timer_handle, self.timer_pointer, _] = _rclpy.rclpy_create_timer(timer_period_ns)
+        # TODO(sloretz) Allow passing clocks in via timer constructor
+        self._clock = Clock(clock_type=ClockType.STEADY_TIME)
+        [self.timer_handle, self.timer_pointer] = _rclpy.rclpy_create_timer(
```
sloretz: The returned clock was unused. Since a ROSTime clock may need to be retrieved off a Node it makes sense to pass the clock in, and reuse the clock creation code in `Clock`.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -2073,25 +2082,11 @@ rclpy_destroy_entity(PyObject * Py_UNUSED(self), PyObject * args)
...
-      rcl_reset_error();
-      return NULL;
-    }
-
     ret = rcl_timer_fini(timer);
     PyMem_Free(timer);
-    PyMem_Free(clock);
+  } else if (PyCapsule_IsValid(pyentity, "rcl_clock_t")) {
+    PyCapsule_SetDestructor(pyentity, NULL);
+    _rclpy_destroy_clock(pyentity);
```
sloretz: The pycapsule destructor will fini the clock if the user doesn't. If the clock is destroyed explicitly then fini the clock, but set the destructor to NULL so the garbage collector doesn't do it again.

sloretz:
@dirk-thomas if the changes look OK to you feel free to merge this PR. It's targeted at branch `timer-with-clock`

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1182,45 +1182,40 @@ rclpy_publish(PyObject * Py_UNUSED(self), PyObject * args)
...
  * \return NULL on failure
  */
 static PyObject *
 rclpy_create_timer(PyObject * Py_UNUSED(self), PyObject * args)
 {
   unsigned PY_LONG_LONG period_nsec;
+  PyObject * pyclock;
 
-  if (!PyArg_ParseTuple(args, "K", &period_nsec)) {
+  if (!PyArg_ParseTuple(args, "KO", &period_nsec, &pyclock)) {
```
dirk-thomas: Should the clock be based first and the period second (as in the other languages)?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -2046,6 +2041,20 @@ rclpy_destroy_node_entity(PyObject * Py_UNUSED(self), PyObject * args)
   Py_RETURN_NONE;
 }
 
+/// Destructor for a clock
+void
+_rclpy_destroy_clock(PyObject * pycapsule)
+{
+  rcl_clock_t * clock = (rcl_clock_t *)PyCapsule_GetPointer(pycapsule, "rcl_clock_t");
```
dirk-thomas: The code should check for `!clock`.

rclpy/rclpy/node.py
dirk-thomas: Instead of accessing a "protected" member (`tmr._clock`) should this be exposed through a getter?

rclpy/src/rclpy/_rclpy.c
sloretz: https://github.com/ros2/rclpy/pull/212/commits/d82ca21d7c11e41d42b553791b2365698fed3715

rclpy/src/rclpy/_rclpy.c
sloretz: https://github.com/ros2/rclpy/pull/212/commits/61fac0c9f86212373ee4e2a4426de3fd3da219ae

rclpy/rclpy/node.py
sloretz: https://github.com/ros2/rclpy/pull/212/commits/d4c43c63ecedb30ffe6de7888f13c14f99a7533e

sloretz:
@dirk-thomas CI testing `rclpy` with this branch + rcl `timer-with-clock`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4967/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1784/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4127/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4983/)

rclpy/src/rclpy/_rclpy.c
dirk-thomas: So currently it silently returns?

dirk-thomas: Sorry, I still don't understand it. If `clock` is `NULL` that means a wrong parameter was passed - not a `rcl_clock_t` capsule. How is this early return different from the case where `clock` is not `null` and correctly finalized?

sloretz: > How is this early return different from the case where clock is not null and correctly finalized?

If `PyCapsule_GetPointer()` returns `NULL` then it raised an exception. If the clock is correctly finalized then no exception is raised.

>  If clock is NULL that means a wrong parameter was passed

I don't see a way to pass a wrong parameter to this function at the moment. It shouldn't get the wrong type of pointer when using it as a capsule destructor.

https://github.com/ros2/rclpy/blob/5ed7ca2755ad8dbee293858f50f2a2707342624d/rclpy/src/rclpy/_rclpy.c#L3184

And the new call in this PR calls `PyCapsule_IsValid()` first. `PyCapsule_GetPointer()` is guaranteed not to return `NULL` if that passes.

dirk-thomas: Thanks for the detailed description. I guess I was kind of confused by the comment - I updated it in a new commit: a6934ca5013637942f75c23150af8597c2f99343

dirk-thomas:
> Connext not installed?

Fixed in https://github.com/ros2/ci/pull/212#issuecomment-408554193

PR #214, opened 2018-08-06
Implement parameter services

nuclearsandwich:
This adds parameter services to rclpy nodes. Like rclcpp nodes, parameter services are started by default and can be turned off with the create_node/Node constructor kwarg `start_parameter_services=False`.

Parameters are stored on the Node class via a `_parameters` dict attribute. The methods following methods have been added to the Node API.
* `get_parameter(name)`
* `get_parameters(names)`
* `set_parameters(parameters)`
* `set_parameters_atomically(parameters)`

A Parameter class provides a python interface to individual parameters that is (hopefully) a bit more ergonomic than using the Parameter-related message types from rcl_interfaces.


rclpy/rclpy/parameter.py
```diff
@@ -0,0 +1,99 @@
...
+    def value(self):
+        return self._value
+
+    def get_descriptor(self):
+        return ParameterDescriptor(name=self.name, type=self.type.value)
+
+    def get_parameter_value(self):
+        parameter_value = ParameterValue(type=self.type.value)
+        if Parameter.Type.BOOL == self.type:
+            parameter_value.bool_value = self.value
```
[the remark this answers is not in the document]
nuclearsandwich: I think it's possible to enforce types on the Parameters. An earlier implementation of the Parameter class determined the type from the value given to the constructor but when I added the Parameter.Type enum to wrap the interface enum values I dropped all of that.

rclpy/rclpy/parameter.py
```diff
@@ -0,0 +1,99 @@
...
+        elif Parameter.Type.INTEGER_ARRAY == type_:
+            value = rcl_param.value.integer_array_value
+        elif Parameter.Type.DOUBLE_ARRAY == type_:
+            value = rcl_param.value.double_array_value
+        elif Parameter.Type.STRING_ARRAY == type_:
+            value = rcl_param.value.string_array_value
+        return cls(rcl_param.name, type_, value)
+
+    def __init__(self, name, type_, value):
+        assert isinstance(type_, Parameter.Type)
```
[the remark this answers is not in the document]
nuclearsandwich: I wasn't certain if `TypeError`s were the correct exception and there were other `assert isinstance(...` instances throughout. I'd be inclined to flip these to TypeErrors if that's the usual pattern.

rclpy/rclpy/parameter.py
```diff
@@ -0,0 +1,99 @@
...
+        return self._type
+
+    @property
+    def value(self):
+        return self._value
+
+    def get_descriptor(self):
+        return ParameterDescriptor(name=self.name, type=self.type.value)
+
+    def get_parameter_value(self):
```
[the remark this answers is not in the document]
nuclearsandwich: If per your later question we end up enforcing parity between declared type and given value that closes this odd behavior.

I ended up creating a positional args constructor for Parameter but maybe I should change it to be a kwarg constructor to make it possible to omit the value for a 'not set' parameter. Deleting parameters isn't currently part of the rclcpp parameter API but we could make it part of the rclpy API if we want.

rclpy/rclpy/parameter_service.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class ParameterService:
+
+    def __init__(self, node):
+        self._node = node
+        nodename = node.get_name()
+
+        describe_parameters_service_name = '/'.join((nodename, 'describe_parameters'))
+        node.create_service(
```
mikaelarguedas: I think we should use the parameter qos profile by default for all these services (I believe that would match rclcpp's implementation): https://github.com/ros2/rclpy/blob/98cdadeb9befa64eac3915b17e65f8ac02e1bec4/rclpy/rclpy/qos.py#L179

And the parameter event publisher should use the corresponding profile as well (but I think this will be in a follow-up PR)

mikaelarguedas:
This looks great !

I didn't get a chance to thoroughly review it yet so just dropped a note for the questions I had. I'll give a review tomorrow morning if noone gets to it by then

rclpy/rclpy/parameter.py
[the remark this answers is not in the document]
nuclearsandwich: per today's offline discussion we'll add code to parameter initialization to prevent setting values that don't match declared type (including setting any value for a parameter of type NOT_SET.

rclpy/rclpy/parameter.py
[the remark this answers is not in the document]
nuclearsandwich: per today's offline discussion we'll add code to parameter initialization to prevent setting values that don't match declared type (including setting any value for a parameter of type NOT_SET.

rclpy/rclpy/parameter.py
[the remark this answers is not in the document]
nuclearsandwich: per today's offline discussion we'll move type assertions to raise TypeErrors in this PR and I'll try to send a followup PR migrating assertions to type errors in the existing code that I used as an example.

nuclearsandwich:
> Where will the testing of the callbacks happen ? in a separate PR here ? or just in a higher level integration test?

During our sync meeting we talked about unit tests going here and integration tests happening in system tests (I haven't written any there yet either) but I guess there's nothing stopping me from unit testing the callback functions.

mikaelarguedas:
> During our sync meeting we talked about unit tests going here and integration tests happening in system tests (I haven't written any there yet either) but I guess there's nothing stopping me from unit testing the callback functions.

Good point, I remember that now.
Looking at rclcpp it looks like the only test that's performed is ensuring that it throws on invalid service name. But that the actual functionality testing of the callbacks is done at the `test_rclcpp` level.
So let's do this in `system_test/test_rclpy` to match what is done for rclcpp :+1:

rclpy/rclpy/parameter_service.py
```diff
@@ -0,0 +1,117 @@
...
+                        full_prefix = '.'.join(name.split('.')[0:-1])
+                        if full_prefix not in response.result.prefixes:
+                            response.result.prefixes.append(full_prefix)
+                        if prefix not in response.result.prefixes:
+                            response.result.prefixes.append(prefix)
+            else:
+                prefix = '.'.join(name.split('.')[0:-1])
+                if prefix not in response.result.prefixes:
+                    response.result.prefixes.append(prefix)
+                response.result.names.append(name)
```
mikaelarguedas: The definition of the result message is ambiguous IMO: https://github.com/ros2/rcl_interfaces/blob/f00fe22a36c8fdf6145d873173a4f9de0fdff131/rcl_interfaces/msg/ListParametersResult.msg.

It's unclear to me if the name of the "parameters under the prefix" should be:
- the full parameter names (of the parameters that have a prefix matching one of the requested ones)
- parameter names stripped of the prefix (of the parameters that have a prefix matching one of the requested ones)

The current implementation seems to match the implementation of rclcpp so I think it's good :+1:. But maybe it would be worth rephrasing or addressing the todo in the message definition. Or clarify what we expect the `prefixes` field in the response to be used for.

mikaelarguedas: maybe @wjwwood has more context regarding the use of the `prefixes` field in the result message?

nuclearsandwich:
@mikaelarguedas this is ready for another round of review.

nuclearsandwich:
Running some CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5031/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1838/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4184/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5039/)

rclpy/rclpy/parameter_service.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def _get_parameter_types_callback(self, request, response):
+        for name in request.names:
+            response.types.append(self._node.get_parameter(name).get_parameter_type())
+        return response
+
+    def _list_parameters_callback(self, request, response):
+        names_with_prefixes = []
+        for name in self._node._parameters.keys():
+            if '.' in name:
```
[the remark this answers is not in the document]
mikaelarguedas: It looks like this comment is still pending (same for other uses of `'.'` below)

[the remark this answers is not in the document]
nuclearsandwich: Whoops, thought I got all of them. Addressed by 67436b0

rclpy/rclpy/parameter.py
```diff
@@ -0,0 +1,134 @@
...
+        self._type = type_
+        self._name = name
+        self._value = value
+
+    @property
+    def name(self):
+        return self._name
+
+    @property  # noqa: A003
+    def type(self):
```
mikaelarguedas: not sure what to do about this, it feels like we should avoid using builtin names here.
@ros2/team how do you feel about using builtin names as method names ?

rclpy/rclpy/parameter_service.py
```diff
@@ -0,0 +1,123 @@
...
+from rclpy.qos import qos_profile_parameters
+
+
+class ParameterService:
+
+    def __init__(self, node):
+        self._node = node
+        nodename = node.get_name()
+
+        describe_parameters_service_name = '/'.join((nodename, 'describe_parameters'))
```
mikaelarguedas: Same as for the parameter separator, it would be better to not redefine this hardcoded separator in every service name definition. Though it looks like that's what's done [currently in rclcpp](https://github.com/ros2/rclcpp/blob/6b34bcc94c12e2213b6de7af57c3d10109402906/rclcpp/src/rclcpp/parameter_service.cpp#L36) so it's not a blocker for getting this merged

rclpy/rclpy/parameter.py
sloretz: I think a property is similar enough to a public attribute and a builtin is similar enough to a reserved keyword that a trailing underscore would be appropriate

https://legacy.python.org/dev/peps/pep-0008/#names-to-avoid

>If your public attribute name collides with a reserved keyword, append a single trailing underscore to your attribute name. This is preferable to an abbreviation or corrupted spelling.

rclpy/test/test_node.py
```diff
@@ -175,6 +177,70 @@ def test_set_executor_clear_executor(self):
         finally:
             node.destroy_node()
 
+    def test_node_set_parameters(self):
```
mikaelarguedas: Based on feedback from https://github.com/ros2/rclpy/pull/216#discussion_r208750095 it looks like we now prefer using directly the assert statement rather than the ones provided by unittest (same multiple times below). Though not a blocker for this PR IMO

rclpy/rclpy/parameter.py
nuclearsandwich: We could lengthen the field name to `parameter_type`. It's redundant and burns a few extra milliseconds typing but avoids the ambiguity.

rclpy/rclpy/parameter_service.py
nuclearsandwich: Constantizing this is straightforward and we've got to start it somewhere. What's a good name for the constant? `TOPIC_SEPARATOR_STRING`?

mikaelarguedas: That sounds good to me. It's consistent with the other constant defined in this PR and the naming used in otherp laces like logging https://github.com/ros2/rcutils/blob/4cbd05d9c09886754f41747afa705f15d02c6818/include/rcutils/logging.h#L34

rclpy/rclpy/parameter.py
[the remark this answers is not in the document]
nuclearsandwich: This line of review is resolved by adding creation-time type checking in 39698fb.

rclpy/rclpy/parameter_service.py
```diff
@@ -0,0 +1,117 @@
...
+            p = self._node.get_parameter(name)
+            response.values.append(p.get_parameter_value())
+        return response
+
+    def _get_parameter_types_callback(self, request, response):
+        for name in request.names:
+            response.types.append(self._node.get_parameter(name).get_parameter_type())
+        return response
+
+    def _list_parameters_callback(self, request, response):
```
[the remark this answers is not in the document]
nuclearsandwich: See https://github.com/ros2/rclpy/pull/214#issuecomment-412939986 for test cases used.

rclpy/rclpy/validate_topic_name.py
```diff
@@ -16,6 +16,8 @@
 from rclpy.exceptions import InvalidTopicNameException
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
+TOPIC_SEPARATOR_STRING = '/'
```
nuclearsandwich: I tried to find a "sensible" place for the topic separator constant because it seems like it could be used for more than just parameter services and this was the best I could come up with based on module names. But I can either put it in parameter_services for now or if there's a place more "sensible" than this I can move it to I'm happy to field suggestions.

rclpy/rclpy/parameter.py
```diff
@@ -0,0 +1,134 @@
...
+        self._type_ = type_
+        self._name = name
+        self._value = value
+
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def type_(self):
```
nuclearsandwich: @mikaelarguedas @sloretz the property is now accessed as `type_`. I had initially let this slide since the generated interface code uses `type` in python currently and I know that surfaced in discussion recently but can't find it to figure out where we came down.

nuclearsandwich:
New round of CI with increased timeout for rclpytests.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5036/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1840/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4186/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5041/)

rclpy/rclpy/parameter.py
nuclearsandwich: Makes sense. And the change is in so all's well.

nuclearsandwich:
I'm able to reproduce the hang that is happening on ARM and am now investigating.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5056/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1852/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4198/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5050/)

A full CI run with the timeout I originally set and changes to the node creation tests so they clean up properly.

mikaelarguedas:
>  full CI run with the timeout I originally set

Now that we know that the timeout issues on arm were related to a hang and not because we were getting close to the timeout limit. Is bumping the existing timeout still necessary ?
As the [last timeout bump](https://github.com/ros2/rclpy/pull/184) introduced a significant margin, I would expect it to be sufficient

nuclearsandwich:
> Is bumping the existing timeout still necessary ?

I could be mistaken about failures like: https://ci.ros2.org/job/ci_linux/5031/testReport/junit/(root)/projectroot/rclpytests/

Which I thought you had mentioned were related to earlier repeats of a test run timing out.  It seems like we should comfortably get by with no timeout. What's a good way to test that well? Run just the rclpy tests with a retest until fail?

nuclearsandwich:
Running just rclpy and tests with connext checking that the unmodified timeout length is still appropriate. [Build Status](https://ci.ros2.org/job/ci_linux/5058/)

mikaelarguedas:
Job with only connext using https://github.com/ros2/ci/pull/214: [Build Status](https://ci.ros2.org/job/ci_linux/5060/)

PR #216, opened 2018-08-08
executor.add_node() returns a bool

sloretz:
The documentation for `executor.add_node()` says it returns `True` if the node was added; however it currently just returns `None`. This fixes the function, and also makes it so the guard condition to rebuild the waitset is only triggered if the node really was added.

### DIFF
--- a/rclpy/test/test_executor.py
+++ b/rclpy/test/test_executor.py
@@ -40,7 +40,7 @@ class TestExecutor(unittest.TestCase):
 
         tmr = self.node.create_timer(0.1, timer_callback)
 
-        executor.add_node(self.node)
+        assert executor.add_node(self.node)
         executor.spin_once(timeout_sec=1.23)
         # TODO(sloretz) redesign test, sleeping to workaround race condition between test cleanup
         # and MultiThreadedExecutor thread pool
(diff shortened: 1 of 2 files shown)


sloretz:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5019/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1829/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4175/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5030/)

rclpy/test/test_executor.py
```diff
@@ -230,6 +230,12 @@ def __await__(self):
         rclpy.spin_once(self.node, timeout_sec=0)
         self.assertTrue(did_return)
 
+    def test_executor_add_node(self):
+        self.assertIsNotNone(self.node.handle)
```
dirk-thomas: I see that this matches the style in this file but in general we should use a plain assert instead (the recommended way with `pytest`). Nothing needs to happen for this patch though...

sloretz: Roger. I'll make sure to use a plain `assert` in the future.

PR #218, opened 2018-08-09
Node has a reference to its executor

sloretz:
This adds a property `node.executor` which accesses the executor the node has been added to. Internally the node stores the executor as a weak reference to avoid a reference cycle.

The purpose is to allow something with a node reference call executor functions. Specifically I want to make another PR where `TimeSource` calls `node.executor.create_task()` with time jump callbacks.

A nice side effect is `node.executor` enforces that a node has only been added to one executor. This should prevent the same entities from being waited on in multiple wait sets.

CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5021/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1831/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4177/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5032/)

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -17,6 +17,7 @@ import inspect
 import multiprocessing
 from threading import Condition
 from threading import Lock
+from threading import RLock
 
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 from rclpy.task import Task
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
Note for future enhancements: We [mentioned recently](https://github.com/ros2/rclcpp/issues/519) the motivation for having multiple executors operating on a node, they would process different callback groups of a given node. This makes the assumption that the executor is unique for a node.

wjwwood:
Yeah, at some point we would change the relationship to be between executors and callback groups, rather than nodes, at which point this would need to change as well. For now, however, it aligns Python and C++.

PR #220, opened 2018-08-10
Allow Duration to be negative

sloretz:
This allows `Duration` instances to be negative as talked about in ros2/rclcpp#525. `Time` is unchanged and cannot have negative values in this PR.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5029/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1836/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4182/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5037/)

### DIFF
--- a/rclpy/test/test_time.py
+++ b/rclpy/test/test_time.py
@@ -55,9 +55,14 @@ class TestTime(unittest.TestCase):
         duration = Duration(nanoseconds=2**63 - 1)
         assert duration.nanoseconds == 2**63 - 1
 
-        duration = Duration(seconds=-1)
+        assert Duration(seconds=-1).nanoseconds == -1 * 1000 * 1000 * 1000
         with self.assertRaises(ValueError):
             duration = Duration(nanoseconds=-1)
+
+        assert Duration(seconds=-2**63 / 1e9).nanoseconds == -2**63
+        with self.assertRaises(OverflowError):
+            # Much smaller number because float to integer conversion of seconds is imprecise
+            duration = Duration(seconds=-2**63 / 1e9 - 1)
 
     def test_time_operators(self):
         time1 = Time(nanoseconds=1, clock_type=ClockType.STEADY_TIME)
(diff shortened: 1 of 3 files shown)


rclpy/test/test_time.py
```diff
@@ -51,12 +51,11 @@ def test_duration_construction(self):
...
-            duration = Duration(nanoseconds=2**64)
-        duration = Duration(nanoseconds=2**64 - 1)
-        assert duration.nanoseconds == 2**64 - 1
+            duration = Duration(nanoseconds=2**63)
+        duration = Duration(nanoseconds=2**63 - 1)
+        assert duration.nanoseconds == 2**63 - 1
 
-        with self.assertRaises(ValueError):
-            duration = Duration(seconds=-1)
+        duration = Duration(seconds=-1)
```
dirk-thomas: It would be good to assert the resulting `duration.nanoseconds` in this case. Maybe also try the lowest possible value?

dhood:
thanks for following up on this @sloretz. For context, this is a follow up to https://github.com/ros2/rclpy/pull/209#issuecomment-408251413

rclpy/test/test_time.py
sloretz: Added check on resulting `duration.nanoseconds` plus a check on lowest possible value 7697213d798ce99b50f13dddf6936fbc90be4d13

The float seconds to integer nanoseconds precision isn't very good at that range. On my machine `Duration(seconds=-2**63 / 1e9 - 954e-9)` is the first negative floating point number to `OverflowError`, but I'm not sure how portable that is so I did `-2**63 / 1e9 - 1` instead.

sloretz:
New CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5030/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1837/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4183/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5038/)

PR #223, opened 2018-08-15
disable 1kHz test on all platforms

mikaelarguedas:
these tests fail more and more often on all platforms (e.g. [here](https://ci.ros2.org/view/nightly/job/nightly_osx_repeated/1216/testReport/junit/rclpy.test/test_timer/test_timer_cancel_reset_1000hertz/))

Disabling the 1kHz test on all platforms for now until we decide how to refactor them
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5045/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1842/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4189/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5042/)

### DIFF
--- a/rclpy/test/test_timer.py
+++ b/rclpy/test/test_timer.py
@@ -13,8 +13,9 @@
 # limitations under the License.
 
 import multiprocessing
-import os
-import platform
+# TODO(mikaelarguedas) comment back in if we reenable 1kHz test for some platforms
+# import os
+# import platform
 import sys
 import time
 import traceback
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
* macOS job with `--retest-umtil-fail 20` [Build Status](http://ci.ros2.org/job/ci_osx/4190/)

dirk-thomas:
I would suggest to use an annotation instead.

rclpy/test/test_timer.py
```diff
@@ -15,6 +15,7 @@
 import multiprocessing
 import os
 import platform
+import pytest
```
dirk-thomas: This might need a test dependency?

mikaelarguedas: Absolutely. I would expect it to always be provided by `ament_cmake_pytest` but it's better to be explicit. Added in https://github.com/ros2/rclpy/pull/223/commits/599edd7cf0f977c57ce443524c67101a456c4c58

mikaelarguedas:
CI as of https://github.com/ros2/rclpy/pull/223/commits/a850ae02f5c19de8ed3ae041abd3421ab45544a6  [Build Status](http://ci.ros2.org/job/ci_linux/5050/)

Issue #202, opened 2018-07-09
Support for parameters


dirk-thomas:
* [x] Nodes contains parameters #214 
* [x] Nodes providing ROS services to interact with the parameters #214
* [x] Parameters can be passed through CLI #225 
* [x] Parameter events are published on updates. #226 
* [x] Nodes can have a parameters_callback set which can arbitrarily accept and reject setting parameters. https://github.com/ros2/rclpy/pull/228
* [x] System tests for rclpy which exercise the parameters API, events, and services ros2/system_tests#293

nuclearsandwich:
All rclpy parameters related PRs have been merged. Unless there's a reason to keep this tracking issue open, I believe it can be closed.

PR #225, opened 2018-08-17
Take initial parameters from parameters yaml files and constructor arguments.

nuclearsandwich:
With this pull request initial parameters can be passed via yaml parameters file with the `__params:=` and as a list of parameters in the `initial_parameters` kwarg of the Node constructor.

Parameters in the constructor will override parameters from a parameters file.

If at least one reviewer could keep a close eye on my use of Py_DECREF and make sure that I have employed everywhere I should and nowhere I shouldn't have I'd be grateful..  I'm fairly confident in my usage but it's also my first time in Python's C API. 

To test this PR I've been using the params.yaml in this gist https://gist.github.com/nuclearsandwich/8753121711671bfa8d9cb5f718ce09bc and the talker node from demo_nodes_py as well as the parameter services behavior just merged from #214 

Connects to #202

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -104,15 +104,6 @@ class Node:
         self._parameter_event_publisher = self.create_publisher(
             ParameterEvent, 'parameter_events', qos_profile=qos_profile_parameter_events)
 
-        for param in _rclpy.rclpy_get_node_parameters(Parameter, self.handle):
-            self._parameters[param.name] = param
-        for param in initial_parameters:
-            if not isinstance(param, Parameter):
-                raise TypeError(
-                    'initial_parameters must be a list of {}'.format(repr(Parameter))
-                )
-            self._parameters[param.name] = param
-
         node_parameters = _rclpy.rclpy_get_node_parameters(Parameter, self.handle)
         # Combine parameters from params files with those from the node constructor and
         # use the set_parameters_atomically API so a parameter event is published.
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -61,7 +61,7 @@ class Node:
 
     def __init__(
         self, node_name, *, cli_args=None, namespace=None, use_global_arguments=True,
-        start_parameter_services=True, initial_parameters=[]
+        start_parameter_services=True, initial_parameters=None
     ):
         self._handle = None
         self._parameters = {}
@@ -112,6 +112,13 @@ class Node:
                     'initial_parameters must be a list of {}'.format(repr(Parameter))
                 )
             self._parameters[param.name] = param
+
+        node_parameters = _rclpy.rclpy_get_node_parameters(Parameter, self.handle)
+        # use the set_parameters API so parameter events are publish.
+        if node_parameters:
+            self.set_parameters(node_parameters)
+        if initial_parameters is not None:
+            self.set_parameters(initial_parameters)
 
         if start_parameter_services:
             self._parameter_service = ParameterService(self)
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -115,15 +115,10 @@ class Node:
 
         node_parameters = _rclpy.rclpy_get_node_parameters(Parameter, self.handle)
         # Combine parameters from params files with those from the node constructor and
-        # use the set_parameters API so parameter events are published.
-        if node_parameters and initial_parameters:
-            params_by_name = {p.name: p for p in node_parameters}
-            params_by_name.update({p.name: p for p in initial_parameters})
-            self.set_parameters(params_by_name.values())
-        elif node_parameters:
-            self.set_parameters(node_parameters)
-        elif initial_parameters:
-            self.set_parameters(initial_parameters)
+        # use the set_parameters_atomically API so a parameter event is published.
+        if initial_parameters is not None:
+            node_parameters.update({p.name: p for p in initial_parameters})
+        self.set_parameters_atomically(node_parameters.values())
 
         if start_parameter_services:
             self._parameter_service = ParameterService(self)
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -114,10 +114,15 @@ class Node:
             self._parameters[param.name] = param
 
         node_parameters = _rclpy.rclpy_get_node_parameters(Parameter, self.handle)
-        # use the set_parameters API so parameter events are publish.
-        if node_parameters:
+        # Combine parameters from params files with those from the node constructor and
+        # use the set_parameters API so parameter events are published.
+        if node_parameters and initial_parameters:
+            params_by_name = {p.name: p for p in node_parameters}
+            params_by_name.update({p.name: p for p in initial_parameters})
+            self.set_parameters(params_by_name.values())
+        elif node_parameters:
             self.set_parameters(node_parameters)
-        if initial_parameters is not None:
+        elif initial_parameters:
             self.set_parameters(initial_parameters)
 
         if start_parameter_services:
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -61,7 +61,7 @@ class Node:
 
     def __init__(
         self, node_name, *, cli_args=None, namespace=None, use_global_arguments=True,
-        start_parameter_services=True
+        start_parameter_services=True, initial_parameters=[]
     ):
         self._handle = None
         self._parameters = {}
@@ -103,6 +103,15 @@ class Node:
 
         self._parameter_event_publisher = self.create_publisher(
             ParameterEvent, 'parameter_events', qos_profile=qos_profile_parameter_events)
+
+        for param in _rclpy.rclpy_get_node_parameters(Parameter, self.handle):
+            self._parameters[param.name] = param
+        for param in initial_parameters:
+            if not isinstance(param, Parameter):
+                raise TypeError(
+                    'initial_parameters must be a list of {}'.format(repr(Parameter))
+                )
+            self._parameters[param.name] = param
 
         if start_parameter_services:
             self._parameter_service = ParameterService(self)
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -21,6 +21,7 @@
 #include <rcl/rcl.h>
 #include <rcl/time.h>
 #include <rcl/validate_topic_name.h>
+#include <rcl_yaml_param_parser/parser.h>
 #include <rcutils/strdup.h>
 #include <rcutils/types.h>
 #include <rmw/error_handling.h>
@@ -3369,6 +3370,213 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
   Py_RETURN_NONE;
 }
 
+/// Populate an rcl_params_t with parameters parsed from arguments.
+/**
+ * On failure a Python exception is raised and false is returned if:
+ *
+ * Raises RuntimeError if param_files cannot be extracted from arguments.
+ * Raises RuntimeError if yaml files do not parse succesfully.
+ *
+ * \param[in] args The arguments to parse for parameter files
+ * \param[in] allocator Allocator to use for allocating and deallocating within the function.
+ * \param[out] params An initialized rcl params struct to place parsed parameters into.
+ *
+ * Returns true when parameters are parsed successfully (including the trivial case)
+ *         false when there was an error during parsing and a Python exception was raised.
+ *
+ */
+static bool _parse_param_files(
+  const rcl_arguments_t * args, rcl_allocator_t allocator,
+  rcl_params_t * params)
+{
+  char ** param_files;
+  int param_files_count = rcl_arguments_get_param_files_count(args);
+  rcl_ret_t ret;
+  if (param_files_count > 0) {
+    ret = rcl_arguments_get_param_files(args, allocator, &param_files);
+    if (RCL_RET_OK != ret) {
+      PyErr_Format(PyExc_RuntimeError, "Failed to get initial parameters: %s",
+        rcl_get_error_string_safe());
+      return false;
+    }
+    for (int i = 0; i < param_files_count; i++) {
+      if (!rcl_parse_yaml_file(param_files[i], params)) {
+        /* failure to parse will automatically fini the params struct */
+        PyErr_Format(PyExc_RuntimeError, "Failed to parse yaml params file: %s", param_files[i]);
+        return false;
+      }
+      allocator.deallocate(param_files[i], allocator.state);
+    }
+  }
+  return true;
+}
+
+/// Create an rclpy.parameter.Parameter from an rcl_variant_t
+/**
+ * On failure a Python exception is raised and false is returned if:
+ *
+ * Raises ValueError if the variant points to no data.
+ *
+ * \param[in] name The name of the parameter
+ * \param[in] variant The variant object to create a Parameter from
+ * \param[in] parameter_cls The PythonObject for the Parameter class.
+ * \param[in] parameter_cls The PythonObject for the Parameter.Type class.
+ *
+ * Returns a pointer to an rclpy.parameter.Parameter with the name, type, and value from
+ * the variant or NULL when raising a python exception.
+ */
+static PyObject * _parameter_from_rcl_variant(
+  const char * name, rcl_variant_t * variant, PyObject * parameter_cls,
+  PyObject * parameter_type_cls)
+{
+  int type_enum_value;
+  PyObject * value;
+  if (variant->bool_value) {
+    type_enum_value = 1;
+    value = variant->bool_value ? Py_True : Py_False;
+  } else if (variant->integer_value) {
+    type_enum_value = 2;
+    value = PyLong_FromLong(*(variant->integer_value));
+  } else if (variant->double_value) {
+    type_enum_value = 3;
+    value = PyFloat_FromDouble(*(variant->double_value));
+  } else if (variant->string_value) {
+    type_enum_value = 4;
+    value = PyUnicode_FromString(variant->string_value);
+  } else if (variant->byte_array_value) {
+    type_enum_value = 5;
+    value = PyList_New(variant->byte_array_value->size);
+    for (size_t i = 0; i < variant->byte_array_value->size; i++) {
+      PyList_SetItem(value, i, PyBytes_FromFormat("%u", variant->byte_array_value->values[i]));
+    }
+  } else if (variant->bool_array_value) {
+    type_enum_value = 6;
+    value = PyList_New(variant->bool_array_value->size);
+    for (size_t i = 0; i < variant->bool_array_value->size; i++) {
+      PyList_SetItem(value, i, variant->bool_array_value->values[i] ? Py_True : Py_False);
+    }
+  } else if (variant->integer_array_value) {
+    type_enum_value = 7;
+    value = PyList_New(variant->integer_array_value->size);
+    for (size_t i = 0; i < variant->integer_array_value->size; i++) {
+      PyList_SetItem(value, i, PyLong_FromLong(variant->integer_array_value->values[i]));
+    }
+  } else if (variant->double_array_value) {
+    type_enum_value = 8;
+    value = PyList_New(variant->double_array_value->size);
+    for (size_t i = 0; i < variant->double_array_value->size; i++) {
+      PyList_SetItem(value, i, PyFloat_FromDouble(variant->double_array_value->values[i]));
+    }
+  } else if (variant->string_array_value) {
+    type_enum_value = 9;
+    value = PyList_New(variant->string_array_value->size);
+    for (size_t i = 0; i < variant->string_array_value->size; i++) {
+      PyList_SetItem(value, i, PyUnicode_FromString(variant->string_array_value->data[i]));
+    }
+  }
+
+  PyObject * type = PyObject_CallObject(parameter_type_cls, Py_BuildValue("(i)", type_enum_value));
+  PyObject * param = PyObject_CallObject(parameter_cls, Py_BuildValue("sOO", name, type, value));
+
+  Py_DECREF(type);
+  return param;
+}
+
+/// Get a list of parameters for the current node from rcl_yaml_param_parser
+/**
+ * On failure, an exception is raised and NULL is returned if:
+ *
+ * Raises ValueError if the argument is not a node handle.
+ * Raises RuntimeError if the parameters file fails to parse
+ *
+ * \param[in] parameter_cls The rclpy.parameter.Parameter class object.
+ * \param[in] node_handle Capsule pointing to the node handle
+ * \return NULL on failure
+ *         A list of rclpy.parameter.Parameter on success (may be empty).
+ *
+ */
+static PyObject *
+rclpy_get_node_parameters(PyObject * Py_UNUSED(self), PyObject * args)
+{
+  PyObject * parameter_cls;
+  PyObject * node_capsule;
+  if (!PyArg_ParseTuple(args, "OO", &parameter_cls, &node_capsule)) {
+    return NULL;
+  }
+
+  rcl_node_t * node = (rcl_node_t *)PyCapsule_GetPointer(node_capsule, "rcl_node_t");
+  const rcl_node_options_t * node_options = rcl_node_get_options(node);
+  const rcl_allocator_t allocator = node_options->allocator;
+  if (!node) {
+    PyErr_Format(PyExc_RuntimeError, "Failed to retrieve rcl node from capsule");
+    return NULL;
+  }
+
+  rcl_params_t * params = rcl_yaml_node_struct_init(allocator);
+  if (NULL == params) {
+    PyErr_Format(PyExc_RuntimeError, "Failed to allocate initial parameters");
+    return NULL;
+  }
+
+  if (node_options->use_global_arguments) {
+    if (!_parse_param_files(rcl_get_global_arguments(), allocator, params)) {
+      return NULL;
+    }
+  }
+
+  if (!_parse_param_files(&(node_options->arguments), allocator, params)) {
+    return NULL;
+  }
+
+  int node_index = -1;
+  const char * node_namespace = rcl_node_get_namespace(node);
+  char * node_name_with_namespace;
+  if ('/' == node_namespace[strlen(node_namespace) - 1]) {
+    node_name_with_namespace = rcutils_format_string(allocator, "%s%s",
+        node_namespace, rcl_node_get_name(node));
+  } else {
+    node_name_with_namespace = rcutils_format_string(allocator, "%s/%s",
+        node_namespace, rcl_node_get_name(node));
+  }
+
+  char * fully_qualified_param_node_name;
+  for (size_t i = 0; i < params->num_nodes; i++) {
+    if (params->node_names[i][0] != '/') {
+      fully_qualified_param_node_name = rcutils_format_string(allocator,
+          "/%s", params->node_names[i]);
+    } else {
+      fully_qualified_param_node_name = params->node_names[i];
+    }
+    if (0 == strcmp(fully_qualified_param_node_name, node_name_with_namespace)) {
+      node_index = i;
+    }
+    if (params->node_names[i][0] != '/') {
+      allocator.deallocate(fully_qualified_param_node_name, allocator.state);
+    }
+  }
+  allocator.deallocate(node_name_with_namespace, allocator.state);
+
+  if (-1 == node_index) {
+    /* No parameters for current node.*/
+    rcl_yaml_node_struct_fini(params);
+    return PyList_New(0);
+  }
+
+  PyObject * parameter_type_cls = PyObject_GetAttrString(parameter_cls, "Type");
+  rcl_node_params_t node_params = params->params[node_index];
+  PyObject * parameter_list = PyList_New(node_params.num_params);
+  for (size_t i = 0; i < node_params.num_params; i++) {
+    PyList_SetItem(parameter_list, i, _parameter_from_rcl_variant(
+        node_params.parameter_names[i], &(node_params.parameter_values[i]),
+        parameter_cls, parameter_type_cls));
+  }
+
+  Py_DECREF(parameter_type_cls);
+  rcl_yaml_node_struct_fini(params);
+  return parameter_list;
+}
+
+
 /// Define the public methods of this module
 static PyMethodDef rclpy_methods[] = {
   {
@@ -3596,6 +3804,10 @@ static PyMethodDef rclpy_methods[] = {
   {
     "rclpy_get_node_names", rclpy_get_node_names, METH_VARARGS,
     "Get node names list from graph API."
+  },
+  {
+    "rclpy_get_node_parameters", rclpy_get_node_parameters, METH_VARARGS,
+    "Get the initial parameters for a node from the command line."
   },
   {
     "rclpy_get_topic_names_and_types", rclpy_get_topic_names_and_types, METH_VARARGS,
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -3436,7 +3436,10 @@ static PyObject * _parameter_from_rcl_variant(
   const char * name, rcl_variant_t * variant, PyObject * parameter_cls,
   PyObject * parameter_type_cls)
 {
-  int type_enum_value;
+
+  /* Default to 0 (not set) to suppress warnings. A python error will raise if
+   * type and value don't agree */
+  int type_enum_value = 0;
   PyObject * value;
   PyObject * member_value;
   if (variant->bool_value) {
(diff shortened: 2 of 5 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,213 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+        rcl_get_error_string_safe());
+      return false;
+    }
+    for (int i = 0; i < param_files_count; i++) {
+      if (!rcl_parse_yaml_file(param_files[i], params)) {
+        /* failure to parse will automatically fini the params struct */
+        PyErr_Format(PyExc_RuntimeError, "Failed to parse yaml params file: %s", param_files[i]);
+        return false;
+      }
+      allocator.deallocate(param_files[i], allocator.state);
```
sloretz: I think this needs a call to `allocator.deallocate(param_files, allocator.state)` after the array members have been deallocated.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,213 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+ */
+static PyObject * _parameter_from_rcl_variant(
+  const char * name, rcl_variant_t * variant, PyObject * parameter_cls,
+  PyObject * parameter_type_cls)
+{
+  int type_enum_value;
+  PyObject * value;
+  if (variant->bool_value) {
+    type_enum_value = 1;
+    value = variant->bool_value ? Py_True : Py_False;
```
sloretz: Need to `Py_INCREF(Py_True)` or `Py_INCREF(Py_False)`.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,213 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+    value = PyLong_FromLong(*(variant->integer_value));
+  } else if (variant->double_value) {
+    type_enum_value = 3;
+    value = PyFloat_FromDouble(*(variant->double_value));
+  } else if (variant->string_value) {
+    type_enum_value = 4;
+    value = PyUnicode_FromString(variant->string_value);
+  } else if (variant->byte_array_value) {
+    type_enum_value = 5;
+    value = PyList_New(variant->byte_array_value->size);
```
sloretz: Need to check for `NULL` return here

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,213 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+    for (size_t i = 0; i < variant->byte_array_value->size; i++) {
+      PyList_SetItem(value, i, PyBytes_FromFormat("%u", variant->byte_array_value->values[i]));
```
sloretz: Since list is new you can use [`PyList_SET_ITEM`](https://docs.python.org/3.7/c-api/list.html#c.PyList_SET_ITEM) here

sloretz: [The docs](https://docs.python.org/3/c-api/bytes.html#c.PyBytes_FromFormat) don't mention it but it looks like [`PyBytes_FromFormat` can error and return NULL](https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Objects/bytesobject.c#L181).

rclpy/src/rclpy/_rclpy.c
sloretz: Same comment in a couple places below

rclpy/src/rclpy/_rclpy.c
sloretz: Same comments about `PyList_SET_ITEM` below

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,213 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+    }
+  } else if (variant->bool_array_value) {
+    type_enum_value = 6;
+    value = PyList_New(variant->bool_array_value->size);
+    for (size_t i = 0; i < variant->bool_array_value->size; i++) {
+      PyList_SetItem(value, i, variant->bool_array_value->values[i] ? Py_True : Py_False);
```
sloretz: Same comment about needing to incref True and false. [`PyList_SetItem` and `PyList_SET_ITEM` steal a reference](https://docs.python.org/3.7/c-api/list.html#c.PyList_SetItem)

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,213 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+    }
+  } else if (variant->string_array_value) {
+    type_enum_value = 9;
+    value = PyList_New(variant->string_array_value->size);
+    for (size_t i = 0; i < variant->string_array_value->size; i++) {
+      PyList_SetItem(value, i, PyUnicode_FromString(variant->string_array_value->data[i]));
+    }
+  }
+
+  PyObject * type = PyObject_CallObject(parameter_type_cls, Py_BuildValue("(i)", type_enum_value));
```
sloretz: Need to check `Py_BuildValue` didn't return `NULL`, as well as `PyObject_CallObject`.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,213 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+rclpy_get_node_parameters(PyObject * Py_UNUSED(self), PyObject * args)
+{
+  PyObject * parameter_cls;
+  PyObject * node_capsule;
+  if (!PyArg_ParseTuple(args, "OO", &parameter_cls, &node_capsule)) {
+    return NULL;
+  }
+
+  rcl_node_t * node = (rcl_node_t *)PyCapsule_GetPointer(node_capsule, "rcl_node_t");
+  const rcl_node_options_t * node_options = rcl_node_get_options(node);
```
sloretz: This should be moved below the `if (!node)` check

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,213 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+  const rcl_allocator_t allocator = node_options->allocator;
+  if (!node) {
+    PyErr_Format(PyExc_RuntimeError, "Failed to retrieve rcl node from capsule");
```
sloretz: `PyCapsule_GetPointer` would have already set an exception here. Returning `NULL` is enough.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,213 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+  }
+  allocator.deallocate(node_name_with_namespace, allocator.state);
+
+  if (-1 == node_index) {
+    /* No parameters for current node.*/
+    rcl_yaml_node_struct_fini(params);
+    return PyList_New(0);
+  }
+
+  PyObject * parameter_type_cls = PyObject_GetAttrString(parameter_cls, "Type");
```
sloretz: Need to check for `NULL`

rclpy/src/rclpy/_rclpy.c
nuclearsandwich: NULL check is now just below.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5065/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1858/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4205/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5056/)

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,263 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+  allocator.deallocate(node_name_with_namespace, allocator.state);
+
+  if (-1 == node_index) {
+    /* No parameters for current node.*/
+    rcl_yaml_node_struct_fini(params);
+    return PyList_New(0);
+  }
+
+  if (!PyObject_HasAttrString(parameter_cls, "Type")) {
+    PyErr_Format(PyExc_RuntimeError, "Parameter class is missing 'Type' attribute");
```
sloretz: need to fini `params` here and a few places below

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,263 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+  }
+  PyObject * parameter_type_cls = PyObject_GetAttrString(parameter_cls, "Type");
+  if (NULL == parameter_type_cls) {
+    PyErr_Format(PyExc_RuntimeError, "Error getting 'Type' attribute from Parameter class");
+    return NULL;
+  }
+  rcl_node_params_t node_params = params->params[node_index];
+  PyObject * parameter_list = PyList_New(node_params.num_params);
+  if (NULL == parameter_list) {
+    return NULL;
```
sloretz: `Py_DECREF(parameter_type_cls)` and in another place below

rclpy/rclpy/node.py
```diff
@@ -60,7 +60,7 @@ class Node:
 
     def __init__(
         self, node_name, *, cli_args=None, namespace=None, use_global_arguments=True,
-        start_parameter_services=True
+        start_parameter_services=True, initial_parameters=[]
```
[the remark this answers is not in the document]
nuclearsandwich: Of course. :grin: I think we were even talking about this together last week.

rclpy/src/rclpy/_rclpy.c
nuclearsandwich: > It might be more convenient for users if this was one bytes() object with all the bytes in it.

It would be more convenient but it would violate the interface specification which states that a byte should be represented as a `bytes` of length one and that any of the array types should be represented by a Python list of the scalar type. http://design.ros2.org/articles/generated_interfaces_python.html

@mikaelarguedas and I talked about this and they may have a link to further discussion material.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,263 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+    return PyList_New(0);
+  }
+
+  if (!PyObject_HasAttrString(parameter_cls, "Type")) {
+    PyErr_Format(PyExc_RuntimeError, "Parameter class is missing 'Type' attribute");
+    return NULL;
+  }
+  PyObject * parameter_type_cls = PyObject_GetAttrString(parameter_cls, "Type");
+  if (NULL == parameter_type_cls) {
+    PyErr_Format(PyExc_RuntimeError, "Error getting 'Type' attribute from Parameter class");
```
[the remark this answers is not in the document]
nuclearsandwich: You would appear to be correct. https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Objects/object.c#L880

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,263 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+  const rcl_allocator_t allocator = node_options->allocator;
+  rcl_params_t * params = rcl_yaml_node_struct_init(allocator);
+  if (NULL == params) {
+    PyErr_Format(PyExc_RuntimeError, "Failed to allocate initial parameters");
+    return NULL;
+  }
+
+  if (node_options->use_global_arguments) {
+    if (!_parse_param_files(rcl_get_global_arguments(), allocator, params)) {
+      return NULL;
```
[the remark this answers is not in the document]
nuclearsandwich: I thought so too but `rcl_parse_yaml_file` will actually fini the passed in struct on error https://github.com/ros2/rcl/blob/adc0190259a4eb725480e7b32a90b902bc5279b0/rcl_yaml_param_parser/src/parser.c#L1382.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,263 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+  char ** param_files;
+  int param_files_count = rcl_arguments_get_param_files_count(args);
+  rcl_ret_t ret;
+  bool successful = true;
+  if (param_files_count > 0) {
+    ret = rcl_arguments_get_param_files(args, allocator, &param_files);
+    if (RCL_RET_OK != ret) {
+      PyErr_Format(PyExc_RuntimeError, "Failed to get initial parameters: %s",
+        rcl_get_error_string_safe());
+      return false;
```
sloretz: Replying to https://github.com/ros2/rclpy/pull/225/files#r211049548 , since the calling code doesn't fini `params` if this errors then I think `params` needs to be fini'd here.

rclpy/src/rclpy/_rclpy.c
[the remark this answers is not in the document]
nuclearsandwich: c08b236 makes `_parse_param_files` fini the params struct in both error situations so that the caller does not need to.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5066/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1859/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4206/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5057/)

Next CI after fixing lint and some warnings. I couldn't track the MSBuild warnings so I'll get those next wave.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,299 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+    if (!_parse_param_files(rcl_get_global_arguments(), allocator, params)) {
+      return NULL;
+    }
+  }
+
+  if (!_parse_param_files(&(node_options->arguments), allocator, params)) {
+    return NULL;
+  }
+
+  int node_index = -1;
```
nuclearsandwich: MSBuild (rightly) complains that `int`might lose information when assigned from a `size_t` below. But changing to a `size_t` makes the sigil value `-1` invalid. Do we have a pattern we use for this sort of situation like storing a second node_index_found value and setting and checking it along with the index?

nuclearsandwich:
While setting up system tests I discovered that the yaml parameter parser doesn't preserve any parameters currently in the struct as I though. So I'll need to refactor this to extract parameters one file as a time and store them in a Python dict rather than a list.

nuclearsandwich:
> refactor this to extract parameters one file at a time and store them in a Python dict rather than a list.

This refactor is complete and the implementation is now working against the existing tests in system_tests/test_cli.

rclpy/rclpy/node.py
```diff
@@ -100,6 +100,13 @@ def __init__(
 
         self.__executor_weakref = None
 
+        node_parameters = _rclpy.rclpy_get_node_parameters(Parameter, self.handle)
+        # use the set_parameters API so parameter events are publish.
```
wjwwood: I tried to read the code to understand this but I'm a bit lost I think. In this case aren't the parameters potentially set twice, once as the "node parameters" and then again using the initial parameters?

And wouldn't that generate two events? Is that the intended behavior? I could see it either way: "I'd like to see the evolution of the values on startup" or "I just care about what it was initialized with not what went into determining the initial value".

wjwwood:
Looks like this comment is still pending right?

https://github.com/ros2/rclpy/pull/225#discussion_r211050534

rclpy/src/rclpy/_rclpy.c
nuclearsandwich: With da7da1a the implementation no longer takes an rcl_params_t as input and instead creates and uses the data structure internally only if parameter arguments are found without error so the params object no longer exists at this point.

nuclearsandwich:
> Looks like this comment is still pending right? [#225 (comment)](https://github.com/ros2/rclpy/pull/225#discussion_r211050534)

It was addressed indirectly. Added an explanation.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5093/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1877/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4223/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5074/)

rclpy/rclpy/node.py
```diff
@@ -100,6 +100,18 @@ def __init__(
...
+        # Combine parameters from params files with those from the node constructor and
+        # use the set_parameters API so parameter events are published.
+        if node_parameters and initial_parameters:
+            params_by_name = {p.name: p for p in node_parameters}
+            params_by_name.update({p.name: p for p in initial_parameters})
+            self.set_parameters(params_by_name.values())
+        elif node_parameters:
+            self.set_parameters(node_parameters)
+        elif initial_parameters:
+            self.set_parameters(initial_parameters)
```
mikaelarguedas: At first glance, it looks like this could be simplified quite a bit by setting default values for node_parameters and initial_parameters.

(I didnt check the implementation so this is hypothetical)
But if:
`initial_parameters` defaulted to an empty list and `node_parameters` to an empty dict.
Could we get rid of the conditional logic and just have the following?
```
params_by_name = {p.name: p for p in node_parameters}
params_by_name.update({p.name: p for p in initial_parameters})
self.set_parameters(params_by_name.values())
```
I wouldnt expect a big performance hit if these are empty.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,385 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+static bool
+_parse_param_files(
+  const rcl_arguments_t * args, rcl_allocator_t allocator, PyObject * parameter_cls,
+  PyObject * parameter_type_cls, PyObject * params_by_node_name)
+{
+  char ** param_files;
+  int param_files_count = rcl_arguments_get_param_files_count(args);
+  rcl_ret_t ret;
+  bool successful = true;
+  if (param_files_count > 0) {
```
mikaelarguedas: we could reduce nesting by just returning `false` if `param_files_count <= 0`

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,385 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+    ret = rcl_arguments_get_param_files(args, allocator, &param_files);
```
mikaelarguedas: as `ret` is not used anywhere else we could just check the return value of `rcl_arguments_get_param_files` and return on failure

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,385 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+        rcl_get_error_string_safe());
+      return false;
+    }
+    for (int i = 0; i < param_files_count; i++) {
+      if (successful) {
+        rcl_params_t * params = rcl_yaml_node_struct_init(allocator);
+        if (!rcl_parse_yaml_file(param_files[i], params)) {
+          /* failure to parse will automatically fini the params struct */
+          PyErr_Format(PyExc_RuntimeError, "Failed to parse yaml params file: %s", param_files[i]);
+          successful = false;
```
mikaelarguedas: should we just cleanup and break out of the loop in this case?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,263 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+      return NULL;
+    }
+    for (size_t i = 0; i < variant->string_array_value->size; i++) {
+      PyList_SET_ITEM(value, i, PyUnicode_FromString(variant->string_array_value->data[i]));
+    }
+  }
+
+  PyObject * args = Py_BuildValue("(i)", type_enum_value);
+  if (NULL == args) {
+    return NULL;
```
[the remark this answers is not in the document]
mikaelarguedas: I think this comment still holds

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,385 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+  }
+  PyObject * type = PyObject_CallObject(parameter_type_cls, args);
+  Py_DECREF(args);
+  args = Py_BuildValue("OOO", name, type, value);
+  if (NULL == args) {
```
mikaelarguedas: should Py_DECREF type and value

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,263 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+  }
+  PyObject * type = PyObject_CallObject(parameter_type_cls, args);
+  Py_DECREF(args);
+  args = Py_BuildValue("sOO", name, type, value);
+  if (NULL == args) {
+    return NULL;
+  }
+  PyObject * param = PyObject_CallObject(parameter_cls, args);
+  Py_DECREF(args);
+  Py_DECREF(type);
```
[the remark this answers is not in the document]
mikaelarguedas: :+1:

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,385 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+
+/// Get a list of parameters for the current node from rcl_yaml_param_parser
+/**
+ * On failure, an exception is raised and NULL is returned if:
+ *
+ * Raises ValueError if the argument is not a node handle.
+ * Raises RuntimeError if the parameters file fails to parse
+ *
+ * \param[in] parameter_cls The rclpy.parameter.Parameter class object.
+ * \param[in] node_handle Capsule pointing to the node handle
```
mikaelarguedas: Nit: node_handle -> node_capsule

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,385 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+rclpy_get_node_parameters(PyObject * Py_UNUSED(self), PyObject * args)
+{
+  PyObject * parameter_cls;
+  PyObject * node_capsule;
+  if (!PyArg_ParseTuple(args, "OO", &parameter_cls, &node_capsule)) {
+    return NULL;
+  }
+
+  rcl_node_t * node = (rcl_node_t *)PyCapsule_GetPointer(node_capsule, "rcl_node_t");
+  if (!node) {
```
mikaelarguedas: Nit: `if (NULL == node)`

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,385 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+  }
+
+  if (!PyObject_HasAttrString(parameter_cls, "Type")) {
+    PyErr_Format(PyExc_RuntimeError, "Parameter class is missing 'Type' attribute");
+    Py_DECREF(params_by_node_name);
+    return NULL;
+  }
+  PyObject * parameter_type_cls = PyObject_GetAttrString(parameter_cls, "Type");
+  if (NULL == parameter_type_cls) {
+    /* PyObject_GetAttrString raises AttributeError on failure. */
```
mikaelarguedas: Nit: please use `//` for single line comments

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3370,385 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+  if (!_parse_param_files(&(node_options->arguments), allocator, parameter_cls,
+    parameter_type_cls, params_by_node_name))
+  {
+    Py_DECREF(parameter_type_cls);
+    Py_DECREF(params_by_node_name);
+    return NULL;
+  }
+  Py_DECREF(parameter_type_cls);
+
+  const char * node_namespace = rcl_node_get_namespace(node);
```
mikaelarguedas: is this deallocated anywhere?

nuclearsandwich: No. Since this is a `const` "borrowed" reference to the internal node string. It's lifetime is bound to the node it is from.

rclpy/src/rclpy/_rclpy.c
nuclearsandwich: I think this might be a matter of preference, and I'm flexible. The current loop deallocates as it goes. We could either preserve that and have a short circuit finish cleanup with an inner for loop before returning or instead of cleaning as we go iterate a second time to do cleanup and break to that case on failure. I tend to prefer iterating a minimal number of times but the *n* of parameter files is potentially never going to be large enough that we notice *2n* > *n*.

mikaelarguedas: Yeah as the loop doesn't do anything else than cleaning up as soon a a failure occurs it's fine to keep as is :+1:

---

> We could either preserve that and have a short circuit finish cleanup with an inner for loop before returning or instead of cleaning as we go iterate a second time to do cleanup and break to that case on failure

Just to clarify, I was suggesting to clean up the "remaining ones" and breaking, not to change "the cleanup as we go" and requiring to iterate again over the entire set

rclpy/src/rclpy/_rclpy.c
nuclearsandwich: Now decref'd as of b5fd549bd0a422a9b9aa73a739df1e0c2897f432

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3613,7 +3994,7 @@ static PyMethodDef rclpy_methods[] = {
 
   {
     "rclpy_convert_from_py_qos_policy", rclpy_convert_from_py_qos_policy, METH_VARARGS,
-    "Convert a QoSPolicy python object into a rmw_qos_profile_t."
+    "Convert a QoSPolicy Python object into a rmw_qos_profile_t."
```
nuclearsandwich: @mikaelarguedas had me turn all my `python` references to `Python` so I brought this one along for the ride. Has nothing else to do with this PR.

nuclearsandwich:
FYI for any reviewers who have pulled locally I've just rebased and pushed in order to resolve merge conflicts.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5106/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1887/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4233/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5084/)

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3371,382 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+/**
+ * On failure a Python exception is raised and false is returned if:
+ *
+ * Raises RuntimeError if param_files cannot be extracted from arguments.
+ * Raises RuntimeError if yaml files do not parse succesfully.
+ *
+ * \param[in] args The arguments to parse for parameter files
+ * \param[in] allocator Allocator to use for allocating and deallocating within the function.
+ * \param[out] params_by_node_name A Python dict object to place parsed parameters into.
+ *
```
sloretz: nit, missing documentation of `parameter_cls`, and `parameter_type_cls`

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3371,382 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+  bool successful = true;
+  if (param_files_count <= 0) {
+    return successful;
+  }
+  if (RCL_RET_OK != rcl_arguments_get_param_files(args, allocator, &param_files)) {
+    PyErr_Format(PyExc_RuntimeError, "Failed to get initial parameters: %s",
+      rcl_get_error_string_safe());
+    return false;
+  }
+  for (int i = 0; i < param_files_count; i++) {
```
sloretz: nit, `++i`

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3371,382 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+      if (!rcl_parse_yaml_file(param_files[i], params)) {
+        // failure to parse will automatically fini the params struct
+        PyErr_Format(PyExc_RuntimeError, "Failed to parse yaml params file: %s", param_files[i]);
+        successful = false;
+      } else {
+        if (!_populate_node_parameters_from_rcl_params(
+            params, allocator, parameter_cls, parameter_type_cls, params_by_node_name))
+        {
+          PyErr_Format(PyExc_RuntimeError,
+            "Failed to fill params dict from file: %s", param_files[i]);
```
sloretz: Except as noted about `PyDict_GetItem`, an exception is already set here.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3371,382 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+          rcl_yaml_node_struct_fini(params);
+          return false;
```
sloretz: I think `param_files` needs to be deallocated before returning.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3371,382 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+        node_namespace, rcl_node_get_name(node));
+  }
+
+  PyObject * py_node_name_with_namespace = PyUnicode_FromString(node_name_with_namespace);
+  if (NULL == py_node_name_with_namespace) {
+    allocator.deallocate(node_name_with_namespace, allocator.state);
+    Py_DECREF(params_by_node_name);
+    return NULL;
+  }
+  allocator.deallocate(node_name_with_namespace, allocator.state);
```
sloretz: Nit, could deduplicate this line by putting it before the `NULL` check above.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3371,382 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+  PyObject * node_params = PyDict_GetItem(params_by_node_name, py_node_name_with_namespace);
+  if (NULL == node_params) {
+    Py_DECREF(params_by_node_name);
+    Py_DECREF(py_node_name_with_namespace);
+    return NULL;
+  }
+  // PyDict_GetItem is a borrowed reference. INCREF so we can return a new one.
+  Py_INCREF(node_params);
+
+  Py_DECREF(parameter_type_cls);
```
sloretz: I think this was already decref'd

sloretz: Need to `Py_DECREF(py_node_name_with_namespace);`

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3369,6 +3371,382 @@ rclpy_clock_set_ros_time_override(PyObject * Py_UNUSED(self), PyObject * args)
...
+      py_node_name = PyUnicode_FromString(
+        rcutils_format_string(allocator, "/%s", params->node_names[i]));
+    } else {
+      py_node_name = PyUnicode_FromString(params->node_names[i]);
+    }
+    if (NULL == py_node_name) {
+      return false;
+    }
+    PyObject * parameter_dict;
+    if (!PyDict_Contains(node_params_dict, py_node_name)) {
```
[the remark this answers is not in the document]
nuclearsandwich: I did a bit of empirical testing and it can return -1 when the item you pass in as a dict isn't. Since we create the dict that doesn't seem like a case worth checking at runtime.

nuclearsandwich:
Testing again using CMAKE_CURRENT_LIST_DIR to get the absolute source directory path rather than leaving it to the chance of the current directory.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5118/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1898/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4244/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5096/)

nuclearsandwich:
Thank you @sloretz and @mikaelarguedas. This PR wouldn't have been realized without the time you spent providing clear feedback and meticulous review.

PR #226, opened 2018-08-17
Publish parameter events

nuclearsandwich:
This is a recreation of a change that was accidentally authored in a temporary container :sob:. 

Adds a parameter events publisher to rclpy nodes and publishes events when parameters are successfully set atomically.

I did some basic testing of the earlier version of this using `ros2 topic echo` and comparing the results with rclcpp nodes but I need to do a bit more before I feel good putting this in review.

Connects to #202

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -14,7 +14,7 @@
 
 import weakref
 
-from rcl_interfaces.msg import SetParametersResult
+from rcl_interfaces.msg import ParameterEvent, SetParametersResult
 from rclpy.callback_groups import MutuallyExclusiveCallbackGroup
 from rclpy.client import Client
 from rclpy.clock import ROSClock
@@ -28,7 +28,8 @@ from rclpy.logging import get_logger
 from rclpy.parameter import Parameter
 from rclpy.parameter_service import ParameterService
 from rclpy.publisher import Publisher
-from rclpy.qos import qos_profile_default, qos_profile_services_default
+from rclpy.qos import qos_profile_default, qos_profile_parameter_events
+from rclpy.qos import qos_profile_services_default
 from rclpy.service import Service
 from rclpy.subscription import Subscription
 from rclpy.time_source import TimeSource
@@ -100,6 +101,9 @@ class Node:
 
         self.__executor_weakref = None
 
+        self._parameter_event_publisher = self.create_publisher(
+            ParameterEvent, 'parameter_events', qos_profile=qos_profile_parameter_events)
+
         if start_parameter_services:
             self._parameter_service = ParameterService(self)
 
@@ -166,8 +170,23 @@ class Node:
             result = SetParametersResult(successful=True)
 
         if result.successful:
+            parameter_event = ParameterEvent()
             for param in parameter_list:
-                self._parameters[param.name] = param
+                if Parameter.Type.NOT_SET == param.type_:
+                    if Parameter.Type.NOT_SET != self.get_parameter(param.name):
+                        # Parameter deleted. (Parameter had value and new value is not set)
+                        parameter_event.deleted_parameters.append(param.get_rcl_parameter())
+                        del self._parameters[param.name]
+                else:
+                    if Parameter.Type.NOT_SET == self.get_parameter(param.name):
+                        #  Parameter is new. (Parameter had no value and new value is set)
+                        parameter_event.new_parameters.append(param.get_rcl_parameter())
+                    else:
+                        # Parameter changed. (Parameter had a value and new value is set)
+                        parameter_event.changed_parameters.append(param.get_rcl_parameter())
+                    self._parameters[param.name] = param
+                self._parameter_event_publisher.publish(parameter_event)
+
         return result
 
     def _validate_topic_or_service_name(self, topic_or_service_name, *, is_service=False):
@@ -338,6 +357,11 @@ class Node:
         ret = True
         if self.handle is None:
             return ret
+
+        # TODO: is this needed/desired?
+        # Drop extra reference to parameter event publisher.
+        # It will be destroyed with other publishers below.
+        self._parameter_event_publisher = None
 
         while self.publishers:
             pub = self.publishers.pop()
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -358,7 +358,6 @@ class Node:
         if self.handle is None:
             return ret
 
-        # TODO: is this needed/desired?
         # Drop extra reference to parameter event publisher.
         # It will be destroyed with other publishers below.
         self._parameter_event_publisher = None
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -175,15 +175,17 @@ class Node:
                 if Parameter.Type.NOT_SET == param.type_:
                     if Parameter.Type.NOT_SET != self.get_parameter(param.name).type_:
                         # Parameter deleted. (Parameter had value and new value is not set)
-                        parameter_event.deleted_parameters.append(param.get_rcl_parameter())
+                        parameter_event.deleted_parameters.append(
+                            param.to_rcl_interface_parameter())
                         del self._parameters[param.name]
                 else:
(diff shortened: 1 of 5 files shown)


rclpy/rclpy/node.py
```diff
@@ -339,6 +358,11 @@ def destroy_node(self):
         if self.handle is None:
             return ret
 
+        # TODO: is this needed/desired?
```
tfoote: Explicit cleanup can't hurt.

rclpy/rclpy/node.py
```diff
@@ -166,8 +170,23 @@ def set_parameters_atomically(self, parameter_list):
...
             for param in parameter_list:
-                self._parameters[param.name] = param
+                if Parameter.Type.NOT_SET == param.type_:
+                    if Parameter.Type.NOT_SET != self.get_parameter(param.name).type_:
+                        # Parameter deleted. (Parameter had value and new value is not set)
+                        parameter_event.deleted_parameters.append(param.get_rcl_parameter())
+                        del self._parameters[param.name]
+                else:
+                    if Parameter.Type.NOT_SET == self.get_parameter(param.name).type_:
+                        #  Parameter is new. (Parameter had no value and new value is set)
```
mikaelarguedas: > (Parameter had no value and new value is set)

This is not a request for changes just a remark:
This is more: "Parameter existed with value not set or didn't exist on the node" (as the API doesn't allow us to distinguish ATM)
Pointing it out as it may be useful to distinguish these cases in some contexts.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5067/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1860/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4207/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5058/)

rclpy/rclpy/node.py
```diff
@@ -166,8 +170,23 @@ def set_parameters_atomically(self, parameter_list):
...
 
         if result.successful:
+            parameter_event = ParameterEvent()
             for param in parameter_list:
-                self._parameters[param.name] = param
+                if Parameter.Type.NOT_SET == param.type_:
+                    if Parameter.Type.NOT_SET != self.get_parameter(param.name).type_:
+                        # Parameter deleted. (Parameter had value and new value is not set)
+                        parameter_event.deleted_parameters.append(param.get_rcl_parameter())
+                        del self._parameters[param.name]
```
mikaelarguedas: What happens in the else case?

(I didnt try it but by reading the code I wonder)

As a user does that mean that if I:
- run list parameters: see parameter `foo`
- send a request to delete `foo` and receive success
- I run list_parameters again:
  - if the parameter was of any type but "NOT_SET": it has been deleted and is not part of the list
  - if the parameter was of type "NOT_SET": it will still be in the list
    - to actually delete it I should:
      - set a value
      - request a deletion

Is that correct?

rclpy/rclpy/parameter.py
```diff
@@ -132,3 +133,6 @@ def get_parameter_value(self):
         elif Parameter.Type.STRING_ARRAY == self.type_:
             parameter_value.string_array_value = self.value
         return parameter_value
+
+    def get_rcl_parameter(self):
```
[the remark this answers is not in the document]
nuclearsandwich: yeah the rcl prefix was to distinguish it from the current Parameter class. I'm happy to change it if another method name will be clearer.

rclpy/rclpy/node.py
nuclearsandwich: For this behavior I mirrored the rclcpp API which drops deleted parameters from the internal data structure on deletion as well. When getting parameters via ROS service NOT_SET is the type assigned to parameters with no current value, whether they have ever been assigned a value or not.

While it might be useful to treat an explicitly NOT_SET parameter different from a parameter absent from internal storage it isn't a state that is representable via the current interfaces and I don't think we should make it part of the local API without amending the interface definitions to convey both different potential meanings of NOT_SET.

rclpy/rclpy/node.py
nuclearsandwich: That's correct but the final subcase "if the parameter was of type "NOT_SET": it will still be in the list" is somewhat vacuous as in order for a NOT_SET parameter to get in the list it would need to be added directly to the parameters map circumventing our API.

mikaelarguedas: What does vacuous mean?
Do you mean that scenario should not happen and should be handled as such (warning, error, something else than just ignoring it?) ?

nuclearsandwich: > What does vacuous mean?

If they use the node parameters API, there won't be a Parameter of type NOT_SET stored and thus it wouldn't be enumerated in a list_parameters service call.

> Do you mean that scenario should not happen and should be handled as such (warning, error, something else than just ignoring it?) ?

The scenario shouldn't happen but I also don't know that it needs to be explicitly warned or errored any more than any other behavior that would arise from manual manipulation of Node class internals. If you're concerned about seeing NOT_SET parameters in the list_parameters results I think an explicit check in that service handler would be the way to achieve it.

All that said, if there's an IllegalStateException or similar in python that we want to raise here I'm not opposed.

rclpy/rclpy/parameter.py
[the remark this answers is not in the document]
mikaelarguedas: I think `to_rcl_interface_parameter` to match the `from_rcl_interface_parameter` method from the same class would be clearer.

rclpy/rclpy/node.py
mikaelarguedas: I guess it's just unclear to me how setting "NOT_SET" is handled in both languages.
As you say setting a parameter to "NOT_SET" and value "None" has no effect on the node (no storage allocated and not added to the list of parameters) I think it's fine to merge as is (same below).

We can revisit when we go further on the "parameter declaration" and "read-only" parameters where the distinction between "a parameter that exist but doesnt have a value yet" and a "parameter that doesnt exist", will be needed.

nuclearsandwich: Setting NOT_SET in rclcpp will erase the parameter from internal storage if it was previously allocated: https://github.com/ros2/rclcpp/blob/25a9b4e339867bacb8ba5b92b955420cec86596a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L216-L221

mikaelarguedas: then I guess we should do the same here for consistency? aka unconditionally delete it if it exists regardless of its type/value, but add it to the `deleted_parameters` list only if it had a value

rclpy/rclpy/parameter.py
```diff
@@ -14,6 +14,7 @@
 
 from enum import Enum
 
+from rcl_interfaces.msg import Parameter as RCLParameter
```
wjwwood: This is a strange rename to me, at first glance, since it's clearly a message a not a type defined within `rcl`. I see that it comes from `rcl_interfaces` which make some sense, but it was surprising to me. `ParameterMsg` is a clearer name to me, since it will always be a message but my not always be in the rcl interfaces or have anything to do with rcl in the future.

This is a pure nitpick, feel free to ignore it.

mikaelarguedas: :+1: I meant to scope that with the [comment with about renaming the method](https://github.com/ros2/rclpy/pull/226#discussion_r210959738) and apparently didn't post it. I'm also in favor of a name expressing the fact that this is a message and not a structure coming from rcl.

Following that logic, the method converting to and from parameter message could also be named to refer to messages rather than rcl_interfaces

nuclearsandwich: > This is a strange rename to me, at first glance

I agree with both of you given the explanation and have pushed 6ddfbba to use ParameterMsg and to/from_parameter_msg. As Mikael suspected the rcl prefix was being used because it was an rcl_interfaces type but using just rcl adds confusing with rcl types. ParameterMsg is apt and available so it's what we'll use.

PR #228, opened 2018-08-20
Add Node API method for setting the parameters_callback.

nuclearsandwich:
In #214 I wired up support for invoking a callback function to accept or reject a set of parameters but did not add the API to set that callback or tests for that functionality.

This pull request adds a `set_parameters_callback` method to rclpy.Node and some unit tests verifying that the callbacks are used and that the parameter values are set or not appropriately.

Every test case I used has been put into the unit tests below.

Connects to #202

### DIFF
--- a/rclpy/test/test_parameters_callback.py
+++ b/rclpy/test/test_parameters_callback.py
@@ -1,0 +1,129 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


rclpy/test/test_parameters_callback.py
```diff
@@ -0,0 +1,129 @@
...
+        rclpy.shutdown()
+
+    def setUp(self):
+        self.node = rclpy.create_node('parameters_callback_node')
+
+    def tearDown(self):
+        self.node.destroy_node()
+
+    def test_set_callback_accepting_all(self):
+        global callback_called
```
mikaelarguedas: is it necessary for this to be global? could it be local to each test case and nonlocal to the callbacks?

nuclearsandwich: TIL about `nonlocal` That's precisely what I needed. Thanks.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5095/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1879/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4225/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5076/)

rclpy/test/test_parameters_callback.py
nuclearsandwich: resolved by 73f6cf2

Issue #224, opened 2018-08-15
Missing exception states in rclpy


hemantku:
## Feature request
rclpy seems to missing ROS Exceptions from rospy

#### Feature description
rospy has the following exceptions:
1. ROSInterruptException/ROSException - used in joint_state_publisher
2. ROSInitException - used in joint_state_publisher
3. ROSTimeMovedBackwardsException
4. ROSInternalException
5. TransportException/TransportTerminated/TransportInitError

rclpy seems to missing all the above exceptions. The following are more relevant to the migration
1. ROSInterruptException/ROSException - used in joint_state_publisher 
2. ROSInitException - used in joint_state_publisher

Additionally, it might be useful to have ROSTimeMovedBackwardsException, I am not sure about 4 and 5

https://github.com/ros/ros_comm/blob/60be2504d5341eb3bc80d86e5d16f1c53dc97ccd/clients/rospy/src/rospy/exceptions.py

#### Implementation considerations
1.  There is existing exceptions.py which could be extended to encase the missing exception 
2. Create rosexception.py which entails all these exceptions

wjwwood:
Reference: https://discourse.ros.org/t/mapping-between-rospy-and-rclpy/5737?u=wjwwood

sloretz:
Hi @hemantku, are there more exceptions than `ROSInterruptExcption` that you would like to see ported? Would you mind adding to the issue description the exceptions you would like to see and the use cases that needs them? I think if the ticket was more specific it would be easier to ask for help on and could be completed quicker.

hemantku:
Hi @sloretz, I have made the ticket more specific(sorry about making it generic). I can add these exceptions if you guys agree with what I see I can make change and send you guys the PR

clalancette:
At this point I think we are far enough away from rospy that we don't need to do this port.  So I'm going to close this one out.

PR #232, opened 2018-08-22
Missing exception states in rclpy

hemantku:
#224

### DIFF
--- a/rclpy/rclpy/exceptions.py
+++ b/rclpy/rclpy/exceptions.py
@@ -65,3 +65,42 @@ class InvalidServiceNameException(NameValidationException):
 
     def __init__(self, name, error_msg, invalid_index, *args):
         NameValidationException.__init__(self, 'service name', name, error_msg, invalid_index)
+
+
+class AlreadyInitalizedException(Exception):
+    """Raised when the rclpy.init() has already been."""
+
+    def __init__(self, *args):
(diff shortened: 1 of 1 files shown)


sloretz:
I agree with @wjwwood that exceptions should be added with code raising them. Without that I'm not sure if any given exception will be necessary.

hemantku:
I have used these exceptions in the joint_state_publisher package which I migrated to ROS 2. I would be happy to include this along with appropriate CR's for initiate and shutdown as they are made

wjwwood:
@hemantku how are you using these exceptions in `joint_state_publisher`? are you just `try-except`'ing them or are you throwing them yourself?

hemantku:
@wjwwood  Try-except'ing them

wjwwood:
Well, if nothing is raising them then why do you need to try-except them? 😄

wjwwood:
Are you trying to have one Python file that supports ROS 1 and ROS 2 maybe?

If not, then I'd really say let's leave this pr (perhaps close it but leave the branch/fork) and reintroduce the exceptions when they're being used.

It's also possible that there are already functions which _should_ raise them already but are not. In which case updating this pr to replace existing `RuntimeError` raises with more specific exceptions would be acceptable as well.

hemantku:
Alright, I will close this PR out and introduce them along when I replace the RuntimeError

PR #235, opened 2018-09-07
add getter for tuple with seconds and nanoseconds

dirk-thomas:
Exposing the tuple avoids calling code to perform the same computation in numerous places.

### DIFF
--- a/rclpy/rclpy/time.py
+++ b/rclpy/rclpy/time.py
@@ -41,7 +41,7 @@ class Time:
         return _rclpy.rclpy_time_point_get_nanoseconds(self._time_handle)
 
     def seconds_nanoseconds(self):
-        nanoseconds = self.nanoseconds()
+        nanoseconds = self.nanoseconds
         return (int(nanoseconds / 1e9), nanoseconds % 1e9)
 
     @property
--- a/rclpy/rclpy/time.py
+++ b/rclpy/rclpy/time.py
@@ -41,6 +41,12 @@ class Time:
         return _rclpy.rclpy_time_point_get_nanoseconds(self._time_handle)
 
     def seconds_nanoseconds(self):
+        """
+        Get time as separate seconds and nanoseconds components.
+
+        :returns: 2-tuple seconds and nanoseconds
+        :rtype: tuple(int, int)
+        """
         nanoseconds = self.nanoseconds
         return (int(nanoseconds / 1e9), nanoseconds % 1e9)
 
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/time.py
```diff
@@ -40,6 +40,10 @@ def __init__(self, *, seconds=0, nanoseconds=0, clock_type=ClockType.SYSTEM_TIME
     def nanoseconds(self):
         return _rclpy.rclpy_time_point_get_nanoseconds(self._time_handle)
 
+    def seconds_nanoseconds(self):
+        nanoseconds = self.nanoseconds()
+        return (int(nanoseconds / 1e9), nanoseconds % 1e9)
```
yechun1: The 'nanosec' field must of type 'int',  when assign to msg.header.stamp.nanosec.

rclpy/rclpy/time.py
```diff
@@ -40,6 +40,10 @@ def __init__(self, *, seconds=0, nanoseconds=0, clock_type=ClockType.SYSTEM_TIME
     def nanoseconds(self):
         return _rclpy.rclpy_time_point_get_nanoseconds(self._time_handle)
 
+    def seconds_nanoseconds(self):
+        nanoseconds = self.nanoseconds()
```
yechun1: TypeError: 'int' object is not callable
Should be `self.nanoseconds` instead of `self.nanoseconds()` ?

yechun1:
below changes could work on my system:

     def seconds_nanoseconds(self):
-        nanoseconds = self.nanoseconds()
-        return (int(nanoseconds / 1e9), nanoseconds % 1e9)
+        nanoseconds = self.nanoseconds
+        return (int(nanoseconds / 1e9), int(nanoseconds % 1e9))

rclpy/rclpy/time.py
dirk-thomas: Absolutely, fixed in 4e4255bd9b07aca611af743c4ca76b52490b2c56.

rclpy/rclpy/time.py
dirk-thomas: I am pretty sure the result of the modulo operator is already an integer. Or do you have a case where the result ended up being not an integer?

dirk-thomas:
> I added a doc string and test, hope you don't mind. LGTM

I don't mind at all. That is great. One docstring in the class is certainly better than none :+1:

PR #222, opened 2018-08-10
Callbacks when time jumps

sloretz:
This adds callbacks when ROSTime jumps. The implementation closely mimics the implementation in `rclcpp`.

Time jump callbacks use the `rcl_clock_t` time jump callbacks. Python code is called in the callbacks using `PyObject_CallObject()`.

Requires ros2/rcl#284

PR #237, opened 2018-09-13
Add methods on Mock class for Python 3.5 compatibility

sloretz:
This fixes some test failures on xenial caused by ros2/rclpy#222. The tests used methods on `unittest.mock.Mock` that were added in python 3.6. This PR ~~~adds a subclass with those methods defined~~~ monkey patches those methods onto the mock class if they are not present.

Xenial CI just rclpy
[Build Status](https://ci.ros2.org/job/ci_linux/5177/)

CI bionic just rclpy
[Build Status](https://ci.ros2.org/job/ci_linux/5178/)

Will run full CI post review

Connects to ros2/build_cop#147

### DIFF
--- a/rclpy/test/test_clock.py
+++ b/rclpy/test/test_clock.py
@@ -23,7 +23,7 @@ from rclpy.clock import ROSClock
 from rclpy.duration import Duration
 from rclpy.time import Time
 
-from .mock_compat import *  # noqa
+from .mock_compat import __name__ as _  # noqa: ignore=F401
 
 
 class TestClock(unittest.TestCase):
--- a/rclpy/test/test_clock.py
+++ b/rclpy/test/test_clock.py
@@ -14,7 +14,6 @@
 
 import time
 import unittest
-from unittest.mock import Mock
 
 from rclpy.clock import Clock
 from rclpy.clock import ClockType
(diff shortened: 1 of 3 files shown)


sloretz:
Looks like CI is quick, Full CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5181/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1953/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4305/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5161/)

dirk-thomas:
You don't need a subclass you could just add the methods to the existing class.

sloretz:
@dirk-thomas monkey patch instead of subclass in acc20659bdc78088c2748b33edb23d6d474bb773

rclpy/test/test_clock.py
```diff
@@ -23,6 +23,8 @@
 from rclpy.duration import Duration
 from rclpy.time import Time
 
+from .mock_compat import *  # noqa
```
sloretz: Can't do `import .mock_compat` becase [PEP 328](https://www.python.org/dev/peps/pep-0328/) forbids it, and can't do `import mock_compat` because pytest doesn't put the test directory into `sys.path`.

dirk-thomas: Only suppress the specific category rather than a blank `noqa`.

PR #238, opened 2018-09-14
Avoid repeated fini-ing on failure to parse yaml params

dhood:
I was getting a segfault parsing an invalid yaml file.

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -3798,7 +3798,9 @@ _parse_param_files(
       rcl_params_t * params = rcl_yaml_node_struct_init(allocator);
       if (!rcl_parse_yaml_file(param_files[i], params)) {
         // failure to parse will automatically fini the params struct
-        PyErr_Format(PyExc_RuntimeError, "Failed to parse yaml params file: %s", param_files[i]);
+        PyErr_Format(PyExc_RuntimeError, "Failed to parse yaml params file '%s': %s",
+          param_files[i], rcl_get_error_string_safe());
+        rcl_reset_error();
         successful = false;
       } else {
         if (!_populate_node_parameters_from_rcl_params(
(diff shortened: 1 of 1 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -3804,11 +3804,10 @@ _parse_param_files(
         if (!_populate_node_parameters_from_rcl_params(
             params, allocator, parameter_cls, parameter_type_cls, params_by_node_name))
         {
-          rcl_yaml_node_struct_fini(params);
           successful = false;
         }
+        rcl_yaml_node_struct_fini(params);
       }
-      rcl_yaml_node_struct_fini(params);
```
mikaelarguedas: In the successful case, shouldn't we fini the params structure?

dhood: that's done at https://github.com/ros2/rclpy/pull/238/files#diff-80dda03110b5be606b823d6b0558b5c2R3811 no?

mikaelarguedas: Yeah I misread and didn't edit my commen in time :smile: 
I got surprised by the fact that we went from fini'ing twice to not at all in case of failure. But it looks like the structure is deallocated by `rcl_parse_yaml_file`  on failure: https://github.com/ros2/rcl/blob/0a6d9186e8fbd133308d0ad4bc3b75d61078b04d/rcl_yaml_param_parser/src/parser.c#L1382

mikaelarguedas: I'll rephrase:

Before we were deallocating twice in both possible failure cases. The case you faced where it was deallocating in `rcl_parse_yaml_file` on parsing failure. And in the case of `_populate_node_parameters_from_rcl_params` failing.

The current state of the PR looks good :+1:, deallocating only once and only if `rcl_parse_yaml_file` succeeds (regardless of the success of `_populate_node_parameters_from_rcl_params`)

dhood:
[Build Status](https://ci.ros2.org/job/ci_windows/5166/)

PR #242, opened 2018-10-08
rclpy raw subscriptions

josephduchesne:
Note: This implementation is incomplete and at the moment this PR exists primarily for getting feedback.
This is being done so that [ros2cli#132](https://github.com/ros2/ros2cli/issues/132) can implement "ros2 topic bw".

- [x] Initial prototype 
- [x] Correct memory allocation implementation
- [x] Add tests
- [x] Add example to demos/demo_nodes_py

The general idea is to allow creating a raw subscription like so:
```
class ListenerRaw(Node):

    def __init__(self):
        super().__init__('listener')
        self.sub = self.create_subscription(String, 'chatter', self.chatter_callback, raw=True)

    def chatter_callback(self, msg):
        self.get_logger().info('I heard: [%s]' % msg)
```
This will call rcl_take_serialized_message rather than rcl_take internally, and return a python byte array representing the serialised message.

> [INFO] [listener]: I heard: [b'\x00\x01\x00\x00\x05\x00\x00\x00Test\x00\x00\x00\x00']
(For the message "Test").

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -2642,6 +2642,8 @@ rclpy_take_raw(rcl_subscription_t * subscription)
   rmw_ret_t r_fini = rmw_serialized_message_fini(&msg);
   if (r_fini != RMW_RET_OK) {
     PyErr_Format(PyExc_RuntimeError, "Failed to deallocate message buffer: %d", r_fini);
+    Py_DECREF(python_bytes);
+    return NULL;
   }
   return python_bytes;
 }
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -2251,23 +2251,23 @@ rclpy_wait_set_add_entity(PyObject * Py_UNUSED(self), PyObject * args)
   if (0 == strcmp(entity_type, "subscription")) {
     rcl_subscription_t * subscription =
       (rcl_subscription_t *)PyCapsule_GetPointer(pyentity, "rcl_subscription_t");
-    ret = rcl_wait_set_add_subscription(wait_set, subscription);
+    ret = rcl_wait_set_add_subscription(wait_set, subscription, NULL);
   } else if (0 == strcmp(entity_type, "client")) {
     rcl_client_t * client =
       (rcl_client_t *)PyCapsule_GetPointer(pyentity, "rcl_client_t");
-    ret = rcl_wait_set_add_client(wait_set, client);
+    ret = rcl_wait_set_add_client(wait_set, client, NULL);
   } else if (0 == strcmp(entity_type, "service")) {
     rcl_service_t * service =
       (rcl_service_t *)PyCapsule_GetPointer(pyentity, "rcl_service_t");
-    ret = rcl_wait_set_add_service(wait_set, service);
+    ret = rcl_wait_set_add_service(wait_set, service, NULL);
   } else if (0 == strcmp(entity_type, "timer")) {
     rcl_timer_t * timer =
       (rcl_timer_t *)PyCapsule_GetPointer(pyentity, "rcl_timer_t");
-    ret = rcl_wait_set_add_timer(wait_set, timer);
+    ret = rcl_wait_set_add_timer(wait_set, timer, NULL);
   } else if (0 == strcmp(entity_type, "guard_condition")) {
     rcl_guard_condition_t * guard_condition =
       (rcl_guard_condition_t *)PyCapsule_GetPointer(pyentity, "rcl_guard_condition_t");
-    ret = rcl_wait_set_add_guard_condition(wait_set, guard_condition);
+    ret = rcl_wait_set_add_guard_condition(wait_set, guard_condition, NULL);
   } else {
     ret = RCL_RET_ERROR;  // to avoid a linter warning
     PyErr_Format(PyExc_RuntimeError,
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -2431,14 +2431,20 @@ rclpy_take(PyObject * Py_UNUSED(self), PyObject * args)
 {
   PyObject * pysubscription;
   PyObject * pymsg_type;
+  PyObject * pyraw;
 
-  if (!PyArg_ParseTuple(args, "OO", &pysubscription, &pymsg_type)) {
+  if (!PyArg_ParseTuple(args, "OOO", &pysubscription, &pymsg_type, &pyraw)) {
     return NULL;
   }
   if (!PyCapsule_CheckExact(pysubscription)) {
     PyErr_Format(PyExc_TypeError, "Argument pysubscription is not a valid PyCapsule");
     return NULL;
   }
+  bool raw = false;
+  if (PyObject_IsTrue(pyraw) == 1) {
+    raw = true;
+  }
+
   rcl_subscription_t * subscription =
     (rcl_subscription_t *)PyCapsule_GetPointer(pysubscription, "rcl_subscription_t");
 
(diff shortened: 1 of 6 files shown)


josephduchesne:
I've updated the implementation to use rmw's allocator (I believe) properly. Let me know if this is a reasonable way to go about this and I'll round it out with tests and an example.

clalancette:
The idea here looks good to me.  This needs a rebase, and as you mentioned tests and examples, but I think we can go ahead with this.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -2459,7 +2468,47 @@ rclpy_take(PyObject * Py_UNUSED(self), PyObject * args)
...
   }
 
-  rcl_ret_t ret = rcl_take(subscription, taken_msg, NULL);
+  rcl_ret_t ret;
+  if (raw) {
+    destroy_ros_message(taken_msg);
+    Py_DECREF(pymetaclass);
+
+    // Create a serialized message object
+    // Not 100% sure that 0u is the best size, but it will do
```
Karsten1987: it will do as the message then gets dynamically resized down the stack when the actual message is taken from the middleware. 
I am not sure where to put this best in the rclpy API, but I guess it would be nice to have some sort of memory allocation system where this can eventually become a parameter to configure.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -2459,7 +2468,47 @@ rclpy_take(PyObject * Py_UNUSED(self), PyObject * args)
     return PyErr_NoMemory();
   }
 
-  rcl_ret_t ret = rcl_take(subscription, taken_msg, NULL);
+  rcl_ret_t ret;
+  if (raw) {
```
Karsten1987: shouldn't that check happen before you call `create_ros_message()` in line 2465? 
It seems like you create a message and then immediately destroy it if it's not `raw`.

Karsten1987: also I feel that this `if` `else` part could deserve its own function for clarity. But that's just my opinion maybe.

rclpy/src/rclpy/_rclpy.c
josephduchesne: For now I think it's ok that it just uses the default allocator. If we need a more specialized one, it can be added but this functionality will probably always be fairly niche (debugging and diagnostics most likely).

josephduchesne:
I think this PR is done now. I don't believe that this functionality requires a demo node, although I can happily add one. It's fairly esoteric and demo nodes are aimed at beginners. 

Unit tests pass locally for the package, however the build farm test is failing. I think it's due to the build farm compiling against bouncy debs (specifically bouncy rcl), when this is built against and branched off of the latest master branch. Let me know if you believe that is my fault and I'll dig further to fix it.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -24,10 +24,13 @@
 #include <rcl_yaml_param_parser/parser.h>
 #include <rcl_interfaces/msg/parameter_type__struct.h>
 #include <rcutils/format_string.h>
+#include <rcutils/allocator.h>
```
Karsten1987: nitpick: alphabetic order.

rclpy/test/test_node.py
```diff
@@ -66,6 +66,30 @@ def test_create_subscription(self):
...
+        primitives_msg = Primitives()
+        cycle_count = 0
+        while rclpy.ok() and cycle_count < 5 and self.raw_subscription_msg is None:
+            primitives_pub.publish(primitives_msg)
+            cycle_count += 1
+            rclpy.spin_once(self.node, timeout_sec=1)
+        self.assertIsNotNone(self.raw_subscription_msg, 'raw subscribe timed out')
+        self.assertIs(type(self.raw_subscription_msg), bytes, 'raw subscribe did not return bytes')
+        # The length might be implementation dependant, but shouldn't be zero
+        self.assertNotEqual(len(self.raw_subscription_msg), 0, 'raw subscribe invalid length')
```
Karsten1987: Does this really test that the incoming message is complete and correct?

Karsten1987:
@josephduchesne thanks for all the work done here.
I ran a first CI build on that:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5640/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2258/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4662/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5529/)

In the meantime, I am actually in favor of having a demo similar to the C++ version of it so that the demo packages don't diverge too much. I don't consider them only aiming at beginners, but also to literally demonstrate the capabilities of ROS 2 and their respective APIs.

josephduchesne:
> One remark I'd have though is that the test is maybe not necessarily complete here. Just because the received message has a non-zero amount of bytes, doesn't necessarily mean the message is complete and correct.
This is true, though it seems unlikely that it's incorrect given that the rclpy code does not manipulate the array. If a non-zero array is passed up, it means that rcl_take_serialized_message returned bad data, and it should have its own tests. I understand the desire for completeness though, and if we do expose deserialization to rclpy 

> In your opinion, how much effort would it be to expose the rcl_serialize and rcl_deserialize functions to this PR? This would allow to instantiate the message being published, serialize it and byte-wise compare the received message.

I poked around the rcl code and it appears that this doesn't exist in rcl. The functions you are thinking of are probably rmw_serialize and rmw_deserialize. That raises some design questions.

How would these look:
```c
rmw_ret_t
rmw_deserialize(
  const rmw_serialized_message_t * serialized_message,
  const rosidl_message_type_support_t * type_support,
void * ros_message)
```
```python
string_message = rclpy.deserialize(input_bytes, std_msgs.msg.String)
```

```c
rmw_ret_t
rmw_serialize(
  const void * ros_message,
  const rosidl_message_type_support_t * type_support,
  rmw_serialized_message_t * serialized_message)
```
```python
output_bytes = rclpy.serialize(std_msgs_string_object)
```

Should I go ahead with this, or is this too much of a diversion from this original ticket?

It's not prohibitively complex, but I'm not sure the use case justifies the effort at the moment.

This would be required if someone wanted to implement something like ros1's rosbag (ros2bag doesn't use the rmw serialization system AFIK, and for good reasons- although it appears to me to result in an inevitable performance hit due to the abstraction). Other than that I don't think it provides much value.

rclpy/test/test_node.py
josephduchesne: It only partially tests the incoming message. 

That said, it seems unlikely that it's incorrect given that the rclpy code does not manipulate the array. If a non-zero array is passed up, it means that rcl_take_serialized_message returned bad data, and it should have its own tests. I understand the desire for completeness though, and if we do expose deserialization to rclpy we can make this test more robust.

rclpy/src/rclpy/_rclpy.c
josephduchesne: Fixed

Karsten1987:
> Should I go ahead with this, or is this too much of a diversion from this original ticket?
> It's not prohibitively complex, but I'm not sure the use case justifies the effort at the moment.

I believe it makes sense to add this here, e.g. once you've received binary data, eventually the binary data should be converted to ROS messages again. This use case is also helpful for the tests of this PR.
We would appreciate if you spend a little bit more time on this, but it's entirely up to you :)

In case, the proposed python API for the `rmw_serialize` and `rmw_deserialize` look good to me. You could maybe stay consistent with the `rclpy` API and swap the arguments similar like here:
``` python
self.pub = self.create_publisher(String, 'chatter')
```
https://github.com/ros2/demos/blob/master/demo_nodes_py/demo_nodes_py/topics/talker.py#L26

With that the proposed API:
``` python
string_message = rclpy.deserialize(std_msgs.msg.String, input_bytes)
binary_message = rclpy.serialize(std_msgs.msg.String, message)
```

Let us know if you want to work a little bit more on this. Otherwise, we could go ahead and merge this and ticket the remaining tasks.

dirk-thomas:
> `rclpy.serialize(std_msgs.msg.String, message)`

For `rclpy.serialize` the first argument isn't necessary since it is implicitly available as `type(message)`.

josephduchesne:
Unfortunately, I don't think I'll have time in the next few weeks to work on further extending rclpy, and it would be nice to unblock the `rostopic bw` work that I originally wrote this for. If it's possible for this to be merged as is or with minimal changes, that would be nice. 

I can try to do the rclpy.deserialize/rclpy.serialize work around the holidays, but that's probably about as much work as this PR by itself, again.

josephduchesne:
@Karsten1987 can this be merged?

Karsten1987:
The current branch does not compile with the latest master version and has to be rebased as far as I can tell.
Then further, there was a recent change in the crystal API which takes the serialized messages with a buffer of type `uint8_t` and not as `char *`. This change has to be addressed here as well.

josephduchesne:
Latest master branch merged in, compiled and tested in a clean source install of ros2 crystal (with the minor fixes performed). Also updated the sister request in ros2/demos#287 in the same way although it worked without changes after merging.

Karsten1987:
Thanks for iterating over this.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5992/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2501/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4944/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5868/)

rclpy/src/rclpy/_rclpy.c
```diff
@@ -2601,6 +2604,48 @@ rclpy_wait(PyObject * Py_UNUSED(self), PyObject * args)
...
+    rmw_ret_t r_fini = rmw_serialized_message_fini(&msg);
+    if (r_fini != RMW_RET_OK) {
+      PyErr_Format(PyExc_RuntimeError, "Failed to deallocate message buffer: %d", r_fini);
+    }
+    return NULL;
+  }
+  PyObject * python_bytes = PyBytes_FromStringAndSize((char *)(msg.buffer), msg.buffer_length);
+  rmw_ret_t r_fini = rmw_serialized_message_fini(&msg);
+  if (r_fini != RMW_RET_OK) {
+    PyErr_Format(PyExc_RuntimeError, "Failed to deallocate message buffer: %d", r_fini);
```
Karsten1987: should it return `NULL` in this case?

josephduchesne: It now returns NULL after decremented the `python_bytes` refcount in this error condition. If this is an adequate solution, I think this might finally be merge-able :)

Karsten1987:
Thanks for iterating over this.

PR #243, opened 2018-10-08
Add library path hook.

nuclearsandwich:
Set library path environment hook for this package. This package is one of a handful which installs libraries without setting an environment hook to add them to the platform library path.

Connects to ros2/ros_workspace#10


nuclearsandwich:
CI for this change was run successfully and linked in https://github.com/ros2/ros_workspace/pull/10#issuecomment-428005194

PR #249, opened 2018-11-19
refactor to allow for non-global init

wjwwood:
Notable this pr, unlike the rclcpp one, does not provide access to the ingoing init options, partially because I think it's not as useful right now and partially because I'm completely out of time to work on this set of changes.

Connects to ros2/rmw#154

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -14,17 +14,17 @@
 
 import sys
 
+from rclpy.utilities import get_default_context
 from rclpy.utilities import get_rmw_implementation_identifier  # noqa
-from rclpy.utilities import ok
 from rclpy.utilities import shutdown as _shutdown
 from rclpy.utilities import try_shutdown  # noqa
 
 
-def init(*, args=None):
+def init(*, args=None, context=None):
+    context = get_default_context if context is None else context
     # imported locally to avoid loading extensions on module import
     from rclpy.impl.implementation_singleton import rclpy_implementation
-    return rclpy_implementation.rclpy_init(
-        args if args is not None else sys.argv)
+    return rclpy_implementation.rclpy_init(args if args is not None else sys.argv, context)
 
 
 # The global spin functions need an executor to do the work
@@ -97,7 +97,7 @@ def spin_once(node, *, timeout_sec=None):
         executor.remove_node(node)
 
 
-def spin(node):
+def spin(node, executor=None):
     """
     Execute work blocking until the library is shutdown.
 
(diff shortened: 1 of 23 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -402,6 +485,11 @@ rclpy_init(PyObject * Py_UNUSED(self), PyObject * args)
   }
   int num_args = (int)pysize_num_args;
 
+  rcl_context_t * context = (rcl_context_t *)PyCapsule_GetPointer(pycontext, "rcl_context_t");
+  if (NULL == context) {
+    return NULL;
```
sloretz: Need to `Py_DECREF(pyargs)` before returning or move this above line 476

rclpy/rclpy/__init__.py
```diff
@@ -97,37 +97,40 @@ def spin_once(node, *, timeout_sec=None):
...
-def spin(node):
+def spin(node, executor=None):
     """
     Execute work blocking until the library is shutdown.
 
     Callbacks will be executed in a SingleThreadedExecutor until shutdown() is called.
     This method blocks.
 
     :param node: A node to add to the executor to check for work.
+    :param executor: The executor to use, of the global executor if None is passed.
```
jacobperron: typo: "**or** the global executor"

rclpy/rclpy/__init__.py
```diff
@@ -97,37 +97,40 @@ def spin_once(node, *, timeout_sec=None):
...
+def spin_until_future_complete(node, future, executor=None):
     """
     Execute work until the future is complete.
 
     Callbacks and other work will be executed in a SingleThreadedExecutor until future.done()
     returns True or rclpy is shutdown.
 
+    :param node: A node to add to the executor to check for work.
     :param future: The future object to wait on.
+    :param executor: The executor to use, of the global executor if None is passed.
```
jacobperron: **or**

rclpy/rclpy/__init__.py
```diff
@@ -97,37 +97,40 @@ def spin_once(node, *, timeout_sec=None):
         executor.remove_node(node)
 
 
-def spin(node):
+def spin(node, executor=None):
```
jacobperron: Should we also be optionally passing the executor to `spin_once()`?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -76,6 +76,66 @@ static void * get_capsule_pointer(PyObject * pymetaclass, const char * attr)
...
+  if (NULL == context) {
+    PyErr_Format(PyExc_MemoryError, "Failed to allocate memory for context");
+    return NULL;
+  }
+  *context = rcl_get_zero_initialized_context();
+  // if it fails, error is set and NULL is returned as it should
+  return PyCapsule_New(context, "rcl_context_t", _rclpy_context_capsule_destructor);
+}
+
+
```
jacobperron: nit: remove extra empty line

gonzodepedro:
LGTM

rclpy/src/rclpy/_rclpy.c
wjwwood: Does this need to be decref'ed on 484 (just above too)?

sloretz: Yeah, it should be decref'd there too

wjwwood: According to this, `PyArg_ParseTuple` returns borrowed refs and does not need to be decref'ed:

https://grokbase.com/t/python/python-list/0033005faw/pyarg-parsetuple-py-decref

I can't find anything definitive in the Python docs for `PyArg_ParseTuple` for `PyObject` entries.

wjwwood: Oh but that's really old... I'll keep looking.

sloretz: > According to this, PyArg_ParseTuple returns borrowed refs and does not need to be decref'ed:

Correct. After that on line 476 `pyargs` is set to the return of `PySequence_List()` which returns a new reference that does need to be decref'd.

wjwwood: So, there's this:

> Note that any Python object references which are provided to the caller are borrowed references; do not decrement their reference count!
-- https://docs.python.org/3/c-api/arg.html#other-objects

So I'm not sure decrefing this is right. In fact, what I've been seeing is examples with `PyArgs_ParseTuple` that do not decref and even warnings that decrefing borrowed references is dangerous.

Perhaps I'm misunderstand? @sloretz what do you think?

wjwwood: So `PySequence_List` returns a _new reference_, which does need decref, I'll make sure that's happening correctly.

rclpy/rclpy/__init__.py
wjwwood: Yeah, good idea, I added that in 04398ba

Issue #257, opened 2019-01-09
Action implementation


jacobperron:
#### Feature description

Add support for ROS Actions as described in ros2/design#193.

Includes:
- [x] Action client #262
- [ ] Action server #270 

Metaticket to ros2/ros2#583
Requires ros2/rosidl_python#14

#### Implementation considerations

We'd like to add this feature to Crystal, so we should avoid disruptions to the existing code (if possible).
Isolating implementation to a new `rclpy.action` module would be ideal.

Usage examples can be found in ros2/examples#222 (subject to change).

PR #269, opened 2019-02-05
Abstract type conversions into functions

jacobperron:
This helps with readability and maintainability.
Also eliminated use of assertions during conversions, defering to exceptions.

I plan to make use of these functions for the Action extension module (#257) in a follow-up PR.

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -83,13 +83,18 @@ target_include_directories(rclpy_common
   src/rclpy_common/include
   ${PythonExtra_INCLUDE_DIRS}
 )
+ament_target_dependencies(rclpy_common
+  "rmw"
+)
+
+# Causes the visibility macros to use dllexport rather than dllimport,
+# which is appropriate when building the dll but not consuming it.
+target_compile_definitions(rclpy_common PRIVATE "RCLPY_COMMON_BUILDING_DLL")
+
 install(TARGETS rclpy_common
   ARCHIVE DESTINATION lib
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
-)
-ament_target_dependencies(rclpy_common
-  "rmw"
 )
 
 add_library(
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -75,10 +75,19 @@ endfunction()
 add_library(rclpy_common
   SHARED src/rclpy_common/src/common.c
 )
-target_include_directories(rclpy_common
-  PUBLIC src/rclpy_common/include
+target_link_libraries(rclpy_common
+  ${PythonExtra_LIBRARIES}
 )
-configure_python_c_extension_library(rclpy_common)
+target_include_directories(rclpy_common
+  PUBLIC
+  src/rclpy_common/include
+  ${PythonExtra_INCLUDE_DIRS}
+)
+install(TARGETS rclpy_common
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
+)
 ament_target_dependencies(rclpy_common
   "rmw"
 )
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -72,10 +72,25 @@ function(configure_python_c_extension_library _library_name)
   )
 endfunction()
 
+add_library(rclpy_common
+  SHARED src/rclpy_common/src/common.c
+)
+target_include_directories(rclpy_common
+  PUBLIC src/rclpy_common/include
+)
+configure_python_c_extension_library(rclpy_common)
+ament_target_dependencies(rclpy_common
+  "rmw"
+)
+
 add_library(
   rclpy
   SHARED src/rclpy/_rclpy.c
 )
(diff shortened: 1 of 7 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6126/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2599/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5053/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5994/)

jacobperron:
The warnings are due to the new functions not being used in `_rclpy_action.c`. This is addressed in #270.

wjwwood:
Though if it requires #270 to avoid warnings, then maybe we should merge them into one pr.

jacobperron:
> Though if it requires #270 to avoid warnings, then maybe we should merge them into one pr.

That works for me. #270 already has this commit. Before I was aware of the warnings, I thought it was nice to make this separate. I'll leave this hear for visibility, until #270 is merged.

rclpy/src/rclpy/impl/common.h
```diff
@@ -124,4 +124,93 @@ get_capsule_pointer(PyObject * pymetaclass, const char * attr)
...
+
+/// Convert a ROS message from a C type to a Python type.
+/**
+ * Raises AttributeError if the Python type is missing a required attribute.
+ *
+ * \param[in] message The C message to convert to a Python type
+ * \param[in] pyclass An instance of the Python type to convert to.
+ * \return The Python version of the input ROS message.
+ */
+static PyObject *
```
sloretz: Since these functions are used in multiple files, how about removing `static` and creating a common library that the libraries for `_rclpy.c` and `_rclpy_action.c` link to? Maybe that will solve the unused function warning?

jacobperron: Yeah, that works: 0ab62b7

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6154/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2621/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5075/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6016/)

rclpy/CMakeLists.txt
```diff
@@ -72,10 +72,25 @@ function(configure_python_c_extension_library _library_name)
   )
 endfunction()
 
+add_library(rclpy_common
+  SHARED src/rclpy_common/src/common.c
+)
+target_include_directories(rclpy_common
+  PUBLIC src/rclpy_common/include
+)
+configure_python_c_extension_library(rclpy_common)
```
sloretz: Since `rclpy_common` is a regular library (not a cpython extension) I think the properties set by this function may not be the right thing to do. Maybe CI will pass if this call is replaced with

```cmake
target_link_libraries(rclpy_common
  ${PythonExtra_LIBRARIES}
)
target_include_directories(rclpy_common
  PUBLIC
  ${PythonExtra_INCLUDE_DIRS}
)
install(
  TARGETS rclpy_common
  ARCHIVE DESTINATION lib
  LIBRARY DESTINATION lib
  RUNTIME DESTINATION bin
)
```

jacobperron: Thanks, I'll give it a shot.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6181/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2637/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5091/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6032/)

rclpy/CMakeLists.txt
jacobperron: Updated 251175d

Works locally, let's see if CI agrees.

jacobperron:
Rebuilding for Windows after adding visibility control: [Build Status](https://ci.ros2.org/job/ci_windows/6033/)

rclpy/src/rclpy_common/include/rclpy_common/visibility_control.h
```diff
@@ -0,0 +1,48 @@
...
+
+#if defined _WIN32 || defined __CYGWIN__
+  #ifdef __GNUC__
+    #define RCLPY_COMMON_EXPORT __attribute__ ((dllexport))
+    #define RCLPY_COMMON_IMPORT __attribute__ ((dllimport))
+  #else
+    #define RCLPY_COMMON_EXPORT __declspec(dllexport)
+    #define RCLPY_COMMON_IMPORT __declspec(dllimport)
+  #endif
+  #ifdef RCLPY_COMMON_BUILDING_DLL
```
sloretz: I think this needs a `target_compile_definitions(rclpy_common PRIVATE RCLPY_COMMON_BUILDING_DLL)` to work

jacobperron: Yep, not the first time I've forgot that. Added 0eded64

Issue #268, opened 2019-02-05
Tests with MultiThreadedExecutor time out on some CI instances


jacobperron:
## Bug report

**Required Info:**

- Operating System:
  - Any
- Installation type:
  - Source
- Version or commit hash:
  - master
- DDS implementation:
  - N/A
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

Unable to reproduce locally on Linux.

Example CI job failing: https://ci.ros2.org/job/ci_linux/6113/

Disabling this test resolves the issue:

https://github.com/ros2/rclpy/blob/b6c6ca89c5e6e78a37d07d0cb17f8221ac5bda4e/rclpy/test/action/test_client.py#L257-L279

PR #270, opened 2019-02-06
Add ActionServer

### LINKED ISSUE #257: Action implementation

jacobperron:
Resolves #257 

Similar to `ActionClient`, `ActionServer` is implemented as a `Waitable` and the user interacts with goals via `ServerGoalHandle` objects. 

This is a work in progress. Still need to implement the "get result" logic, add documentation, and some more tests. I separated some changes to `_rclpy.c` into #269 to avoid adding more to this large change. Since this PR is based on that branch, #269 should be resolved first.

### DIFF
--- a/rclpy/rclpy/action/server.py
+++ b/rclpy/rclpy/action/server.py
@@ -179,7 +179,8 @@ class ActionServer(Waitable):
         result_service_qos_profile=qos_profile_services_default,
         cancel_service_qos_profile=qos_profile_services_default,
         feedback_pub_qos_profile=qos_profile_default,
-        status_pub_qos_profile=qos_profile_action_status_default
+        status_pub_qos_profile=qos_profile_action_status_default,
+        result_timeout=900
     ):
         """
         Constructor.
@@ -198,6 +199,8 @@ class ActionServer(Waitable):
         :param cancel_service_qos_profile: QoS profile for the cancel service.
         :param feedback_pub_qos_profile: QoS profile for the feedback publisher.
         :param status_pub_qos_profile: QoS profile for the status publisher.
+        :param result_timeout: Goals that have results longer than this number of seconds
+            are discarded.
         """
         if callback_group is None:
             callback_group = node.default_callback_group
@@ -222,6 +225,7 @@ class ActionServer(Waitable):
             cancel_service_qos_profile.get_c_qos_profile(),
             feedback_pub_qos_profile.get_c_qos_profile(),
             status_pub_qos_profile.get_c_qos_profile(),
+            result_timeout,
         )
 
         # key: UUID in bytes, value: GoalHandle
@@ -349,6 +353,11 @@ class ActionServer(Waitable):
         self._goal_handles[bytes(goal_uuid)]._result_future.add_done_callback(
             functools.partial(self._send_result_response, request_header))
 
+    async def _execute_expire_goals(self, expired_goals):
+        for goal in expired_goals:
+            goal_uuid = bytes(goal.goal_id.uuid)
+            del self._goal_handles[goal_uuid]
+
     def _send_result_response(self, request_header, future):
         _rclpy_action.rclpy_action_send_result_response(
             self._handle,
@@ -370,23 +379,28 @@ class ActionServer(Waitable):
         """Take stuff from lower level so the wait set doesn't immediately wake again."""
         data = {}
         if self._is_goal_request_ready:
-            take_result = _rclpy_action.rclpy_action_take_goal_request(
-                self._handle, self._action_type.GoalRequestService.Request)
-            data['goal'] = take_result
+            data['goal'] = _rclpy_action.rclpy_action_take_goal_request(
+                self._handle,
+                self._action_type.GoalRequestService.Request,
+            )
 
         if self._is_cancel_request_ready:
-            take_result = _rclpy_action.rclpy_action_take_cancel_request(
-                self._handle, self._action_type.CancelGoalService.Request)
-            data['cancel'] = take_result
+            data['cancel'] = _rclpy_action.rclpy_action_take_cancel_request(
+                self._handle,
+                self._action_type.CancelGoalService.Request,
+            )
 
         if self._is_result_request_ready:
-            take_result = _rclpy_action.rclpy_action_take_result_request(
-                self._handle, self._action_type.GoalResultService.Request)
-            data['result'] = take_result
+            data['result'] = _rclpy_action.rclpy_action_take_result_request(
+                self._handle,
+                self._action_type.GoalResultService.Request,
+            )
 
         if self._is_goal_expired:
-            expired_goals = _rclpy_action.rclpy_action_expire_goals(self._handle)
-            data['expired'] = expired_goals
+            data['expired'] = _rclpy_action.rclpy_action_expire_goals(
+                self._handle,
+                len(self._goal_handles),
+            )
 
         if not any(data):
             return None
@@ -409,8 +423,7 @@ class ActionServer(Waitable):
             await self._execute_get_result_request(taken_data['result'])
 
         if 'expired' in taken_data:
-            expired_goals = taken_data['expired']
-            # TODO(jacobperron): Do something
+            await self._execute_expire_goals(taken_data['expired'])
 
     def get_num_entities(self):
         """Return number of each type of entity used in the wait set."""
--- a/rclpy/rclpy/action/server.py
+++ b/rclpy/rclpy/action/server.py
@@ -129,7 +129,6 @@ class ServerGoalHandle:
             if not _rclpy_action.rclpy_action_goal_handle_is_active(self._handle):
                 self._action_server.notify_goal_done()
 
-
     def publish_feedback(self, feedback_msg):
         with self._lock:
             # Ignore for already destructed goal handles
--- a/rclpy/rclpy/action/server.py
+++ b/rclpy/rclpy/action/server.py
@@ -124,7 +124,12 @@ class ServerGoalHandle:
             if self._handle is None:
                 return
 
+            # Update state
             _rclpy_action.rclpy_action_update_goal_state(self._handle, event.value)
+
+            # Publish state change
+            _rclpy_action.rclpy_action_publish_status(self._action_server._handle)
+
             # If it's a terminal state, then also notify the action server
             if not _rclpy_action.rclpy_action_goal_handle_is_active(self._handle):
                 self._action_server.notify_goal_done()
@@ -134,6 +139,11 @@ class ServerGoalHandle:
             # Ignore for already destructed goal handles
             if self._handle is None:
                 return
+
+            # Ensure the feedback message contains metadata about this goal
+            feedback_msg.action_goal_id = self.goal_id
+
+            # Publish
             _rclpy_action.rclpy_action_publish_feedback(self._action_server._handle, feedback_msg)
 
     def set_succeeded(self):
--- a/rclpy/rclpy/action/server.py
+++ b/rclpy/rclpy/action/server.py
@@ -96,6 +96,13 @@ class ServerGoalHandle:
         return self._goal_info.goal_id
 
     @property
+    def is_active(self):
+        with self._lock:
+            if self._handle is None:
+                return False
+            return _rclpy_action.rclpy_action_goal_handle_is_active(self._handle)
+
+    @property
     def is_cancel_requested(self):
         with self._lock:
             return self._cancel_requested
@@ -122,18 +129,13 @@ class ServerGoalHandle:
             if not _rclpy_action.rclpy_action_goal_handle_is_active(self._handle):
                 self._action_server.notify_goal_done()
 
-    def is_active(self):
-        with self._lock:
-            if self._handle is None:
-                return False
-            return _rclpy_action.rclpy_action_goal_handle_is_active(self._handle)
 
     def publish_feedback(self, feedback_msg):
         with self._lock:
             # Ignore for already destructed goal handles
             if self._handle is None:
                 return
-            _rclpy_action.rclpy_action_publish_feedback(self._action_sever._handle, feedback_msg)
+            _rclpy_action.rclpy_action_publish_feedback(self._action_server._handle, feedback_msg)
 
     def set_succeeded(self):
         self._update_state(GoalEvent.SET_SUCCEEDED)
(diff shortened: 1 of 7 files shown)


jacobperron:
Rooting out any cross-platform issues:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6136/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2608/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5062/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6003/)

jacobperron:
Since this is a large PR, I think it can start being reviewed while I try to resolve two outstanding issues:

1. Feedback messages aren't making it to action clients (but the mock action client used in the tests appears to work).
2. Tests involving a `MultiThreadedExecutor` timeout (#268)

jacobperron:
> 1. Feedback messages aren't making it to action clients (but the mock action client used in the tests appears to work).

This is because I forgot to set the goal ID when publishing feedback messages.

jacobperron:
@sloretz Please take a look at c2026b5. It is an attempt to fix a scenario where 'done callbacks' for futures were not being called, specifically when result responses are ready for Action Servers. I'm not sure if this is the right thing to do.

sloretz:
> @sloretz Please take a look at [c2026b5](https://github.com/ros2/rclpy/commit/c2026b565f461ea65369823a98df2fb5b1d8883d). It is an attempt to fix a scenario where 'done callbacks' for futures were not being called,

@jacobperron Looking at it now. Mind moving it to a separate PR with a test?

jacobperron:
> Looking at it now. Mind moving it to a separate PR with a test?

Done #272

rclpy/rclpy/action/server.py
```diff
@@ -0,0 +1,534 @@
...
+        if not accepted:
+            self._node.get_logger().debug('New goal rejected: {0}'.format(goal_uuid.uuid))
+            return
+
+        self._node.get_logger().debug('New goal accepted: {0}'.format(goal_uuid.uuid))
+
+        goal_handle._update_state(GoalEvent.EXECUTE)
+        # Call user execute callback
+        execute_result = await await_or_execute(self._execute_callback, goal_handle)
+        # If user did not trigger a terminal state, assume success
```
sloretz: If an execute callback returns without reaching a terminal state I would assume `aborted`.

rclpy/rclpy/action/server.py
```diff
@@ -0,0 +1,534 @@
...
+            response_msg,
+        )
+
+        if not accepted:
+            self._node.get_logger().debug('New goal rejected: {0}'.format(goal_uuid.uuid))
+            return
+
+        self._node.get_logger().debug('New goal accepted: {0}'.format(goal_uuid.uuid))
+
+        goal_handle._update_state(GoalEvent.EXECUTE)
```
sloretz: This looks like it is setting the goal handle to `EXECUTE` even if the user chose `ACCEPT_AND_DEFER`. In `rclcpp` instead of an `execute_callback` there is `handle_accepted`. If the user says `ACCEPT_AND_DEFER` then it arrives at that callback in state `ACCEPTED`.

If an `execute_callback` is provided then it should be separate from `handle_accepted`. Maybe add the `execute_callback` as a task to the executor if `handle_accepted` returns and the goal handle is not at a terminal state?

jacobperron: I... deferred implementing the defer logic. I meant to make a note of it.

Originally I was thinking that we could provide an API via the `GoalHandle` to execute a deferred goal. E.g. `def execute(callback=None)` where the callback will optionally override a previous established callback. This function will schedule the callback (if there is one) with the executor.

I like the idea of the `execute_callback` as a way to let the action server handle scheduling execution with the executor. I think this puts less burden on the user; not having to manage threads. Do you think this is something `rclcpp_action` is missing?

Reflecting on the `rclcpp` action server, I think it makes sense to take a `handle_accepted` callback in `rclpy`'s aciton server as well. The `handle_accepted` makes a nice venue for providing the user with a reference to the goal handle.

sloretz: > I like the idea of the `execute_callback` as a way to let the action server handle scheduling execution with the executor. I think this puts less burden on the user; not having to manage threads. Do you think this is something `rclcpp_action` is missing?

I agree an `execute_callback` is convenient. I've assumed in `rclcpp_action` an execute callback would be provided by a class like ROS 1's `SimpleActionServer`. It should only be invoked after a goal transitions to `EXECUTING` though. Right now I don't see a way to make a python action server queue goals instead of executing them all in parallel.

Looking back, I'm not sure `ACCEPT_AND_EXECUTE` and `ACCEPT_AND_DEFER` are necessary. The same thing could be accomplished by whether or not `handle_goal` changed the state to `EXECUTING` .

rclpy/test/action/test_server.py
```diff
@@ -0,0 +1,491 @@
...
+
+    def tearDown(self):
+        self.node.destroy_node()
+        self.executor.shutdown()
+        rclpy.shutdown(context=self.context)
+
+    def timed_spin(self, duration):
+        start_time = time.time()
+        current_time = start_time
+        while (current_time - start_time) < duration:
```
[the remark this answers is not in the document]
jacobperron: Updated 876b60f

rclpy/rclpy/action/server.py
jacobperron: > Looking back, I'm not sure `ACCEPT_AND_EXECUTE` and `ACCEPT_AND_DEFER` are necessary.

Agreed.

I've made an update in 9a15695 that adds a `handle_accepted_callback` in which the user can optionally call `goal_handle.execute()`, or hold onto the goal handle to execute later (ie. defer).

jacobperron:
Rebased on #275, which is required for creating tasks associated with user callbacks with the executor. 
The `Waitable.add_future` mechanism is not as reliable since it needs something in the wait set to be ready for `rcl_wait` to wake up.

I'm considering replacing the `add_future` API introduced in #262 with direct calls to `Executor.create_task()`.

rclpy/src/rclpy/_rclpy_action.c
```diff
@@ -850,12 +1248,451 @@ rclpy_action_take_status(PyObject * Py_UNUSED(self), PyObject * args)
...
+  if (!pygoal_info_type) {
+    free(expired_goals);
+    return NULL;
+  }
+
+  // Create a tuple of GoalInfo instances to return
+  PyObject * result_tuple = PyTuple_New(num_expired);
+  if (!result_tuple) {
+    free(expired_goals);
+    return NULL;
```
sloretz: `Py_DECREF(pygoal_info_type)` before returning

jacobperron: Nice catch. Fixed in b037186

rclpy/rclpy/action/server.py
```diff
@@ -0,0 +1,595 @@
...
+                )
+
+        if self._is_goal_expired:
+            with self._lock:
+                data['expired'] = _rclpy_action.rclpy_action_expire_goals(
+                    self._handle,
+                    len(self._goal_handles),
+                )
+
+        if not any(data):
```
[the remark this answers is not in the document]
jacobperron: Done b037186

rclpy/rclpy/action/server.py
```diff
@@ -0,0 +1,595 @@
...
+                self._node.get_logger().error(
+                    'Failed to accept new goal with ID {0}: {1}'.format(goal_uuid.uuid, e))
+                accepted = False
+            else:
+                self._goal_handles[bytes(goal_uuid.uuid)] = goal_handle
+
+        # Send response
+        response_msg = self._action_type.GoalRequestService.Response()
+        response_msg.accepted = accepted
+        response_msg.stamp = self._node.get_clock().now().to_msg()
```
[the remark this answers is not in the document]
jacobperron: Done b037186

rclpy/rclpy/action/server.py
```diff
@@ -0,0 +1,595 @@
...
+        self._node.add_waitable(self)
+        # import rclpy
+        # self._node.get_logger().set_level(rclpy.logging.LoggingSeverity.DEBUG)
+
+    async def _execute_goal_request(self, request_header_and_message):
+        request_header, goal_request = request_header_and_message
+        goal_uuid = goal_request.action_goal_id
+        goal_info = GoalInfo()
+        goal_info.goal_id = goal_uuid
+        goal_info.stamp = self._node.get_clock().now().to_msg()
```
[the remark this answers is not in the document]
jacobperron: Done b037186

rclpy/rclpy/action/server.py
```diff
@@ -0,0 +1,595 @@
...
+        :param goal_callback: Callback function for handling new goal requests.
+        :param handle_accepted_callback: Callback function for handling newly accepted goals.
+            Passes an instance of `ServerGoalHandle` as an argument.
+        :param cancel_callback: Callback function for handling cancel requests.
+        :param goal_service_qos_profile: QoS profile for the goal service.
+        :param result_service_qos_profile: QoS profile for the result service.
+        :param cancel_service_qos_profile: QoS profile for the cancel service.
+        :param feedback_pub_qos_profile: QoS profile for the feedback publisher.
+        :param status_pub_qos_profile: QoS profile for the status publisher.
+        :param result_timeout: Goals that have results longer than this number of seconds
```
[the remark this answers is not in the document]
jacobperron: Done (thanks!):  b037186

rclpy/rclpy/action/server.py
```diff
@@ -0,0 +1,595 @@
...
+        self._update_state(GoalEvent.SET_SUCCEEDED)
+
+    def set_aborted(self):
+        self._update_state(GoalEvent.SET_ABORTED)
+
+    def set_canceled(self):
+        self._update_state(GoalEvent.SET_CANCELED)
+
+    def destroy(self):
+        if self._handle is None:
```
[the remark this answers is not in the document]
jacobperron: I don't think it hurts, and is probably the safer thing to do. b037186

rclpy/rclpy/action/server.py
```diff
@@ -0,0 +1,534 @@
...
+            if goal_uuid not in self._goal_handles:
+                # Possibly the user doesn't care to track the goal handle
+                # Remove from response
+                cancel_response.goals_canceling.remove(goal_info)
+                continue
+            goal_handle = self._goal_handles[goal_uuid]
+            response = await await_or_execute(self._cancel_callback, goal_handle)
+
+            accepted = CancelResponse.ACCEPT == response
+            if accepted:
```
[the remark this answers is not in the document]
jacobperron: Done b037186

jacobperron:
@sloretz Thank you for the reviews :bowing_man: 

In addition to addressing your latest comments, I've resolved an issue for the scenario where a goal is deferred, canceled, and then executed in 093d52d.

I've also pushed an example of a goal being deferred in ros2/examples#222.

rclpy/rclpy/action/server.py
jacobperron: > Right now I don't see a way to make a python action server queue goals instead of executing them all in parallel

With the updated examples, I believe the equivalent of a `SimpleActionServer` could be implemented in Python by adapting the ["single goal" example](https://github.com/ros2/examples/blob/0ea84a050d5025fde8ab65000dc3579937ee2cf7/rclpy/actions/minimal_action_server/examples_rclpy_minimal_action_server/server_single_goal.py). Instead of aborting newly received goals, it could queue them instead.

rclpy/rclpy/action/server.py
```diff
@@ -0,0 +1,597 @@
...
+
+            # If it's a terminal state, then also notify the action server
+            if not _rclpy_action.rclpy_action_goal_handle_is_active(self._handle):
+                self._action_server.notify_goal_done()
+
+    def execute(self, execute_callback=None):
+        # It's possible that there has been a request to cancel the goal prior to executing.
+        # In this case we want to avoid the illegal state transition to EXECUTING
+        # but still call the users execute callback to let them handle canceling the goal.
+        if not self.is_cancel_requested:
```
sloretz: Does `is_cancel_requested` mean the cancellation was accepted? I think the transition is allowed then. If this only means it was requested then what happens if: a goal to be requested to be canceled, the execute callback is called but the state transition to `EXECUTE` is skipped, then the cancellation request rejected, and finally the execute callback tries to set a terminal state? It looks like it might try to transition from `ACCEPTED -> SUCCEEDED/ABORTED` which isn't an allowed transition.

jacobperron: > Does `is_cancel_requested` mean the cancellation was accepted?

Yes. The cancellation request was accepted and the goal is not in the `CANCELING` state.

https://github.com/ros2/rclpy/blob/b03718678f5a60948175f7e7da8b9a6e7a6e95c3/rclpy/rclpy/action/server.py#L102-L103

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6204/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2655/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5116/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6056/)

jacobperron:
Rebased on master.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6213/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2663/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5125/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6079/)

jacobperron:
Hopefully, fixed the Windows build :crossed_fingers: 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6244/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2681/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5141/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6089/)

jacobperron:
@sloretz Test failures are unrelated.
This is ready for another round of review :)

PR #271, opened 2019-02-09
Replace PyUnicode_1BYTE_DATA() with PyUnicode_AsUTF8()

### LINKED ISSUE #266: Replace PyUnicode_1BYTE_DATA with PyUnicode_AsUTF8()
There are a few places where string data is read using a macro `PyUnicode_1BYTE_DATA()`. However, this method is not always the right call. First it requires `PyUnicode_READY()` to be called on the string first in case it was created with deprecated APIs. Second, cpython may not have chosen to store the string as UTF8, in which case the APIs `PyUnicode_<2|4>BYTE_DATA()` should have been called instead.

[PyUnicode_1BYTE_DATA documentation](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_1BYTE_DATA)

Since the code using this wants a `char *` string, the API that should be used instead is [`PyUnicode_AsUTF8`](https://docs.python.org/3.3/c-api/unicode.html#PyUnicode_AsUTF8)

One place that needs to be updated (there are more)
https://github.com/ros2/rclpy/blob/0fafaec6b3764654413d0ce37839d7c7178245af/rclpy/src/rclpy/_rclpy.c#L812-L816

jacobperron:
Resolves #266 

Ensure return value is `const char *` for compatibility with Python v3.7+.
Also, renamed PyObject variables for consistency with the rest of the file.

Update expected error message in unit test.

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -789,17 +789,16 @@ rclpy_count_subscribers(PyObject * Py_UNUSED(self), PyObject * args)
 static PyObject *
 rclpy_get_validation_error_for_topic_name(PyObject * Py_UNUSED(self), PyObject * args)
 {
-  PyObject * topic_name_py;
+  PyObject * pytopic_name;
 
-  if (!PyArg_ParseTuple(args, "O", &topic_name_py)) {
+  if (!PyArg_ParseTuple(args, "O", &pytopic_name)) {
     return NULL;
   }
 
-  if (!PyUnicode_Check(topic_name_py)) {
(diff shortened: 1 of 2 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6146/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2617/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5071/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6012/)

PR #272, opened 2019-02-11
Fix Executor not executing tasks if there are no ready entities in the wait set

jacobperron:
If a task is created, then trigger the Executors guard condition. This will wake any blocking call to `rcl_wait()`.
In this scenario, no work is yielded, so we also have to move the check for 'in-progress' tasks into the wait loop.

Added a unit test to reproduce the issue.

This resolves an issue in some cases where the result response from the action server is never processed, therefore never reaching the action client.

### DIFF
--- a/rclpy/test/test_executor.py
+++ b/rclpy/test/test_executor.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import asyncio
+import threading
 import time
 import unittest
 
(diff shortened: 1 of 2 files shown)


rclpy/test/test_executor.py
```diff
@@ -221,6 +222,40 @@ def test_create_task_dependent_coroutines(self):
...
+    def test_create_task_during_spin(self):
+        self.assertIsNotNone(self.node.handle)
+        executor = SingleThreadedExecutor(context=self.context)
+        executor.add_node(self.node)
+
+        future = None
+
+        def spin_until_task_done(executor):
+            nonlocal future
+            while future is None or future.done():
```
sloretz: `... or not future.done()`?

rclpy/test/test_executor.py
```diff
@@ -221,6 +222,40 @@ def test_create_task_dependent_coroutines(self):
...
+
+        # Start spinning in a separate thread
+        thr = threading.Thread(target=spin_until_task_done, args=(executor, ), daemon=True)
+        thr.start()
+
+        def func():
+            return 'Sentinel Result'
+
+        # Create a task
+        future = executor.create_task(func)
```
sloretz: I think there might be a race where this test could have a false negative if `spin_until_task_done` has not reached the `while` loop in `_wait_for_ready_callbacks` prior to this task being created.

rclpy/test/test_executor.py
jacobperron: oops, yeah.

rclpy/test/test_executor.py
jacobperron: Yeah, I think you're right. I'm not sure how to avoid it, but it can be mitigated by sleeping for a short time before creating the task.

rclpy/test/test_executor.py
jacobperron: Fixed  1ee031a

rclpy/test/test_executor.py
jacobperron: See 1ee031a

sloretz:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6149/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2619/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5073/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6014/)

PR #275, opened 2019-02-13
Fix Node's reference to executor

jacobperron:
Previously, if a `Node` was added to an `Executor` using the executors `add_node` method, then the nodes `executor` property would return `None`.

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -112,11 +112,14 @@ class Node:
     def executor(self, new_executor):
         """Set or change the executor the node belongs to."""
         current_executor = self.executor
+        if current_executor == new_executor:
+            return
         if current_executor is not None:
             current_executor.remove_node(self)
         if new_executor is None:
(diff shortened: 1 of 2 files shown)


jacobperron:
> Looks reasonable and changes look good, but don't have a whole lot of context for the change.

As part of the `ActionServer`, I want to schedule tasks for user callbacks using the executor. Something like:

```python
class ActionServer:
    ...
    self.node.executor.create_task(user_execute_callback)
```

But, without this change `self.node.executor` is `None` (which I presume is a bug).

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6175/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2636/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5090/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6031/)

PR #277, opened 2019-02-26
Improve documentation

jacobperron:
Adding Sphinx documentation that makes use of the [autodoc](http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) and [sphinx-autodoc-typehints](https://pypi.org/project/sphinx-autodoc-typehints/) extensions for auto-generating documentation from docstrings and annotations.


jacobperron:
So far I've only documented the Node module. I'm looking for early feedback on the proposed method for documentation before I continue documenting more `rclpy` modules.

sloretz:
Does documentation generated this way show up at https://index.ros.org/p/rclpy/ ?

jacobperron:
> Does documentation generated this way show up at https://index.ros.org/p/rclpy/ ?

I would like it to be, but I'm not sure what is required to make that happen. I didn't find any examples of ROS 2 Python packages that had their documentation available on ROS Index, but I can look into it.

Karsten1987:
I might be wrong here, but I believe this has to be backported to the crystal branch and re-released.

jacobperron:
After some investigation, I don't think code API documentation generated this way will show up on ROS index. It doesn't appear that we have the infrastructure up and running to facilitate this. I found a [Makefile](https://github.com/ros2/docs.ros2.org/tree/doc_gen) with some instructions, indicating the current process for generating API docs is done manually and only for a small set of packages (excluding rclpy).

@nuclearsandwich Does this sound right?

I'll proceed with adding more Sphinx documentation under the assumption that we'll eventually integrate a documentation tool like [rosdoc_lite](http://wiki.ros.org/rosdoc_lite) (from ROS 1).

nuclearsandwich:
> @nuclearsandwich Does this sound right?

Yeah. It doesn't take much but a ROS 2 workspace and push access to update the documentation and it's probably do for a refresh from Crystal patch releases (Off-topic, I also recall you mentioning the desire to add rcl{,cpp}_action docs). If you've got a way to build documentation locally, there's no reason we can't expand the makefile to include rclpy documentation.

jacobperron:
This is ready for review.

I've only documented what I considered the major components of rclpy, and plan to open a follow-up PR adding more docstrings and annotations. I figured there's enough here to review for now.
In particular, I've held off on updating the docs for Actions until after the IDL work has landed #274.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -2979,14 +2979,15 @@ rclpy_get_service_names_and_types_by_node(PyObject * Py_UNUSED(self), PyObject *
   return pyservice_names_and_types;
 }
 
-/// Get the list of published topics discovered by the provided node for the remote node name
+/// Get the list of subscription topics discovered by the provided node for the remote node name
```
wjwwood: I would say `Get the list of subscribed topics`, though the whole brief is confusing. It really is doing something like:

> Get the list of topic names and types for which there is at least one subscription associated with the given node name.

But updating this one would mean updating the publisher version as well and may be out of scope for this pr.

rclpy/rclpy/node.py
```diff
@@ -39,26 +52,61 @@
...
-        self.guards = []
-        self.waitables = []
+        self._parameters = {}  # type: dict
+        self.publishers = []  # type: List[Publisher]
+        self.subscriptions = []  # type: List[Subscription]
+        self.clients = []  # type: List[Client]
+        self.services = []  # type: List[Service]
+        self.timers = []  # type: List[WallTimer]
+        self.guards = []  # type: List[GuardCondition]
+        self.waitables = []  # type: List[Waitable]
```
wjwwood: Since we're no longer supporting Python <3.6, you can (and imo should) use the non-comment based syntax, e.g. `self.waitables: List[Waitable] = []`.

rclpy/src/rclpy/_rclpy.c
jacobperron: Clarified briefs 935c054

rclpy/rclpy/node.py
jacobperron: Updated: 0cabd58

jacobperron:
Confirming I didn't break any linters:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6387/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2779/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5255/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6227/)

PR #278, opened 2019-03-02
Improve error handling to avoid memory leaks in C extension

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6305/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2721/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5182/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6143/)

Edit: test failures are unrelated.

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -1435,17 +1435,23 @@ rclpy_create_timer(PyObject * Py_UNUSED(self), PyObject * args)
 
   PyObject * pylist = PyList_New(2);
   if (!pylist) {
+    ret = rcl_timer_fini(timer);
+    (void)ret;
     PyMem_Free(timer);
     return NULL;
   }
   PyObject * pytimer = PyCapsule_New(timer, "rcl_timer_t", NULL);
   if (!pytimer) {
+    ret = rcl_timer_fini(timer);
+    (void)ret;
     PyMem_Free(timer);
     Py_DECREF(pylist);
     return NULL;
   }
   PyObject * pytimer_impl_reference = PyLong_FromUnsignedLongLong((uint64_t)&timer->impl);
   if (!pytimer_impl_reference) {
+    ret = rcl_timer_fini(timer);
+    (void)ret;
     PyMem_Free(timer);
     Py_DECREF(pylist);
     Py_DECREF(pytimer);
@@ -1857,11 +1863,15 @@ rclpy_create_subscription(PyObject * Py_UNUSED(self), PyObject * args)
 
   PyObject * pylist = PyList_New(2);
   if (!pylist) {
+    ret = rcl_subscription_fini(subscription, node);
+    (void)ret;
     PyMem_Free(subscription);
     return NULL;
   }
   PyObject * pysubscription = PyCapsule_New(subscription, "rcl_subscription_t", NULL);
   if (!pysubscription) {
+    ret = rcl_subscription_fini(subscription, node);
+    (void)ret;
     PyMem_Free(subscription);
     Py_DECREF(pylist);
     return NULL;
@@ -1869,6 +1879,8 @@ rclpy_create_subscription(PyObject * Py_UNUSED(self), PyObject * args)
   PyObject * pysubscription_impl_reference =
     PyLong_FromUnsignedLongLong((uint64_t)&subscription->impl);
   if (!pysubscription_impl_reference) {
+    ret = rcl_subscription_fini(subscription, node);
+    (void)ret;
     PyMem_Free(subscription);
     Py_DECREF(pylist);
     Py_DECREF(pysubscription);
@@ -1979,17 +1991,23 @@ rclpy_create_client(PyObject * Py_UNUSED(self), PyObject * args)
   }
   PyObject * pylist = PyList_New(2);
   if (!pylist) {
+    ret = rcl_client_fini(client, node);
+    (void)ret;
     PyMem_Free(client);
     return NULL;
   }
   PyObject * pyclient = PyCapsule_New(client, "rcl_client_t", NULL);
   if (!pyclient) {
+    ret = rcl_client_fini(client, node);
+    (void)ret;
     PyMem_Free(client);
     Py_DECREF(pylist);
     return NULL;
   }
   PyObject * pyclient_impl_reference = PyLong_FromUnsignedLongLong((uint64_t)&client->impl);
   if (!pyclient_impl_reference) {
+    ret = rcl_client_fini(client, node);
+    (void)ret;
     PyMem_Free(client);
     Py_DECREF(pylist);
     Py_DECREF(pyclient);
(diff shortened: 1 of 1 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -2834,31 +3141,50 @@ rclpy_get_node_names_and_namespaces(PyObject * Py_UNUSED(self), PyObject * args)
...
+    PyTuple_SET_ITEM(pytuple, 1, pynode_namespace);
+    // Steals the reference
+    PyList_SET_ITEM(pynode_names_and_namespaces, idx, pytuple);
+  }
+
+  rcutils_ret_t fini_names_ret;
+  rcutils_ret_t fini_namespaces_ret;
+cleanup:
+  fini_names_ret = rcutils_string_array_fini(&node_names);
+  fini_namespaces_ret = rcutils_string_array_fini(&node_namespaces);
```
hidmic: @jacobperron nit: can't we declare and initialize on the same statement?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3193,28 +3540,48 @@ rclpy_get_service_names_and_types(PyObject * Py_UNUSED(self), PyObject * args)
...
+    PyObject * pyservice_name = PyUnicode_FromString(service_names_and_types.names.data[i]);
+    if (!pyservice_name) {
+      Py_DECREF(pytuple);
+      goto cleanup;
+    }
+    PyTuple_SetItem(pytuple, 0, pyservice_name);
+    PyObject * pytypes_list = PyList_New(service_names_and_types.types[i].size);
+    if (!pytypes_list) {
+      Py_DECREF(pytuple);
+      Py_DECREF(pyservice_name);
```
hidmic: @jacobperron I _think_ `PyTuple_SetItem` on [line 3557](https://github.com/ros2/rclpy/pull/278/files#diff-80dda03110b5be606b823d6b0558b5c2R3557) steals `pyservice_name` reference.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3193,28 +3540,48 @@ rclpy_get_service_names_and_types(PyObject * Py_UNUSED(self), PyObject * args)
...
-    PyObject * types_list = PyList_New(service_names_and_types.types[i].size);
+    if (!pytuple) {
+      goto cleanup;
+    }
+    PyObject * pyservice_name = PyUnicode_FromString(service_names_and_types.names.data[i]);
+    if (!pyservice_name) {
+      Py_DECREF(pytuple);
+      goto cleanup;
+    }
+    PyTuple_SetItem(pytuple, 0, pyservice_name);
```
hidmic: @jacobperron why using this variant instead of `PyTuple_SET_ITEM`?

rclpy/src/rclpy/_rclpy.c
jacobperron: Apparently, this is not supported and results in compiler errors. This [SO post](https://stackoverflow.com/a/8384505) suggest we could add an empty statement as a workaround. I don't feel strongly either way, but I'm leaning towards leaving it as-is because `cleanup:;` looks odd to me.

wjwwood: I'm actually surprised you don't have warnings above where you declare local variables _after_ the use of a `goto`. In the `rmw` implementations we declare all local variables just before the first use of `goto`, which is essentially what the compiler does even when you declare them throughout and I gather the use of goto might cause strange order of side-effects if you don't declare everything ahead of their use.

For example:

https://github.com/ros2/rmw_connext/blob/43d6833977ddfc3d8fa599a51c7b01bcb3008fa3/rmw_connext_shared_cpp/src/node.cpp#L119-L140

rclpy/src/rclpy/_rclpy.c
jacobperron: Good catch. Fixed in 914ff23

rclpy/src/rclpy/_rclpy.c
jacobperron: This was what was being used before the change, but see no reason why we shouldn't use the `SET_ITEM` variant here (914ff23).

rclpy/src/rclpy/_rclpy.c
jacobperron: Yeah, in retrospect this is odd.
I am able to produce an error if I try to both declare and initialize *after* a goto statement. But the compiler is fine if it's only a declaration. I've moved the declarations to the top so code reads better 67d3660

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1881,9 +2140,27 @@ rclpy_create_service(PyObject * Py_UNUSED(self), PyObject * args)
     rcl_reset_error();
     return NULL;
   }
+
   PyObject * pylist = PyList_New(2);
-  PyList_SET_ITEM(pylist, 0, PyCapsule_New(service, "rcl_service_t", NULL));
-  PyList_SET_ITEM(pylist, 1, PyLong_FromUnsignedLongLong((uint64_t)&service->impl));
+  if (!pylist) {
+    PyMem_Free(service);
```
sloretz: Should fini `service` too

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1756,8 +1978,25 @@ rclpy_create_client(PyObject * Py_UNUSED(self), PyObject * args)
     return NULL;
   }
   PyObject * pylist = PyList_New(2);
-  PyList_SET_ITEM(pylist, 0, PyCapsule_New(client, "rcl_client_t", NULL));
-  PyList_SET_ITEM(pylist, 1, PyLong_FromUnsignedLongLong((uint64_t)&client->impl));
+  if (!pylist) {
+    PyMem_Free(client);
```
sloretz: should `fini` client too

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1667,9 +1854,28 @@ rclpy_create_subscription(PyObject * Py_UNUSED(self), PyObject * args)
     PyMem_Free(subscription);
     return NULL;
   }
+
   PyObject * pylist = PyList_New(2);
-  PyList_SET_ITEM(pylist, 0, PyCapsule_New(subscription, "rcl_subscription_t", NULL));
-  PyList_SET_ITEM(pylist, 1, PyLong_FromUnsignedLongLong((uint64_t)&subscription->impl));
+  if (!pylist) {
+    PyMem_Free(subscription);
```
sloretz: should `fini` subscription too

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1283,9 +1432,27 @@ rclpy_create_timer(PyObject * Py_UNUSED(self), PyObject * args)
     PyMem_Free(timer);
     return NULL;
   }
+
   PyObject * pylist = PyList_New(2);
-  PyList_SET_ITEM(pylist, 0, PyCapsule_New(timer, "rcl_timer_t", NULL));
-  PyList_SET_ITEM(pylist, 1, PyLong_FromUnsignedLongLong((uint64_t)&timer->impl));
+  if (!pylist) {
+    PyMem_Free(timer);
```
sloretz: should `fini` timer too

rclpy/src/rclpy/_rclpy.c
jacobperron: Nice catch 906af1b

rclpy/src/rclpy/_rclpy.c
jacobperron: 906af1b

rclpy/src/rclpy/_rclpy.c
jacobperron: 906af1b

rclpy/src/rclpy/_rclpy.c
jacobperron: 906af1b

jacobperron:
Thanks for the reviews!

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6512/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2859/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5334/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6308/)

(I expect the `test_params_yaml` to fail ros2/build_cop#166)

PR #281, opened 2019-03-07
Guard against failed take when taking action messages

jacobperron:
Some middlewares (e.g. Connext and OpenSplice) appear to send empty messages after a successful service call which results in a 'failed take'.
It is unclear if this is correct behavior from the rmw/rcl layers, but it doesn't hurt to guard against this scenario.

Connects to ros2/system_tests#332.

### DIFF
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -231,27 +231,40 @@ class ActionClient(Waitable):
         """Take stuff from lower level so the wait set doesn't immediately wake again."""
         data = {}
         if self._is_goal_response_ready:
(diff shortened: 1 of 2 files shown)


wjwwood:
For reference, rclcpp also allows take to fail, but only if the reason is "RET_TAKE_FAILED" (or similar):

https://github.com/ros2/rclcpp/blob/d2d9ad8796cb0528351482b1f55ba36fffaa9728/rclcpp/src/rclcpp/executor.cpp#L324-L332

This can happen if the sample is invalid (DDS's way of indicating the instance has been disposed), but we don't know that until we try to take (i.e. we don't know that is the reason rather than data being available after being woken up from the waitset).

PR #294, opened 2019-03-26
Update tests to include namespace in ROS types

jacobperron:
Connects to ros2/ros2#677

### DIFF
--- a/rclpy/test/test_action_graph.py
+++ b/rclpy/test/test_action_graph.py
@@ -101,7 +101,7 @@ class TestActionGraph(unittest.TestCase):
         assert name1 == TEST_NAMESPACE1 + '/' + TEST_ACTION0
         assert isinstance(types1, list)
         assert len(types1) == 1
-        assert types1[0] == 'test_msgs/Fibonacci'
+        assert types1[0] == 'test_msgs/action/Fibonacci'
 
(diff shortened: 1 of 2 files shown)


sloretz:
The code looks fine.

Is `action`, `msg`, and `srv` indicating the type of interface, or the namespace the namespace in the idl file? If a struct in an `idl` file has arbitrary namespace `test_msgs::foo::bar::Fibonacci`, what will the type name be here? `test_msgs/foo/bar/Fibonacci`?

jacobperron:
> Is action, msg, and srv indicating the type of interface, or the namespace the namespace in the idl file? If a struct in an idl file has arbitrary namespace test_msgs::foo::bar::Fibonacci, what will the type name be here? test_msgs/foo/bar/Fibonacci?

Indicating the namespace of the IDL type.
So, in theory, a type defined with namespace `test_msgs::foo::bar::Fibonacci` would have the identifying name `test_msgs/foo/bar/Fibonacci`. I'm not sure that generating code from arbitrary IDL files will work right now, but I believe that this the long term goal. By convention, namespaces of legacy ROS interfaces (`msg`, `srv`, and `action`) will coincide with the type, ie `<pkg_name>/<msg|srv|action>/<type_name>`.

Issue #240, opened 2018-09-20
Respect the use_sim_time parameter for rclpy nodes


dhood:
There is support for Clocks that use ROS Time as described in [this design document](http://design.ros2.org/articles/clock_and_time.html). It was added in [this PR](https://github.com/ros2/rclpy/pull/210).

Clocks can be set to have ROS time "active" (will use the value seen on the /clock topic as its time) by calling `time_source.ros_time_is_active = True`.

A node has a time source associated with it (added in https://github.com/ros2/rclpy/pull/210). At the time of that PR, parameters were not available in rclpy, so the node's time source does not respond to the `use_sim_time` parameter:
https://github.com/ros2/rclpy/blob/62012d3650de790d89c725232451c4e998b396fb/rclpy/rclpy/node.py#L97

As of https://github.com/ros2/rclpy/pull/214, parameters are available in rclpy.

Tasks to get rclpy nodes to respond to the `use_sim_time` parameter:

- [ ] Add a [parameter callback](https://github.com/ros2/rclpy/pull/228) that detects if `use_sim_time` has been set on the node.
- [ ] Add tests that the node's clock output matches what's expected based on the combination of `use_sim_time` parameter and the presence of a `/clock` publisher. [These are the tests in rclcpp](https://github.com/ros2/rclcpp/pull/559/files#diff-8b3e7c8b544d7848db66291e08959f94R138).

shivangg:
@dhood I'd like to work on this issue. Can you assign it to me?

sloretz:
@shivangg Thanks!

I can't assign the issue to you (isaacs/github#100), but your contribution is still very welcome. When you make a pull request add the text `connects to ros2/rclpy#240` to the PR description. That will link the PR and issue on the [waffle board](https://waffle.io/ros2/ros2).

mkhansen-intel:
@shivangg - I'm in need of this functionality now for testing ROS2 Navigation, did you get this to work?

jacobperron:
Resolved by #297

PR #297, opened 2019-03-29
Add param callback to time_source

vinnamkim:
Targets to #240

Since it seems there have been no response on #240 for a long time, I made this PR. However, It seems that the callback function should be changed from [parameter callback](https://github.com/ros2/rclpy/pull/228) to something like rclcpp::SyncParametersClient or rclcpp::AsyncParametersClient.

Signed-off-by: vinnamkim <vinnam.kim@gmail.com>

### DIFF
--- a/rclpy/rclpy/time_source.py
+++ b/rclpy/rclpy/time_source.py
@@ -34,21 +34,6 @@ class TimeSource:
         if node is not None:
             self.attach_node(node)
 
-        use_sim_time_param = node.get_parameter('use_sim_time')
-        if use_sim_time_param.type_ != Parameter.Type.NOT_SET:
-            if use_sim_time_param.type_ == Parameter.Type.BOOL:
-                self._ros_time_is_active = use_sim_time_param.value
-            else:
-                node.get_logger().error(
-                    'Invalid type for parameter \'use_sim_time\' ' +
-                    str(use_sim_time_param.type_) + ' should be bool')
-        else:
-            node.get_logger().debug(
-                '\'use_sim_time\' parameter not set, \
-                    using wall time by default')
-
-        node.set_parameters_callback(self.on_parameter_event)
-
     @property
     def ros_time_is_active(self):
         return self._ros_time_is_active
@@ -65,6 +50,7 @@ class TimeSource:
         else:
             if self._clock_sub is not None and self._node is not None:
                 self._node.destroy_subscription(self._clock_sub)
+                self._clock_sub = None
 
     def _subscribe_to_clock_topic(self):
         if self._clock_sub is None and self._node is not None:
@@ -82,8 +68,20 @@ class TimeSource:
         if self._node is not None:
             self.detach_node()
         self._node = node
-        if self.ros_time_is_active:
-            self._subscribe_to_clock_topic()
+
+        use_sim_time_param = node.get_parameter('use_sim_time')
+        if use_sim_time_param.type_ != Parameter.Type.NOT_SET:
+            if use_sim_time_param.type_ == Parameter.Type.BOOL:
+                self.ros_time_is_active = use_sim_time_param.value
+            else:
+                node.get_logger().error(
+                    "Invalid type for parameter 'use_sim_time' {!r} should be bool"
+                    .format(use_sim_time_param.type_))
+        else:
+            node.get_logger().debug(
+                "'use_sim_time' parameter not set, using wall time by default")
+
+        node.set_parameters_callback(self._on_parameter_event)
 
     def detach_node(self):
         # Remove the subscription to the clock topic.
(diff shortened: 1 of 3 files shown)


rclpy/test/test_time_source.py
```diff
@@ -67,6 +68,19 @@ def publish_reversed_clock_messages(self):
             rclpy.spin_once(self.node, timeout_sec=1, executor=executor)
             time.sleep(1)
 
+    def set_use_sim_time_parameter(self, value):
+        self.node.set_parameters(
+            [Parameter('use_sim_time', Parameter.Type.BOOL, value)])
+        executor = rclpy.executors.SingleThreadedExecutor(context=self.context)
+
+        while rclpy.ok(context=self.context):
```
jacobperron: In case of an error, it might be good to add a timeout here (e.g. 5 seconds) so that we don't wait indefinitely.

rclpy/test/test_time_source.py
```diff
@@ -67,6 +68,19 @@ def publish_reversed_clock_messages(self):
...
+            use_sim_time_param = self.node.get_parameter('use_sim_time')
+
+            if use_sim_time_param._type_ == Parameter.Type.BOOL:
```
jacobperron: `use_sim_time_param.type_`

rclpy/test/test_time_source.py
```diff
@@ -67,6 +68,19 @@ def publish_reversed_clock_messages(self):
...
+                break
```
jacobperron: Suggestion: return a bool indicating if this method call was successful, then we can assert there wasn't an error in the tests. E.g. instead of breaking,

```python
return use_sim_time_param.value == value
```

Then when calling in the tests:

```python
assert set_use_sim_time_parameter(False)
```

rclpy/rclpy/time_source.py
```diff
@@ -88,3 +108,14 @@ def clock_callback(self, msg):
         self._last_time_set = time_from_msg
         for clock in self._associated_clocks:
             clock.set_ros_time_override(time_from_msg)
+
+    def on_parameter_event(self, parameter_list):
```
jacobperron: This method doesn't need to be public and can be hidden by adding a `_` prefix.

rclpy/rclpy/time_source.py
```diff
@@ -88,3 +108,14 @@ def clock_callback(self, msg):
...
+        for parameter in parameter_list:
+            if parameter.name == 'use_sim_time':
```
jacobperron: nit: in this case we can break/return early.

rclpy/rclpy/time_source.py
```diff
@@ -32,6 +34,21 @@ def __init__(self, *, node=None):
         if node is not None:
             self.attach_node(node)
 
+        use_sim_time_param = node.get_parameter('use_sim_time')
```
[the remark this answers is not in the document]
vinnamkim: Fixed

rclpy/rclpy/time_source.py
```diff
@@ -32,6 +34,21 @@ def __init__(self, *, node=None):
...
+        if use_sim_time_param.type_ != Parameter.Type.NOT_SET:
+            if use_sim_time_param.type_ == Parameter.Type.BOOL:
+                self._ros_time_is_active = use_sim_time_param.value
```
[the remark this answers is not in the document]
vinnamkim: Fixed

rclpy/rclpy/time_source.py
```diff
@@ -32,6 +34,21 @@ def __init__(self, *, node=None):
...
+            else:
+                node.get_logger().error(
+                    'Invalid type for parameter \'use_sim_time\' ' +
+                    str(use_sim_time_param.type_) + ' should be bool')
```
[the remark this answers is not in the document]
vinnamkim: Fixed

rclpy/rclpy/time_source.py
```diff
@@ -32,6 +34,21 @@ def __init__(self, *, node=None):
...
+        else:
+            node.get_logger().debug(
+                '\'use_sim_time\' parameter not set, \
```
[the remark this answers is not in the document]
vinnamkim: Fixed

rclpy/test/test_time_source.py
vinnamkim: Fixed

rclpy/test/test_time_source.py
vinnamkim: Fixed

rclpy/rclpy/time_source.py
vinnamkim: Fixed

rclpy/rclpy/time_source.py
vinnamkim: Fixed

rclpy/test/test_time_source.py
vinnamkim: Fixed

jacobperron:
Test failures are unrelated.

PR #307, opened 2019-04-04
Refactor QoS Python-C conversion into less error-prone pattern (pre-QoS, standalone)

emersonknapp:
The manual ordering of DECREFs is unwieldy - refactor to a simpler pattern that is easier to extend.

I am adding new policy types for QoS - which will utilize this change to make it much easier to do so. 

### DIFF
--- a/rclpy/src/rclpy_common/src/common.c
+++ b/rclpy/src/rclpy_common/src/common.c
@@ -15,12 +15,10 @@
 #include "rclpy_common/common.h"
 
 /**
- * Mirrors the struct rmw_qos_profile_t from rmw/types.h,
- * with the addition of member `profile` to to hold the Python QoSProfile
+ * Mirrors the struct rmw_qos_profile_t from rmw/types.h
  */
 typedef struct rclpy_qos_profile
 {
-  PyObject * profile;
   PyObject * depth;
   PyObject * history;
   PyObject * reliability;
@@ -29,18 +27,14 @@ typedef struct rclpy_qos_profile
 } rclpy_qos_profile_t;
 
 void
-init_rclpy_qos_profile(rclpy_qos_profile_t * rclpy_profile, PyObject * py_profile)
+init_rclpy_qos_profile(rclpy_qos_profile_t * rclpy_profile)
 {
   memset(rclpy_profile, 0, sizeof(*rclpy_profile));
-  rclpy_profile->profile = py_profile;
 }
 
 void
-cleanup_rclpy_qos_profile(rclpy_qos_profile_t * profile, bool profile_also)
+cleanup_rclpy_qos_profile(rclpy_qos_profile_t * profile)
 {
-  if (profile_also) {
-    Py_XDECREF(profile->profile);
-  }
   Py_XDECREF(profile->depth);
   Py_XDECREF(profile->history);
   Py_XDECREF(profile->reliability);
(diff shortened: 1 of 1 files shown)


emersonknapp:
@thomas-moulard - please run the following CI job:

* Gist file: https://gist.githubusercontent.com/emersonknapp/136ce7fc45add2a8553c08dad819d635/raw/950e24815a567dffac2eb4f7a4fb45c9f5da9ee7/ros2.repos
* Additional test flags: --packages-up-to rclpy
* CI Jobs: ci_launcher

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6630/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2956/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5440/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6452/)

rclpy/src/rclpy_common/src/common.c
```diff
@@ -34,63 +34,56 @@ rclpy_convert_to_py_qos_policy(void * profile)
...
   set_result += PyObject_SetAttrString(pyqos_profile,
       "avoid_ros_namespace_conventions", pyqos_avoid_ros_namespace_conventions);
 
-  Py_DECREF(pyqos_depth);
-  Py_DECREF(pyqos_history);
-  Py_DECREF(pyqos_reliability);
-  Py_DECREF(pyqos_durability);
-  Py_DECREF(pyqos_avoid_ros_namespace_conventions);
+cleanup_qos_setting:
+  if (pyqos_depth) {Py_DECREF(pyqos_depth);}
```
thomas-moulard: You can drop the `if (...)` and use `Py_XDECREF` instead.

rclpy/src/rclpy_common/src/common.c
```diff
@@ -34,63 +34,56 @@ rclpy_convert_to_py_qos_policy(void * profile)
...
 
+  // start qos setting
   rmw_qos_profile_t * qos_profile = (rmw_qos_profile_t *)profile;
-
-  PyObject * pyqos_depth = PyLong_FromSize_t(qos_profile->depth);
+  PyObject * pyqos_depth = NULL;
+  PyObject * pyqos_history = NULL;
+  PyObject * pyqos_reliability = NULL;
+  PyObject * pyqos_durability = NULL;
+  PyObject * pyqos_avoid_ros_namespace_conventions = NULL;
```
thomas-moulard: You could declare a `rclpy_row_profile_t` which would be a 1:1 mapping of `rmw_qos_profile_t` with PyObject.
This would allow you to e.g. pass it to a cleaning function and avoid the goto.

rclpy/src/rclpy_common/src/common.c
emersonknapp: Cool! Didn't know about that

rclpy/src/rclpy_common/src/common.c
emersonknapp: I'm not positive I understand - you mean something like

```
struct rclpy_qos_profile_t {
  PyObject * depth;
  PyObject * history;
  // etc ...
};

void clean_rclpy_qos_profile(rclpy_qos_profile_t * p) {
  Py_XDECREF(p->depth);
  Py_XDECREF(p->history);
  // etc ...
}

PyObject * rclpy_convert_to_py_qos_policy(void * profile) {
  rclpy_qos_profile_t pyprof;
  // etc ...

  pyprof->depth = PyLong_FromSize_t(qos_profile->depth);
  if (!pyprof->depth) {
    clean_rclpy_qos_profile(pyprof);
    return NULL;
  }

  //etc ...
}

```

thomas-moulard: Yeah, having a separate clean-up function would spare you the need to use goto.
It's also easier to test, easier to extend, easier to split into multiple functions, etc.

thomas-moulard:
It's hard to know when we need to decrease the ref count.
Here is what I found: https://mail.python.org/pipermail/python-list/2009-November/557842.html

IIUC this code is correct because the cleanup code (Py_XDECREF) is called even if the function succeeds which is what we want.
Why? Creating the Python object increases the ref count by one, PyObject_SetAttrString increases the ref count by one so we need to decrease by one to make sure the objects are not leaker when the Python object holding those attributes die.

rclpy/src/rclpy_common/src/common.c
```diff
@@ -14,6 +14,33 @@
...
+{
+  PyObject * depth;
+  PyObject * history;
+  PyObject * reliability;
+  PyObject * durability;
+  PyObject * avoid_ros_namespace_conventions;
+} rclpy_qos_profile_t;
+
+void init_rclpy_qos_profile(rclpy_qos_profile_t * profile) {
+  profile->depth = NULL;
```
thomas-moulard: `memset` ?

emersonknapp:
@tfoote this PR is prerequisite for upcoming QoS work, I'm not sure who the best person is to review rclpy changes - can we ping them?

rclpy/src/rclpy_common/src/common.c
```diff
@@ -34,63 +59,53 @@ rclpy_convert_to_py_qos_policy(void * profile)
...
 
+  // start qos setting
   rmw_qos_profile_t * qos_profile = (rmw_qos_profile_t *)profile;
+  rclpy_qos_profile_t pyqos;
+  init_rclpy_qos_profile(&pyqos);
+  int set_result = -1;
 
-  PyObject * pyqos_depth = PyLong_FromSize_t(qos_profile->depth);
-  if (!pyqos_depth) {
-    Py_DECREF(pyqos_profile);
```
jacobperron: We should still decref `pyqos_profile` if there is an error between here and any return.

rclpy/src/rclpy_common/src/common.c
```diff
@@ -34,63 +59,53 @@ rclpy_convert_to_py_qos_policy(void * profile)
     return NULL;
   }
 
+  // start qos setting
   rmw_qos_profile_t * qos_profile = (rmw_qos_profile_t *)profile;
+  rclpy_qos_profile_t pyqos;
```
jacobperron: nit: Since it is a C struct, I would name this `rclpy_qos` to better distinguish it from the Python object `pyqos_profile`. We typically prefix Python objects with `py`.

emersonknapp: Yeah good catch

rclpy/src/rclpy_common/src/common.c
emersonknapp: Yeah good catch - updating

rclpy/src/rclpy_common/src/common.c
```diff
@@ -34,63 +68,54 @@ rclpy_convert_to_py_qos_policy(void * profile)
...
+  init_rclpy_qos_profile(&rclpy_qos, pyqos_profile);
 
-  PyObject * pyqos_depth = PyLong_FromSize_t(qos_profile->depth);
-  if (!pyqos_depth) {
-    Py_DECREF(pyqos_profile);
+  int set_result = -1;
+
+  rclpy_qos.depth = PyLong_FromSize_t(qos_profile->depth);
+  if (!rclpy_qos.depth) {
+    cleanup_rclpy_qos_profile(&rclpy_qos, true);
```
jacobperron: It's a personal preference, but I think it would be more readable to keep `Py_DECREF(pyqos)` at the same scope level where it is initialized (not in the `rclpy_qos` cleanup function). This would be more in line with pattern we see throughout the rest of the C extension. I don't feel too strongly about it though.

emersonknapp: I don't feel strongly about it either - changed

jacobperron: thanks :)

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6656/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2980/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5467/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6488/)

Issue #287, opened 2019-03-18
confuse about the sigint_gc's lifecycle


reed-lau:
https://github.com/ros2/rclpy/blob/e9cc80ca752fb16c059a994dc0dbe6ec79051b35/rclpy/rclpy/executors.py#L466

In my point of view ,the signt_gc (short for sigint guard condition) is an global behaviour. why each time create a new one?

this may be related with https://github.com/ros2/rclpy/issues/253

Issue #192, opened 2018-06-19
Threads should have their own signal handler


dhood:
Before `rcl_wait` is called, an existing signal handler is stored so that it can be called from within rclpy's signal handler. The existing signal handler is restored after waking up.

This is not currently robust to `rclpy_wait` being called from multiple threads; each thread ought to store its own signal handler to restore.

wjwwood:
Well, I'm not sure that's the right fix for this. Instead, maybe there should be a thread-agnostic signal handler in the Python layer that acts like the rclcpp signal handler which triggers the sigint guard condition for all wait sets:

- https://github.com/ros2/rclcpp/blob/d33a46c3b69a3c5307a2ddf3029493279a6981f3/rclcpp/src/rclcpp/utilities.cpp#L161
- https://github.com/ros2/rclcpp/blob/d33a46c3b69a3c5307a2ddf3029493279a6981f3/rclcpp/src/rclcpp/utilities.cpp#L112-L129

So to recap, I think the issue is that `rclpy_wait` (which as you said calls `rcl_wait`) is not thread-safe and should be. I think the solution is to have one signal handler which is registered during "rclpy init" (either in C or Python, not sure which yet), rather than within `rclpy_wait`, and then multiple guard conditions, one for each wait set just like `rclcpp`.

I'm not sure it ever make sense to have one "stored signal handler to be restored" for each thread. I did mention that when we were talking in person, but after thinking about it more, I'm not sure that would work correctly.

pbaughman:
@dejanpan

I think we're getting burned by this right now.  We have a python process with multiple nodes and multiple executors on different threads, but our process is still blowing up with 
>"Executor.wait_for_ready_callbacks() called concurrently in multiple threads"
Apparently in c++ it's OK to run multiple executors on multiple threads, but in python we're being prevented from doing this by the global g_wait_set_spinning bool defined at the module https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/executors.py

esteve:
I think the following might work:

- when a `WaitSet` is created, create a `GuardCondition` and add it as a
`PyCapsule` to the Python object for the `WaitSet`
- in `rclpy_wait`, retrieve the `GuardCondition` from the `PyCapsule`
- wait on the `GuardCondition` before calling `rcl_wait`, after
https://github.com/ros2/rclpy/blob/master/rclpy/src/rclpy/_rclpy.c#L2572
[...]

Issue #253, opened 2018-12-14
rclpy crash sometimes when hitting Ctrl-C


clalancette:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04 AMD64
- Installation type:
  - Source
- Version or commit hash:
  - Release versions of crystal, from https://github.com/ros2/ros2/pull/624
- DDS implementation:
  - RTI Connext
- Client library (if applicable):
  - rclpy-ish

#### Steps to reproduce issue
Terminal 1:
```
ros2 launch topic_monitor depth_demo.launch.py
```

Terminal 2:
```
ros2 run topic_monitor topic_monitor
(hit Ctrl-C here)
```

#### Expected behavior
Topic monitor quits cleanly.

#### Actual behavior
Topic monitor quits with various different behaviors.  One particularly nasty one:

```
[INFO] [topic_monitor]: /topic1_data: 159
[INFO] [topic_monitor]: /topic1_data: 160
[INFO] [topic_monitor]: /topic1_data: 161
[INFO] [topic_monitor]: /topic1_data: 162
[INFO] [topic_monitor]: /topic1_data: 163
[INFO] [topic_monitor]: /topic1_data: 164
[INFO] [topic_monitor]: /topic1_data: 165
[INFO] [topic_monitor]: /topic1_data: 166
[INFO] [topic_monitor]: /topic1_data: 167
[INFO] [topic_monitor]: /topic1_data: 168
[INFO] [topic_monitor]: /topic1_data: 169
^CRuntimeError: Failed to trigger guard_condition: guard condition handle implementation '�[�'(0x7f553e777db0) does not match rmw implementation 'rmw_connext_cpp'(0x7f553bf7c1a0), at /home/ubuntu/ros2_ws/src/ros2/rmw_connext/rmw_connext_shared_cpp/src/trigger_guard_condition.cpp:35, at /home/ubuntu/ros2_ws/src/ros2/rcl/rcl/src/rcl/guard_condition.c:160

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
File "/home/ubuntu/ros2_ws/build/topic_monitor/topic_monitor/scripts/topic_monitor.py", line 420, in main
while data_receiving_thread.isAlive():
SystemError: PyEval_EvalFrameEx returned a result with an error set

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #308, opened 2019-04-05
Every executor gets its own SIGINT guard condition

sloretz:
Implements @wjwwood's comment https://github.com/ros2/rclpy/issues/192#issuecomment-398251659

This moves signal handling code to a separate module `_rclpy_signal_handler`. Every executor adds a guard condition to a global list. The signal handler triggers all guard conditions in that list.

Resolves ros2/rclpy#192, and because of that this PR also allows multiple executors to wait in parallel again. This also resolves ros2/rclpy#287 by having each executor reuse the same guard condition.

It does not change the updated behavior of ros2/rclpy#253 described here: https://github.com/ros2/rclpy/issues/253#issuecomment-475613097 . Maybe that should be an issue downstream of rclpy?

### DIFF
--- a/rclpy/rclpy/signals.py
+++ b/rclpy/rclpy/signals.py
@@ -1,0 +1,37 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+from rclpy.impl.implementation_singleton import rclpy_signal_handler_implementation as _signals
+from rclpy.utilities import get_default_context
+
+
+class SignalHandlerGuardCondition:
+
+    def __init__(self, context=None):
+        if context is None:
+            context = get_default_context()
+        self.guard_handle, self.guard_pointer = _rclpy.rclpy_create_guard_condition(context.handle)
+        _signals.rclpy_register_sigint_guard_condition(self.guard_handle)
+
+    def __del__(self):
+        self.destroy()
+
+    def destroy(self):
+        if self.guard_handle is None:
+            return
+        _signals.rclpy_unregister_sigint_guard_condition(self.guard_handle)
+        _rclpy.rclpy_destroy_entity(self.guard_handle)
+        self.guard_handle = None
+        self.guard_pointer = None
--- a/rclpy/rclpy/signals.py
+++ b/rclpy/rclpy/signals.py
@@ -22,7 +22,7 @@ class SignalHandlerGuardCondition:
     def __init__(self, context=None):
         if context is None:
             context = get_default_context()
-        self.guard_handle, self.guard_pointer = _rclpy.rclpy_create_guard_condition(context.handle)
+        self.guard_handle, _ = _rclpy.rclpy_create_guard_condition(context.handle)
         _signals.rclpy_register_sigint_guard_condition(self.guard_handle)
 
     def __del__(self):
@@ -34,4 +34,3 @@ class SignalHandlerGuardCondition:
         _signals.rclpy_unregister_sigint_guard_condition(self.guard_handle)
         _rclpy.rclpy_destroy_entity(self.guard_handle)
         self.guard_handle = None
-        self.guard_pointer = None
--- a/rclpy/src/rclpy/_rclpy_signal_handler.c
+++ b/rclpy/src/rclpy/_rclpy_signal_handler.c
@@ -98,7 +98,8 @@ call_original_signal_handler(int signo)
  */
 static void catch_function(int signo)
 {
-  rcl_guard_condition_t ** guard_conditions = atomic_load(&g_guard_conditions);
+  rcl_guard_condition_t ** guard_conditions;
+  rcutils_atomic_load(&g_guard_conditions, guard_conditions);
   if (NULL == guard_conditions || NULL == guard_conditions[0]) {
     // There may have been another signal handler chaining to this one when we last tried to
     // restore the old signal handler.
@@ -151,7 +152,8 @@ rclpy_register_sigint_guard_condition(PyObject * Py_UNUSED(self), PyObject * arg
     return NULL;
   }
 
-  rcl_guard_condition_t ** guard_conditions = atomic_load(&g_guard_conditions);
+  rcl_guard_condition_t ** guard_conditions;
+  rcutils_atomic_load(&g_guard_conditions, guard_conditions);
 
   // Figure out how big the list currently is
   size_t count_gcs = 0;
@@ -179,7 +181,8 @@ rclpy_register_sigint_guard_condition(PyObject * Py_UNUSED(self), PyObject * arg
   new_gcs[count_gcs + 1] = NULL;
 
   // Swap the lists and free the old
-  rcl_guard_condition_t ** old_gcs = atomic_exchange(&g_guard_conditions, new_gcs);
+  rcl_guard_condition_t ** old_gcs;
+  rcutils_atomic_exchange(&g_guard_conditions, old_gcs, new_gcs);
   if (NULL != old_gcs) {
     allocator.deallocate(old_gcs, allocator.state);
   }
@@ -213,7 +216,8 @@ rclpy_unregister_sigint_guard_condition(PyObject * Py_UNUSED(self), PyObject * a
     return NULL;
   }
 
-  rcl_guard_condition_t ** guard_conditions = atomic_load(&g_guard_conditions);
+  rcl_guard_condition_t ** guard_conditions;
+  rcutils_atomic_load(&g_guard_conditions, guard_conditions);
 
   // Figure out how big the list currently is
   size_t count_gcs = 0;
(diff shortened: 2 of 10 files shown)


sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6619/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2949/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5434/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6443/)

sloretz:
CI looks good, just failures from ros2/build_cop#166 and ros-visualization/rqt#192

rclpy/src/rclpy/_rclpy_signal_handler.c
```diff
@@ -0,0 +1,288 @@
...
+  for (size_t i = 0; i < count_gcs; ++i) {
+    new_gcs[i] = g_guard_conditions[i];
+  }
+  new_gcs[count_gcs] = gc;
+  new_gcs[count_gcs + 1] = NULL;
+
+  // Swap the lists and free the old
+  rcl_guard_condition_t ** old_gcs = g_guard_conditions;
+  // Assumes this assignment is atomic
+  g_guard_conditions = new_gcs;
```
hidmic: @sloretz hmm, I'm not sure this is always true. It depends a lot on how this is compiled for each of our supported processor architectures. And that's even before adding SMP and cache coherence to the equation. 

Why not using OS specific atomics (i.e. abstracted CPU specific atomics)?

sloretz: I didn't find an atomic for pointer types. Maybe I could use a different type that is at least as big as an `rcl_guard_condition **` and cast to it?

sloretz:
> how's that this is both thread- and signal-safe? Specially when it comes to manipulating a global list. Am I missing something?

This relies on the GIL for thread safety. Neither function that edits the list releases the GIL. As long as they don't invoke anything that can call arbitrary python code (like getting an attribute off a python object, because it could have a custom `__getattr__` that releases the GIL), there's no way these could touch the list at the same time.

Since the signal handler only reads the list, this tries to make sure `g_guard_conditions` can be read no matter where the code is interrupted. Edits are done by copying the current list, manipulating the copy, then swapping the original with the copy (assumes assigning to the global pointer is atomic so the signal handler doesn't fetch some invalid memory address).

rclpy/src/rclpy/_rclpy_signal_handler.c
```diff
@@ -0,0 +1,288 @@
...
+    ++pgc;
+  }
+
+  // Chain signal handlers
+  call_original_signal_handler(signo);
+}
+
+/// Register a guard condition to be triggered when SIGINT is received.
+/**
+ * On failure, an exception is raised and NULL is returned if:
```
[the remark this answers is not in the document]
sloretz: Removed in d48b19502950eab73ce178474a5e8a84219c86f2

rclpy/src/rclpy/_rclpy_signal_handler.c
```diff
@@ -0,0 +1,288 @@
...
+    rcl_guard_condition_t ** new_gcs =
+      allocator.allocate(sizeof(rcl_guard_condition_t *) * (count_gcs), allocator.state);
+
+    // Put remaining guard conditions in the list, ending with a sentinel of NULL
+    size_t offset = 0;
+    for (size_t i = 0; i < count_gcs; ++i) {
+      // assumes guard condition was only added to list once
+      if (g_guard_conditions[i + offset] == gc) {
+        offset = 1;
+      }
```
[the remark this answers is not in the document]
sloretz: Split loop in 506fb44

rclpy/src/rclpy/_rclpy_signal_handler.c
sloretz: Gah, I didn't realize any type could be made atomic. 35388745423aad221846dfc7e1d0195da9e3c0ae uses an atomic type and methods. IIUC the signal handler could deadlock if this type is not lock-free, so the code issues a python warning once if that's the case.

sloretz:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6641/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2967/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5453/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6466/)

rclpy/src/rclpy/_rclpy_signal_handler.c
hidmic: Well, `gcc` has some [intrinsics](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html) that `llvm` seems to support as well. As usual, `msvc` has it's [own stuff](https://docs.microsoft.com/en-us/windows/desktop/api/winnt/nf-winnt-interlockedexchangepointer). We should verify against CI though.

hidmic:
> This relies on the GIL for thread safety.

I see. I thought we were not relying on the GIL on purpose. I've seen locks here and there (not in this PR) that seemed redundant with the GIL. Anyways, sounds good.

rclpy/src/rclpy/_rclpy_signal_handler.c
hidmic: Oh, I didn't know we had an `rcutils` helper, nice! Though in that case I think we should be using `rcutils_atomic_*` macros/functions instead of `atomic_*` ones.

sloretz: Using `rcutils_atomic_*` in a2b1aaf79663a74ce168fe3d270ad69d279b99a8

sloretz:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6642/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2968/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5454/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6467/)

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6644/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2970/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5456/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6469/)

rclpy/src/rclpy/_rclpy_signal_handler.c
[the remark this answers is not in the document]
sloretz: There was one extra assignment in the second loop, fixed it in 
081c921

sloretz:
CI (test packages above rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6655/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2979/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5466/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6487/)

sloretz:
Blocked by ros2/rcutils#150

sloretz:
Quick check on windows with ros2/rcutils#150 to make sure warnings are gone [Build Status](https://ci.ros2.org/job/ci_windows/6491/)

sloretz:
CI and moving back to review
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6664/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2984/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5471/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6493/)

rclpy/src/rclpy/_rclpy_signal_handler.c
```diff
@@ -0,0 +1,316 @@
...
+    while (NULL != guard_conditions[count_gcs]) {
+      if (gc == guard_conditions[count_gcs]) {
+        found_gc = true;
+        found_index = count_gcs;
+      }
+      ++count_gcs;
+    }
+  }
+
+  if (count_gcs == 0 || !found_gc) {
```
dirk-thomas: Nitpick: isn't the second condition sufficient on its own?

rclpy/rclpy/signals.py
```diff
@@ -0,0 +1,37 @@
...
+from rclpy.impl.implementation_singleton import rclpy_signal_handler_implementation as _signals
+from rclpy.utilities import get_default_context
+
+
+class SignalHandlerGuardCondition:
+
+    def __init__(self, context=None):
+        if context is None:
+            context = get_default_context()
+        self.guard_handle, self.guard_pointer = _rclpy.rclpy_create_guard_condition(context.handle)
```
dirk-thomas: Is there any need to store the `guard_pointer`?

sloretz: No need, removed in 28d9f55b802f29c6a587d5d26308b5cd6889851b

rclpy/src/rclpy/_rclpy_signal_handler.c
sloretz: It is, removed first part of condition in 3d25d7f3efe8fe96f72d9c5e41c454a903d7722a

sloretz:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6745/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3053/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5543/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6570/)

sloretz:
CI looks good. Since the only changes were addressing CI comments I'll assume @dirk-thomas's approval still applies and merge this.

PR #309, opened 2019-04-05
Executor exits immediately when shut down

sloretz:
This makes `executor.shutdown()` trigger the executor's guard condition. Without this the executor could block forever if no more work comes in.

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -125,6 +125,12 @@ class TimeoutException(Exception):
     pass
 
 
+class ShutdownException(Exception):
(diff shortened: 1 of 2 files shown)


sloretz:
CI (Packages above rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6634/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2960/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5444/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6457/)

PR #310, opened 2019-04-05
Refresh nodes when executor is woken

sloretz:
This fixes bug where executor would rebuild wait_set without adding or
removing nodes. This happened when an executor was waiting and had not
executed any work yet at the time the node was added or removed.

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -389,11 +389,13 @@ class Executor:
         if timeout_nsec > 0:
             timeout_timer = WallTimer(None, None, timeout_nsec)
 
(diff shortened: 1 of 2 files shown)


sloretz:
CI (test packages above rclpy)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6635/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2961/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5445/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6458/)

PR #311, opened 2019-04-10
Set QoS profile to default values (pre-QoS, standalone)

emersonknapp:
Set allocated `rmw_qos_profile_t` to the default, to future-proof against uninitialized data if new fields are added. E.g. if a field is added to `rmw_qos_profile_t` and that field is not immediately exposed in `rclpy`, the value of that field will be garbage data, and be passed all the way to the RMW implementation.


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6653/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2977/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5464/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6484/)

PR #312, opened 2019-04-10
QoS - Expose Lifespan, Deadline, and Liveliness policy settings

emersonknapp:
Depends on ros2/rmw#173

Expose the new QoS policy settings for the Python client. Needs `rmw/types.h` from the above mentioned PR

### DIFF
--- a/rclpy/rclpy/duration.py
+++ b/rclpy/rclpy/duration.py
@@ -76,3 +76,6 @@ class Duration:
         if not isinstance(msg, builtin_interfaces.msg.Duration):
             raise TypeError('Must pass a builtin_interfaces.msg.Duration object')
         return cls(seconds=msg.sec, nanoseconds=msg.nanosec)
+
+    def get_c_duration(self):
+        return self._duration_handle
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -159,7 +159,7 @@ class QoSProfile:
 
     @liveliness.setter
     def liveliness(self, value):
-        assert isinstance(value, QoSLivelinessPolicy) or isinstance(value, int)
+        assert isinstance(value, (QoSLivelinessPolicy, int))
         self._liveliness = QoSLivelinessPolicy(value)
 
     @property
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -205,6 +205,13 @@ class QoSProfile:
             self.avoid_ros_namespace_conventions,
         )
 
+    def __eq__(self, other):
+        if not isinstance(other, QoSProfile):
+            return False
+        return all(
+            self.__getattribute__(slot) == other.__getattribute__(slot)
+            for slot in self.__slots__)
+
 
 class QoSHistoryPolicy(IntEnum):
     """
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -14,6 +14,7 @@
 
 from enum import IntEnum
 
+from rclpy.duration import Duration
 from rclpy.impl.implementation_singleton import rclpy_action_implementation as _rclpy_action
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
@@ -26,6 +27,10 @@ class QoSProfile:
         '_depth',
         '_reliability',
         '_durability',
+        '_lifespan',
+        '_deadline',
+        '_liveliness',
+        '_liveliness_lease_duration',
         '_avoid_ros_namespace_conventions',
     ]
 
@@ -42,6 +47,12 @@ class QoSProfile:
         self.durability = kwargs.get(
             'durability',
             QoSDurabilityPolicy.RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT)
+        self.lifespan = kwargs.get('lifespan', Duration())
+        self.deadline = kwargs.get('deadline', Duration())
+        self.liveliness = kwargs.get(
+            'liveliness',
+            QoSLivelinessPolicy.RMW_QOS_POLICY_LIVELINESS_SYSTEM_DEFAULT)
+        self.liveliness_lease_duration = kwargs.get('liveliness_lease_duration', Duration())
         self.avoid_ros_namespace_conventions = kwargs.get(
             'avoid_ros_namespace_conventions',
             False)
@@ -107,6 +118,66 @@ class QoSProfile:
         self._depth = value
 
     @property
+    def lifespan(self):
+        """
+        Get field 'lifespan'.
+
+        :returns: lifespan attribute
+        :rtype: Duration
+        """
+        return self._lifespan
+
+    @lifespan.setter
+    def lifespan(self, value):
+        assert isinstance(value, Duration)
+        self._lifespan = value
+
+    @property
+    def deadline(self):
+        """
+        Get field 'deadline'.
+
+        :returns: deadline attribute.
+        :rtype: Duration
+        """
+        return self._deadline
+
+    @deadline.setter
+    def deadline(self, value):
+        assert isinstance(value, Duration)
+        self._deadline = value
+
+    @property
+    def liveliness(self):
+        """
+        Get field 'liveliness'.
+
+        :returns: liveliness attribute
+        :rtype: QoSLivelinessPolicy
+        """
+        return self._liveliness
+
+    @liveliness.setter
+    def liveliness(self, value):
+        assert isinstance(value, QoSLivelinessPolicy) or isinstance(value, int)
+        self._liveliness = QoSLivelinessPolicy(value)
+
+    @property
+    def liveliness_lease_duration(self):
+        """
+        Get field 'liveliness_lease_duration'.
+
+        :returns: liveliness_lease_duration attribute.
+        :rtype: Duration
+        """
+        return self._liveliness_lease_duration
+
+    @liveliness_lease_duration.setter
+    def liveliness_lease_duration(self, value):
+        assert isinstance(value, Duration)
+        self._liveliness_lease_duration = value
+
+    @property
     def avoid_ros_namespace_conventions(self):
         """
         Get field 'avoid_ros_namespace_conventions'.
@@ -127,6 +198,10 @@ class QoSProfile:
             self.depth,
             self.reliability,
             self.durability,
+            self.lifespan.get_c_duration(),
+            self.deadline.get_c_duration(),
+            self.liveliness,
+            self.liveliness_lease_duration.get_c_duration(),
             self.avoid_ros_namespace_conventions,
         )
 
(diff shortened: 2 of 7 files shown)


emersonknapp:
@wjwwood this is ready for review, can we update the label/waffle column

rclpy/rclpy/qos.py
```diff
@@ -106,6 +117,66 @@ def depth(self, value):
...
+        Get field 'liveliness'.
+
+        :returns: liveliness attribute
+        :rtype: QoSLivelinessPolicy
+        """
+        return self._liveliness
+
+    @liveliness.setter
+    def liveliness(self, value):
+        assert isinstance(value, QoSLivelinessPolicy) or isinstance(value, int)
```
sloretz: nit, could be shorter

```python3
assert isinstance(value, (QoSLivelinessPolicy, int))
```

rclpy/rclpy/duration.py
```diff
@@ -76,3 +76,6 @@ def from_msg(cls, msg):
         if not isinstance(msg, builtin_interfaces.msg.Duration):
             raise TypeError('Must pass a builtin_interfaces.msg.Duration object')
         return cls(seconds=msg.sec, nanoseconds=msg.nanosec)
+
+    def get_c_duration(self):
```
sloretz: nit, recommend using `@property` instead of a `get_` method in python.

```python3
@property
def duration_handle(self):
    return self._duration_handle
```

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3583,6 +3595,21 @@ rclpy_convert_from_py_qos_policy(PyObject * Py_UNUSED(self), PyObject * args)
   qos_profile->depth = pyqos_depth;
   qos_profile->reliability = pyqos_reliability;
   qos_profile->durability = pyqos_durability;
+
+  rcl_duration_t * lifespan =
+    (rcl_duration_t *)PyCapsule_GetPointer(pyqos_lifespan, "rcl_duration_t");
+  qos_profile->lifespan = NANOSECONDS_TO_RMW_TIME(lifespan->nanoseconds);
```
sloretz: Need to check if lifespan is `NULL`. If so, this function must return `NULL` because an exception will be set.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3583,6 +3595,21 @@ rclpy_convert_from_py_qos_policy(PyObject * Py_UNUSED(self), PyObject * args)
...
+
+  rcl_duration_t * deadline =
+    (rcl_duration_t *)PyCapsule_GetPointer(pyqos_deadline, "rcl_duration_t");
+  qos_profile->deadline = NANOSECONDS_TO_RMW_TIME(deadline->nanoseconds);
```
sloretz: Need to check if `deadline` is `NULL`

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3583,6 +3595,21 @@ rclpy_convert_from_py_qos_policy(PyObject * Py_UNUSED(self), PyObject * args)
...
+
+  qos_profile->liveliness = pyqos_liveliness;
+  rcl_duration_t * liveliness_lease_duration =
+    (rcl_duration_t *)PyCapsule_GetPointer(pyqos_liveliness_lease_duration, "rcl_duration_t");
+  qos_profile->liveliness_lease_duration =
+    NANOSECONDS_TO_RMW_TIME(liveliness_lease_duration->nanoseconds);
```
sloretz: please check if `liveliness_lease_duration` is `NULL` too

rclpy/src/rclpy_common/src/common.c
```diff
@@ -109,6 +117,31 @@ rclpy_convert_to_py_names_and_types(rcl_names_and_types_t * names_and_types)
...
+  if (!pyduration_module) {
+    return NULL;
+  }
+  PyObject * pyduration_class = PyObject_GetAttrString(pyduration_module, "Duration");
+  Py_DECREF(pyduration_module);
+  if (!pyduration_class) {
+    return NULL;
+  }
+
+  PyObject * args = PyTuple_New(0);
```
sloretz: Need to check if `args` is `NULL`, and if so `Py_DECREF(pyduration_class)` and return `NULL`

sloretz: Need to check if `args` is null, `PyObject_Call` [forbids it being  from `NULL`](https://github.com/python/cpython/blob/d6bf6f2d0c83f0c64ce86e7b9340278627798090/Include/object.h#L127)

rclpy/src/rclpy_common/src/common.c
```diff
@@ -109,6 +117,31 @@ rclpy_convert_to_py_names_and_types(rcl_names_and_types_t * names_and_types)
...
+  PyObject * kwargs = Py_BuildValue("{s:i}", "nanoseconds", total_nanoseconds);
```
sloretz: Need to check if `kwargs` is `NULL`. While `PyObject_Call()` allows kwargs to be `NULL`, that's not desired here. If `kwargs` is null it means `Py_BuildValue` set an exception, which would cause an [`assert()`  in `PyObject_Call`](https://github.com/python/cpython/blob/888f37bc2826d9ab2cbec6f153e7f58a34785c4a/Objects/call.c#L223).

emersonknapp:
Thanks for the feedback! I will address the comments. There are no other dependencies than the one listed in the description, and that's been merged. This PR stands alone.

rclpy/rclpy/duration.py
emersonknapp: I was copying a "c-handle pattern" that I observed here https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/qos.py#L124 - do you think I should change both of these to use properties instead?

sloretz: Rodger, Feel free to ignore this comment and use the same pattern. If the pattern should be changed it doesn't need to block this PR.

emersonknapp:
@sloretz I've updated to address your comments - did a bit of restructuring to make the code less error prone. Additionally, exposed an API and used that to add some unit tests for the new type conversions.

emersonknapp:
@thomas-moulard please trigger CI for this PR
Gist: https://gist.githubusercontent.com/emersonknapp/0c3f022bd8104fa9ed39610501e24dfd/raw/001c650b781a8cccbff48dc801aad94128798075/ros2.repos
Additional BUILD flags: --packages-up-to rclpy
Additional TEST flags: --packages-select rclpy
CI Job: ci_launcher

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6792/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3089/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5584/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6618/)

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3545,6 +3545,28 @@ rclpy_get_service_names_and_types(PyObject * Py_UNUSED(self), PyObject * args)
...
+  * \param[in] pyobject Python Object that should be a Duration, error checking is done
+  * \param[out] out_time Valid destination for time data
+  * \return Whether or not conversion succeeded.
+  */
+bool
+_convert_py_duration_to_rmw_time(PyObject * pyobject, rmw_time_t * out_time)
+{
+  rcl_duration_t * duration = (rcl_duration_t *)PyCapsule_GetPointer(pyobject, "rcl_duration_t");
+  if (!duration) {
+    PyErr_Format(PyExc_TypeError,
```
sloretz: If `duration` is `NULL` then `GetPointer` already raised `ValueError`. If that exception type is fine then it's enough to return here, but if you'd like to change the type of exception set then call `PyErr_Clear()` first.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3587,11 +3618,49 @@ rclpy_convert_from_py_qos_policy(PyObject * Py_UNUSED(self), PyObject * args)
...
+{
+  PyObject * pyqos_profile = NULL;
+  if (!PyArg_ParseTuple(args, "O", &pyqos_profile)) {
+    return NULL;
+  }
+
+  rmw_qos_profile_t * profile = (rmw_qos_profile_t *)PyCapsule_GetPointer(
+    pyqos_profile, "rmw_qos_profile_t");
+  if (!profile) {
+    PyErr_Format(PyExc_TypeError, "Capsule was not a valid rmw_qos_profile_t.");
```
sloretz: Same as above, `PyCapsule_GetPointer` already raised `ValueError`

rclpy/src/rclpy/_rclpy.c
sloretz: cd6e8a5

rclpy/src/rclpy/_rclpy.c
sloretz: cd6e8a5

rclpy/src/rclpy/_rclpy.c
emersonknapp: Ah - thanks for the heads up on these, now I will know next time. Removed the redundant error messaging.

emersonknapp:
Windooooowwwwwwwws - I'll have to spin up a windows instance tomorrow and try this out... @sloretz any ideas about the 32-bit int issue in converting between layers?

rclpy/src/rclpy_common/src/common.c
```diff
@@ -109,8 +117,48 @@ rclpy_convert_to_py_names_and_types(rcl_names_and_types_t * names_and_types)
...
+  }
+  pyduration_class = PyObject_GetAttrString(pyduration_module, "Duration");
+  if (!pyduration_class) {
+    goto cleanup;
+  }
+  args = PyTuple_New(0);
+  if (!args) {
+    goto cleanup;
+  }
+  kwargs = Py_BuildValue("{s:l}", "nanoseconds", total_nanoseconds);
```
sloretz: @emersonknapp changing [lowercase `l` to uppercase `K`](https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue) may fix the windows issue. Using gcc 7.3 `long` and `long long` are both 8 bytes, but on windows `long` is only 4 bytes https://docs.microsoft.com/en-us/cpp/cpp/data-type-ranges?view=vs-2019

sloretz: Also, the documentation for `Py_BuildValue` doesn't include a `:` for dictionaries, it just says consecutive C values become key/value pairs. I think this should read `{sK}`.

emersonknapp: Thanks - I forgot about the `K` specifier, C99 spec only guarantees 32 bit for `long`, so that makes perfect sense, I did want a `long long`

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6803/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3099/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5594/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6628/)

PR #305, opened 2019-04-03
QoS - Minimal change for build against new rcl API

emersonknapp:
Connects to ros2/rmw#171

Allows this package to build when new QoS APIs are checked in, but exposes none of the new functionality

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

PR #316, opened 2019-04-18
QoS - API and implementation for Liveliness and Deadline event callbacks

emersonknapp:
Depends on ros2/rmw#171
Depends on ros2/rmw_fastrtps#276
Depends on ros2/rcl#432

* Present the Python-side API for receiving QoS event callbacks for Liveliness and Deadline QoS policies.
* Provide the C-side translation layer to `rcl` to support that API.

Unit tests are still in progress - but the code is working according to manual testing and is ready for review.

### DIFF
--- a/rclpy/rclpy/qos_event.py
+++ b/rclpy/rclpy/qos_event.py
@@ -1,4 +1,4 @@
-# Copyright 2019 Open Source Robotics Foundation, Inc.
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/rclpy/rclpy/qos_event.py
+++ b/rclpy/rclpy/qos_event.py
@@ -1,0 +1,226 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from enum import IntEnum
+from typing import Callable
+from typing import List
+from typing import NamedTuple
+
+from rclpy.callback_groups import CallbackGroup
+from rclpy.handle import Handle
+from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+from rclpy.waitable import NumberOfEntities
+from rclpy.waitable import Waitable
+
+
+class QoSPublisherEventType(IntEnum):
+    """
+    Enum for types of QoS events that a Publisher can receive.
+
+    This enum matches the one defined in rcl/event.h
+    """
+
+    RCL_PUBLISHER_OFFERED_DEADLINE_MISSED = 0
+    RCL_PUBLISHER_LIVELINESS_LOST = 1
+
+
+class QoSSubscriptionEventType(IntEnum):
+    """
+    Enum for types of QoS events that a Subscription can receive.
+
+    This enum matches the one defined in rcl/event.h
+    """
+
+    RCL_SUBSCRIPTION_REQUESTED_DEADLINE_MISSED = 0
+    RCL_SUBSCRIPTION_LIVELINESS_CHANGED = 1
+
+
+"""
+Payload type for Subscription Deadline callback.
+
+Mirrors rmw_requested_deadline_missed_status_t from rmw/types.h
+"""
+QoSRequestedDeadlineMissedInfo = NamedTuple(
+    'QoSRequestedDeadlineMissedInfo', [
+        ('total_count', 'int'),
+        ('total_count_change', 'int'),
+    ])
+
+"""
+Payload type for Subscription Liveliness callback.
+
+Mirrors rmw_liveliness_changed_status_t from rmw/types.h
+"""
+QoSLivelinessChangedInfo = NamedTuple(
+    'QoSLivelinessChangedInfo', [
+        ('alive_count', 'int'),
+        ('not_alive_count', 'int'),
+        ('alive_count_change', 'int'),
+        ('not_alive_count_change', 'int'),
+    ])
+
+"""
+Payload type for Publisher Deadline callback.
+
+Mirrors rmw_offered_deadline_missed_status_t from rmw/types.h
+"""
+QoSOfferedDeadlineMissedInfo = NamedTuple(
+    'QoSOfferedDeadlineMissedInfo', [
+        ('total_count', 'int'),
+        ('total_count_change', 'int'),
+    ])
+
+"""
+Payload type for Publisher Liveliness callback.
+
+Mirrors rmw_liveliness_lost_status_t from rmw/types.h
+"""
+QoSLivelinessLostInfo = NamedTuple(
+    'QoSLivelinessLostInfo', [
+        ('total_count', 'int'),
+        ('total_count_change', 'int'),
+    ])
+
+
+class QoSEventHandler(Waitable):
+    """Waitable type to handle QoS events."""
+
+    def __init__(
+        self,
+        *,
+        callback_group,
+        callback,
+        event_type,
+        parent_handle,
+    ):
+        # Waitable init adds self to callback_group
+        super().__init__(callback_group)
+        self.event_type = event_type
+        self.callback = callback
+
+        self._parent_handle = parent_handle
+        self._event_handle = _rclpy.rclpy_create_event(event_type, parent_handle)
+        self._ready_to_take_data = False
+        self._event_index = None
+
+    # Start Waitable API
+    def is_ready(self, wait_set):
+        """Return True if entities are ready in the wait set."""
+        if self._event_index is None:
+            return False
+        if _rclpy.rclpy_wait_set_is_ready('event', wait_set, self._event_index):
+            self._ready_to_take_data = True
+        return self._ready_to_take_data
+
+    def take_data(self):
+        """Take stuff from lower level so the wait set doesn't immediately wake again."""
+        if self._ready_to_take_data:
+            self._ready_to_take_data = False
+            return _rclpy.rclpy_take_event(
+                self._event_handle, self._parent_handle, self.event_type)
+        return None
+
+    async def execute(self, taken_data):
+        """Execute work after data has been taken from a ready wait set."""
+        if not taken_data:
+            return
+        self.callback(taken_data)
+
+    def get_num_entities(self):
+        """Return number of each type of entity used."""
+        return NumberOfEntities(num_events=1)
+
+    def add_to_wait_set(self, wait_set):
+        """Add entites to wait set."""
+        self._event_index = _rclpy.rclpy_wait_set_add_entity('event', wait_set, self._event_handle)
+    # End Waitable API
+
+
+class SubscriptionEventCallbacks:
+    """Container to provide middleware event callbacks for a Subscription."""
+
+    def __init__(
+        self,
+        *,
+        deadline: Callable[[QoSRequestedDeadlineMissedInfo], None] = None,
+        liveliness: Callable[[QoSLivelinessChangedInfo], None] = None,
+    ) -> None:
+        """
+        Constructor.
+
+        :param deadline: A user-defined callback that is called when a topic misses our
+            requested Deadline.
+        :param liveliness: A user-defined callback that is called when the Liveliness of
+            a Publisher on subscribed topic changes.
+        """
+        self.deadline = deadline
+        self.liveliness = liveliness
+
+    def create_event_handlers(
+        self, callback_group: CallbackGroup, subscription_handle: Handle
+    ) -> List[QoSEventHandler]:
+        event_handlers = []
+        with subscription_handle as subscription_capsule:
+            if self.deadline:
+                event_handlers.append(QoSEventHandler(
+                    callback_group=callback_group,
+                    callback=self.deadline,
+                    event_type=QoSSubscriptionEventType.RCL_SUBSCRIPTION_REQUESTED_DEADLINE_MISSED,
+                    parent_handle=subscription_capsule))
+            if self.liveliness:
+                event_handlers.append(QoSEventHandler(
+                    callback_group=callback_group,
+                    callback=self.liveliness,
+                    event_type=QoSSubscriptionEventType.RCL_SUBSCRIPTION_LIVELINESS_CHANGED,
+                    parent_handle=subscription_capsule))
+        return event_handlers
+
+
+class PublisherEventCallbacks:
+    """Container to provide middleware event callbacks for a Publisher."""
+
+    def __init__(
+        self,
+        *,
+        deadline: Callable[[QoSOfferedDeadlineMissedInfo], None] = None,
+        liveliness: Callable[[QoSLivelinessLostInfo], None] = None
+    ) -> None:
+        """
+        Constructor.
+
+        :param deadline: A user-defined callback that is called when the Publisher misses
+            its offered Deadline.
+        :param liveliness: A user-defined callback that is called when this Publisher
+            fails to signal its Liveliness and is reported as not-alive.
+        """
+        self.deadline = deadline
+        self.liveliness = liveliness
+
+    def create_event_handlers(
+        self, callback_group: CallbackGroup, publisher_handle
+    ) -> List[QoSEventHandler]:
+        event_handlers = []
+        if self.deadline:
+            event_handlers.append(QoSEventHandler(
+                callback_group=callback_group,
+                callback=self.deadline,
+                event_type=QoSPublisherEventType.RCL_PUBLISHER_OFFERED_DEADLINE_MISSED,
+                parent_handle=publisher_handle))
+        if self.liveliness:
+            event_handlers.append(QoSEventHandler(
+                callback_group=callback_group,
+                callback=self.liveliness,
+                event_type=QoSPublisherEventType.RCL_PUBLISHER_LIVELINESS_LOST,
+                parent_handle=publisher_handle))
+        return event_handlers
--- a/rclpy/rclpy/qos_event.py
+++ b/rclpy/rclpy/qos_event.py
@@ -16,7 +16,9 @@ from enum import IntEnum
 from typing import Callable
 from typing import List
 from typing import NamedTuple
+from typing import Optional
 
+import rclpy
 from rclpy.callback_groups import CallbackGroup
 from rclpy.handle import Handle
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
@@ -99,10 +101,10 @@ class QoSEventHandler(Waitable):
     def __init__(
         self,
         *,
-        callback_group,
-        callback,
-        event_type,
-        parent_handle,
+        callback_group: CallbackGroup,
+        callback: Callable,
+        event_type: IntEnum,
+        parent_handle: Handle,
     ):
         # Waitable init adds self to callback_group
         super().__init__(callback_group)
@@ -110,7 +112,10 @@ class QoSEventHandler(Waitable):
         self.callback = callback
 
         self._parent_handle = parent_handle
-        self._event_handle = _rclpy.rclpy_create_event(event_type, parent_handle)
+        with parent_handle as parent_capsule:
+            event_capsule = _rclpy.rclpy_create_event(event_type, parent_capsule)
+        self._event_handle = Handle(event_capsule)
+        self._event_handle.requires(self._parent_handle)
         self._ready_to_take_data = False
         self._event_index = None
 
@@ -127,15 +132,15 @@ class QoSEventHandler(Waitable):
         """Take stuff from lower level so the wait set doesn't immediately wake again."""
         if self._ready_to_take_data:
             self._ready_to_take_data = False
-            return _rclpy.rclpy_take_event(
-                self._event_handle, self._parent_handle, self.event_type)
+            with self._parent_handle as parent_capsule, self._event_handle as event_capsule:
+                return _rclpy.rclpy_take_event(event_capsule, parent_capsule, self.event_type)
         return None
 
     async def execute(self, taken_data):
         """Execute work after data has been taken from a ready wait set."""
         if not taken_data:
             return
-        self.callback(taken_data)
+        await rclpy.executors.await_or_execute(self.callback, [taken_data])
 
     def get_num_entities(self):
         """Return number of each type of entity used."""
@@ -143,7 +148,8 @@ class QoSEventHandler(Waitable):
 
     def add_to_wait_set(self, wait_set):
         """Add entites to wait set."""
-        self._event_index = _rclpy.rclpy_wait_set_add_entity('event', wait_set, self._event_handle)
+        with self._event_handle as event_capsule:
+            self._event_index = _rclpy.rclpy_wait_set_add_entity('event', wait_set, event_capsule)
     # End Waitable API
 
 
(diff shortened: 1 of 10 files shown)


sloretz:
@emersonknapp Looks like there are merge conflicts

emersonknapp:
Added new commit to refactor to use the new Subscription+Handle stuff properly

emersonknapp:
@sloretz when you get a chance, would love feedback on this PR

rclpy/rclpy/qos_event.py
```diff
@@ -0,0 +1,226 @@
...
+            self._ready_to_take_data = False
+            return _rclpy.rclpy_take_event(
+                self._event_handle, self._parent_handle, self.event_type)
+        return None
+
+    async def execute(self, taken_data):
+        """Execute work after data has been taken from a ready wait set."""
+        if not taken_data:
+            return
+        self.callback(taken_data)
```
sloretz: Recommend `await rclpy.executors.await_or_execute(callback, [taken_data])` to support coroutines as callbacks.

rclpy/rclpy/qos_event.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def get_num_entities(self):
+        """Return number of each type of entity used."""
+        return NumberOfEntities(num_events=1)
```
sloretz: Since there is just one entity, this might fit better as an `Event` class instead of using the `Waitable` API. Though, that makes adding it to the executor harder so I'm fine with this as is.

rclpy/rclpy/qos_event.py
```diff
@@ -0,0 +1,226 @@
...
+    # End Waitable API
+
+
+class SubscriptionEventCallbacks:
+    """Container to provide middleware event callbacks for a Subscription."""
+
+    def __init__(
+        self,
+        *,
+        deadline: Callable[[QoSRequestedDeadlineMissedInfo], None] = None,
```
sloretz: Nit, `Optional[Callable[[QoSRequestedDeadlineMissedInfo], None]]` since the argument may be `None`. Same below

rclpy/rclpy/qos_event.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+class PublisherEventCallbacks:
+    """Container to provide middleware event callbacks for a Publisher."""
+
+    def __init__(
+        self,
+        *,
+        deadline: Callable[[QoSOfferedDeadlineMissedInfo], None] = None,
+        liveliness: Callable[[QoSLivelinessLostInfo], None] = None
```
sloretz: Same nitpick about `Optional`

rclpy/rclpy/subscription.py
```diff
@@ -66,6 +70,10 @@ def __init__(
         self.qos_profile = qos_profile
         self.raw = raw
 
+        self.event_callbacks = event_callbacks
```
sloretz: It looks like `event_callbacks` is never used again. Recommend not storing it

rclpy/src/rclpy/_rclpy.c
```diff
@@ -2315,6 +2316,10 @@ rclpy_destroy_entity(PyObject * Py_UNUSED(self), PyObject * args)
       pyentity, "rcl_guard_condition_t");
     ret = rcl_guard_condition_fini(guard_condition);
     PyMem_Free(guard_condition);
+  } else if (PyCapsule_IsValid(pyentity, "rcl_event_t")) {
```
sloretz: Would you mind using `rclpy.handle.Handle` instead of adding to `rclpy_destroy_entity`? This method has been removed in #319.

rclpy/test/test_qos_event.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def test_publisher_event_create_destroy(self):
+        # Publisher event types
+        publisher = self.node.create_publisher(EmptyMsg, 'test_topic')
+
+        event_handle = _rclpy.rclpy_create_event(
+            QoSPublisherEventType.RCL_PUBLISHER_OFFERED_DEADLINE_MISSED,
+            publisher.publisher_handle)
+        self.assertIsNotNone(event_handle)
+        _rclpy.rclpy_destroy_entity(event_handle)
```
sloretz: Please wrap in `rclpy.handle.Handle` and call `Handle.destroy()`. This function will be removed shortly

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -0,0 +1,334 @@
...
+  rmw_liveliness_changed_status_t liveliness_changed;
+  // Publisher events
+  rmw_offered_deadline_missed_status_t offered_deadline_missed;
+  rmw_liveliness_lost_status_t liveliness_lost;
+} _qos_event_callback_data_t;
+
+typedef PyObject * (* _qos_event_data_filler_function)(_qos_event_callback_data_t *);
+
+bool
+_check_rcl_return(rcl_ret_t ret, const char * error_msg)
```
sloretz: nit, `static` on the functions which should only be usable by this file. I think this avoids possible name conflicts.

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -0,0 +1,334 @@
...
+  return true;
+}
+
+void
+_destroy_event_capsule(PyObject * pycapsule)
+{
+  rcl_event_t * event = (rcl_event_t *)PyCapsule_GetPointer(pycapsule, "rcl_event_t");
+  rcl_ret_t ret = rcl_event_fini(event);
+  PyMem_Free(event);
+  _check_rcl_return(ret, "Failed to fini 'rcl_event_t'");
```
sloretz: Please avoid raising exceptions in pycapsule destructors. If the capsule is garbage collected this results in tracebacks in random places. Instead I recommend following the pattern used by the subscription where it warns if an error occurs.

https://github.com/ros2/rclpy/blob/897123438ca05834e6ed943045d871c568c26448/rclpy/src/rclpy/_rclpy.c#L1671-L1695

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -0,0 +1,334 @@
...
+  if (!event) {
+    PyErr_Format(PyExc_MemoryError, "Failed to allocate memory for event");
+    return NULL;
+  }
+  *event = rcl_get_zero_initialized_event();
+  return event;
+}
+
+
+PyObject * _create_py_qos_event(const char * class_name, PyObject * args)
```
sloretz: Recommend commenting that this steals the reference to `args` instead of just borrowing it.

emersonknapp:
@sloretz thanks for the feeback, I am trying to address these as quickly as possible today because we want to get this through for the API freeze today. The part that will take the most time is using the new Handle type, it is a good change but definitely caught me unprepared.

1) Are you planning on merging any more changes that would affect this feature?
2) Will you be available to review this later this afternoon so that we can push it through once the dependency is merged?

rclpy/rclpy/qos_event.py
emersonknapp: Noted - maybe we will change it in a future revision, for now I will leave it as a Waitable

rclpy/src/rclpy/_rclpy.c
emersonknapp: Yes, I am refactoring the code that uses capsules to use `Handle`

rclpy/test/test_qos_event.py
emersonknapp: ok

rclpy/src/rclpy/_rclpy_qos_event.c
emersonknapp: Actually was hoping for your feedback on this pattern - I've `#include "./_rclpy_qos_event.c"` as source, instead of from a header, so the rcply C layer remains a single compilation unit, just separated out these two APIs and their helper functions by file only.

I will definitely mark all the helper functions as static though, don't want their names to leak anywhere.

rclpy/src/rclpy/_rclpy_qos_event.c
emersonknapp: Ok I will use this warning pattern.

sloretz:
> Are you planning on merging any more changes that would affect this feature?

No changes are planned. I would like to add API to the `Waitable` interface to make destroying them thread safe too, but I doubt I will be able to work on that before Dashing is released.

> Will you be available to review this later this afternoon so that we can push it through once the dependency is merged?

Yes

emersonknapp:
@sloretz I've updated, tests are passing locally. I will have to do a rebase when https://waffle.io/ros2/ros2/cards/5c9e5639f2c40c0066440aba is merged (because of  https://github.com/ros2/rclpy/pull/305) but it should affect very little of what's here - a few lines.

Will I need to do any further restructuring of this?

rclpy/src/rclpy/_rclpy_qos_event.c
sloretz: > Actually was hoping for your feedback on this pattern - I've #include "./_rclpy_qos_event.c" as source

Whatever builds is fine with me ;-) . Separate cpython extensions could make it possible to only load extensions if their features are being used, but we woudn't get that savings now because `rclpy.impl` loads all of them anyways.

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -0,0 +1,363 @@
...
+  * \return Python object from rclpy.qos_event containing callback data.
+  * \return NULL on failure.
+  */
+static PyObject *
+rclpy_take_event(PyObject * Py_UNUSED(self), PyObject * args)
+{
+  // Arguments
+  PyObject * pyevent = NULL;
+  PyObject * pyparent = NULL;
+  unsigned PY_LONG_LONG pyevent_type;
```
sloretz: nit, usually we only use `py...` for PyObjects.

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -0,0 +1,363 @@
...
+  PyObject * pyqos_event = NULL;
+  _qos_event_callback_data_t event_data;
+  _qos_event_data_filler_function event_filler = NULL;
+
+  if (!PyArg_ParseTuple(args, "OOK", &pyevent, &pyparent, &pyevent_type)) {
+    return NULL;
+  }
+
+  if (!_is_pycapsule_rcl_event(pyevent)) {
+    PyErr_Format(PyExc_TypeError, "Capsule was not a valid rcl_event_t");
```
sloretz: nit, would rather check if `_pycapsule_to_rcl_event` returned NULL instead of calling `PyCapsule_IsValid` since it's shorter and the capsule name only has to be given in one place. Though one thing to note if you're catking this exception is `PyCapsule_GetPointer` raises `ValueError` instead of `TypeError`.

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  ret = rcl_take_event(event, &event_data);
+  if (!_check_rcl_return(ret, "Failed to take event")) {
+    return NULL;
+  }
+
+  pyqos_event = event_filler(&event_data);
+  if (!pyqos_event) {
+    return NULL;
```
sloretz: Nit, no need to check. This can just `return pyqos_event`

emersonknapp:
@sloretz thanks - it seems like it'd be hard to run CI on this change until https://github.com/ros2/rmw/pull/171#issuecomment-488871259 this batch passes and is merged first, since these features depend on those new APIs from `rcl` and `rmw`, and I'm not sure I'd trust a CI run against the pre-merged PR branches.

Got my fingers crossedon that build.

If you need to get some other things under this that are small (saw the comment from the other one), it should be fine - I will be doing the one final rebase for this, and that conflict looks trivial to resolve.

emersonknapp:
While wringing my hands over the other ci - addressed the nitpicks

emersonknapp:
@thomas-moulard - please run the following CI job:
NOTE - it contains the changes for this PR plus its two new dependencies
https://github.com/ros2/rcl/pull/432
https://github.com/ros2/rmw_fastrtps/pull/276

 * Gist file: https://gist.githubusercontent.com/emersonknapp/67d93f9e90183987832715b5979640e8/raw/e75240e9745803f9f8fa09632f997cd07ce95f83/qos-4-callbacks.repos
* Additional BUILD flags: --packages-up-to rclpy
* Additional TEST flags: --packages-select rclpy
* CI Job: ci_launcher

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  event_filler = _get_qos_event_data_filler_function_for(pyparent, event_type);
+  if (!event_filler) {
+    return NULL;
+  }
+
+  ret = rcl_take_event(event, &event_data);
+  fprintf(stderr, "THE RET IS %d\n", ret);
```
sloretz: should remove before merging

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6952/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3206/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5703/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6775/)

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -0,0 +1,357 @@
...
+  if (!PyArg_ParseTuple(args, "OOK", &pyevent, &pyparent, &event_type)) {
+    return NULL;
+  }
+
+  event = _pycapsule_to_rcl_event(pyevent);
+  if (!event) {
+    return NULL;
+  }
+
+  event_filler = _get_qos_event_data_filler_function_for(pyparent, event_type);
```
sloretz: Mind looking at the windows warning? https://ci.ros2.org/job/ci_windows/6775/warnings43Result/new/

sloretz: That was fast, I assume this solves it 107baa6d7349f179a0cd789e11d7194c653eeb25

emersonknapp: Yeah - I just pushed a commit but unfortunately I don't have a windows dev env set up right now. It doesn't introduce warnings on linux (still building mac - it's a lot slower)

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -0,0 +1,357 @@
...
+_pycapsule_to_rcl_subscription(PyObject * pycapsule)
+{
+  return (rcl_subscription_t *)PyCapsule_GetPointer(pycapsule, "rclpy_subscription_t");
+}
+
+static
+rcl_publisher_t *
+_pycapsule_to_rcl_publisher(PyObject * pycapsule)
+{
+  return (rcl_publisher_t *)PyCapsule_GetPointer(pycapsule, "rclpy_publisher_t");
```
sloretz: Just noting that this works because the first member of the `rclpy_publisher_t` struct is an `rcl_publisher_t`. It would be more explicit to use a temporary and return the address of the publisher in that struct instead.

rclpy/src/rclpy/_rclpy_qos_event.c
emersonknapp: @sloretz mac build looks good, this change also does not introduce any warnings there - should we just kick off another build? i've updated the gist to use the master version of everything except this, since the dependencies merged

https://gist.githubusercontent.com/emersonknapp/67d93f9e90183987832715b5979640e8/raw/7eb11515e048841d603d35ce0a5e6fecfdf9296f/qos-4-callbacks.repos

rclpy/src/rclpy/_rclpy_qos_event.c
emersonknapp: Noted - I was aware of that - you're right that it would be more explicit to do it the other way. I'd like to do a bit of refactoring after this, but for this change let's leave it the way it is.

sloretz: Since this needs another round of CI to check the windows warning, mind updating this one before that?

emersonknapp: Ok - I am just running the tests locally first - and I have updated the usage from the other PR as well - should be just a minute

emersonknapp: @sloretz OK - i've updated the code and it's looking good locally - the same repos list gist is accurate https://gist.githubusercontent.com/emersonknapp/67d93f9e90183987832715b5979640e8/raw/7eb11515e048841d603d35ce0a5e6fecfdf9296f/qos-4-callbacks.repos

sloretz:
CI (testing just rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6967/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3218/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5715/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6789/)

sloretz:
@emersonknapp Looks like `test_qos_event.py` is crashing on OSX https://ci.ros2.org/job/ci_osx/5715/testReport/junit/(root)/projectroot/test_qos_event/ . Mind taking a look?

rutvih:
> @emersonknapp Looks like `test_qos_event.py` is crashing on OSX https://ci.ros2.org/job/ci_osx/5715/testReport/junit/(root)/projectroot/test_qos_event/ . Mind taking a look?

@sloretz @emersonknapp is on vacation today. @mm318 from our team will take a look at this and address comments.

emersonknapp:
Well - it seems to be that trying on OSX to `PyArg_ParseTuple("OOk")` on an `IntEnum` type from Python overflows 32 bits into the second `O` from the `k`. That makes it much more clear why all the other enum-based APIs use `PY_LONG_LONG` instead of native int types directly. (thanks @nburek for the overflow insight). Pushed a new commit to follow that pattern. @thomas-moulard or @sloretz can you please re-trigger CI with the latest version?

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7005/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3250/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5745/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6825/)

emersonknapp:
Gist rotted due to upstream changes and deleted branches. Will run again

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7008/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3253/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5748/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6828/)

emersonknapp:
@thomas-moulard - recreated a fresh gist since the update seems to have not worked for some reason

Gist file: https://gist.githubusercontent.com/emersonknapp/ec686aca278eb8896b1d3415a19ec169/raw/5a135ec8266374e6b9e7ae3a2e54b4d0a7fa67b9/ros2.repos
Additional BUILD flags: --packages-up-to rclpy
Additional TEST flags: --packages-select rclpy
CI Job: ci_launcher

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7009/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3254/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5749/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6829/)

emersonknapp:
@sloretz @wjwwood Can we merge this PR? The fix was the right one so we're all verified now - this is the final API for the Dashing QoS features.

wjwwood:
Yes, but I want to run a full CI (not stopping at `rclpy`) on at least Linux to make sure nothing else is broken. As you can imagine, breaking CI right now is high cost.

This CI builds everything, but without Connext, so it should be a bit faster than normal, and tests `--packages-above rclpy`:

[Build Status](https://ci.ros2.org/job/ci_linux/7010/)

emersonknapp:
@wjwwood Looks like the only failure is the cli params known issue, is that correct?

wjwwood:
That's right, but I'm going to hold merging this until tomorrow, since we have some other actual API changes in the pipe. This only adds to the API as far as I can tell.

emersonknapp:
That's right, it just adds optical kwargs to the create_ functions

emersonknapp:
@wjwwood ok to merge now? Looks like the API freeze board has been closed out

wjwwood:
lgtm, but a lot has been merged since the last CI, so I'm running Linux again:

[Build Status](https://ci.ros2.org/job/ci_linux/7058/)

It's possible this needs to be rebased or merged with master.

emersonknapp:
@wjwwood Ok - rebased - built and tested locally on linux, looks good, can you run that again?

wjwwood:
Sure thing, thanks for the rebase: [Build Status](https://ci.ros2.org/job/ci_linux/7059/)

suab321321:
@emersonknapp could you please brief me what did you do on the backend side(`.c` files) to achieve?
I m trying to implement this rclnodejs issue https://github.com/RobotWebTools/rclnodejs/issues/492.
:)

emersonknapp:
@suab321321 do you have a specific question about the code? Most of it is just data type conversions between `rcl` and the `Python` layer. The functionality part of it is exposing the new Waitable type `rcl_event_t` and adding the polling for events.

You can see this PR https://github.com/ros2/rclcpp/pull/695 for how this feature was added to `rclcpp`, for more context on the pattern.

suab321321:
@emersonknapp thanks for helping me :).If you could tell me about the hiarchey of this implementation from its core to python side user side api,example how `_rclpy.c` calls `_rclpy_qos_event.c` like this a brief word about you did, will be very much helpful :)

emersonknapp:
https://github.com/ros2/rclpy/pull/316/files?file-filters%5B%5D=.c#diff-80dda03110b5be606b823d6b0558b5c2R40 you can see on this line that the source file is included directly, they are only split for readability.

suab321321:
@emersonknapp why `qos events` are `waitables type`

suab321321:
```
class QoSEventHandler(Waitable):
    """Waitable type to handle QoS events."""

    def __init__(
        self,
        *,
        callback_group: CallbackGroup,
        callback: Callable,
        event_type: IntEnum,
        parent_handle: Handle,
    ):
        # Waitable init adds self to callback_group
        super().__init__(callback_group)
        self.event_type = event_type
        self.callback = callback

        self._parent_handle = parent_handle
        with parent_handle as parent_capsule:
            event_capsule = _rclpy.rclpy_create_event(event_type, parent_capsule)
        self._event_handle = Handle(event_capsule)
        self._event_handle.requires(self._parent_handle)
        self._ready_to_take_data = False
        self._event_index = None
```
file : `qos_event.py`
@emersonknapp could you please help me understand this part of code

suab321321:
@emersonknapp and one more thing :) what is the difference between a `callback` and `callback_group`.

emersonknapp:
Waitables are used for asynchronous communication. You "wait" for things to happen on it. If you familiarize yourself with the `rclpy` code, you will see that Action Servers and Clients are also Waitables. You add them to a "wait set" that is checked during "spin" to see if the event has occurred. 

For your second question, please look at https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/handle.py#L103 - `requires` on a Python Capsule Handle means that the Handle does not work without the other entity. In this case, the `parent` is the `subscription` or `publisher` that the QoSEventHandler is attached to. If that parent entity is deleted, this event handler is no longer valid.

suab321321:
@emersonknapp I just want to know that in `_rclpy.c` for a timer callback to get executed there is method `rcl_timer_call` taking `rcl_timer_t*` as a paramter,so is there any method like this which is used for calling the qos callback? :)

emersonknapp:
Timers, Subscriptions, Guards, Clients, and Services use an old way of asynchronous communication, they all have special lists that they are held in (https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/executors.py#L589)

QoS Events use the generic Waitable pattern which can hold many types of objects, not just QoS Events. It is checked right before that special cases loop - (https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/executors.py#L579) -
you can see that it checks on every `spin` loop, if an event `is_ready` (has an event happened?), then it uses `take_data` to grab the event data if one is ready. [Python side](https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/qos_event.py#L123) and [C side](https://github.com/ros2/rclpy/blob/master/rclpy/src/rclpy/_rclpy_qos_event.c#L311)

Issue #255, opened 2018-12-19
executor.spin_once crashes after destroy_subscription called in separate thread


brawner:
This is a follow up from https://github.com/ros-visualization/rqt/issues/182. Destroying subscriptions is not thread safe. Per this comment https://github.com/ros-visualization/rqt/issues/181#issuecomment-446820387 I suppose rmw is not designed to be threadsafe. It would be very helpful for overall stability if there was a thread management solution in the client libraries.
## Bug report

**Required Info:**

- Operating System:
  - MacOS 10.13.6
- Installation type:
  - Source
- Version or commit hash:
  - 0.6.1 91f7782619b1ccd9c812012ff85f6336186baab9
- DDS implementation:
  - Default (Fast-RTPS)

#### Steps to reproduce issue
Minimal example here https://github.com/brawner/crashing_subscription_destroyer
```
> python3 publisher.py &
> python3 subscriber.py
```

Change the code in ```subscriber.py``` to see an example that does not crash
```
if __name__ == '__main__':
    dont_segfault()
    #segfault()
```

#### Expected behavior
Node continues to spin and subscription is ended

#### Actual behavior
Locking/or thread management is required to safely destroy a subscription

### Stacktrace

```
    Thread 6 Crashed:: RclpySpinner
0   librmw_fastrtps_shared_cpp.dylib	0x0000000104de81ae SubListener::hasData() + 126
1   librmw_fastrtps_shared_cpp.dylib	0x0000000104de7f43 check_wait_set_for_data(rmw_subscriptions_t const*, rmw_guard_conditions_t const*, rmw_services_t const*, rmw_clients_t const*) + 115
2   librmw_fastrtps_shared_cpp.dylib	0x0000000104de97c7 rmw_fastrtps_shared_cpp::__rmw_wait(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_wait_set_t*, rmw_time_t const*)::$_0::operator()() const + 39
[...]

PR #318, opened 2019-04-23
Make destroy_subscription thread safe

sloretz:
Resolves ros2/rclpy#255

This makes `node.destroy_subscription` thread safe by wrapping the pycapsule with a class that keeps track of who is using it. The subscription is destroyed when it is no longer being used.

## Potentially breaking changes

* `_rclpy.rclpy_create_subscription()` returns just the pycapsule instead of a tuple. However, I would be surprised if any downstream projects are using this method.
* `_rclpy.rclpy_destroy_node_entity()` no longer destroys subscriptions. Instead `rclpy_pycapsule_destroy()` does the work. Again I would be surprised if any downstream code is using this.
* The class `Subscription` no longer has attributes `subscription_handle` and `subscription_pointer`. Instead it just has `handle`. I would expect those attributes to have been for internal use only.

## Next steps

In order to keep PRs small, I would like this PR to get review and be merged after addressing feedback. Afterwards I'll create a new PR to fix the same thread safety issues for the other destroy methods:

* destroy_publisher
* destroy_client
* destroy_service
* destroy_timer
* destroy_guard_condition

Finally I want to defer making `destroy_node` thread safe, and instead create an issue for it. It requires more thought since it can't be destroyed until every entity created from is no longer being used and has been destroyed.


sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6771/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3072/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5565/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6598/)

rclpy/rclpy/handle.py
```diff
@@ -0,0 +1,127 @@
...
+        return self.__capsule
+
+    def _return_capsule(self):
+        """
+        Return the pycapsule that was previously gotten with :meth:`_get_capsule`.
+
+        :return: None
+        """
+        with self.__lock:
+            self.__use_count -= 1
```
wjwwood: Can `__use_count` be `0` at this point?

rclpy/src/rclpy/_rclpy_pycapsule.c
```diff
@@ -0,0 +1,161 @@
...
+  const char * name = PyCapsule_GetName(pycapsule);
+
+  if (PyErr_Occurred()) {
+    return NULL;
+  }
+
+  if (NULL == name) {
+    Py_RETURN_NONE;
+  }
+  return PyUnicode_FromString(name);
```
wjwwood: I assume returning `NULL` here is the right thing to do if `FromString` fails?

rclpy/src/rclpy/_rclpy_pycapsule.c
```diff
@@ -0,0 +1,161 @@
...
+    return NULL;
+  }
+
+  void * pointer = PyCapsule_GetPointer(pycapsule, name);
+
+  if (PyErr_Occurred()) {
+    return NULL;
+  }
+
+  return PyLong_FromUnsignedLongLong((uint64_t)pointer);
```
wjwwood: This assumes the size of a pointer, which will probably "just work", but maybe there's a better way to handle this?

rclpy/src/rclpy/_rclpy_pycapsule.c
```diff
@@ -0,0 +1,161 @@
...
+  {
+    "rclpy_pycapsule_destroy", rclpy_pycapsule_destroy,
+    METH_VARARGS,
+    "Destroy a pycapsule and clear its destructor"
+  },
+  {NULL, NULL, 0, NULL}  /* sentinel */
+};
+
+PyDoc_STRVAR(rclpy_pycapsule__doc__,
+  "RCLPY module for working with PyCapsule objects.");
```
wjwwood: nitpick: `rclpy`?

rclpy/src/rclpy/_rclpy_pycapsule.c
sloretz: Correct,  [`PyUnicode_FromString`](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_FromString) returns `NULL` if it sets an exception ([second paragraph](https://docs.python.org/3/c-api/intro.html#exceptions)), and this function is required to do the same: https://docs.python.org/3/c-api/structures.html#c.PyCFunction

rclpy/src/rclpy/_rclpy_pycapsule.c
sloretz: Added `_Static_assert` in 5e40a23. Hopefully visual studio supports this, but if not I think I can `#ifdef` and use [_STATIC_ASSERT](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/static-assert-macro?view=vs-2019)

rclpy/src/rclpy/_rclpy_pycapsule.c
sloretz: I'M JUST REALLY EXCITED ABOUT RCLPY

-> rclpy 5f72443

rclpy/rclpy/handle.py
sloretz: Only if there's a bug. Added assert in 2870fee

sloretz:
CI (testing just rclpy, mostly to check that `_Static_assert` is supported on all platforms)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6772/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3073/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5567/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6600/)

rclpy/src/rclpy/_rclpy_pycapsule.c
sloretz: Did you mean better way to represent the pointer address in python, or better way to work with entities than rely on the pointer address?

If the latter, it would be nice if the rclpy wait set API said which entity was ready without having to look it up the pycapsule's address.

wjwwood: Static assert is good.

Both I guess? I was thinking like using `void *` would be better, and having like a `PyPointer` like thing (not sure that exists) would be better, but maybe that's sort of what a capsule is?

sloretz: Yeah [`PyCapsule`](https://docs.python.org/3/c-api/capsule.html) is the only pointer thing I know of, but it doesn't expose the address it holds to python code.

sloretz: Aha, there is better. The builtin function `id` returns the address in memory of an object. [It uses `PyLong_FromVoidPtr()`.](https://github.com/python/cpython/blob/d246a6766b9d8cc625112906299c4cb019944300/Python/bltinmodule.c#L1206-L1211)

sloretz: 69399f5

sloretz:
CI with 69399f5 testing packages above rclpy to make sure the new commits didn't break anything
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6775/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3076/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5570/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6603/)

PR #319, opened 2019-04-24
Thread safe node.destroy_*

sloretz:
Follow up to ros2/rclpy#318 and ros2/rclpy#330. This PR applies the changes done to `Subscription` to the rest of the entities on the node so they can all be destroyed safely.

- [x] destroy_publisher
- [x] destroy_client
- [x] destroy_service
- [x] destroy_timer
  - [x] Clock used by timer safely destroyed
- [x] destroy_guard_condition

Not done in this PR. These might require a change to the waitable API.

* action server
* action client

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -328,8 +328,8 @@ class Executor:
                 future.set_result(response)
 
     def _take_service(self, srv):
-        request_and_header = _rclpy.rclpy_take_request(
-            srv.service_handle, srv.srv_type.Request)
+        with srv.handle as capsule:
+            request_and_header = _rclpy.rclpy_take_request(capsule, srv.srv_type.Request)
         return request_and_header
 
     async def _execute_service(self, srv, request_and_header):
@@ -491,6 +491,13 @@ class Executor:
                     except InvalidHandle:
                         entity_count.num_clients -= 1
 
+                service_capsules = []
+                for srv in services:
+                    try:
+                        service_capsules.append(context_stack.enter_context(srv.handle))
+                    except InvalidHandle:
+                        entity_count.num_services -= 1
+
                 _rclpy.rclpy_wait_set_init(
                     wait_set,
                     entity_count.num_subscriptions,
@@ -502,7 +509,6 @@ class Executor:
 
                 entities = {
                     'guard_condition': (guards, 'guard_handle'),
-                    'service': (services, 'service_handle'),
                     'timer': (timers, 'timer_handle'),
                 }
                 _rclpy.rclpy_wait_set_clear_entities(wait_set)
@@ -515,6 +521,8 @@ class Executor:
                     _rclpy.rclpy_wait_set_add_entity('subscription', wait_set, sub_capsule)
                 for cli_capsule in client_capsules:
                     _rclpy.rclpy_wait_set_add_entity('client', wait_set, cli_capsule)
+                for srv_capsule in service_capsules:
+                    _rclpy.rclpy_wait_set_add_entity('service', wait_set, srv_capsule)
                 for waitable in waitables:
                     waitable.add_to_wait_set(wait_set)
 
@@ -588,7 +596,7 @@ class Executor:
                             yield handler, client, node
 
                 for srv in node.services:
-                    if srv.service_pointer in services_ready:
+                    if srv.handle.pointer in services_ready:
                         if srv.callback_group.can_execute(srv):
                             handler = self._make_handler(
                                 srv, node, self._take_service, self._execute_service)
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -312,7 +312,8 @@ class Executor:
             await await_or_execute(sub.callback, msg)
 
     def _take_client(self, client):
-        return _rclpy.rclpy_take_response(client.client_handle, client.srv_type.Response)
+        with client.handle as capsule:
+            return _rclpy.rclpy_take_response(capsule, client.srv_type.Response)
 
     async def _execute_client(self, client, seq_and_response):
         sequence, response = seq_and_response
@@ -483,6 +484,13 @@ class Executor:
                     except InvalidHandle:
                         entity_count.num_subscriptions -= 1
 
+                client_capsules = []
+                for cli in clients:
+                    try:
+                        client_capsules.append(context_stack.enter_context(cli.handle))
+                    except InvalidHandle:
+                        entity_count.num_clients -= 1
+
                 _rclpy.rclpy_wait_set_init(
                     wait_set,
                     entity_count.num_subscriptions,
@@ -494,7 +502,6 @@ class Executor:
 
                 entities = {
                     'guard_condition': (guards, 'guard_handle'),
-                    'client': (clients, 'client_handle'),
                     'service': (services, 'service_handle'),
                     'timer': (timers, 'timer_handle'),
                 }
@@ -506,6 +513,8 @@ class Executor:
                         )
                 for sub_capsule in sub_capsules:
                     _rclpy.rclpy_wait_set_add_entity('subscription', wait_set, sub_capsule)
+                for cli_capsule in client_capsules:
+                    _rclpy.rclpy_wait_set_add_entity('client', wait_set, cli_capsule)
                 for waitable in waitables:
                     waitable.add_to_wait_set(wait_set)
 
@@ -571,7 +580,7 @@ class Executor:
                             yield handler, gc, node
 
                 for client in node.clients:
-                    if client.client_pointer in clients_ready:
+                    if client.handle.pointer in clients_ready:
                         if client.callback_group.can_execute(client):
                             handler = self._make_handler(
                                 client, node, self._take_client, self._execute_client)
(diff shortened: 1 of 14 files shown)


sloretz:
I think I have to include `destroy_node` in this PR. This approach can change the order that entities get shut down.

```python3
# In another thread an executor is spinning
# srv destruction delayed because executor is waiting on service
node.destroy_service(srv)

# node is destroyed instantly; service still not destroyed because it's being used by executor
node.destroy_node()

# When executor shuts down the service is no longer being used, so it's destroyed
# since the node was destroyed first the service will segfault in rmw_fastrtps
executor.shutdown()
```

I'm thinking of chaining handles so that `node.destroy_node()` destroy's all node entities before allowing itself to be destroyed.

rclpy/rclpy/executors.py
```diff
@@ -151,9 +151,8 @@ def __init__(self, *, context: Context = None) -> None:
         self._tasks: List[Tuple[Task, Optional[WaitableEntityType], Optional[Node]]] = []
         self._tasks_lock = Lock()
         # This is triggered when wait_for_ready_callbacks should rebuild the wait list
-        gc, gc_handle = _rclpy.rclpy_create_guard_condition(self._context.handle)
-        self._guard_condition = gc
-        self._guard_condition_handle = gc_handle
+        self._guard = GuardCondition(
+            callback=None, callback_group=None, context=self._context)
```
sloretz: Reusing `GuardCondition` to reuse executor code that uses  `guard_condition.handle`

sloretz:
Ready for review.

CI (testing rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6847/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3139/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5630/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6668/)

rclpy/src/rclpy/_rclpy.c
```diff
@@ -2221,141 +2294,32 @@ rclpy_service_server_is_available(PyObject * Py_UNUSED(self), PyObject * args)
...
-  Py_RETURN_NONE;
-}
-
 /// Destructor for a clock
-void
+static void
 _rclpy_destroy_clock(PyObject * pycapsule)
 {
   rcl_clock_t * clock = (rcl_clock_t *)PyCapsule_GetPointer(pycapsule, "rcl_clock_t");
   if (NULL == clock) {
```
clalancette: I know you didn't add this, but for consistency it would probably be better to use `!clock` like all of the rest of the PyCapsulte_GetPointer checks.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -2502,13 +2466,13 @@ rclpy_wait_set_add_entity(PyObject * Py_UNUSED(self), PyObject * args)
       (rclpy_subscription_t *)PyCapsule_GetPointer(pyentity, "rclpy_subscription_t");
     ret = rcl_wait_set_add_subscription(wait_set, &(sub->subscription), &index);
   } else if (0 == strcmp(entity_type, "client")) {
-    rcl_client_t * client =
-      (rcl_client_t *)PyCapsule_GetPointer(pyentity, "rcl_client_t");
-    ret = rcl_wait_set_add_client(wait_set, client, &index);
+    rclpy_client_t * client =
+      (rclpy_client_t *)PyCapsule_GetPointer(pyentity, "rclpy_client_t");
```
clalancette: All other uses of `PyCapsule_GetPointer` check for a NULL pointer after it, but this one and the one below seem to elide it.  I know that isn't new to this PR, but we should probably be more careful here.

sloretz: Check that pointers are not `NULL` before using them in 5f7b7c8

rclpy/rclpy/node.py
```diff
@@ -648,27 +651,12 @@ def destroy_node(self) -> bool:
...
-            while self.timers:
-                tmr = self.timers.pop()
-                _rclpy.rclpy_destroy_entity(tmr.timer_handle)
-                # TODO(sloretz) Store clocks on node and destroy them separately
-                _rclpy.rclpy_destroy_entity(tmr.clock._clock_handle)
-            while self.guards:
-                gc = self.guards.pop()
-                _rclpy.rclpy_destroy_entity(gc.guard_handle)
-
+        self.publishers = ()
```
clalancette: It's a little strange that `self.publishers` (and the rest below) are declared as `List` in the constructor, but are then set to empty tuples here.  I would have guessed mypy would complain about that, but maybe not.  Is this intended, or just an oversight?

rclpy/rclpy/node.py
```diff
@@ -554,12 +567,10 @@ def destroy_publisher(self, publisher: Publisher) -> bool:
...
         """
-        for pub in self.publishers:
-            if pub.publisher_handle == publisher.publisher_handle:
-                with self.handle as capsule:
-                    _rclpy.rclpy_destroy_node_entity(pub.publisher_handle, capsule)
-                self.publishers.remove(pub)
-                return True
+        if publisher in self.publishers:
+            self.publishers.remove(publisher)
+            publisher.destroy()
```
clalancette: For all of these destroy calls, do we need to check for the `InvalidHandle` exception?  Is that a case that can happen?

rclpy/rclpy/executors.py
```diff
@@ -544,14 +564,15 @@ def _wait_for_ready_callbacks(
...
-                    if tmr.timer_pointer in timers_ready:
-                        # Check that a timer is ready to workaround rcl issue with cancelled timers
-                        if _rclpy.rclpy_is_timer_ready(tmr.timer_handle):
-                            if tmr.callback_group.can_execute(tmr):
-                                handler = self._make_handler(
-                                    tmr, node, self._take_timer, self._execute_timer)
-                                yielded_work = True
-                                yield handler, tmr, node
+                    if tmr.handle.pointer in timers_ready:
+                        with tmr.handle as capsule:
```
clalancette: It almost seems like this line and the one above it should be swapped.  That is, we effectively want to "protect" the timer as a capsule, and *then* dereference it to get the pointer to ensure that the handle hasn't gone away in the meantime.  On the  other hand, this may not matter.  Thoughts?

rclpy/src/rclpy/_rclpy.c
sloretz: It looks like `_rclpy` is inconsistent about `NULL == foo` vs `!foo` after `PyCapsule_GetPointer`.
https://github.com/ros2/rclpy/blob/1b4728b8bc75d5e8ae765e938ed2513d24f7e45d/rclpy/src/rclpy/_rclpy.c#L655-L656
https://github.com/ros2/rclpy/blob/1b4728b8bc75d5e8ae765e938ed2513d24f7e45d/rclpy/src/rclpy/_rclpy.c#L4457-L4458

 After jumping through the calls it feels like `!foo` is more popular, but there are a lot of `NULL== foo`. If it's alright, I'd like to open an issue instead changing them in this PR.

sloretz: Opened #331

rclpy/rclpy/executors.py
```diff
@@ -571,15 +592,15 @@ def _wait_for_ready_callbacks(
                             yield handler, gc, node
 
                 for client in node.clients:
-                    if client.client_pointer in clients_ready:
+                    if client.handle.pointer in clients_ready:
```
clalancette: Do we not need to get the capsules for `client` here and `server` below?

rclpy/rclpy/node.py
sloretz: It is. Maybe this should catch it and return `False`?

rclpy/rclpy/executors.py
sloretz: There's no need to protect a call to `handle.pointer`. It is returning a cached integer. Even if it was calling `PyCapsule_GetPointer` it would still be safe to call because the address held by a pycapsule is immutable.

rclpy/rclpy/executors.py
sloretz: The capsules are only needed when they're used. The timer had to get its capsule to call `rclpy_is_timer_ready` above. The client and service capsules are gotten in `executor._take_client` and `executor._take_service`.

rclpy/rclpy/node.py
clalancette: Yeah, that seems reasonable.

sloretz: `destroy_<entity>()` catch `InvalidHandle` and return `False` in e630404

rclpy/rclpy/node.py
sloretz: Calls `.clear()` to remove the elements in the lists instead of setting them to an empty tuple in 4bcfe99

sloretz:
CI (testing `rclpy`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6864/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3154/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5641/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6684/)

sloretz:
> My only remaining question is whether we should open an issue for reducing the complexity here by introducing rmw_shutdown_but_not_fini for a future enhancement. Thoughts on that?

I've gone back and forth between writing the issue on `rmw` vs making it a higher level feature of `rcl`. I opened ros2/rcl#429, but the solution seems a lot more complicated than I first thought. I put it in the backlog since it feels like it needs more thought.

PR #323, opened 2019-04-26
Remove __eq__ and __hash__ from Subscription

### LINKED ISSUE #322: Unexpected new stack trace in rclpy
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04 - running nightly docker image
- Installation type:
  - binaries
- Version or commit hash:
  - Using docker image sha256:14cb0769c64b280b5c609a9bef7f52393d4f306197bb48201399865486f4c6f1 for osrf/ros2:nightly
- DDS implementation:
  - 
  - rclpy

#### Steps to reproduce issue

One of my nightly CI jobs for apex_launchtest (soon to be launch_testing) started failing with an rclpy stack trace.  See issue here: https://github.com/ApexAI/apex_rostest/issues/42.

I'm attempting to put together a simpler repro that doesn't use launch_testing

#### Actual behavior
This stack trace when calling rclpy.spin:
```
======================================================================
FAIL: test_talker_transmits (talker_listener.test.py.TestTalkerListenerLink)
----------------------------------------------------------------------
  File "/apex_rostest/apex_launchtest_ros/examples/talker_listener.test.py", line 126, in test_talker_transmits
    self.spin_rclpy(1.0)
  File "/apex_rostest/apex_launchtest_ros/examples/talker_listener.test.py", line 109, in spin_rclpy
    executor.spin_once(timeout_sec=timeout_sec)
  File "/opt/ros/crystal/lib/python3.6/site-packages/rclpy/executors.py", line 631, in spin_once
  File "/opt/ros/crystal/lib/python3.6/site-packages/rclpy/executors.py", line 617, in wait_for_ready_callbacks
    return next(self._cb_iter)
  File "/opt/ros/crystal/lib/python3.6/site-packages/rclpy/executors.py", line 558, in _wait_for_ready_callbacks
    if sub.callback_group.can_execute(sub):
  File "/opt/ros/crystal/lib/python3.6/site-packages/rclpy/callback_groups.py", line 103, in can_execute
    assert weakref.ref(entity) in self.entities
AssertionError

----------------------------------------------------------------------
```

#### Additional information
This started failing in a nightly CI job, so I think it's been triggered by a recent change.  I've never done a deep dive into rclpy callback_groups though, so I'm not sure what's going on in there quite yet.  It's possible I'm doing something wrong.

sloretz:
Fixes #322

**TL;DR** the default `__eq__` and `__hash__` work just fine.

The Subscription class was using the address held by the pycapsule as the hash. However, that address is `free`'d when `Subscription.destroy()` is called. This makes it possible for a new subscription to be created using the same address, and therefore having the same hash as a `destroy()`'d subscription.

This is a problem in the unit tests that reuse the same node for multiple tests. After the subscription is destroyed in one test, a new test can create a subscription that has the same memory address. Callback groups keep a `set()` of weak references to entities (including subscriptions) that are in them. This prevents the new subscription from being added to the node's default callback group, because a weak reference with that hash already exists. The weak reference to the `destroy()`'d entity stays in the callback group's `set()` until it is garbage collected.

After the garbage collector runs the callback group's weakref to the old subscription is deleted. This leads to an assertion error in `callback_group.can_execute` because there is no weak reference with the same hash as the newly created subscription.

### DIFF
--- a/rclpy/rclpy/subscription.py
+++ b/rclpy/rclpy/subscription.py
@@ -72,9 +72,3 @@ class Subscription:
 
     def destroy(self):
         self.handle.destroy()
-
-    def __eq__(self, other):
-        return self.handle == other.handle
-
-    def __hash__(self):
-        return self.handle.pointer


sloretz:
CI (testing just rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6798/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3094/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5589/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6623/)

dirk-thomas:
Are there any cases where we create two instances of `Subscription` with the same subscription handle? If not, I agree these functions can be removed. If yes, we need to make sure that we don't rely on them being equal and instead compare the handle property explicitly.

sloretz:
> Are there any cases where we create two instances of Subscription with the same subscription handle?

@dirk-thomas No cases that I'm aware of.

rclpy/rclpy/subscription.py
```diff
@@ -72,9 +72,3 @@ def handle(self):
 
     def destroy(self):
         self.handle.destroy()
```
dirk-thomas: Should we set the handle to `None` after it has been destroyed?

sloretz: It's meant to be kept around. Trying to use the handle after this raises `InvalidHandle`.

PR #325, opened 2019-04-26
Parameter API enhancements

### LINKED ISSUE #321: Enhance parameters API - declarations and descriptors
The purpose of this issue is to describe proposed changes to `rclpy` considering changes in `rclcpp` regarding parameter handling: https://github.com/ros2/rclcpp/pull/495, and gather feedback if necessary.

## Changes to rclcpp: summary
Considering the changes in the PR mentioned above, this is what the API in `rclcpp` looks like for parameter handling (`Node.hpp`):

- `declare_parameter`: declares and initialize a parameter, returns the effective value.
  - Has overload to specify type.
- `declare_parameters`: declares and initializes several parameters with the same namespace and type.
  - Has overload with map + key / value pairs.
- `undeclare_parameter`: undeclares previously declared parameter.
- `has_parameter`: returns true if a parameter is declared. Doesn't throw exceptions if not.
- `set_parameter`: sets the given parameter and then returns result of the set action.
- `set_parameters`: sets multiple parameters, one at the time.
- `set_parameters_atomically`: sets multiple parameters at once.
- `get_parameter`: returns parameter.
  - Has overload that doesn't throw; returns false when it's not possible to get the parameter.
- `get_parameter_or`: gets parameter, or returns a given default.
- `get_parameters`: gets multiple parameters at once.
  -  Has overload for prefixes.
- `describe_parameter`: returns parameter descriptor.
- `describe_parameters`: returns vector of descriptors.
- `get_parameter_types`: returns vector of parameter types.
- `list_parameters`: returns list of parameter with any of given prefixes
- `set_on_parameters_set_callback`: registers callback anytime a parameter is about to be changed.

As a general rule, any method that gets or sets undeclared parameters might throw an exception if undeclared parameters are not allowed, or return a sane default if undeclared parameters are allowed. Methods that don't throw exceptions are explicity documented as such. The list above doesn't consider deprecated functions.

## Current API in rclpy
The current API in `rclpy` is briefer:
- `get_parameter`.
- `set_parameters`.
- `set_parameters_atomically`.
- `set_parameters_callback`.

## Proposal

Changes to `rclpy` should:
- Allow the user to declare and undeclare parameters.
- Allow getting and setting undeclared parameters, throwing exceptions in the case the node doesn't allow it.
- Handle parameter descriptors.

### Methods to add

```python

def declare_parameter(parameter: Parameter) -> Parameter:
    """Declares parameter with a given descriptor; returns effective parameter value assigned to the parameter."""
 
def declare_parameters(namespace: str, parameters: List[Parameter])) -> List[Parameter]:
    """ 
    Same as declare_parameter, for multiple parameters. Returns list of parameters containing the effective values assigned to each of them.
    """

def undeclare_parameter(name: str):
    """Undeclares a given parameter"""

def describe_parameter(name: str) -> ParameterDescriptor:
    """Gets parameter descriptor for a given parameter"""
    
def describe_parameters(name: List[str]) -> List[ParameterDescriptor]:
    """Same as describe_parameter, for multiple parameters at once."""

def has_parameter(name: str) -> bool:
    """Returns true if a parameter has been declared; false otherwise (doesn't throw)."""

def get_parameter_or(name: str, alternative_value: Parameter) -> Parameter:
    """Returns a parameter if declared or the alternative value otherwise.
    Option B: return Tuple(bool, Parameter), with the boolean indicating whether the parameter was declared or not.
    """
```

## Methods to modify

These methods will have the same API, but their behavior will be modified:
- `set_parameters(List[Parameter]) -> List[SetParametersResult]`
  - Will throw if parameter is not declared and undeclared parameters are not allowed.
  - If one fails beacuse it was not declared, stop there. If the callback fails, that will be reflected in the returned list as in the current implementation.
- `set_parameters_atomically(List[Parameter]) -> SetParametersResult`
  - Will throw if parameter is not declared and undeclared parameters are not allowed.
  - If one fails, none is set (i.e. all have to be declared for this to succeed).
- `get_parameter(name: str) -> Parameter`
  - Will throw if not declared and undeclared not allowed. Returns PARAMETER_NOT_SET if allowed as in the current implementation.
- `get_parameters(names: List[str]) -> List[Parameter]`
  - Will throw if a parameter is not declared and undeclared are not allowed. Returns PARAMETER_NOT_SET if allowed as in the current implementation in the corresponding position of the list.

### Additional modifications
These changes shouldn't modify the API either if default values are used:
- `allow_undeclared_parameters` flag to shall be added to the node constructor, set to `True` by default for backwards compatibility.
- `Parameter` class should support custom `ParameterDescriptor`s other than the default one.

### To be verified
Regarding callbacks when parameters are set, `rclpy` has the method `set_parameters_callback`. In `rclcpp`, `set_on_parameters_set_callback` was introduced; these callbacks can reject a parameter from being set.
At first sight it looks like the callbacks in `rclpy` can do the same; I need to double check if this is the case.

jubeira:
Closes #321; refer to the issue for details.

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -111,7 +111,8 @@ def create_node(
     namespace: str = None,
     use_global_arguments: bool = True,
     start_parameter_services: bool = True,
-    initial_parameters: List[Parameter] = None
+    initial_parameters: List[Parameter] = None,
+    allow_undeclared_parameters: bool = False
 ) -> 'Node':
     """
     Create an instance of :class:`.Node`.
@@ -126,6 +127,7 @@ def create_node(
         arguments.
     :param start_parameter_services: ``False`` if the node should not create parameter services.
     :param initial_parameters: A list of :class:`.Parameter` to be set during node creation.
+    :param allow_undeclared_parameters: True if undeclared parameters are allowed, False otherwise.
     :return: An instance of the newly created node.
     """
     # imported locally to avoid loading extensions on module import
@@ -134,7 +136,8 @@ def create_node(
         node_name, context=context, cli_args=cli_args, namespace=namespace,
         use_global_arguments=use_global_arguments,
         start_parameter_services=start_parameter_services,
-        initial_parameters=initial_parameters)
+        initial_parameters=initial_parameters,
+        allow_undeclared_parameters=allow_undeclared_parameters)
 
 
 def spin_once(node: 'Node', *, executor: 'Executor' = None, timeout_sec: float = None) -> None:
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -112,7 +112,8 @@ def create_node(
     use_global_arguments: bool = True,
     start_parameter_services: bool = True,
     initial_parameters: List[Parameter] = None,
-    allow_undeclared_parameters: bool = False
+    allow_undeclared_parameters: bool = False,
+    automatically_declare_initial_parameters: bool = True
 ) -> 'Node':
     """
     Create an instance of :class:`.Node`.
@@ -128,6 +129,8 @@ def create_node(
     :param start_parameter_services: ``False`` if the node should not create parameter services.
     :param initial_parameters: A list of :class:`.Parameter` to be set during node creation.
     :param allow_undeclared_parameters: True if undeclared parameters are allowed, False otherwise.
+    :param automatically_declare_initial_parameters: True if initial parameters have to be declared
+        upon node creation, false otherwise.
     :return: An instance of the newly created node.
     """
     # imported locally to avoid loading extensions on module import
@@ -137,7 +140,8 @@ def create_node(
         use_global_arguments=use_global_arguments,
         start_parameter_services=start_parameter_services,
         initial_parameters=initial_parameters,
-        allow_undeclared_parameters=allow_undeclared_parameters)
+        allow_undeclared_parameters=allow_undeclared_parameters,
+        automatically_declare_initial_parameters=automatically_declare_initial_parameters)
 
 
 def spin_once(node: 'Node', *, executor: 'Executor' = None, timeout_sec: float = None) -> None:
--- a/rclpy/rclpy/exceptions.py
+++ b/rclpy/rclpy/exceptions.py
@@ -65,3 +65,48 @@ class InvalidServiceNameException(NameValidationException):
 
     def __init__(self, name, error_msg, invalid_index, *args):
         NameValidationException.__init__(self, 'service name', name, error_msg, invalid_index)
+
+
+class ParameterException(Exception):
+    """Base exception for parameter-related errors."""
+
+    def __init__(self, error_msg, parameters, *args):
+        Exception.__init__(self, '{error_msg}: {parameters}'.format(error_msg, parameters), *args)
+
+
+class ParameterNotDeclaredException(ParameterException):
+    """Raised when handling an undeclared parameter when it is not allowed."""
+
+    def __init__(self, parameters, *args):
+        Exception.__init__(self, 'Invalid access to undeclared parameter(s)', parameters, *args)
+
+
+class ParameterAlreadyDeclaredException(ParameterException):
+    """Raised when declaring a parameter that had been declared before."""
+
+    def __init__(self, parameters, *args):
+        Exception.__init__(self, 'Parameter(s) already declared', parameters, *args)
+
+
+class InvalidParameterException(ParameterException):
+    """Raised when a parameter to be declared has an invalid name."""
+
+    def __init__(self, parameter, *args):
+        Exception.__init__(self, 'Invalid parameter name', parameter, *args)
+
+
+class InvalidParameterValueException(ParameterException):
+    """Raised when a parameter to be declared is rejected by a callback."""
+
+    def __init__(self, parameter, value, *args):
+        Exception.__init__(
+            self,
+            'Invalid parameter value ({value}) for parameter'.format(
+                value=value), parameter, *args)
+
+
+class ParameterImmutableException(ParameterException):
+    """Raised when a read-only parameter is modified."""
+
+    def __init__(self, parameter, *args):
+        Exception.__init__(self, 'Attempted to modify read-only parameter', parameter, *args)
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -405,6 +405,11 @@ class Node:
         if self.has_parameter(name):
             return self._parameters[name]
         elif self._allow_undeclared_parameters:
+            # If undeclared parameters are allowed, the parameter might be in the initial set.
+            # If that's the case, first set and then return.
+            if name in self._initial_parameters:
+                self._set_parameters_atomically([self._initial_parameters[name]])
+                return self._parameters[name]
             return Parameter(name, Parameter.Type.NOT_SET, None)
         else:
             raise ParameterNotDeclaredException(name)
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -321,7 +321,7 @@ class Node:
             assert isinstance(name, str)
             # Get value from initial parameters, of from tuple if it doesn't exist.
             if name in self._initial_parameters:
-                value = self._initial_parameters[name]
+                value = self._initial_parameters[name].get_parameter_value()
             elif parameter_tuple[1] is None:
                 value = ParameterValue()
             else:
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -92,7 +92,8 @@ class Node:
         use_global_arguments: bool = True,
         start_parameter_services: bool = True,
         initial_parameters: List[Parameter] = None,
-        allow_undeclared_parameters: bool = False
+        allow_undeclared_parameters: bool = False,
+        automatically_declare_initial_parameters: bool = True
     ) -> None:
         """
         Constructor.
@@ -110,6 +111,8 @@ class Node:
         :param initial_parameters: A list of parameters to be set during node creation.
         :param allow_undeclared_parameters: True if undeclared parameters are allowed.
             This flag affects the behavior of parameter-related operations.
+        :param automatically_declare_initial_parameters: True if initial parameters have to be
+            declared upon node creation, false otherwise.
         """
         self.__handle = None
         self._context = get_default_context() if context is None else context
@@ -124,6 +127,8 @@ class Node:
         self._default_callback_group = MutuallyExclusiveCallbackGroup()
         self._parameters_callback = None
         self._allow_undeclared_parameters = allow_undeclared_parameters
+        self._initial_parameters = {}
+        self._descriptors = {}
 
         namespace = namespace or ''
         if not self._context.ok():
@@ -162,10 +167,12 @@ class Node:
         if initial_parameters is not None:
             self._initial_parameters.update({p.name: p for p in initial_parameters})
 
+        if automatically_declare_initial_parameters:
+            self._parameters.update(self._initial_parameters)
+            self._descriptors.update({p: ParameterDescriptor() for p in self._parameters})
+
         if start_parameter_services:
             self._parameter_service = ParameterService(self)
-
-        self._descriptors = {}
 
     @property
     def publishers(self) -> Iterator[Publisher]:
@@ -408,13 +415,15 @@ class Node:
             # If undeclared parameters are allowed, the parameter might be in the initial set.
             # If that's the case, first set and then return.
             if name in self._initial_parameters:
-                self._set_parameters_atomically([self._initial_parameters[name]])
+                self._parameters.update({name: self._initial_parameters[name]})
+                self._descriptors.update({name: ParameterDescriptor()})
                 return self._parameters[name]
             return Parameter(name, Parameter.Type.NOT_SET, None)
         else:
             raise ParameterNotDeclaredException(name)
 
-    def get_parameter_or(self, name: str, alternative_value: Parameter = None) -> Parameter:
+    def get_parameter_or(
+            self, name: str, alternative_value: Optional[Parameter] = None) -> Parameter:
         """
         Get a parameter or the alternative value.
 
@@ -581,6 +590,7 @@ class Node:
                 parameter_event.node = self.get_namespace() + self.get_name()
             else:
                 parameter_event.node = self.get_namespace() + '/' + self.get_name()
+
             for param in parameter_list:
                 if Parameter.Type.NOT_SET == param.type_:
                     if Parameter.Type.NOT_SET != self.get_parameter_or(param.name).type_:
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -21,7 +21,10 @@ from typing import TypeVar
 
 import weakref
 
+from rcl_interfaces.msg import Parameter as ParameterMsg
+from rcl_interfaces.msg import ParameterDescriptor
 from rcl_interfaces.msg import ParameterEvent
+from rcl_interfaces.msg import ParameterValue
 from rcl_interfaces.msg import SetParametersResult
 from rclpy.callback_groups import CallbackGroup
 from rclpy.callback_groups import MutuallyExclusiveCallbackGroup
@@ -30,7 +33,11 @@ from rclpy.clock import Clock
 from rclpy.clock import ROSClock
 from rclpy.constants import S_TO_NS
 from rclpy.context import Context
+from rclpy.exceptions import InvalidParameterValueException
 from rclpy.exceptions import NotInitializedException
+from rclpy.exceptions import ParameterAlreadyDeclaredException
+from rclpy.exceptions import ParameterImmutableException
+from rclpy.exceptions import ParameterNotDeclaredException
 from rclpy.executors import Executor
 from rclpy.expand_topic_name import expand_topic_name
 from rclpy.guard_condition import GuardCondition
@@ -54,6 +61,7 @@ from rclpy.utilities import get_default_context
 from rclpy.validate_full_topic_name import validate_full_topic_name
 from rclpy.validate_namespace import validate_namespace
 from rclpy.validate_node_name import validate_node_name
+from rclpy.validate_parameter_name import validate_parameter_name
 from rclpy.validate_topic_name import validate_topic_name
 from rclpy.waitable import Waitable
 
@@ -83,7 +91,8 @@ class Node:
         namespace: str = None,
         use_global_arguments: bool = True,
         start_parameter_services: bool = True,
-        initial_parameters: List[Parameter] = None
+        initial_parameters: List[Parameter] = None,
+        allow_undeclared_parameters: bool = False
     ) -> None:
         """
         Constructor.
@@ -99,6 +108,8 @@ class Node:
         :param start_parameter_services: ``False`` if the node should not create parameter
             services.
         :param initial_parameters: A list of parameters to be set during node creation.
+        :param allow_undeclared_parameters: True if undeclared parameters are allowed.
+            This flag affects the behavior of parameter-related operations.
         """
         self.__handle = None
         self._context = get_default_context() if context is None else context
@@ -112,6 +123,7 @@ class Node:
         self.__waitables: List[Waitable] = []
         self._default_callback_group = MutuallyExclusiveCallbackGroup()
         self._parameters_callback = None
+        self._allow_undeclared_parameters = allow_undeclared_parameters
 
         namespace = namespace or ''
         if not self._context.ok():
@@ -144,15 +156,16 @@ class Node:
             ParameterEvent, 'parameter_events', qos_profile=qos_profile_parameter_events)
 
         with self.handle as capsule:
-            node_parameters = _rclpy.rclpy_get_node_parameters(Parameter, capsule)
+            self._initial_parameters = _rclpy.rclpy_get_node_parameters(Parameter, capsule)
         # Combine parameters from params files with those from the node constructor and
         # use the set_parameters_atomically API so a parameter event is published.
         if initial_parameters is not None:
-            node_parameters.update({p.name: p for p in initial_parameters})
-        self.set_parameters_atomically(node_parameters.values())
+            self._initial_parameters.update({p.name: p for p in initial_parameters})
 
         if start_parameter_services:
             self._parameter_service = ParameterService(self)
+
+        self._descriptors = {}
 
     @property
     def publishers(self) -> Iterator[Publisher]:
@@ -232,7 +245,7 @@ class Node:
 
         Cannot be modified after node creation.
 
-        :raises AttributeError: if modified after creation.
+        :raises: AttributeError if modified after creation.
         """
         return self.__handle
 
@@ -258,12 +271,121 @@ class Node:
         """Get the nodes logger."""
         return self._logger
 
+    def declare_parameter(
+        self,
+        name: str,
+        value: ParameterValue = ParameterValue(),
+        descriptor: ParameterDescriptor = ParameterDescriptor()
+    ) -> Parameter:
+        """
+        Declare and initialize a parameter.
+
+        This method, if successful, will result in any callback registered with
+        :func:`set_parameters_callback` to be called.
+
+        :param name: Fully-qualified name of the parameter, including its namespace.
+        :param value: Value of the parameter to declare.
+        :param descriptor: Descriptor for the parameter to declare.
+        :return: Parameter with the effectively assigned value.
+        :raises: ParameterAlreadyDeclaredException if the parameter had already been declared.
+        :raises: InvalidParameterException if the parameter name is invalid.
+        :raises: InvalidParameterValueException if the registered callback rejects the parameter.
+        """
+        return self.declare_parameters('', [(name, value, descriptor)])[0]
+
+    def declare_parameters(
+        self,
+        namespace: str,
+        parameters: List[Tuple[str, Optional[ParameterValue], Optional[ParameterDescriptor]]]
+    ) -> List[Parameter]:
+        """
+        Declare a list of parameters.
+
+        This method, if successful, will result in any callback registered with
+        :func:`set_parameters_callback` to be called once for each parameter.
+        If one of those calls fail, an exception will be raised and the remaining parameters will
+        not be declared.
+        Parameters declared up to that point will not be undeclared.
+
+        :param namespace: Namespace for parameters.
+        :param parameters: Tuple with parameters to declare, with a name, value and descriptor.
+        :return: Parameter list with the effectively assigned values for each of them.
+        :raises: ParameterAlreadyDeclaredException if the parameter had already been declared.
+        :raises: InvalidParameterException if the parameter name is invalid.
+        :raises: InvalidParameterValueException if the registered callback rejects any parameter.
+        """
+        parameter_list = []
+        descriptor_list = []
+        for parameter_tuple in parameters:
+            name = parameter_tuple[0]
+            assert isinstance(name, str)
+            # Get value from initial parameters, of from tuple if it doesn't exist.
+            if name in self._initial_parameters:
+                value = self._initial_parameters[name]
+            elif parameter_tuple[1] is None:
+                value = ParameterValue()
+            else:
+                value = parameter_tuple[1]
+            assert isinstance(value, ParameterValue)
+            descriptor = parameter_tuple[2]
+            if descriptor is None:
+                descriptor = ParameterDescriptor()
+            assert isinstance(descriptor, ParameterDescriptor)
+
+            # Note(jubeira): declare_parameters verifies the name, but set_parameters doesn't.
+            full_name = namespace + name
+            validate_parameter_name(full_name)
+
+            parameter_list.append(Parameter.from_parameter_msg(
+                ParameterMsg(name=full_name, value=value)))
+            descriptor_list.append(descriptor)
+
+        parameters_already_declared = [
+            parameter.name for parameter in parameter_list if parameter.name in self._parameters
+        ]
+        if any(parameters_already_declared):
+            raise ParameterAlreadyDeclaredException(parameters_already_declared)
+
+        # Call the callback once for each of the parameters, using method that doesn't
+        # check whether the parameter was declared beforehand or not.
+        self._set_parameters(
+            parameter_list, descriptor_list=descriptor_list, raise_on_failure=True)
+        return self.get_parameters([parameter.name for parameter in parameter_list])
+
+    def undeclare_parameter(self, name: str):
+        """
+        Undeclare a previously declared parameter.
+
+        This method will not cause a callback registered with
+        :func:`set_parameters_callback` to be called.
+
+        :param name: Fully-qualified name of the parameter, including its namespace.
+        :raises: ParameterNotDeclaredException if parameter had not been declared before.
+        :raises: ParameterImmutableException if the parameter was created as read-only.
+        """
+        if self.has_parameter(name):
+            if self._descriptors[name].read_only:
+                raise ParameterImmutableException(name)
+            else:
+                del self._parameters[name]
+                del self._descriptors[name]
+        else:
+            raise ParameterNotDeclaredException(name)
+
+    def has_parameter(self, name: str) -> bool:
+        """Return True if parameter is declared; False otherwise."""
+        return name in self._parameters
+
     def get_parameters(self, names: List[str]) -> List[Parameter]:
         """
         Get a list of parameters.
 
-        :param names: The names of the parameters to get.
+        :param names: Fully-qualified names of the parameters to get, including their namespaces.
         :return: The values for the given parameter names.
+            A default Parameter will be returned for undeclared parameters if
+            undeclared parameters are allowed.
+        :raises: ParameterNotDeclaredException if undeclared parameters are not allowed,
+            and at least one parameter hadn't been declared beforehand.
         """
         if not all(isinstance(name, str) for name in names):
             raise TypeError('All names must be instances of type str')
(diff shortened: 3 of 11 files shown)


jubeira:
This is a draft for now; there are still multiple TODOs around the code and testing to be done before promoting it to a PR.
If anyone's interested in taking a preliminary look and providing feedback, it is more than welcome!

rclpy/rclpy/exceptions.py
```diff
@@ -65,3 +65,29 @@ class InvalidServiceNameException(NameValidationException):
 
     def __init__(self, name, error_msg, invalid_index, *args):
         NameValidationException.__init__(self, 'service name', name, error_msg, invalid_index)
+
+
+# TODO(jubeira): Polish parameter-related exceptions; improve error messages.
+class ParameterNotDeclaredException(Exception):
```
jacobperron: Consider adding a `ParameterImmutableException` (analogous to rclcpp) for read only parameters.

rclpy/rclpy/node.py
```diff
@@ -239,6 +308,18 @@ def get_parameter(self, name: str) -> Parameter:
...
         return self._parameters[name]
 
+    def get_parameter_or(self, name: str, alternative_value: Parameter) -> Parameter:
+        """
+        Get a parameter or the alternative value if it had not been declared before.
+
+        :param name: Name of the parameter to get.
+        :alternative_value: Alternative parameter to get if it had not been declared before.
+        """
+        if has_parameter(name):
```
jacobperron: This can be reduced to
```python
return self._parameters.get(name, alternative_value)
```

rclpy/rclpy/node.py
```diff
@@ -248,14 +329,29 @@ def set_parameters(self, parameter_list: List[Parameter]) -> List[SetParametersR
...
         published.
 
+        If setting a parameter fails due to not being declared, then the
+        parameters which have already been set will stay set, and no attempt will
+        be made to set the parameters which come after.
+
         :param parameter_list: The list of parameters to set.
         :return: A list of SetParametersResult messages.
         """
+        return self._set_parameters(parameter_list, self.set_parameters_atomically)
```
jacobperron: I was confused for a minute. Maybe it should be clarified that the `setter_func` parameter is applied individually for each parameter. Otherwise, this looks like we might be setting the parameter list atomically, contrary to the docstring.

rclpy/rclpy/node.py
```diff
@@ -248,14 +329,29 @@ def set_parameters(self, parameter_list: List[Parameter]) -> List[SetParametersR
...
+        if setter_func is None:
+            setter_func = self._set_parameters_atomically
+
         results = []
         for param in parameter_list:
-            if not isinstance(param, Parameter):
-                raise TypeError("parameter must be instance of type '{}'".format(repr(Parameter)))
-            results.append(self.set_parameters_atomically([param]))
+            result = setter_func([param])
+            if raise_on_failure and not result.successful:
```
jacobperron: What is the imagined use case for `raise_on_failure`? I don't see it being used anywhere.

rclpy/rclpy/node.py
jubeira: Good idea.
I still need to work a bit on the documentation; it still needs a second pass.
In general terms I'm reusing the existing functionality as much as possible. Following that line, these are some ideas behind this approach:
- The setter methods will be split in two: a "public" part and a "private" part (with trailing underscore).
- "Private" setters won't check whether the parameters were declared or not, so they will be used to actually declare the parameters.
- "Public" setters will comply with the new logic: raise exceptions if undeclared parameters are not allowed, or implicitly declare and set the parameters if undeclared parameters are allowed.

rclpy/rclpy/node.py
jubeira: Good catch; this is used in `declare_parameters`. The callback may reject the parameter values, and in this case it should raise an exception. Added in https://github.com/ros2/rclpy/pull/325/commits/4440e852ddf7bdfc84bcc09ba9ace73c94769fa8.

jubeira:
Code updated and rebased with the current `master`.
This is ready for review; the only pending tasks are adding proper test cases for the new API, and improving exception handling. Existing 

Note that this PR should get in before the API freeze. The implementation reused the existing codebase as much as possible; the intention was to comply with the new API changing it as little as possible. Following that line, I would suggest to correct implementation details that may appear after the API freeze.

hidmic:
CI (full run):

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/6841/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3134/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5625/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6663/)

rclpy/rclpy/node.py
```diff
@@ -233,41 +316,163 @@ def get_parameter(self, name: str) -> Parameter:
...
+    ) -> List[SetParametersResult]:
+        """
+        Set parameters for the node, and return the result for the set action.
+
+        Method for internal usage; applies a setter method for each parameters in the list.
+        By default it doesn't check if the parameters were declared, and both declares and sets
+        the given list.
+
+        If a callback was registered previously with :func:`set_parameters_callback`, it will be
+        called prior to setting the parameters for the node once for each parameter.
```
hidmic: @jubeira consider
```suggestion
        called prior to setting the parameters for the node, once for each parameter.
```

rclpy/test/test_create_node.py
```diff
@@ -43,7 +43,8 @@ def test_create_node_with_namespace(self):
     def test_create_node_with_empty_namespace(self):
         node_name = 'create_node_test'
         namespace = ''
-        node = rclpy.create_node(node_name, namespace=namespace, context=self.context)
+        node = rclpy.create_node(
+            node_name, namespace=namespace, context=self.context)
```
hidmic: @jubeira nit: why the change?

rclpy/rclpy/node.py
```diff
@@ -20,6 +20,8 @@
 
 import weakref
 
+# TODO(jubeira): check for missing imports.
```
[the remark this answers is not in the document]
jubeira: Removed in https://github.com/ros2/rclpy/pull/325/commits/4d2429d48296b76742448dccf74d8f4450c43ca9; this note wasn't applicable anymore.

rclpy/rclpy/exceptions.py
```diff
@@ -65,3 +65,39 @@ class InvalidServiceNameException(NameValidationException):
...
+
+    def __init__(self, *args):
+        Exception.__init__(self, 'Invalid parameter value.', *args)
+
+
+class ParameterImmutableException(Exception):
+    """Raised when a read-only parameter is modified."""
+
+    def __init__(self, *args):
+        Exception.__init__(self, 'Attempted to modify an read-only parameter.', *args)
```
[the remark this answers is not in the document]
jubeira: Addressed in https://github.com/ros2/rclpy/pull/325/commits/4d2429d48296b76742448dccf74d8f4450c43ca9.
The exception messages were improved as well to provide information about which parameter(s) are failing in particular for each case.

rclpy/rclpy/node.py
```diff
@@ -217,12 +230,82 @@ def get_logger(self):
...
+        This method will not cause a callback registered with
+        :func:`set_parameters_callback` to be called.
+
+        :param name: name of the parameter.
+        :raises: ParameterNotDeclaredException if parameter had not been declared before.
+        :raises: ParameterImmutableException if the parameter was created as read-only.
+        """
+        if self.has_parameter(name):
+            if self._parameters[name].descriptor.read_only:
+                raise ParameterImmutableException()
```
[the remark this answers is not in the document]
jubeira: Addressed in https://github.com/ros2/rclpy/pull/325/commits/4d2429d48296b76742448dccf74d8f4450c43ca9.
The exception messages were improved as well to provide information about which parameter(s) are failing in particular for each case.

rclpy/rclpy/node.py
```diff
@@ -217,12 +230,82 @@ def get_logger(self):
...
+            else:
+                del self._parameters[name]
+        else:
+            raise ParameterNotDeclaredException()
```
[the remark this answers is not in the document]
jubeira: Addressed in https://github.com/ros2/rclpy/pull/325/commits/4d2429d48296b76742448dccf74d8f4450c43ca9.
The exception messages were improved as well to provide information about which parameter(s) are failing in particular for each case.

rclpy/rclpy/node.py
jubeira: Addressed in https://github.com/ros2/rclpy/pull/325/commits/4d2429d48296b76742448dccf74d8f4450c43ca9.

rclpy/test/test_create_node.py
jubeira: Reverted in https://github.com/ros2/rclpy/pull/325/commits/4d2429d48296b76742448dccf74d8f4450c43ca9. This unwanted change had got through the rebase.

rclpy/rclpy/node.py
```diff
@@ -217,12 +230,82 @@ def get_logger(self):
...
+
+        :param namespace: Namespace for parameters.
+        :param parameters: The parameters to declare, with a name, value and descriptor.
+        :return: Parameter list with the effectively assigned values for each of them.
+        :raises: ParameterAlreadyDeclaredException if the parameter had already been declared.
+        :raises: InvalidParameterException if the parameter name is invalid.
+        :raises: InvalidParameterValueException if the registered callback rejects the parameter.
+        """
+        for parameter in parameters:
+            parameter.name = namespace + parameter.name
```
jubeira: Writing tests I just realized that this is not possible (name can't be set after creating `Parameter` object).
I can think of these alternatives, in order of preference:
- Creating new parameters with same value, type and descriptor but with a new name containing the namespace.
- Suppressing `namespace` parameter. This is introduced to match the behavior in `rclcpp`, but on the other hand the current `set_parameters` method in `master` doesn't have a namespace option.
- Adding a setter to the name.
- Using the private `_name` member (argh).

@hidmic thoughts?
/cc @wjwwood input on this is welcome!

rclpy/rclpy/node.py
```diff
@@ -217,12 +229,85 @@ def get_logger(self):
...
+        return self.declare_parameters('/', [parameter])
+
+    def declare_parameters(self, namespace: str, parameters: List[Parameter]) -> List[Parameter]:
+        """
+        Declare a list of parameters.
+
+        This method, if successful, will result in any callback registered with
+        :func:`set_parameters_callback` to be called once for each parameter.
+        If one of those calls fail, an exception will be raised and the remaining parameters will
+        not be declared. Parameters declared up to that point will not be undeclared.
```
wjwwood: nitpick: one sentence per line

rclpy/rclpy/node.py
```diff
@@ -217,12 +229,85 @@ def get_logger(self):
...
+
+        :param namespace: Namespace for parameters.
```
wjwwood: What happens if the namespace is an empty string?

rclpy/rclpy/node.py
wjwwood: Why not have a version of `set_parameters` or `self._set_parameters` which takes a tuple of `(name, value)`, then you can take apart the `Parameter` and use it how you want.

I guess that's the same as creating a temporary `Parameter` with the name you want...

rclpy/rclpy/node.py
```diff
@@ -233,41 +318,164 @@ def get_parameter(self, name: str) -> Parameter:
...
+        called prior to setting the parameters for the node only once for all parameters.
+        If the callback prevents the parameters from being set, then it will be reflected in the
+        returned result; no exceptions will be raised in this case.
+        For each successfully set parameter, a :class:`ParameterEvent` message is
+        published.
+
+        If the value type of the parameter is NOT_SET, and the existing parameter type is
+        something else, then the parameter will be implicitly undeclared.
+
+        :param parameter_list: The list of parameters to set.
```
wjwwood: I see a problem here (I think). If I declare a parameter as read_only or with a range as part of the descriptor, and then I set it without a descriptor, doesn't the descriptor default to `ParameterDescriptor()` and therefore overwrite the descriptor that I set when declaring?

rclpy/rclpy/node.py
```diff
@@ -217,12 +229,85 @@ def get_logger(self):
...
+        return self.get_parameters([parameter.name for parameter in parameters])
+
+    def undeclare_parameter(self, name: str):
+        """
+        Undeclare a previously declared parameter.
+
+        This method will not cause a callback registered with
+        :func:`set_parameters_callback` to be called.
+
+        :param name: name of the parameter.
```
jacobperron: I realize it wasn't explicit in `rclcpp`, but it might be good to clarify here that the name is fully-qualified (including any namespace provided when declared).

rclpy/rclpy/node.py
jacobperron: Considering that other parameter-related methods typically reference parameters by name (undeclare, get, set), it might make sense to also accept names when declaring parameters as well. Like @wjwwood suggests, we could change `set_parameters` to also take a tuple of `(name, value)`. This would be more akin to the implementation in `rclcpp`.

We use a temporary parameter `Parameter` and create a new one with the provided name, e.g

```python
def declare_parameter(self, name: str, parameter: Parameter) -> Parameter:
    ...
def declare_parameters(self, namespace: str, parameters: List[Tuple[str, Parameter]]) -> List[Parameter]:
    ...
```

Or consider using a `ParameterValue` instead (like `rclcpp`):

```python
def declare_parameter(self, name: str, parameter: ParameterValue) -> Parameter:
    ...
def declare_parameters(self, namespace: str, parameters: List[Tuple[str, ParameterValue]]) -> List[Parameter]:
    ...
```

jubeira: See a proposed fix in https://github.com/ros2/rclpy/pull/325/commits/a39ea62cb69685e437a5935cfc206a648808d26a.

In `rclcpp`, `declare_parameter` takes a name, a value and a descriptor, but `set_parameter` takes a `Parameter_T`. Following that line, I left `set_parameters` with the same signature, and changed `declare_parameters` to take the pieces of the parameter as separate arguments in a tuple.
In other words, its @jacobperron 's second suggestion, considering `ParameterDescriptor` as well.
```
def declare_parameter(self, name: str, parameter: ParameterValue, descriptor: ParameterDescriptor) -> Parameter:
    ...
def declare_parameters(self, namespace: str, parameters: List[Tuple[str, ParameterValue, ParameterDescriptor]]) -> List[Parameter]:
    ...
```
/cc @hidmic

rclpy/rclpy/node.py
jubeira: - Attempting to set a `read_only` parameter should fail with an exception. That was not implemented here, and was not implemented in `rclcpp` yet AFAIK. If I'm not mistaken, that should be taken into account in a follow-up PR.
- Yes, the descriptor would override the existing one; I'm not sure if that would be expected behavior or not (it could be as long as the existing descriptor is not `read_only`?).
I don't think it's a good idea to change `set_parameters` signature at this point to take just `name` and `value`. On the other hand, the `Parameter` class as of `master` branch only supports the default `ParameterDescriptor`, so I think it's a good idea to extend it as in this PR to support custom ones. A possible compromise solution would be to ignore the `ParameterDescriptor` sent to this method, and use the existing one instead if it exists if we don't want to override the descriptor.

rclpy/rclpy/node.py
```diff
@@ -217,12 +229,85 @@ def get_logger(self):
...
+        This method, if successful, will result in any callback registered with
+        :func:`set_parameters_callback` to be called.
+
+        :param parameter: The parameter to declare, with a name, value and descriptor.
+        :return: Parameter with the effectively assigned value.
+        :raises: ParameterAlreadyDeclaredException if the parameter had already been declared.
+        :raises: InvalidParameterException if the parameter name is invalid.
+        :raises: InvalidParameterValueException if the registered callback rejects the parameter.
+        """
+        return self.declare_parameters('/', [parameter])
```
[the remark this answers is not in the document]
jubeira: Good catch, thanks! Addressed in https://github.com/ros2/rclpy/pull/325/commits/a39ea62cb69685e437a5935cfc206a648808d26a.

rclpy/rclpy/validate_parameter_name.py
```diff
@@ -0,0 +1,39 @@
...
+def validate_parameter_name(name):
+    """
+    Validate a given parameter name, and raise an exception if invalid.
+
+    The name does not have to be fully-qualified and is not expanded.
+
+    If the name is invalid then rclpy.exceptions.InvalidParameterNameException
+    will be raised.
+
+    :param name str: topic or service name to be validated
```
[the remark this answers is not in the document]
jubeira: Thanks; removed in https://github.com/ros2/rclpy/pull/325/commits/b56a25931f252123348348a71be3e1d0233ff396.

rclpy/rclpy/validate_parameter_name.py
```diff
@@ -0,0 +1,39 @@
...
+    :returns: True when it is valid
+    :raises: InvalidParameterNameException: when the name is invalid
+    """
+    # TODO(jubeira): define whether this is an appropriate method to validate
+    # parameter names.
+    result = _rclpy.rclpy_get_validation_error_for_topic_name(name)
+    if result is None:
+        return True
+    error_msg, invalid_index = result
+    raise InvalidParameterException(name, error_msg, invalid_index)
```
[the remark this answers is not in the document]
jubeira: I do agree that `InvalidParameterName` exception is a better name, but in `rclcpp` the exception for invalid name is `InvalidParameterException`, and that's why I'm using that name here as well.

For now I changed the documentation to match this exception here https://github.com/ros2/rclpy/pull/325/commits/b56a25931f252123348348a71be3e1d0233ff396. I'm slightly inclined to using the same name as in `rclcpp` in this case, but I wouldn't mind refactoring it to `InvalidParameterNameException` if you think it's a better option.

rclpy/rclpy/node.py
```diff
@@ -217,12 +229,85 @@ def get_logger(self):
...
+        """
+        Declare and initialize a parameter.
+
+        This method, if successful, will result in any callback registered with
+        :func:`set_parameters_callback` to be called.
+
+        :param parameter: The parameter to declare, with a name, value and descriptor.
+        :return: Parameter with the effectively assigned value.
+        :raises: ParameterAlreadyDeclaredException if the parameter had already been declared.
+        :raises: InvalidParameterException if the parameter name is invalid.
```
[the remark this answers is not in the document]
jubeira: (see comment above).

rclpy/rclpy/validate_parameter_name.py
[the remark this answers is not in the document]
jacobperron: That's okay with me, but in `rclcpp` it's plural: `InvalidParametersExeception` :wink:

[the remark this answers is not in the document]
jubeira: Yes, indeed; I did notice that. I used singular because:
- All the remaining exceptions but this one are singular.
- The exception is only raised when one parameter name in a given list is incorrect when the name is verified; the verification isn't applied to all the parameters at once. This seems to match the [documentation](https://github.com/ros2/rclcpp/blob/a8a0788f81ddd70210df0c38b4434901d3a9ca44/rclcpp/include/rclcpp/node.hpp#L364-L365) for `rclcpp`.

That being said, perhaps we should just stick to the plural right?

rclpy/rclpy/node.py
wjwwood: Setting a read only parameter should fail in rclcpp, but that might be an oversight right now, there is a test for undeclare:

https://github.com/ros2/rclcpp/blob/7ed130cf7ac7b08158d6aa2e8bd808e0cccee215/rclcpp/test/test_node.cpp#L577-L588

It will fail to set it:

https://github.com/ros2/rclcpp/blob/7ed130cf7ac7b08158d6aa2e8bd808e0cccee215/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L392-L397

I don't think it should be an exception necessarily, but maybe.

---

To match C++, we'd need to remove the descriptor from the Parameter class. I think that makes sense, because you want to be able to conveniently create a Parameter as a user in order to set one, e.g. `node.set_parameter(Parameter(name, value))`, and not worry about the description.

I don't know if we should let set change the description or not, in C++ it is not possible, you much undeclare and re-declare to change the description.

Personally I'd like to see something like `node.declare_parameter(Parameter, ParameterDescriptor)`, `node.set_parameter(Parameter)`, and `node.set_parameter_descriptor(ParameterDescriptor)` (the `ParameterDescriptor` contains the name already I believe. We could later add a "set parameter descriptor" function in c++ too.

rclpy/rclpy/validate_parameter_name.py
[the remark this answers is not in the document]
wjwwood: We could change the one in rclcpp if singular makes sense. I don't remember having a strong opinion about that. I probably used singular because at the time it was always thrown from `set_parameters` or `set_parameters_atomically` which were plural as well.

rclpy/rclpy/node.py
wjwwood: For what it is worth, there's no reason C++ doesn't provide a `set_parameters(vector<pair<string, ParameterValue>>)` nor is there a reason against `declare_parameters(string, vector<pair<Parameter, descriptor>>)`, other than that I ran out of time to keep adding convenient signatures. Many of the other signatures that seem less obvious were added by request before my big refactor.

rclpy/rclpy/validate_parameter_name.py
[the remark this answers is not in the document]
jacobperron: Singular makes sense to me. We could update the rclcpp name for consistency.

rclpy/rclpy/node.py
jubeira: Thanks for the feedback; that makes sense.
I was confused by the fact that `Parameter` held a `ParameterDescriptor`; removing it would iron out the confusion and all the intermediate problems that arise from it. Let's go that way; I'll make those changes.

rclpy/rclpy/validate_parameter_name.py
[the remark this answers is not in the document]
jubeira: Alright, then let's keep it `InvalidParameterException` here :+1:

rclpy/rclpy/node.py
```diff
@@ -217,12 +231,109 @@ def get_logger(self):
         """Get the nodes logger."""
         return self._logger
 
+    def declare_parameter(
+        self,
+        name: str,
+        value: ParameterValue,
+        descriptor: ParameterDescriptor
```
jacobperron: I think it makes sense to make `value` and `descriptor` optional, in case we just want to declare the names of parameters and set the value/descriptor later. I believe this is more akin to `rclcpp`.

e.g.

```python
def declare_parameter(
        self,
        name: str,
        value: ParameterValue = ParameterValue(),
        descriptor: ParameterDescriptor = ParameterDescriptor()
```

Or perhaps `None`.

rclpy/rclpy/node.py
```diff
@@ -217,12 +231,109 @@ def get_logger(self):
...
+        :raises: ParameterAlreadyDeclaredException if the parameter had already been declared.
+        :raises: InvalidParameterException if the parameter name is invalid.
+        :raises: InvalidParameterValueException if the registered callback rejects the parameter.
+        """
+        return self.declare_parameters('', [(name, value, descriptor)])[0]
+
+    def declare_parameters(
+        self,
+        namespace: str,
+        parameters: List[Tuple[str, ParameterValue, ParameterDescriptor]]
```
jacobperron: Similar to my previous comment, the value and descriptor could be optional:
```suggestion
        parameters: List[Tuple[str, Optional[ParameterValue], Optional[ParameterDescriptor]]]
```

rclpy/rclpy/validate_parameter_name.py
```diff
@@ -0,0 +1,37 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from rclpy.exceptions import InvalidParameterException
+from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+
+
+def validate_parameter_name(name):
```
jacobperron:
```suggestion
def validate_parameter_name(name: str) -> bool:
```

rclpy/rclpy/validate_parameter_name.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Validate a given parameter name, and raise an exception if invalid.
+
+    The name does not have to be fully-qualified and is not expanded.
+
+    If the name is invalid then rclpy.exceptions.InvalidParameterException
+    will be raised.
+
+    :param name str: topic or service name to be validated.
```
jacobperron:
```suggestion
    :param name: parameter name to be validated.
```

rclpy/rclpy/node.py
jubeira: > Personally I'd like to see something like node.declare_parameter(Parameter, ParameterDescriptor)

That could work for `declare_parameter` but not for `declare_parameters` as we discussed here: https://github.com/ros2/rclpy/pull/325#discussion_r280233234. For that I'd stick with the proposal that we have now (i.e. using separate `name`, `value`, `descriptor`).

rclpy/rclpy/node.py
jubeira: This shouldn't be an issue; the current `set` method for parameters doesn't actually need a namespace.

rclpy/rclpy/validate_parameter_name.py
jubeira: Thanks; addressed in https://github.com/ros2/rclpy/pull/325/commits/0925fe4e4a72a4b32ff1ec58790d73c4ba35838a.

rclpy/rclpy/validate_parameter_name.py
jubeira: Addressed in https://github.com/ros2/rclpy/pull/325/commits/0925fe4e4a72a4b32ff1ec58790d73c4ba35838a.

rclpy/rclpy/node.py
jubeira: Addressed in https://github.com/ros2/rclpy/pull/325/commits/0925fe4e4a72a4b32ff1ec58790d73c4ba35838a.

rclpy/rclpy/node.py
jubeira: Addressed in https://github.com/ros2/rclpy/pull/325/commits/0925fe4e4a72a4b32ff1ec58790d73c4ba35838a.

rclpy/rclpy/node.py
jubeira: Addressed in https://github.com/ros2/rclpy/pull/325/commits/0925fe4e4a72a4b32ff1ec58790d73c4ba35838a.

jubeira:
Thanks for the comments @wjwwood @jacobperron !
I've just pushed some more changes addressing them. I still need to add more test cases, but this seems to be taking shape. I'll rebase to fix the conflicts before proceeding.

jubeira:
CI for linux:
- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/6921/)
- Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/3190/)

MacOS and Windows were not available and didn't finish yet.
@wjwwood any more thoughts or comments? Do you think this can get in the API freeze?

sloretz:
@jubeira It looks like those CI runs tested all packages. For the next run I recommend `--packages-above rclpy` in the tests args to reduce the time CI takes

rclpy/rclpy/node.py
wjwwood: Sounds good to me.

rclpy/test/test_node.py
```diff
@@ -277,6 +285,231 @@ def test_node_has_parameter_services(self):
...
+                    1234,
+                    ParameterDescriptor()
+                )]
+            )
+
+        with self.assertRaises(AssertionError):
+            self.node.declare_parameters(
+                '',
+                [(
+                    'wrong_parameter_descriptor_tpye',
```
jacobperron:
```suggestion
                    'wrong_parameter_descriptor_type',
```

jacobperron:
@jubeira My recent review is just minor nitpicks. Also, I'm okay if we want to put the additional tests in a follow-up PR.

rclpy/rclpy/node.py
```diff
@@ -258,12 +271,121 @@ def get_logger(self):
...
+        :raises: ParameterAlreadyDeclaredException if the parameter had already been declared.
+        :raises: InvalidParameterException if the parameter name is invalid.
+        :raises: InvalidParameterValueException if the registered callback rejects the parameter.
+        """
+        return self.declare_parameters('', [(name, value, descriptor)])[0]
+
+    def declare_parameters(
+        self,
+        namespace: str,
+        parameters: List[Tuple[str, Optional[ParameterValue], Optional[ParameterDescriptor]]]
```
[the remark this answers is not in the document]
jubeira: Great suggestion, I agree!

jubeira:
@jacobperron @wjwwood thanks again for your reviews!
If you are OK with it, I'd suggest merging this one and opening a ticket to track down the missing suggestions and test cases. I can do the follow-up.

/cc @hidmic

rclpy/rclpy/__init__.py
```diff
@@ -111,7 +111,9 @@ def create_node(
     namespace: str = None,
     use_global_arguments: bool = True,
     start_parameter_services: bool = True,
-    initial_parameters: List[Parameter] = None
+    initial_parameters: List[Parameter] = None,
+    allow_undeclared_parameters: bool = False,
+    automatically_declare_initial_parameters: bool = True
```
jacobperron: I don't have strong opinions about it, but this option defaults to `False` in the `rclcpp` implementation. @wjwwood Thoughts?

jubeira:
CI:
- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/6961/)
- Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/3219/)
- Mac OS: [Build Status](https://ci.ros2.org/job/ci_osx/5711/)
- [Build Status](https://ci.ros2.org/job/ci_windows/6784/)

Overall looks good; the failing cases seem to be unrelated.

rclpy/rclpy/__init__.py
jubeira: I know that changing this flag is technically an API change, even though it's rather small compared to the changes this PR introduces.
The only reason I see not to declare initial parameters is to define a descriptor, which would be possible if we add a `set_descriptor` method after the API freeze.
- If undeclared parameters are allowed, I don't see why not to declare initial parameters implicitly. Not doing so actually complicates the logic of `get_parameter` and `get_parameter_or` with little benefits IMO. I'd actually propose to declare them implicitly if undeclared are allowed in a follow-up.
- If undeclared parameters are not allowed and initial parameters are not declared, the user would have to re-declare all of them again. As pointed up above, the only bonus is to define a descriptor when declaring, which can be solved by introducing a `set_descriptor`.

That being said, I'd propose to go ahead and discuss what to do with the flag in the follow-up.

wjwwood: I couldn't disagree more unfortunately, I think it should behave like rclcpp and default to `False`.

> If undeclared parameters are allowed, I don't see why not to declare initial parameters implicitly.

I don't see why we would declare them?

> Not doing so actually complicates the logic of get_parameter and get_parameter_or with little benefits IMO.

I disagree, if you get or set a parameter anywhere in your code, then you should declare it first. Imagine you are supposed to have a parameter which is called `foo` but you accidentally type in your code `get_parameter('fooo')`, declaring allows us to let you know that's probably not what you meant.

It just like programming in any sane language (you have to declare a variable before setting or getting it) vs CMake, where you can `${something}` and that doesn't complain even if `something` doesn't exist...

What actually complicates the logic is support `allow_undeclared_parameters`, which I only agreed to after discussing with the team and so that existing users can flip that switch and get pre-Dashing behavior.

Checking whether or not the parameter has been declared, and throwing or returning `successful=False` if it has not been declared, is easy.

> I'd actually propose to declare them implicitly if undeclared are allowed in a follow-up.

I think these two options should remain separate, and should default to off, as they only benefit the user in special use cases.

> If undeclared parameters are not allowed and initial parameters are not declared, the user would have to re-declare all of them again.

Why would they be declaring again? When did they declare it the first time? The majority of parameters, in ROS 1 anyways, are used in the node but rarely overridden externally. If you didn't declare a parameter first, then you'd always have to use `get_or` like methods to access it because you could never depend on the user to pass you all the require parameters every time. By declaring the parameter, usually in your constructor or before you call `spin`, you can access it freely through the rest of your application without worrying that it may not be set.

> As pointed up above, the only bonus is to define a descriptor when declaring, which can be solved by introducing a `set_descriptor`.

That's not the only bonus, there are also the reasons I described above, though declaring a description is a bonus to using declare parameter, especially if it should be type checked.

By not declaring them automatically, and therefore only considering the initial parameter values when you declare or set, you get to make sure the type given on the command line is valid before setting it. If you declare them automatically, then you cannot add any type checking or constraints (like range) and therefore cannot prevent invalid values from getting initialized.

Then there's the problem of setting parameters that are never used in a node. If you declare them automatically then external observers might see that you node has a parameter like `throttle` and think that it does something, when in fact your code may never even contains the string `throttle`, and in stead it was set because someone accidentally passed that parameter to your node. In a future set of changes we want to have a checker that actually throws an exception if your node is given parameters you didn't expect to help guard against typos in the initial parameter values.

So, in my opinion, there's no point in declaring parameters given by constructor or via the command line automatically, at least most of the time. Because if you don't get or set them, then they don't need to be declared in the first place, and if they're declared you don't need to set the automatically.

The only real case where I think declaring automatically is useful is something like a parameter server, where you'd have a generic executable which could store any number of parameters based on command line input and serve them to other processes, and that use case is the only reason why this option exists, from my perspective.

Maybe you have use cases in mind that we didn't consider when we discussed this in a previous ROS 2 meeting, but without that, I think we've got the right default behavior by disabling automatic declaration and requiring declaration before getting and setting.

> That being said, I'd propose to go ahead and discuss what to do with the flag in the follow-up.

We should really change it soon, because it is not a small change in behavior at all.

jubeira: Thanks for the thorough explanation!
I'll submit a PR with the change for review.

wjwwood: Thanks!

Reading this again, it seems a bit aggressive, sorry about that, I've just been working on this for a while and had a lot to say on the subject. :) Hopefully, the neutral tone wasn't taken as negative.

I think the perspective you had above, is not unreasonable, and in fact we had to go in a few rounds as a team before we settled on the behavior in rclcpp.

jubeira: No worries! I've definitely found the details useful. In the rush of fixing some test cases last week I didn't have time to wrap this up as I wanted to; hopefully the changes in #339 should do the job.

PR #326, opened 2019-04-27
Prevent rcutils_log from accessing invalid memory

thomas-moulard:
Before this fix, rclpy_logging_rcutils_log was using
rcutils_log without a format string (log message was
passed directly as the foramt).

This is unsafe because if the log message contains
a sequence of character which can be interpreted by
snprintf ("%d" for instance), snprintf will look for
this argument value which was never passed.
This leads to reading invalid memory and may cause a crash
or corrupted log messages.

The simple fix is to always pass "%s" when a static log
string needs to be logged.

Signed-off-by: Thomas Moulard <tmoulard@amazon.com>

### DIFF
--- a/rclpy/src/rclpy/_rclpy_logging.c
+++ b/rclpy/src/rclpy/_rclpy_logging.c
@@ -165,7 +165,7 @@ rclpy_logging_rcutils_log(PyObject * Py_UNUSED(self), PyObject * args)
 
   RCUTILS_LOGGING_AUTOINIT
   rcutils_log_location_t logging_location = {function_name, file_name, line_number};
-  rcutils_log(&logging_location, severity, name, message);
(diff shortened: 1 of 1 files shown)


thomas-moulard:
https://ci.ros2.org/job/ci_launcher/3376/

sloretz:
Thanks for the PR! CI LGTM

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6804/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3100/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5595/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6629/)

PR #328, opened 2019-04-29
Don't use multiprocessing in tests

sloretz:
Blocks #319 

Some tests use the `multiprocessing` module to run tests in another process. On linux this uses `fork()`, which is not safe to do when using multiple threads.

From http://man7.org/linux/man-pages/man2/fork.2.html
```
The child process is created with a single thread—the one that
          called fork().  The entire virtual address space of the parent is
          replicated in the child, including the states of mutexes,
          condition variables, and other pthreads objects;
```

This means the child only has one thread, but objects with references to other threads exist. If another test does not call `destroy_node()` then the fork will get a copy of that node, including a copy of an `eprosima::fastrtps::rtps::ChannelResource` which has a reference to a thread that only exists in the parent. When the garbage collector runs in the child, the child's the copy of the node will block forever [when it the `ChannelResource` tries to `join()` the non-existent thread](https://github.com/eProsima/Fast-RTPS/blob/a8691a40be6b8460b01edde36ad8563170a3a35a/src/cpp/transport/ChannelResource.cpp#L59). 

It should be possible to work around this by [changing the `multiprocessing` module to use the `spawn` method instead of `fork`](https://docs.python.org/3.7/library/multiprocessing.html#contexts-and-start-methods), but for some reason `multiprocessing.Pool.apply()` deadlocks in my python environment with trivial functions when using `spawn`, so I rewrote these tests to get rid of `multiprocessing` altogether.


sloretz:
CI (testing just rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6811/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3105/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5600/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6634/)

wjwwood:
The reason for using multiprocessing in the past was to allow us to use more than one rmw implementation. Without multiprocessing you cannot test fastrtps and connext in the same pytest instance. Also you cannot be sure that init/shutdown is isolated between tests.

Is this not a problem of doing some work outside of multiprocessing and the rest inside? Perhaps this could be addressed by moving statements into the multiprocessing functions rather than removing multiprocessing?

Can you say whether or not multiprocessing is no longer required for the use cases I mentioned above (may be the case)?

sloretz:
> Is this not a problem of doing some work outside of multiprocessing and the rest inside? Perhaps this could be addressed by moving statements into the multiprocessing functions rather than removing multiprocessing?

Hmm it could be fixed by making every test use `multiprocessing` instead of just these 3.

> Can you say whether or not multiprocessing is no longer required for the use cases I mentioned above (may be the case)?

I'll look into this. I wasn't aware we were testing multiple RMW implementations starting from the same pytest instance.

wjwwood:
We may not be anymore, but that's why we used it originally, in order to isolate which rmw implementation we imported.

sloretz:
> The reason for using multiprocessing in the past was to allow us to use more than one rmw implementation. Without multiprocessing you cannot test fastrtps and connext in the same pytest instance.

It looks like `rclpy` only runs its tests for a single RMW implementation. If we were going to add this I think the least effort way would be to call `ament_add_pytest_test()` for each RMW, meaning each is a different pytest invocation, and `multiprocessing` isn't needed.

> Also you cannot be sure that init/shutdown is isolated between tests.

It does look like `multiprocessing` is required for assurance that `init/shutdown` is isolated. This only works if all tests use `multiprocessing` otherwise state will be copied to the child process. This standard is unique to `rclpy`; `rclcpp` does not make every test in a test file run in a separate process, but `rclcpp` does have more isolation because each test file is a separate executable.

Rewriting all tests to use multiprocessing looks like a bit of work. There are more tests not using multiprocessing than using it. We could get more isolation in `rclpy` by calling `ament_add_pytest_test` for each test file instead of once for the whole test directory, which would make it more like `rclcpp`.

sloretz:
@wjwwood a3806cc calls `ament_add_pytest_test` for each test file. How does that approach sound?

wjwwood:
That's ok by me.

sloretz:
CI (testing rclpy)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6814/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3108/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5603/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6637/)

Assuming CI comes back green, I'll rebase and merge to preserve the commit renaming the action test files in e65e018

PR #330, opened 2019-04-29
Make `destroy_node` thread safe

sloretz:
Blocks #319

I was going to make this part of #319, but I think it would be easier to review separately. This makes `Node.handle` use the `Handle` class from #308. It also adds `handle.requires(another_handle)` to make sure `handle` is destroyed before `another_handle`.

## Problem this solves

#308 makes `destroy_subscription` thread safe by delaying destruction if the handle is currently being used. This makes it possible for a subscription to be destroyed after the node; which crashes when doing the same strategy with services using fast-rtps. See https://github.com/ros2/rclpy/pull/319#issuecomment-486860562 for a better explanation.

## Breaking changes
* Previously `node.destroy_node()` either returned `True` or raised an exception. I removed the return value since it was only ever `True`.
* Previously `node.handle` was set to `None` after the node was destroyed. Now it raises `InvalidHandle` when someone tries to use it after it is destroyed.
* previously `node.handle` was a pycapsule and could be used directly. Now one must use:
   ```python3
   with node.handle as node_capsule:
      ...
   ```

## How handle.requires() works
Subscriptions must be destroyed before the node that created them, so

```
subscription.handle.requires(node.handle)
```

When `node.handle.destroy()` is called the subscription handle is destroyed first. Once the node handle is no longer being used it will mark itself invalid so no one else can use it. Then the node handle tells the subscription handle to destroy itself. Once the subscription handle is no longer in use it will delete itself, and then tell the node it can destroy itself.

If the user does not call `destroy()` on either handle, the subscription handle holds a reference to the node handle so the subscription handle will be garbage collected first. The node handle only holds a weak reference to the subscription handle.

## Next steps

After incorporating feedback, assuming this approach is OK I would like to merge this PR. Then I'll rebase #319 to make all the other entities require the `node` so everything can be destroyed safely.

### DIFF
--- a/rclpy/rclpy/handle.py
+++ b/rclpy/rclpy/handle.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from threading import Lock
+from contextlib import ExitStack
+from threading import RLock
+import weakref
 
 from rclpy.impl.implementation_singleton import rclpy_pycapsule_implementation as _rclpy_capsule
 
@@ -44,7 +46,10 @@ class Handle:
         self.__use_count = 0
         self.__request_invalidation = False
         self.__valid = True
-        self.__lock = Lock()
+        self.__lock = RLock()
+        self.__required_handles = []
+        self.__dependent_handles = weakref.WeakSet()
+        self.__destroy_callbacks = []
         # Called to give an opportunity to raise an exception if the object is not a pycapsule.
         self.__capsule_name = _rclpy_capsule.rclpy_pycapsule_name(pycapsule)
         self.__capsule_pointer = _rclpy_capsule.rclpy_pycapsule_pointer(pycapsule)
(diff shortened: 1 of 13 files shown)


sloretz:
CI (testing just rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6831/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3126/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5618/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6656/)

rclpy/rclpy/handle.py
```diff
@@ -117,13 +147,48 @@ def __exit__(self, type_, value, traceback):
         self._return_capsule()
 
     def __destroy(self):
-        # Assume lock is held
-        assert not self.__lock.acquire(blocking=False)
-        # Assume capsule has not been destroyed
-        assert self.__valid
```
sloretz: Destruction can be delayed waiting for dependents. Previously `self.__valid = False` meant the capsule itself has been destroyed. Now setting it to false means `destroy()` was called and `__use_count` hit zero. The purpose is to prevent new users of this handle while waiting for dependent handles to be destroyed.

sloretz:
CI (testing rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6835/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3130/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5620/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6659/)

rclpy/rclpy/node.py
```diff
@@ -98,7 +99,7 @@ def __init__(
             services.
         :param initial_parameters: A list of parameters to be set during node creation.
         """
-        self._handle = None
+        self.__handle = None
```
dirk-thomas: Is this rename intentionally to break code using it directly? Or is it only changed to match the same attribute in other classes?

rclpy/rclpy/handle.py
```diff
@@ -44,7 +45,10 @@ def __init__(self, pycapsule):
         self.__use_count = 0
         self.__request_invalidation = False
         self.__valid = True
-        self.__lock = Lock()
+        self.__lock = RLock()
```
dirk-thomas: Maybe renaming to `__rlock` so when reading code using it it is clear that it is a recursive lock?

rclpy/rclpy/node.py
sloretz: The latter. I used `__handle` in `Subscription` so I changed it here too.

rclpy/rclpy/handle.py
sloretz: Renamed `__lock` -> `__rlock` in 8c95ea6

sloretz:
CI (testing rclpy with 8c95ea6)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6837/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3131/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5622/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6660/)

PR #332, opened 2019-05-01
Use new test interface definitions

jacobperron:
Connects to https://github.com/ros2/rcl_interfaces/issues/58

### DIFF
--- a/rclpy/test/test_node.py
+++ b/rclpy/test/test_node.py
@@ -75,7 +75,7 @@ class TestNode(unittest.TestCase):
     def test_create_raw_subscription(self):
         executor = SingleThreadedExecutor(context=self.context)
         executor.add_node(self.node)
-        primitives_pub = self.node.create_publisher(BasicTypes, 'raw_subscription_test')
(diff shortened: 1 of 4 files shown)


rclpy/test/test_node.py
```diff
@@ -75,15 +75,15 @@ def raw_subscription_callback(self, msg):
     def test_create_raw_subscription(self):
         executor = SingleThreadedExecutor(context=self.context)
         executor.add_node(self.node)
-        primitives_pub = self.node.create_publisher(Primitives, 'raw_subscription_test')
+        primitives_pub = self.node.create_publisher(BasicTypes, 'raw_subscription_test')
```
dirk-thomas: Maybe change the name of the variables too?

Same below: `primitives_msg`.

jacobperron: Missed it! Changed be3848c

sloretz:
@jacobperron FYI it looks like there's a small merge conflict (git doesn't know which order to add the new import statements)

jacobperron:
@sloretz I've rebased and also updated the changes from master to use the new service name `BasicTypes`.

PR #333, opened 2019-05-01
Make pub/sub/cli/srv/etc lists use @property on node

sloretz:
~~I would call this API neutral, but it is possible this raises an exception in already broken downstream code.~~ **Edit**: Changes APIs to be generators in 236b20a

Currently `Node` stores `publishers`, `subscriptions`, `clients`, ... in lists on itself. These names are pretty generic, and since we expect users to inherit from `Node` there is a risk of the names colliding. This is a problem because the executor uses these lists to add stuff to a wait set. This PR uses `@property` to at least prevent the attribute from being assigned a different object.

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from typing import Callable
+from typing import Iterator
 from typing import List
 from typing import Optional
 from typing import Tuple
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/node.py
```diff
@@ -153,6 +153,76 @@ def __init__(
         if start_parameter_services:
             self._parameter_service = ParameterService(self)
 
+    @property
+    def publishers(self) -> List[Publisher]:
+        """
+        Return publishers that have been created on this node.
+
+        .. warning::
+        The returned list must not be modified.
```
dirk-thomas: The API could provide a tuple or copied list instead?

sloretz: That means the executor would copy lists every time it rebuilds the wait set.

https://github.com/ros2/rclpy/blob/d672075cd5a821314af2e460af68b90373c5506f/rclpy/rclpy/executors.py#L455-L460

I don't know if that's a big deal; I could profile it.

dirk-thomas: Was just an idea to prevent callers from modifying the lists...

sloretz: Actually, this could `yield from` and be a generator to avoid copying stuff

sloretz: 236b20a makes these generators. This is definitely an API change now though.

dirk-thomas: Interesting idea. I don't think I have seen this combination. :+1:

sloretz:
CI (testing rclpy, I ran a local test using `--packages-above rclpy` and saw no test failures so I assume all downstream code is ok with these being generators)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6884/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3167/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5655/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6707/)

sloretz:
CI after rebasing
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6888/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3170/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5658/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6713/)

sloretz:
Since the rebase didn't change any lines (#332 changed adjacent lines and git didn't know which order to put the changes in) I'll assume that @dirk-thomas's approval still applies and merge.

PR #313, opened 2019-04-16
QoS - Expose the assert_liveliness API for Publishers and Nodes

emersonknapp:
Depends on ros2/rmw#171

Expose the rcl assert_liveliness APIs for Nodes and Publishers

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

PR #340, opened 2019-05-06
Fix assert_liveliness use of node and publisher capsules

sloretz:
I overlooked this when merging #313. The publisher and node pycapsules are managed by `self.handle` instead of using `node_handle` and `publisher_handle`.

There are no tests for these APIs at the moment.


sloretz:
CI (testing just rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6994/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3244/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5737/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6815/)

sloretz:
CI looks good. `test_timer` with `0.01` period failed, which is surprising, but is very unlikely to be caused by this PR.

PR #344, opened 2019-05-10
Encourage users to always provide a QoS history depth

### LINKED ISSUE #342: Encourage users to always specify QoS history depth
#### Feature description

Similar to what was done in `rclcpp` (https://github.com/ros2/rclcpp/pull/713), we want to add additional API that requires providing the history depth and deprecate existing API.

Since there is already a `QoSProfile`, we can probably just add an additional constructor that takes a history depth and log a deprecation warning if the default constructor is used.

jacobperron:
Resolves #342 

* Added a new parameter to methods for creating publishers and subscriptions

  The new parameter is required and must be a QoSProfile object or an integer value for the history depth.
  For now, it is possible to not use the parameter for backwards compatibility, but a deprecation warning is made.

* Add deprecation warning when constructing a QoSProfile without a history depth

  If the history setting is KEEP_ALL, then it is not required to pass the depth setting.

### DIFF
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -44,7 +44,8 @@ class QoSProfile:
             else:
                 warnings.warn(
                     "QoSProfile needs a 'history' and/or 'depth' setting when constructed",
-                    DeprecationWarning)
+                    DeprecationWarning,
+                    stacklevel=2)
         self.history = kwargs.get(
             'history',
             QoSHistoryPolicy.RMW_QOS_POLICY_HISTORY_SYSTEM_DEFAULT)
@@ -54,7 +55,8 @@ class QoSProfile:
         ):
             warnings.warn(
                 'A QoSProfile with history set to KEEP_LAST needs a depth specified',
-                DeprecationWarning)
+                DeprecationWarning,
+                stacklevel=2)
         self.depth = kwargs.get('depth', int())
         self.reliability = kwargs.get(
             'reliability',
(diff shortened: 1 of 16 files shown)


jacobperron:
@wjwwood I'm not sure if this exactly what we discussed. PTAL before I continue with services and actions.

rclpy/rclpy/qos.py
```diff
@@ -37,9 +38,19 @@ class QoSProfile:
     def __init__(self, **kwargs):
         assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
             'Invalid arguments passed to constructor: %r' % kwargs.keys()
+        if 'history' not in kwargs:
```
wjwwood: I think it would be good if this class `QoSProfile` could be created with just a `depth`, which this case doesn't allow for.

rclpy/test/test_node.py
```diff
@@ -123,6 +140,40 @@ def test_create_service(self):
         with self.assertRaisesRegex(ValueError, 'unknown substitution'):
             self.node.create_service(GetParameters, 'foo/{bad_sub}', lambda req: None)
 
+    def test_deprecation_warnings(self):
+        warnings.simplefilter('always')
```
wjwwood: We discussed offline that it would maybe be nice if `DeprecationWarnings` were turned on by default in rclpy itself, if possible only in rclpy. See the last example in:

https://docs.python.org/3/library/warnings.html#overriding-the-default-filter

jacobperron: I couldn't figure out how to get this to work. Including a line similar to that example in the top-level module (or submodule) doesn't seem to work as I expect. Maybe it is sufficient to set a filter to error if there is a warning in our unit tests (e.g. `warnings.simplefilter("error")`).

As for users, we should either [expect them to also be appropriately testing for deprecation warnings](https://docs.python.org/3/library/warnings.html#updating-code-for-new-versions-of-dependencies), or we can use a different type of warning that is not ignored by default.

jacobperron:
@wjwwood Thanks for the review!

I've address your comments and will continue to make similar changes for services and actions.

I've refactored the RMW to rclpy conversion function (ea2de90) so it constructs a `QoSProfile` with keyword arguments. This avoids deprecation warnings when importing `rclpy.qos` due to the construction of the default profiles.

rclpy/rclpy/qos.py
jacobperron: Done 1a47d59

jacobperron:
Rebased on master; ready for review.

jacobperron:
I forgot to find/replace usage of `qos_profile_default` with `QoSProfile(depth=10)`. Doing that now.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7064/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3292/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5788/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6880/)

PR #345, opened 2019-05-13
Parameter handling improvements.

### LINKED ISSUE #341: Follow-up on parameter API
This issue is to track some pending comments in #325, which should not modify existing method signatures:

- [x] Use `Optional` for `declare_parameters` so as not to require `None` values in the tuples ([comment](https://github.com/ros2/rclpy/pull/325#discussion_r280645899)).
- [x] Improve test cases; add tests for the remaining methods added.
- [ ] Check if name validation for parameters is appropriate.
- [x] Check for read-only parameters ([comment](https://github.com/ros2/rclpy/pull/325#discussion_r280239402)).
- [x] Introduce `set_parameter_descriptor` ([comment](https://github.com/ros2/rclpy/pull/325#discussion_r280534305)).
- [ ] Optional signature using `Parameter` as input for `declare_parameter` ([comment](https://github.com/ros2/rclpy/pull/325#discussion_r280534305)).

jubeira:
Closes #341.

There are some details that still need a bit of attention (see https://github.com/ros2/rclpy/issues/341#issuecomment-491292717). The PR includes a proposal for `set_descriptor` as described in the comment.

On the other hand, I'd like some input about the optional signature using `Parameter` as input for `declare_parameter` (see [comment](https://github.com/ros2/rclpy/pull/325#discussion_r280534305)). Is it worth implementing it even if it only applies to `declare_parameter` but not `declare_parameters`?

Other than that, this should be ready for an initial review.

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -18,6 +18,7 @@ from typing import List
 from typing import Optional
 from typing import Tuple
 from typing import TypeVar
+from typing import Union
 
 import weakref
 
@@ -308,7 +309,11 @@ class Node:
     def declare_parameters(
         self,
         namespace: str,
-        parameters: List[Tuple[str, Optional[ParameterValue], Optional[ParameterDescriptor]]]
+        parameters: List[Union[
+            Tuple[str],
+            Tuple[str, ParameterValue],
+            Tuple[str, ParameterValue, ParameterDescriptor],
+        ]]
     ) -> List[Parameter]:
         """
         Declare a list of parameters.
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/node.py
```diff
@@ -325,24 +330,56 @@ def declare_parameters(
...
+                        'is not a str.'.format_map(locals())
+                    )
+
+                # Get value from initial parameters, of from tuple if it doesn't exist.
+                if name in self._initial_parameters:
+                    value = self._initial_parameters[name].get_parameter_value()
+                else:
+                    value = parameter_tuple[1]
+                assert \
+                    isinstance(value, ParameterValue), \
```
wjwwood: Does this work if I do `('foo', 42)`? Or do you always have to do `('foo', ParameterValue(42))`?

rclpy/rclpy/node.py
```diff
@@ -651,6 +735,52 @@ def describe_parameters(self, names: List[str]) -> List[ParameterDescriptor]:
...
+        :param descriptor: New descriptor to apply to the parameter.
+        :param alternative_value: Value to set to the parameter if the existing value does not
+            comply with the new descriptor.
+        :return: ParameterValue for the given parameter name after applying the new descriptor.
+        :raises: ParameterNotDeclaredException if parameter had not been declared before
+            and undeclared parameters are not allowed.
+        :raises: ParameterImmutableException if the parameter exists and is read-only.
+        :raises: ParameterValueException if neither the existing value nor the alternative value
+            complies with the provided descriptor.
+        """
```
wjwwood: In particular, this looks fine.

wjwwood:
> On the other hand, I'd like some input about the optional signature using Parameter as input for declare_parameter (see comment). Is it worth implementing it even if it only applies to declare_parameter but not declare_parameters?

I don't think it is necessary. It just looked more consistent.

rclpy/rclpy/node.py
jubeira: `('foo', 42)` won't work; `ParameterValue` is actually the [message](https://github.com/ros2/rcl_interfaces/blob/master/rcl_interfaces/msg/ParameterValue.msg) defined in `rcl_interfaces`, so you need to specify the type and set the value where it corresponds (see this [test](https://github.com/ros2/rclpy/pull/345/files#diff-960e40f792ccd92281ee56dfb65726bcR671) as an example).

wjwwood: That's not very convenient, we have a separate class for that in C++ (separate from the message type) that can be conversion constructed from common parameter value types.

I guess that doesn't have to be improved in this pr, but it makes this method pretty ugly to use.

jubeira: That's indeed a reason to use `Parameter` directly for `declare_parameter` and `declare_parameters`, but we go back to the problem with the namespace in `declare_parameters` that we found out in the previous PR: https://github.com/ros2/rclpy/pull/325#discussion_r280233234.

I agree that it would be better to have a version that accepts a `Parameter` and a `ParameterDescriptor`, but I don't see a way of achieving it other than creating a copy of the parameter with the given namespace inside the method. Let's discuss this in the meeting, shall we?

jubeira: As discussed in the meeting, we shall move forward with this PR and introduce a way of converting basically any value to a `ParameterValue` message so that these methods are more convenient to use. One possibility is adding a new class (analogous to `Parameter`) that does the job.

jubeira:
CI: 
- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/7061/)
- Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/3303/)
- Mac OS: [Build Status](https://ci.ros2.org/job/ci_osx/5798/)
- Windows: [Build Status](https://ci.ros2.org/job/ci_windows/6890/)

@wjwwood Errors look unrelated; I think we can proceed with this one.

wjwwood:
Thanks @jubeira!

PR #347, opened 2019-05-15
Parameter flexibility enhancements

jubeira:
This PR is another iteration for #341.
- Addresses [this comment](https://github.com/ros2/rclpy/pull/345#discussion_r283552026) about `declare_parameter` and `declare_parameters` usability.
- Allows creating `Parameter` objects without specifying the type, which is deduced in the constructor without changing the API.
- Prevents service calls from raising exceptions for `set` and `describe` parameters (as in https://github.com/ros2/rclcpp/pull/718).

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Any
 from typing import Callable
 from typing import Iterator
 from typing import List
@@ -290,7 +291,7 @@ class Node:
     def declare_parameter(
         self,
         name: str,
-        value: ParameterValue = ParameterValue(),
+        value: Any = None,
         descriptor: ParameterDescriptor = ParameterDescriptor()
     ) -> Parameter:
         """
@@ -314,8 +315,8 @@ class Node:
         namespace: str,
         parameters: List[Union[
             Tuple[str],
-            Tuple[str, ParameterValue],
-            Tuple[str, ParameterValue, ParameterDescriptor],
+            Tuple[str, Any],
+            Tuple[str, Any, ParameterDescriptor],
         ]]
     ) -> List[Parameter]:
         """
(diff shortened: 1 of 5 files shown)


rclpy/rclpy/node.py
```diff
@@ -632,7 +629,19 @@ def _set_parameters_atomically(
...
 
         result = None
-        if self._parameters_callback:
+        # First check if there's a read-only parameter among the ones in the list.
+        # The first time a read-only parameter is declared it has no descriptor
+        # at this point.
+        if any(
+                self.describe_parameter(param.name).read_only for
+                param in parameter_list if
+                param.name in self._descriptors
```
wjwwood: nitpick: Indention looks weird here, shouldn't it only be indent 4 spaces from the column the `if` is in?

rclpy/rclpy/node.py
```diff
@@ -290,7 +291,7 @@ def get_logger(self):
     def declare_parameter(
         self,
         name: str,
-        value: ParameterValue = ParameterValue(),
+        value: Any = None,
```
wjwwood: I think the type checking here could be narrower, since `Parameter()` will not all just any type, there is a limit, though the union of possible types is large.

It would be nice to improve this, but necessary in this pr.

wjwwood: Edited the previous comment, meant to say `_not_ necessary`.

rclpy/rclpy/node.py
jubeira: Thanks; corrected in https://github.com/ros2/rclpy/pull/347/commits/2c9e657d3497c35037e4e80dac362bc12b532b71.

rclpy/rclpy/node.py
jubeira: Yes, I've thought about it but annotating all the types could make it a bit unreadable.
I'll have this in mind for the distro freeze.

jubeira:
CI:
- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/7106/)
- Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/3326/)
- Mac OS: [Build Status](https://ci.ros2.org/job/ci_osx/5820/)
- Windows: [Build Status](https://ci.ros2.org/job/ci_windows/6929/)

There were some failures but they seem unrelated (on `cpp` side).
I think it should be safe to merge.

Issue #353, opened 2019-05-17
Subscription with requested Durability TRANSIENT_LOCAL does not receive events that were published before Subscription created


emersonknapp:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Source
- Version or commit hash:
  - master (b8217b623a11687bbc0f1a558217d8b63c5534c1)
- DDS implementation:
  - rmw_fastrtps_cpp, rmw_connext_cpp
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue
* Create a Publisher with depth > 0, durability=RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL
* The publisher can be either `rclpy` or `rclcpp`, it works correctly either way
* Publish a message
* Create an rclpy Subscription with the same QoS policy as the Publisher
* Spin - you should receive the message that was published. An rclcpp Subscription will receive it, an rclpy Subscription will not
 

rclpy example, Subscription does not receive message
```
import rclpy
from rclpy.node import Node
from rclpy.qos import QoSDurabilityPolicy
from rclpy.qos import QoSProfile

from std_msgs.msg import String


def main(args=None):
    rclpy.init(args=args)
    topic = 'qos_durability_chatter'

    node = Node('durability_demo')
    qos_profile = QoSProfile(
        depth=1,
        durability=QoSDurabilityPolicy.RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL)
    publisher = node.create_publisher(String, topic, qos_profile)

    message = String()
    message.data = 'PUBLISHER SAYS HI'
    publisher.publish(message)

    subscription = node.create_subscription(  # noqa: F841
        String, topic,
        lambda msg: print('Subscription received: [{}]'.format(msg.data)),
        qos_profile)

    rclpy.spin(node)
    rclpy.shutdown()


if __name__ == '__main__':
    main()
```

rclcpp same example, Subscription DOES receive message
```
#include "rclcpp/rclcpp.hpp"

#include "std_msgs/msg/string.hpp"

int main(int argc, char * argv[])
{
  // Force flush of the stdout buffer
  setvbuf(stdout, NULL, _IONBF, BUFSIZ);

  // Configuration and Initialization
  rclcpp::init(argc, argv);
  rclcpp::executors::SingleThreadedExecutor exec;

  std::string topic("qos_durability_chatter");

[...]

PR #356, opened 2019-05-21
Initialize QoSProfile with values from rmw_qos_profile_default

emersonknapp:
To match behavior in rclcpp.

Fixes https://github.com/ros2/rclpy/issues/353

Would like some feedback on the approach, it feels a little kludgy. It's an awkward situation though, where the default profile values we want are only exposed via a QoSProfile instance from _rclpy.c, so we have a circular dependency.

### DIFF
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from argparse import Namespace
 from enum import Enum
 from enum import IntEnum
 import warnings
@@ -19,6 +20,10 @@ import warnings
 from rclpy.duration import Duration
 from rclpy.impl.implementation_singleton import rclpy_action_implementation as _rclpy_action
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+
+# "Forward-declare" this value so that it can be used in the QoSProfile initializer.
+# It will have a value by the end of definitions, before user code runs.
+_qos_default_value_for_init = []
 
 
 class QoSProfile:
@@ -39,6 +44,17 @@ class QoSProfile:
     def __init__(self, **kwargs):
         assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
             'Invalid arguments passed to constructor: %r' % kwargs.keys()
+
+        if not len(_qos_default_value_for_init):
+            # Any of the setters, upon receiving these None values, would assert.
+            # This can't happen though, because this is only used at definition time, when
+            # we are initializing fully-defined preset profiles.
+            from_profile = Namespace(**{
+                slot[1:]: None for slot in self.__slots__
+            })
+        else:
+            from_profile = _qos_default_value_for_init[0]
+
         if 'history' not in kwargs:
             if 'depth' in kwargs:
                 kwargs['history'] = QoSHistoryPolicy.RMW_QOS_POLICY_HISTORY_KEEP_LAST
@@ -46,9 +62,7 @@ class QoSProfile:
                 warnings.warn(
                     "QoSProfile needs a 'history' and/or 'depth' setting when constructed",
                     stacklevel=2)
-        self.history = kwargs.get(
-            'history',
-            QoSHistoryPolicy.RMW_QOS_POLICY_HISTORY_SYSTEM_DEFAULT)
+        self.history = kwargs.get('history', from_profile.history)
         if (
             QoSHistoryPolicy.RMW_QOS_POLICY_HISTORY_KEEP_LAST == self.history and
             'depth' not in kwargs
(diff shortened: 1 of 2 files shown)


emersonknapp:
@thomas-moulard - please run this CI job
Gist: https://gist.githubusercontent.com/emersonknapp/a3bb0ea80650afa69910ebe9d0ac4a06/raw/be2aecf41e6fdfd09e8c41969684627ac4ae7ce0/ros2.repos
BUILD args: --packages-up-to rclpy
TEST args: --packages-select rclpy
Job: ci_launcher

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7190/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3400/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5896/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7026/)

rclpy/rclpy/qos.py
```diff
@@ -39,39 +44,43 @@ class QoSProfile:
     def __init__(self, **kwargs):
         assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
             'Invalid arguments passed to constructor: %r' % kwargs.keys()
+
+        if not len(_qos_default_value_for_init):
+            # Any of the setters, upon receiving these None values, would assert.
+            # This can't happen though, because this is only used at definition time, when
+            # we are initializing fully-defined preset profiles.
+            from_profile = Namespace(**{
```
nburek: It feels dirty that we're relying on an instance of `QoSProfile` in the constructor for `QoSProfile` and that we're essentially mocking it to fake that it exists so we can construct the initial instance. The only options around this I can think of though would be to switch to a Factory pattern, which would be ton of work, or creating a new function in `_rclpy` that would return a more raw representation of the policy, like in the form of a dict, instead of relying on the `QoSProfile` class.

emersonknapp: Like you said, the only way we expose this information to the python layer is via this QoSProfile class

rclpy/rclpy/qos.py
```diff
@@ -325,9 +334,11 @@ def __init__(self, qos_profile: QoSProfile, profile_name: str):
         self.name = profile_name
 
 
-__qos_profile_default = _rclpy.rclpy_get_rmw_qos_profile(
+_qos_profile_default = _rclpy.rclpy_get_rmw_qos_profile(
```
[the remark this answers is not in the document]
emersonknapp: That was the original thought, but then I thought I wasn't getting the updated value... but that was an unrelated problem. I'll go back to just the variable.

rclpy/rclpy/qos.py
```diff
@@ -343,6 +352,7 @@ def __init__(self, qos_profile: QoSProfile, profile_name: str):
 
 
 class QoSPresetProfiles(Enum):
+    DEFAULT = _qos_profile_default
```
nburek: Nitpick: Unless there is a strong reason for providing this I think it just causes confusion on what the difference is between `DEFAULT` and `SYSTEM_DEFAULT`. Not to mention that the settings should be the same for them, with the only difference as far as I can tell being that `SYSTEM_DEFAULT` has a `name` attribute set because it's using the `DeprecatedQoSProfile`.

emersonknapp: That's not quite correct. `DEFAULT` is using `rmw_qos_profile_default`https://github.com/ros2/rmw/blob/master/rmw/include/rmw/qos_profiles.h#L51 - the "chosen default settings for ros2 publishers and subscriptions", whereas `SYSTEM_DEFAULT` is a different qos profile `rmw_qos_profile_system_default` https://github.com/ros2/rmw/blob/master/rmw/include/rmw/qos_profiles.h#L51 - which sets all DDS Policy settings to SYSTEM_DEFAULT, basically using the rmw implementation's default settings.

These two profiles are slightly different - notably (and why this "bug" existed) `rmw_qos_profle_default` chooses `reliability=RELIABLE`, whereas SYSTEM_DEFAULT I think evaluates to `BEST_EFFORT` in Fast-RTPS

emersonknapp: I do agree, however, that it's confusing and not documented. I think it is worth a pass to document the rmw/qos_profiles.h and places that mirror it e.g. here. However, I think that's a separate change from this one.

nburek: You're right. I misread the diff. Also agree that updating documentation is separate from this.

jacobperron: Using the "default" profile is deprecated, and will be removed in the future. I'd rather not add it as a preset profile, since we discourage it's use.
I understand that the default profile is still in use at the rmw/rcl layers, but I would expect this to be removed in the future (I may be wrong about this). At the very least, we've decided not to allow using it at the client library level.

emersonknapp: @jacobperron To try and clarify what you're saying - do you mean we're deprecating _exposing_ `rmw_qos_profile_default` to users? Because it seems that we do intend it to be the implicit default, based on the fairly new QoS class in rclcpp - https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/qos.hpp#L62

jacobperron: Yes, we are deprecating exposing it to be used directly. Regarding the linked rclcpp code, some settings of the rmw profile (ie. history and depth) are overridden by the `QoSInitialization` object passed as the first argument. The QoSInitialization settings need to be explicitly set by the user.

---

So, while I think initializing `QoSProfile` internally with the default profile is fine, we should not enable users to use the default profile directly, without them being explicit about the history and/or depth setting.

emersonknapp: Ok, got it. Removing it from `QoSPresetProfiles`. The history/depth requirement is unchanged by this PR, i.e. it still generates warnings when neither is specified.

rclpy/test/test_qos.py
```diff
@@ -99,8 +100,9 @@ def test_deprecation_warnings(self):
         with warnings.catch_warnings(record=True) as w:
             warnings.simplefilter('always')
             QoSProfile()
-            assert len(w) == 1
+            assert len(w) == 2  # must supply depth or history, _and_ KEEP_LAST needs depth
             assert issubclass(w[0].category, UserWarning)
+
```
jacobperron: nit: this whitespace change isn't necessary.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7211/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3410/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5906/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7036/)

rclpy/rclpy/qos.py
jacobperron: I think it's worth changing the implementation in `_rclpy` so it's not returning a `QoSProfile`. It has been brought up that the current implementation is less than desirable. I've created an issue https://github.com/ros2/rclpy/issues/358 that we can hopefully pick up after in the near future.

PR #357, opened 2019-05-23
Enforcing parameter ranges.

### LINKED ISSUE #350: Range support for parameters
## Feature request
Support enforcing parameter ranges according to `ParameterDescriptor`.
#### Feature description
See https://github.com/ros2/rcl_interfaces/pull/72 for related discussion.

Current parameter API allows setting descriptors, but the ranges are not currently being enforced.

jubeira:
Closes #350.

See https://github.com/ros2/rclcpp/issues/728 for details and related discussion.

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -12,8 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import math
+
 from typing import Any
 from typing import Callable
+from typing import Dict
 from typing import Iterator
 from typing import List
 from typing import Optional
@@ -24,6 +27,8 @@ from typing import Union
 import warnings
 import weakref
 
+from rcl_interfaces.msg import FloatingPointRange
+from rcl_interfaces.msg import IntegerRange
 from rcl_interfaces.msg import Parameter as ParameterMsg
 from rcl_interfaces.msg import ParameterDescriptor
 from rcl_interfaces.msg import ParameterEvent
@@ -80,6 +85,9 @@ SrvTypeResponse = TypeVar('SrvTypeResponse')
 
 
 class Node:
+
+    PARAM_REL_TOL = 1e-6
+
     """
     A Node in the ROS graph.
 
@@ -337,7 +345,7 @@ class Node:
         :raises: TypeError if any tuple in :param:`parameters` does not match the annotated type.
         """
         parameter_list = []
-        descriptor_list = []
+        descriptors = {}
         for index, parameter_tuple in enumerate(parameters):
             if len(parameter_tuple) < 1 or len(parameter_tuple) > 3:
                 raise TypeError(
@@ -387,7 +395,7 @@ class Node:
             validate_parameter_name(full_name)
 
             parameter_list.append(Parameter(full_name, type_, value))
-            descriptor_list.append(descriptor)
+            descriptors.update({full_name: descriptor})
 
         parameters_already_declared = [
             parameter.name for parameter in parameter_list if parameter.name in self._parameters
@@ -398,7 +406,7 @@ class Node:
         # Call the callback once for each of the parameters, using method that doesn't
         # check whether the parameter was declared beforehand or not.
         self._set_parameters(
-            parameter_list, descriptor_list=descriptor_list, raise_on_failure=True)
+            parameter_list, descriptors=descriptors, raise_on_failure=True)
         return self.get_parameters([parameter.name for parameter in parameter_list])
 
     def undeclare_parameter(self, name: str):
@@ -510,7 +518,7 @@ class Node:
     def _set_parameters(
         self,
         parameter_list: List[Parameter],
-        descriptor_list: Optional[List[ParameterDescriptor]] = None,
+        descriptors: Optional[Dict[str, ParameterDescriptor]] = None,
         raise_on_failure=False,
         check_undeclared_parameters=False
     ) -> List[SetParametersResult]:
@@ -528,7 +536,7 @@ class Node:
         according to `raise_on_failure` flag.
 
         :param parameter_list: List of parameters to set.
-        :param descriptor_list: List of descriptors to set to the given parameters.
+        :param descriptors: Descriptors to set to the given parameters.
             If a list is given, it must have the same size as the parameter list.
         :param raise_on_failure: True if InvalidParameterValueException has to be raised when
             the user callback rejects a parameter, False otherwise.
@@ -538,19 +546,19 @@ class Node:
         :raises: InvalidParameterValueException if the user-defined callback rejects the
             parameter value and raise_on_failure flag is True.
         """
-        if descriptor_list is not None:
-            assert len(descriptor_list) == len(parameter_list)
+        if descriptors is not None:
+            assert all(parameter.name in descriptors for parameter in parameter_list)
 
         results = []
-        for index, param in enumerate(parameter_list):
+        for param in parameter_list:
             if check_undeclared_parameters:
                 self._check_undeclared_parameters([param])
             result = self._set_parameters_atomically(
                 [param],
-                None if descriptor_list is None else [descriptor_list[index]]
+                descriptors
             )
             if raise_on_failure and not result.successful:
-                raise InvalidParameterValueException(param.name, param.value)
+                raise InvalidParameterValueException(param.name, param.value, result.reason)
             results.append(result)
         return results
 
@@ -604,7 +612,7 @@ class Node:
     def _set_parameters_atomically(
         self,
         parameter_list: List[Parameter],
-        descriptor_list: Optional[List[ParameterDescriptor]] = None
+        descriptors: Optional[Dict[str, ParameterDescriptor]] = None
     ) -> SetParametersResult:
         """
         Set the given parameters, all at one time, and then aggregate result.
@@ -623,24 +631,16 @@ class Node:
         something else, then the parameter will be implicitly undeclared.
 
         :param parameter_list: The list of parameters to set.
+        :param descriptors: New descriptors to apply to the parameters before setting them.
         :return: Aggregate result of setting all the parameters atomically.
         """
-        if descriptor_list is not None:
-            assert len(descriptor_list) == len(parameter_list)
+        if descriptors is not None:
+            assert all(parameter.name in descriptors for parameter in parameter_list)
 
-        result = None
-        # First check if there's a read-only parameter among the ones in the list.
-        # The first time a read-only parameter is declared it has no descriptor
-        # at this point.
-        if any(
-            self.describe_parameter(param.name).read_only for
-            param in parameter_list if
-            param.name in self._descriptors
-        ):
-            result = SetParametersResult(
-                successful=False,
-                reason='Trying to set a read-only parameter.'
-            )
+        # If new descriptors are not provided, check for read-only.
+        result = self._apply_descriptors(parameter_list, descriptors, descriptors is None)
+        if not result.successful:
+            return result
         elif self._parameters_callback:
             result = self._parameters_callback(parameter_list)
         else:
@@ -654,7 +654,7 @@ class Node:
             else:
                 parameter_event.node = self.get_namespace() + '/' + self.get_name()
 
-            for index, param in enumerate(parameter_list):
+            for param in parameter_list:
                 if Parameter.Type.NOT_SET == param.type_:
                     if Parameter.Type.NOT_SET != self.get_parameter_or(param.name).type_:
                         # Parameter deleted. (Parameter had value and new value is not set)
@@ -670,8 +670,8 @@ class Node:
                     # Update descriptors; set a default if it doesn't exist.
                     # Don't update if it already exists for the current parameter and a new one
                     # was not specified in this method call.
-                    if descriptor_list is not None:
-                        self._descriptors[param.name] = descriptor_list[index]
+                    if descriptors is not None:
+                        self._descriptors[param.name] = descriptors[param.name]
                     elif param.name not in self._descriptors:
                         self._descriptors[param.name] = ParameterDescriptor()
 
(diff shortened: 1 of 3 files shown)


jubeira:
Partial CI:
- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/7210/).

Running remaining jobs now.

jubeira:
Full CI:
- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/7210/)
- Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/3416/)
- Mac OS: [Build Status](https://ci.ros2.org/job/ci_osx/5912/)
- Windows: [Build Status](https://ci.ros2.org/job/ci_windows/7043/)

rclpy/rclpy/node.py
```diff
@@ -528,7 +536,7 @@ def _set_parameters(
         according to `raise_on_failure` flag.
 
         :param parameter_list: List of parameters to set.
-        :param descriptor_list: List of descriptors to set to the given parameters.
+        :param descriptors: Descriptors to set to the given parameters.
             If a list is given, it must have the same size as the parameter list.
```
ivanpauno: If a list is given ... -> If a dictionary is given, each parameter should have an entry in it.

rclpy/rclpy/node.py
```diff
@@ -682,25 +682,180 @@ def _set_parameters_atomically(
...
+                       'Min: {}, Max: {}, value: {}'.format(
+                            parameter.name, min_value, max_value, parameter.value
+                        )
+            )
+
+        if integer_range.step != 0 and (parameter.value - min_value) % integer_range.step != 0:
+            return SetParametersResult(
+                successful=False,
+                reason='The distance between the parameter value for {}'
+                       'is not a multiple of step.\n'
```
ivanpauno: The distance between the parameter value, and what ...?? It sounds strange to me.
Maybe, `the parameter value is not one of the allowed steps`.

rclpy/rclpy/node.py
```diff
@@ -747,6 +902,8 @@ def set_descriptor(
         """
         Set a new descriptor for a given parameter.
 
+        The name in the descriptor is ignored and set to :param:`name`.
```
ivanpauno: I know this is out of the scope of this PR, but what is the motivation of `set_descriptor`?
I don't see why someone would want to use it. Maybe, it shouldn't be allowed (we don't have an equivalent in rclcpp).

rclpy/rclpy/node.py
```diff
@@ -775,14 +932,20 @@ def set_descriptor(
...
         # First try keeping the parameter, then try the alternative one.
-        if (
-            self._apply_descriptor_and_set(current_parameter, descriptor) or
-            self._apply_descriptor_and_set(alternative_parameter, descriptor)
-        ):
-            self._descriptors[name] = descriptor
-            return self.get_parameter(name).get_parameter_value()
-        else:
-            raise InvalidParameterValueException(name)
+        # Don't check for read-only since we are applying a new descriptor now.
```
ivanpauno: This sounds strange

rclpy/test/test_node.py
```diff
@@ -1212,6 +1238,184 @@ def test_set_descriptor_read_only(self):
...
             )
 
+    def test_floating_point_range_descriptor(self):
+        # OK cases; non-floats are not affected by the range.
+        fp_range = FloatingPointRange(from_value=0.0, to_value=10.0, step=0.5)
+        parameters = [
+            ('from_value', 0.0, ParameterDescriptor(floating_point_range=[fp_range])),
+            ('to_value', 10.0, ParameterDescriptor(floating_point_range=[fp_range])),
+            ('in_range', 4.5, ParameterDescriptor(floating_point_range=[fp_range])),
+            ('str_value', 'I am no integer', ParameterDescriptor(floating_point_range=[fp_range])),
```
ivanpauno: I am not integer -> I am not float

rclpy/test/test_node.py
```diff
@@ -1212,6 +1238,184 @@ def test_set_descriptor_read_only(self):
...
+        # Parameters that don't comply with the description will raise an exception.
+        fp_range = FloatingPointRange(from_value=-10.0, to_value=0.0, step=30.0)
+        parameters = [
+            ('from_value_2', -10.0, ParameterDescriptor(floating_point_range=[fp_range])),
+            ('to_value_2', 0.0, ParameterDescriptor(floating_point_range=[fp_range])),
+            ('in_range_bad_step', -4.5, ParameterDescriptor(floating_point_range=[fp_range])),
+            ('out_of_range', 30.0, ParameterDescriptor(floating_point_range=[fp_range]))
+        ]
+        with self.assertRaises(InvalidParameterValueException):
+            self.node.declare_parameters('', parameters)
```
ivanpauno: Should `declare_parameters` raise, or simply return a not successful result (in the corresponding list item).
Doing both things sounds strange to me.

rclpy/test/test_node.py
```diff
@@ -1212,6 +1238,184 @@ def test_set_descriptor_read_only(self):
...
+        # Parameters that don't comply with the description will raise an exception.
+        integer_range = IntegerRange(from_value=-10, to_value=0, step=30)
+        parameters = [
+            ('from_value_2', -10, ParameterDescriptor(integer_range=[integer_range])),
+            ('to_value_2', 0, ParameterDescriptor(integer_range=[integer_range])),
+            ('in_range_bad_step', -4, ParameterDescriptor(integer_range=[integer_range])),
+            ('out_of_range', 30, ParameterDescriptor(integer_range=[integer_range]))
+        ]
+        with self.assertRaises(InvalidParameterValueException):
+            self.node.declare_parameters('', parameters)
```
ivanpauno: same

rclpy/rclpy/node.py
jubeira: This was part of the scope of https://github.com/ros2/rclpy/pull/345; here's the comment where it was proposed: https://github.com/ros2/rclpy/pull/325#discussion_r280534305.

I think it makes sense in some cases -perhaps not most of them- but it's not bad either. You could change the limits for a parameter without requiring to re-declare it, or you could make it read-only. On top of that, you can specify the descriptor after declaring the parameter.

rclpy/rclpy/node.py
jubeira: The whole descriptor is applied in a single method, which makes sense to me.

When you set a new descriptor that is read-only, you need to skip that check. You can't modify a parameter that is read-only (nor modify its descriptor), but if the existing descriptor is read-only but the new one is not, that's a valid use-case. Therefore, that check is skipped for new descriptors.

The same applies when declaring a new parameter. The first (and only) time the value is set, you can't check for read-only attribute.

rclpy/test/test_node.py
jubeira: Declaring parameters is something you should do once in your program, and as a user you should probably know the constraints and the initial value beforehand. Following that line, I think it makes sense to raise an exception.

This is the case in `rclcpp` AFAIK: https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node.hpp#L390-L391.

On the other hand, parameters are declared in order. If the first ones in the list are correct, they will be declared until the 'faulty' one is reached. That part is explicitly documented like that.

rclpy/test/test_node.py
jubeira: Thanks; addressed in https://github.com/ros2/rclpy/pull/357/commits/7381f62aeade16395c33dee943d7a5aa94173062.

rclpy/rclpy/node.py
jubeira: Thanks; addressed in https://github.com/ros2/rclpy/pull/357/commits/7381f62aeade16395c33dee943d7a5aa94173062.

rclpy/rclpy/node.py
jubeira: Thanks; addressed in https://github.com/ros2/rclpy/pull/357/commits/7381f62aeade16395c33dee943d7a5aa94173062.

jubeira:
CI after comments:

- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/7234/)
- Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/3426/)
- Mac OS: [Build Status](https://ci.ros2.org/job/ci_osx/5922/)
- Windows: [Build Status](https://ci.ros2.org/job/ci_windows/7056/)

rclpy/test/test_node.py
```diff
@@ -1135,9 +1135,10 @@ def test_describe_parameters(self):
         self.assertEqual(foo_descriptor.integer_range[0].to_value, 10)
         self.assertEqual(foo_descriptor.integer_range[0].step, 2)
 
+        # The descriptor gets the type of the parameter.
         bar_descriptor = descriptor_list[1]
         self.assertEqual(bar_descriptor.name, 'bar')
-        self.assertEqual(bar_descriptor.type, ParameterType.PARAMETER_INTEGER)
+        self.assertEqual(bar_descriptor.type, ParameterType.PARAMETER_STRING)
```
jacobperron: Can we instead declare the variable with an integer type so not all of the tests have type string?

rclpy/test/test_node.py
```diff
@@ -1212,6 +1238,184 @@ def test_set_descriptor_read_only(self):
...
+
+        # Try to set a parameter out of range (bad step).
+        result = self.node.set_parameters([Parameter('in_range', value=4.25)])
+        self.assertIsInstance(result, list)
+        self.assertIsInstance(result[0], SetParametersResult)
+        self.assertFalse(result[0].successful)
+        self.assertEqual(self.node.get_parameter('in_range').value, 4.5)
+
+        # Change in_range parameter to int; ranges will not apply.
+        result = self.node.set_parameters([Parameter('in_range', value=12)])
```
jacobperron: That's subtle. I hope users don't forget (or accidentally add) a decimal point in their code, otherwise they may experience unexpected behavior. I guess this is a case for type constraints.

rclpy/rclpy/node.py
jacobperron: IIUC, we can currently change a read-only parameter by changing it's descriptor to `read_only=False` and then updating the parameter. My gut reaction is that the descriptor shouldn't be able to change for a read-only parameter, though I'm okay if we leave the behavior as-is for now and update later if we change our mind.

jubeira: Not exactly; the `read_only` condition of the existing descriptor [is checked](https://github.com/ros2/rclpy/pull/357/files#diff-129fafcf34261ae2f52bd43718815308R922) before attempting to set a new descriptor.

> My gut reaction is that the descriptor shouldn't be able to change for a read-only parameter

That's correct, read-only parameters cannot be modified using this method either.

rclpy/test/test_node.py
jubeira: Indeed.
Note that in this case I'm making the `Parameter` guess the type; using the old API where the type had to be explicitly given to the parameter you wouldn't have this problem.

In any case, I also think the solution is eventually enforcing the types (now for Dashing, or later for a patch release).

rclpy/rclpy/node.py
jacobperron: Thanks for the clarification.

rclpy/rclpy/node.py
```diff
@@ -623,24 +631,16 @@ def _set_parameters_atomically(
...
-        ):
-            result = SetParametersResult(
-                successful=False,
-                reason='Trying to set a read-only parameter.'
-            )
+        if descriptors is not None:
+            assert all(parameter.name in descriptors for parameter in parameter_list)
+
+        # If new descriptors are not provided, check for read-only.
+        result = self._apply_descriptors(parameter_list, descriptors, descriptors is None)
```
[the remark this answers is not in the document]
jubeira: Good idea; addressed in https://github.com/ros2/rclpy/pull/357/commits/b093de0109f78044588ab109de2857e343b788d4.

rclpy/rclpy/node.py
```diff
@@ -682,25 +682,178 @@ def _set_parameters_atomically(
...
+        check_read_only: bool = True
+    ) -> SetParametersResult:
+        """
+        Apply descriptors to parameters and return an aggregated result without saving parameters.
+
+        In case no descriptors are provided to the method, existing descriptors shall be used.
+        In any case, if a given parameter doesn't have a descriptor it shall be skipped.
+
+        :param parameter_list: Parameters to be checked.
+        :param descriptors: Descriptors to apply. If None, the stored descriptors for the given
```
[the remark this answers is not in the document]
jubeira: Addressed in https://github.com/ros2/rclpy/pull/357/commits/b093de0109f78044588ab109de2857e343b788d4.

rclpy/rclpy/node.py
```diff
@@ -682,25 +682,178 @@ def _set_parameters_atomically(
...
+        min_value = min(integer_range.from_value, integer_range.to_value)
+        max_value = max(integer_range.from_value, integer_range.to_value)
+
+        # Values in the edge are always OK.
+        if parameter.value == min_value or parameter.value == max_value:
+            return SetParametersResult(successful=True)
 
-        self._parameters[parameter.name] = parameter
-        return True
+        if parameter.value < min_value or parameter.value > max_value:
```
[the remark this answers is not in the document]
jubeira: Nice catch.
Addressed in https://github.com/ros2/rclpy/pull/357/commits/b093de0109f78044588ab109de2857e343b788d4.

rclpy/rclpy/node.py
```diff
@@ -682,25 +682,178 @@ def _set_parameters_atomically(
...
+            return SetParametersResult(
+                successful=False,
+                reason='Parameter {} out of range\n'
```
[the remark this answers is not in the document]
jubeira: Good catch; removed in https://github.com/ros2/rclpy/pull/357/commits/b093de0109f78044588ab109de2857e343b788d4.

rclpy/test/test_node.py
jubeira: Done in https://github.com/ros2/rclpy/pull/357/commits/b093de0109f78044588ab109de2857e343b788d4.

jubeira:
Thanks for the thoughtful comments @jacobperron!
They should be addressed now; I'll run CI once more.

jubeira:
CI:
- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/7250/)
- Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/3431/)
- Mac OS: [Build Status](https://ci.ros2.org/job/ci_osx/5928/)
- Windows: [Build Status](https://ci.ros2.org/job/ci_windows/7063/)

jubeira:
Failure on Windows seems to be unrelated; merging now!

nuclearsandwich:
The Windows CI failure has appeared before and is in test_rclcpp. When [Build Status](https://ci.ros2.org/job/ci_windows/7069/) runs it should tell is whether we made the right call.

PR #339, opened 2019-05-05
Setting automatic declaration for initial parameters to False.

jubeira:
Follow-up on #325; addressing this pending [comment](https://github.com/ros2/rclpy/pull/325#discussion_r280966366).

PR #363, opened 2019-05-28
Rename parameter options

wjwwood:
Connects to ros2/ros2_documentation#242

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -111,7 +111,7 @@ def create_node(
     namespace: str = None,
     use_global_arguments: bool = True,
     start_parameter_services: bool = True,
-    initial_parameters: List[Parameter] = None,
+    parameter_overrides: List[Parameter] = None,
     allow_undeclared_parameters: bool = False,
     automatically_declare_initial_parameters: bool = True
 ) -> 'Node':
@@ -127,7 +127,8 @@ def create_node(
     :param use_global_arguments: ``False`` if the node should ignore process-wide command line
         arguments.
     :param start_parameter_services: ``False`` if the node should not create parameter services.
-    :param initial_parameters: A list of :class:`.Parameter` to be set during node creation.
+    :param parameter_overrides: A list of :class:`.Parameter` which are used to override the
+        initial values of parameters declared on this node.
     :param allow_undeclared_parameters: True if undeclared parameters are allowed, False otherwise.
     :param automatically_declare_initial_parameters: True if initial parameters have to be declared
         upon node creation, false otherwise.
@@ -139,7 +140,7 @@ def create_node(
         node_name, context=context, cli_args=cli_args, namespace=namespace,
         use_global_arguments=use_global_arguments,
         start_parameter_services=start_parameter_services,
-        initial_parameters=initial_parameters,
+        parameter_overrides=parameter_overrides,
         allow_undeclared_parameters=allow_undeclared_parameters,
         automatically_declare_initial_parameters=automatically_declare_initial_parameters)
 
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -129,7 +129,8 @@ def create_node(
     :param start_parameter_services: ``False`` if the node should not create parameter services.
     :param parameter_overrides: A list of :class:`.Parameter` which are used to override the
         initial values of parameters declared on this node.
-    :param allow_undeclared_parameters: True if undeclared parameters are allowed, False otherwise.
+     :param allow_undeclared_parameters: True if undeclared parameters are allowed, False otherwise.
+        This option doesn't affect `parameter_overrides`.
     :param automatically_declare_parameters_from_overrides: If True, the "parameter overrides" will
         be used to implicitly declare parameters on the node during creation.
     :return: An instance of the newly created node.
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -129,10 +129,10 @@ def create_node(
     :param start_parameter_services: ``False`` if the node should not create parameter services.
     :param parameter_overrides: A list of :class:`.Parameter` which are used to override the
         initial values of parameters declared on this node.
-     :param allow_undeclared_parameters: True if undeclared parameters are allowed, False otherwise.
+     :param allow_undeclared_parameters: if True undeclared parameters are allowed, default False.
         This option doesn't affect `parameter_overrides`.
     :param automatically_declare_parameters_from_overrides: If True, the "parameter overrides" will
-        be used to implicitly declare parameters on the node during creation.
+        be used to implicitly declare parameters on the node during creation, default False.
     :return: An instance of the newly created node.
     """
     # imported locally to avoid loading extensions on module import
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -113,7 +113,7 @@ def create_node(
     start_parameter_services: bool = True,
     parameter_overrides: List[Parameter] = None,
     allow_undeclared_parameters: bool = False,
-    automatically_declare_initial_parameters: bool = True
+    automatically_declare_parameters_from_overrides: bool = True
 ) -> 'Node':
     """
     Create an instance of :class:`.Node`.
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/node.py
```diff
@@ -166,14 +167,14 @@ def __init__(
             ParameterEvent, 'parameter_events', qos_profile_parameter_events)
 
         with self.handle as capsule:
-            self._initial_parameters = _rclpy.rclpy_get_node_parameters(Parameter, capsule)
+            self._parameter_overrides = _rclpy.rclpy_get_node_parameters(Parameter, capsule)
```
jubeira: Just a suggestion: consider keeping the original name `_initial_parameters` for this variable. I think it's a bit more clear and it would better match the [description for the argument](https://github.com/ros2/rclpy/pull/363/files#diff-239a24d897ab2e601141491ac46633f9R131) if "initial parameters" retrieved by the node are overridden by "parameter overrides" given when the node is created.

Feel free to disregard this suggestion if you think it's fine like this.

wjwwood: I pretty much just blindly find-replaced it and then looked around each change to update comments and docs, so none of it was super intentional. I'll have another look at this particular case.

wjwwood: I kind of feel like this new name is appropriate, e.g. this snippet reads nicely to me:

https://github.com/ros2/rclpy/blob/fca845ba3113e7f43a5a1de899e13242ee1094d8/rclpy/rclpy/node.py#L364-L370

The one case where it doesn't mesh so well is here:

https://github.com/ros2/rclpy/blob/fca845ba3113e7f43a5a1de899e13242ee1094d8/rclpy/rclpy/node.py#L169-L170

But I think that's mostly due to the vaguely named `rclpy_get_node_parameters` (e.g. is that the current nodes? apparently not, so is it the parameters from the node options? no that's in rclpy only... so I guess its from YAML files, etc? It's not clear at all to me from the name).

I'm inclined to leave it, but if you or someone else still thinks it should be reverted I'm happy to do it since it's not user facing.

jubeira: Sounds good to keep it then; it wasn't a super strong opinion, and this is a good point so I'm also fine with it :+1:

rclpy/rclpy/__init__.py
```diff
@@ -127,10 +127,11 @@ def create_node(
     :param use_global_arguments: ``False`` if the node should ignore process-wide command line
         arguments.
     :param start_parameter_services: ``False`` if the node should not create parameter services.
-    :param initial_parameters: A list of :class:`.Parameter` to be set during node creation.
+    :param parameter_overrides: A list of :class:`.Parameter` which are used to override the
+        initial values of parameters declared on this node.
     :param allow_undeclared_parameters: True if undeclared parameters are allowed, False otherwise.
```
chapulina:
```suggestion
     :param allow_undeclared_parameters: True if undeclared parameters are allowed, False otherwise.
        This option doesn't affect `parameter_overrides`.
```

rclpy/rclpy/node.py
```diff
@@ -113,11 +113,12 @@ def __init__(
             args.
         :param start_parameter_services: ``False`` if the node should not create parameter
             services.
-        :param initial_parameters: A list of parameters to be set during node creation.
+        :param parameter_overrides: A list of overrides for initial values for parameters declared
```
chapulina: Same comment as on `rclcpp`, I think that mentioning "initial values" is misleading, because these behave more like the final values.

wjwwood: They are not final values... They only apply when being declared. If you declare it and then set it afterwards the override doesn't affect the set.

rclpy/rclpy/__init__.py
```diff
@@ -111,9 +111,9 @@ def create_node(
     namespace: str = None,
     use_global_arguments: bool = True,
     start_parameter_services: bool = True,
-    initial_parameters: List[Parameter] = None,
+    parameter_overrides: List[Parameter] = None,
     allow_undeclared_parameters: bool = False,
-    automatically_declare_initial_parameters: bool = True
+    automatically_declare_parameters_from_overrides: bool = True
```
wjwwood: @jubeira how did this end up getting defaulted to `True`? Feel like we went circles around this before merging it. This does not match `rclcpp`'s default behavior and in my opinion needs to change:

- https://github.com/ros2/rclcpp/blob/0723a0a6fc2f46edcbeebf2933c7929822177d56/rclcpp/include/rclcpp/node_options.hpp#L52
- https://github.com/ros2/rclcpp/blob/0723a0a6fc2f46edcbeebf2933c7929822177d56/rclcpp/include/rclcpp/node_options.hpp#L323

rclpy/rclpy/node.py
wjwwood: Here's an example I made just to make sure I was right in my thinking:

```python
import rclpy

rclpy.init()

n1 = rclpy.create_node('test')
print("Parameter 'foo' after declare:", n1.declare_parameter('foo', 'bar').value)
n1.set_parameters([rclpy.Parameter('foo', rclpy.Parameter.Type.STRING, 'ping')])
print("Parameter 'foo' after set:", n1.get_parameter('foo').value)

n2 = rclpy.create_node(
    'test2',
    automatically_declare_parameters_from_overrides=False,
    parameter_overrides={rclpy.Parameter('foo', rclpy.Parameter.Type.STRING, 'baz')})
print("Parameter 'foo' after declare:", n2.declare_parameter('foo', 'bar').value)
n2.set_parameters([rclpy.Parameter('foo', rclpy.Parameter.Type.STRING, 'ping')])
print("Parameter 'foo' after set:", n2.get_parameter('foo').value)
```

Which produces:

```
Parameter 'foo' after declare: bar
Parameter 'foo' after set: ping
Parameter 'foo' after declare: baz
Parameter 'foo' after set: ping
```

----

@jubeira after writing this little toy program to check the behavior @chapulina and I were talking about, I realized the parameter API in rclpy is not that friendly to use...

I'll make a new issue to track this, but I think we'll have some follow up work to do after Dashing on that part of the interface.

chapulina: See https://github.com/ros2/rclcpp/pull/745#discussion_r288344793 and the `foo == baz` behaviour on https://github.com/ros2/rclcpp/issues/730#issuecomment-495801806

wjwwood: I think it's a false alarm, see my response: https://github.com/ros2/rclcpp/pull/745#discussion_r288353024

rclpy/rclpy/__init__.py
chapulina: +1 to matching rclcpp

wjwwood: Ah I found it, the right value is being used in the Node constructor but was left True in the `create_node` method. Fixed in 78bc9f4a6ea49165eb139e5dbfc4c31ad95a6508.

jubeira: Thanks! We must have missed it in #339.

rclpy/rclpy/node.py
jubeira: > @jubeira after writing this little toy program to check the behavior @chapulina and I were talking about, I realized the parameter API in rclpy is not that friendly to use...
> 
> I'll make a new issue to track this, but I think we'll have some follow up work to do after Dashing on that part of the interface.

Sounds good.
From that example, I think there are two things that contribute to that feeling:
- Having to specify the `Parameter.Type` when creating a `Parameter` object, instead of guessing it automatically according to the given value. This was addressed in #347 without breaking the existing API:
```python
    # These two are equivalent:
    rclpy.Parameter('foo', rclpy.Parameter.Type.STRING, 'baz')
    rclpy.Parameter('foo', value='baz')
```
While I think this looks better than before, perhaps there's still room for improvement.
- There is an asymmetry between `declare` and `set` methods: the former ones use a tuple with name, value (and descriptor), and the latter ones use `rclpy.Parameter` as arguments. Following that line, maybe a version of `declare_parameter(s)` accepting a `rclpy.Parameter` and a descriptor would work (as indicated in https://github.com/ros2/rclpy/pull/325#discussion_r280534305). The problem with that approach was that `declare_parameters` accepts a namespace, and the code that actually sets the parameter downstream takes a `Parameter` is not prepared to handle it so it would not be a trivial change.

Let me know if you want to discuss this in a brief meeting; I'd like to hear your opinion about this.

PR #383, opened 2019-07-03
Added clients by node implementation from rcl

skucheria:
Connects to https://github.com/ros2/sros2/pull/141

Requires https://github.com/ros2/rcl/pull/459

PR #385, opened 2019-07-11
remove whitespace

wjwwood:
This is causing all of our nightly jobs to be unstable, e.g.:

https://ci.ros2.org/view/nightly/job/nightly_linux_release/1232/testReport/junit/(root)/projectroot/flake8/

I think it's a follow up of https://github.com/ros2/rclpy/pull/383, but I'm not sure.


wjwwood:
Yeah, it's just weird because I don't see it in the github diff for the pr, and the CI settings for that job looked ok to me.

PR #386, opened 2019-07-11
Adding get_parameters_by_prefix method to Node.

### LINKED ISSUE #374: Add `get_parameters_by_prefix` method
#### Feature description
It should work in the same way as the `rclcpp` counterpart:
https://github.com/ros2/rclcpp/blob/e7c463dc74849aa64cb8a74e13f2a57b4ed59862/rclcpp/include/rclcpp/node.hpp#L783-L787

See [discussion](https://github.com/ros/joint_state_publisher/pull/16#discussion_r296263399).

jubeira:
Closes #374.

### DIFF
--- a/rclpy/test/test_node.py
+++ b/rclpy/test/test_node.py
@@ -614,36 +614,36 @@ class TestNode(unittest.TestCase):
 
     def test_node_get_parameters_by_prefix(self):
         parameters = [
-            ('foo.foo', 43),
-            ('foo.bar', 'hello'),
-            ('foo.baz', 2.41),
-            ('bar.foo', 1),
-            ('bar.bar', 12.3),
-            ('bar.baz', 'world'),
+            ('foo_prefix.foo', 43),
+            ('foo_prefix.bar', 'hello'),
+            ('foo_prefix.baz', 2.41),
+            ('bar_prefix.foo', 1),
+            ('bar_prefix.bar', 12.3),
+            ('bar_prefix.baz', 'world'),
         ]
         self.node.declare_parameters('', parameters)
 
-        parameters = self.node.get_parameters_by_prefix('foo')
+        parameters = self.node.get_parameters_by_prefix('foo_prefix')
         self.assertIsInstance(parameters, dict)
         self.assertEqual(len(parameters), 3)
         self.assertDictEqual(
             parameters,
             {
-                'foo': self.node.get_parameter('foo.foo'),
-                'bar': self.node.get_parameter('foo.bar'),
-                'baz': self.node.get_parameter('foo.baz')
+                'foo': self.node.get_parameter('foo_prefix.foo'),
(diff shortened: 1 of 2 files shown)


jubeira:
Speculative partial CI:
- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/7539/)

jubeira:
I understand the use-case in C++, but considering the parameter name is part of each object in the list, returning a dictionary seemed a bit redundant. Furthermore, the current `get_parameters` method returns a list of parameters, so a list seemed more consistent.

Getting a dictionary with the current approach can be done via dict comprehension like this (it can also be compressed into one line):
```
param_list = node.get_parameters_by_prefix('foo')
param_dict = {param.name: param for param in param_list}
```

That being said, is it still preferable to return a dictionary?
I'm bringing @clalancette in the loop for comments, who was in the discussion referenced above.

clalancette:
The big difference here is that the C++ version of the code explicitly cuts off the prefix: https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L719

From an API consistency standpoint, I'd rather that the Python API more closely mirror the C++ API.  That would imply (to me) a dictionary, where the key is the name stripped of the prefix, and the value is the `Parameter` class (or whatever the equivalent in Python is).

jubeira:
@ivanpauno comment addressed; PTAL.
/cc @clalancette

rclpy/test/test_node.py
```diff
@@ -612,6 +612,60 @@ def test_node_get_undeclared_parameter_or(self):
         self.assertEqual(result.name, 'foo')
         self.assertEqual(result.value, 152)
 
+    def test_node_get_parameters_by_prefix(self):
+        parameters = [
+            ('foo.foo', 43),
+            ('foo.bar', 'hello'),
+            ('foo.baz', 2.41),
+            ('bar.foo', 1),
```
clalancette: I think it would be better to either change the name of `foo.bar`, or the prefix `bar` in this code.  At first glance it is confusing whether we are checking the specific key or the prefix in the code below.

jubeira: Thanks for the comment; addressed in https://github.com/ros2/rclpy/pull/386/commits/d031737d6d658ce15a8f065ad1822e6a206af458.

clalancette: Thanks!

jubeira:
CI:
- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/7576/)
- Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/3708/)
- OSX: [Build Status](https://ci.ros2.org/job/ci_osx/6201/)
- Windows: [Build Status](https://ci.ros2.org/view/manual/job/ci_windows/7401/)

jubeira:
Failures are known and unrelated; merging now!

Issue #387, opened 2019-07-12
Missing "raise"


danielwangksu:
Instantiating an exception, but not raising it, has no effect
https://github.com/ros2/rclpy/blob/29c2baca891dddc93b63371de30cd948fde70c88/rclpy/rclpy/clock.py#L211

sloretz:
Thanks @danielwangksu . Would you be willing to open a PR with a fix?

PR #390, opened 2019-07-17
fix missing 'raise'

danielwangksu:
fix  https://github.com/ros2/rclpy/issues/387


sloretz:
Thanks for the PR @danielwangksu ! It looks like [the DCO check](https://discourse.ros.org/t/starting-to-enforce-developer-certificate-of-origin-dco-for-some-ros-2-repos/7420) is complaining. Would you mind signing off on the commit?

```
git commit --amend --signoff
```

danielwangksu:
Sure thank you for reminding me.

Thanks
-Dan

sloretz:
CI (testing rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7605/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3728/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6221/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7428/)

clalancette:
@danielwangksu Thanks for the patch, looks good and CI is green, so I'll merge.  @sloretz , should we propose this for Dashing patch release 2?  It is a change in behavior, but right now I think Dashing will just "do the wrong thing" if a wrong object type is passed to `set_ros_time_override`.  Thoughts?

sloretz:
@clalancette I added to the dashing patch release because I think the risk is low. IIUC the old behavior would likely be [raising `AttributeError`](https://github.com/danielwangksu/rclpy/blob/29c2baca891dddc93b63371de30cd948fde70c88/rclpy/rclpy/clock.py#L214) if the object doesn't have an attribute called `_time_handle`, or [raising `ValueError`](https://github.com/ros2/rclpy/blob/7462ae98ca64a2693bb094ace51b8a8cb2f184c5/rclpy/src/rclpy/_rclpy.c#L4025-L4029) if it does, but it's not a pycapsule with the right name.

PR #396, opened 2019-07-29
Declaring 'use_sim_time' when attaching node to time source.

jubeira:
Closes #395.

Signed-off-by: Juan Ignacio Ubeira <jubeira@ekumenlabs.com>

PR #401, opened 2019-07-30
[Backport - dashing] Declaring 'use_sim_time' when attaching node to time source

jubeira:
Backports #396 to `dashing`.

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -164,11 +164,6 @@ class Node:
         with self.handle as capsule:
             self._logger = get_logger(_rclpy.rclpy_get_node_logger_name(capsule))
 
-        # Clock that has support for ROS time.
-        self._clock = ROSClock()
-        self._time_source = TimeSource(node=self)
-        self._time_source.attach_clock(self._clock)
-
         self.__executor_weakref = None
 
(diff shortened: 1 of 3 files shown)


jubeira:
Partial CI:
- Linux: [Build Status](https://ci.ros2.org/job/ci_linux/7732/)
- Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/3836/)

I'll post complete CI once executors are ready.

clalancette:
> My one concern with this pull request is that any tests or tools that (incorrectly) assume how many parameters a python node will have, may now have that assumption broken because we're now always declaring this parameter.

From @jubeira 's partial CI above, it's pretty clear that none of the core ROS 2 tests have these expectations.  That doesn't mean that others outside of the core don't, but I agree that this is a bug and we should fix it.  Anything that is relying on certain numbers of parameters is fragile and buggy anyway, I guess they'll find out now :).  So I'm going to approve this as well.

jubeira:
Thanks for your input @wjwwood @clalancette!
I've edited the comment above with full CI results; everything's green.

Just for the record, the only test that I found that had this issue was fixed in https://github.com/ros2/ros2cli/pull/307 (master branch; the test doesn't exist in dashing).

That being said, I'll proceed to merge now.

PR #402, opened 2019-07-31
Check if context is ok when initializing wait set in executor

ivanpauno:
Fixes https://github.com/ros2/rclpy/issues/398.

The context is being shutdown externally https://github.com/ros-visualization/rqt/blob/3911625f95cc14a751c91fa2712d78ab1a6854cc/rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py#L81-L85.
But the executor is using it for creating a wait_set:
https://github.com/ros2/rclpy/blob/1588d0753ecddb149c22db33883763cb4b0de69e/rclpy/rclpy/executors.py#L524-L532
So when the context is shut down externally, this error happens:
```
Traceback (most recent call last):
  File "/Users/karsten/workspace/osrf/ros2_full/build/rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py", line 32, in run
    executor.spin_once(timeout_sec=1.0)
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/executors.py", line 695, in spin_once
    handler, entity, node = self.wait_for_ready_callbacks(timeout_sec=timeout_sec)
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/executors.py", line 649, in wait_for_ready_callbacks
    return next(self._cb_iter)
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/executors.py", line 532, in _wait_for_ready_callbacks
    self._context.handle)
RuntimeError: Failed to initialize wait set: the given context is not valid, either rcl_init() was not called or rcl_shutdown() was called., at /Users/karsten/workspace/osrf/ros2_full/src/ros2/rcl/rcl/src/rcl/wait.c:130
```

This avoids the error, by checking if the context is ok. If not it raises a `ShutdownException`.

----

rclcpp is doing something much cleaner: It doesn't use the executor when waiting for a new executable. It only uses the context when creating/destructing the executor. Why?
Because rclcpp doesn't create a wait set each time, but resizes it:
https://github.com/ros2/rclcpp/blob/9be3e08cd4d04193fb76cb6dd8f5db11d7199e2f/rclcpp/src/rclcpp/executor.cpp#L451-L459
A context is not needed for resizing.

If we want to do the same, we will need to add some functions from rcl to rclpy (I think that's subject of a follow up).

----

[...]

PR #403, opened 2019-08-02
Awake waitables on shutdown, check if context is valid

### LINKED ISSUE #398: Avoid exception when ctrl-c
## Bug report

This issue is mainly motivated by https://github.com/ros-visualization/rqt/issues/198
When quitting a GUI application, such as `rqt_gui` by either closing it through the window button or `ctrl-c` it, I see a traceback of an uncaught exception regarding the waitset.

**Required Info:**

- Operating System:
  - OSX
- Installation type:
  - N/A
- Version or commit hash:
  - dashing and master

#### Steps to reproduce issue
Closing a RQT gui application

#### Expected behavior
GUI application closes correctly.

#### Actual behavior
```
  File "/Users/karsten/workspace/osrf/ros2_full/build/rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py", line 32, in run
    executor.spin_once(timeout_sec=1.0)
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/executors.py", line 695, in spin_once
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/executors.py", line 649, in wait_for_ready_callbacks
    return next(self._cb_iter)
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/executors.py", line 532, in _wait_for_ready_callbacks
    self._context.handle)
RuntimeError: Failed to initialize wait set: the given context is not valid, either rcl_init() was not called or rcl_shutdown() was called., at /Users/karsten/workspace/osrf/ros2_full/src/ros2/rcl/rcl/src/rcl/wait.c:130
```
#### Additional information
There is a try to "fix" this by catching the exception and thus suppressing the traceback. 
See here: https://github.com/ros-visualization/rqt/pull/199

However, I believe the root cause of this issue might lie in rclpy.

ivanpauno:
Replaces https://github.com/ros2/rclpy/pull/402.
Fixes #398.

It solves two problems:

- The executor wasn't awaken when shutdown
- `_wait_for_ready_callbacks` was looping until yielding work, without checking if the context is valid.

Some error messages may still appear when shutting down contexts (similar to https://github.com/ros2/rclcpp/issues/812), but they are harder to reproduce.

### DIFF
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -28,10 +28,31 @@ class Context:
         from rclpy.impl.implementation_singleton import rclpy_implementation
         self._handle = rclpy_implementation.rclpy_create_context()
         self._lock = threading.Lock()
+        self._guard_lock = threading.RLock()
+        self._guard_conditions = {}
 
     @property
     def handle(self):
         return self._handle
+
+    def get_interrupt_guard_condition(self, wait_set):
+        with self._guard_lock:
+            from .guard_condition import GuardCondition
+            if wait_set not in self._guard_conditions:
+                self._guard_conditions[wait_set] = GuardCondition(
+                    callback=None, callback_group=None, context=self)
+            return self._guard_conditions[wait_set]
+
+    def release_interrupt_guard_condition(self, wait_set):
+        with self._guard_lock:
+            if wait_set in self._guard_conditions:
+                self._guard_conditions[wait_set].destroy()
+                del self._guard_conditions[wait_set]
+
+    def interrupt_wait_sets(self):
+        with self._guard_lock:
+            for gc in self._guard_conditions.values():
+                gc.trigger()
 
     def ok(self):
         # imported locally to avoid loading extensions on module import
@@ -43,7 +64,8 @@ class Context:
         # imported locally to avoid loading extensions on module import
         from rclpy.impl.implementation_singleton import rclpy_implementation
         with self._lock:
-            return rclpy_implementation.rclpy_shutdown(self._handle)
+            rclpy_implementation.rclpy_shutdown(self._handle)
+            self.interrupt_wait_sets()
 
     def try_shutdown(self):
         """Shutdown rclpy if not already shutdown."""
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import threading
+from typing import Callable
 
 
 class Context:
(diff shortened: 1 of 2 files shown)


ivanpauno:
An example that was waiting forever after shutdown without this fix:
```python3
import threading
import time
import rclpy
from rclpy.node import Node
from rclpy.executors import MultiThreadedExecutor

class Listener(Node):

    def __init__(self):
        super().__init__('listener')
        self.sub = self.create_subscription(String, 'chatter', self.chatter_callback, 10)

    def chatter_callback(self, msg):
        self.get_logger().info('I heard: [%s]' % msg.data)


def kill():
    print('waiting')
    time.sleep(5)
    print('shutting down')
    rclpy.shutdown()


def main(args=None):
    rclpy.init()
    executor = MultiThreadedExecutor()
    node = Listener()

    thread = threading.Thread(target=kill)
    thread.start()

    executor.add_node(node)

    print('spinning')
    while rclpy.ok():
        executor.spin_once()

    print('done')


if __name__ == '__main__':
    main()
```

ivanpauno:
@sloretz Friendly ping

rclpy/rclpy/context.py
```diff
@@ -28,11 +28,32 @@ def __init__(self):
         from rclpy.impl.implementation_singleton import rclpy_implementation
         self._handle = rclpy_implementation.rclpy_create_context()
         self._lock = threading.Lock()
+        self._guard_lock = threading.RLock()
+        self._guard_conditions = {}
```
sloretz: I recommend against having the `Context` store guard conditions. It seems too specific to one task: waking the executor. Neither wait sets nor guard conditions are things the `Context` needs to know about.

Instead I recommend adding APIs to register a callback to call when the `Context` is shut down. Something like [`atexit.register()` and `atexit.unregister()`](https://docs.python.org/3/library/atexit.html), but on a `Context` instance.

That would allow an executor to register a callback that calls `executor.wake()`. A small benefit is it would reuse the existing guard condition rather than create a second one.

rclpy/rclpy/context.py
```diff
@@ -28,11 +28,32 @@ def __init__(self):
         from rclpy.impl.implementation_singleton import rclpy_implementation
         self._handle = rclpy_implementation.rclpy_create_context()
         self._lock = threading.Lock()
+        self._guard_lock = threading.RLock()
```
sloretz: Could this be a `Lock()` instead?

rclpy/rclpy/context.py
ivanpauno: I've applied this changes in https://github.com/ros2/rclpy/pull/403/commits/2918d3480ade7f84bf838cf370a787623cb53343.

rclpy/rclpy/context.py
```diff
@@ -28,27 +29,49 @@ def __init__(self):
...
     def try_shutdown(self):
-        """Shutdown rclpy if not already shutdown."""
+        """Shutdown this context, if not already shutdown."""
         # imported locally to avoid loading extensions on module import
         from rclpy.impl.implementation_singleton import rclpy_implementation
         with self._lock:
             if rclpy_implementation.rclpy_ok(self._handle):
-                return rclpy_implementation.rclpy_shutdown(self._handle)
+                rclpy_implementation.rclpy_shutdown(self._handle)
+        self._call_on_shutdown_callbacks()
```
sloretz: It looks like this would call callbacks again if the context was previously shutdown.

rclpy/rclpy/context.py
```diff
@@ -28,27 +29,49 @@ def __init__(self):
...
+        """Check if context hasn't been shut down."""
         # imported locally to avoid loading extensions on module import
         from rclpy.impl.implementation_singleton import rclpy_implementation
         with self._lock:
             return rclpy_implementation.rclpy_ok(self._handle)
 
+    def _call_on_shutdown_callbacks(self):
+        with self._callbacks_lock:
+            for callback in self._callbacks:
+                callback()
```
sloretz: Recommend clearing `self._callbacks`. That decreases the reference count of the objects being called, and the callbacks aren't needed anymore since a context can only be shut down once.

sloretz: Alternatively, could put [`weakref.WeakMethod(callable)`](https://docs.python.org/3/library/weakref.html#weakref.WeakMethod) into the list instead of storing the callable directly.

rclpy/rclpy/context.py
ivanpauno: Good catch, solved in https://github.com/ros2/rclpy/pull/403/commits/60c54db9e7ea98f9edb12ecd61d3c573e6df5336

rclpy/rclpy/context.py
ivanpauno: Addressed in https://github.com/ros2/rclpy/pull/403/commits/60c54db9e7ea98f9edb12ecd61d3c573e6df5336

rclpy/rclpy/context.py
```diff
@@ -28,27 +29,50 @@ def __init__(self):
...
+                self._call_on_shutdown_callbacks()
+
+    def on_shutdown(self, callback: Callable[[], None]):
+        """Add a callback to be called on shutdown."""
+        if not callable(callback):
+            raise ValueError('callback should be a callable, got {}', type(callback))
+        with self._callbacks_lock:
+            self._callbacks.append(callback)
+
+    def get_on_shutdown_callbacks(self):
```
sloretz: If this function isn't needed by this PR, I recommend leaving it out for now.

rclpy/rclpy/context.py
```diff
@@ -28,27 +29,50 @@ def __init__(self):
...
         with self._lock:
             if rclpy_implementation.rclpy_ok(self._handle):
-                return rclpy_implementation.rclpy_shutdown(self._handle)
+                rclpy_implementation.rclpy_shutdown(self._handle)
+                self._call_on_shutdown_callbacks()
+
+    def on_shutdown(self, callback: Callable[[], None]):
+        """Add a callback to be called on shutdown."""
+        if not callable(callback):
+            raise ValueError('callback should be a callable, got {}', type(callback))
```
sloretz: nit, recommend `TypeError`

ivanpauno:
CI (testing rclpy, only fastrtps)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7923/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3957/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6450/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7774/)

PR #147, opened 2017-11-14
avoid custom working directory

dirk-thomas:
Follow up of https://github.com/ament/ament_cmake/pull/116#issuecomment-344127884

See http://ci.ros2.org/job/ci_linux/3522/testReport/ with the "nicer" grouping.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3523/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/742/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2874/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3612/)

Ready for review.

PR #413, opened 2019-08-30
Raise custom error when node name is not found

ivanpauno:
Addresses https://github.com/ros2/ros2cli/pull/322#issuecomment-525364023

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -58,10 +58,6 @@ if TYPE_CHECKING:
     from rclpy.executors import Executor  # noqa: F401
     from rclpy.node import Node  # noqa: F401
 
-# `rclpy.node.Node.get_*_names_and_types_by_node` methods may raise this error.
-# It is populated in `init` function.
-NodeNameNonExistentError = None
-
 
 def init(*, args: List[str] = None, context: Context = None) -> None:
     """
@@ -75,8 +71,7 @@ def init(*, args: List[str] = None, context: Context = None) -> None:
     # imported locally to avoid loading extensions on module import
     from rclpy.impl.implementation_singleton import rclpy_implementation
     global NodeNameNonExistentError
-    rclpy_implementation.rclpy_init(args if args is not None else sys.argv, context.handle)
-    NodeNameNonExistentError = rclpy_implementation.NodeNameNonExistentError
+    return rclpy_implementation.rclpy_init(args if args is not None else sys.argv, context.handle)
 
 
 # The global spin functions need an executor to do the work
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -45,6 +45,7 @@ from typing import List
 from typing import TYPE_CHECKING
 
 from rclpy.context import Context
+# from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 from rclpy.parameter import Parameter
 from rclpy.task import Future
 from rclpy.utilities import get_default_context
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/__init__.py
```diff
@@ -70,7 +74,9 @@ def init(*, args: List[str] = None, context: Context = None) -> None:
     context = get_default_context() if context is None else context
     # imported locally to avoid loading extensions on module import
     from rclpy.impl.implementation_singleton import rclpy_implementation
-    return rclpy_implementation.rclpy_init(args if args is not None else sys.argv, context.handle)
+    global NodeNameNonExistentError
+    rclpy_implementation.rclpy_init(args if args is not None else sys.argv, context.handle)
+    NodeNameNonExistentError = rclpy_implementation.NodeNameNonExistentError
```
dirk-thomas: Why should this be the exposed C symbol? Wouldn't an exception defined in Python make more sense here?

hidmic: I'll weight in here, because I had a similar dilemma while figuring out how to raise appropriate exceptions upon failed argument parsing in `rclpy`: either defining an exception on Python and importing it from C or defining it in C and importing it on Python. And the latter seems more natural to me too. 

But then this. Why is that C extensions cannot be imported at module level?

ivanpauno: A little of history, see: https://github.com/ros2/rclpy/pull/147

I would rather not limit C extensions to be loaded after the module. Instead, it should be possible to import them at module level too.
I have a solution in mind, that makes possible to import C extensions at module level, and shows a clearer relative path when testing. I will try if it works or not.

ivanpauno: My current solution has a big problem, the exception is only valid after `rclpy_init`. I will switch to an exception defined in Python if I don't successfully avoid the limitation of importing C extensions at module level.

rclpy/rclpy/node.py
```diff
@@ -83,6 +84,10 @@
 SrvTypeRequest = TypeVar('SrvTypeRequest')
 SrvTypeResponse = TypeVar('SrvTypeResponse')
 
+# Re-export exception defined in _rclpy C extension.
+# `Node.get_*_names_and_types_by_node` methods may raise this error.
+NodeNameNonExistentError = _rclpy.NodeNameNonExistentError
```
hidmic: @ivanpauno there was [a previous comment](https://github.com/ros2/rclpy/pull/413#discussion_r319660573) from @dirk-thomas about not exporting to Python exceptions defined in C but the other way around. Since modules like `rclpy.qos` already import C extensions at the module level, this does not worsen the situation in any way. 

So I'd say that we merge this as-is, to solve https://github.com/ros2/build_cop/issues/223, and then we work towards normalizing C extension use in `rclpy` -- whatever that means. There's a PR, https://github.com/ros2/rclpy/pull/422, for it already. @dirk-thomas do you agree?

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8149/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4135/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6638/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8030/)

ivanpauno:
I forgot to push a commit in https://github.com/ros2/rcl/pull/492, again CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8150/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4136/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6639/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8032/)

ivanpauno:
I'm merging this one, we can continue the discussion on how importing C extensions in https://github.com/ros2/rclpy/pull/422.

PR #417, opened 2019-09-05
Avoid the limitation of not being able to import rclpy C extensions at module level

ivanpauno:
Kind of reverting the old https://github.com/ros2/rclpy/pull/147, plus solving the "not nice grouping" problem in another way.

See discussion: https://github.com/ros2/rclpy/pull/413#discussion_r319660573

----

There's some new nightly failures that are somewhat related with this problem, eg:
https://ci.ros2.org/view/nightly/job/nightly_win_rel/1302/testReport/junit/(root)/projectroot/test_client/

Basically, a pytest plugin is trying to import the C extensions before we "hack" the import in `__init__.py` file of the test directory.
This PR doesn't solve that problem.
I'm not to sure why this is only hitting as on Windows.
There are some possible solutions:
- avoiding importing rclpy at module level in the plugin
- Figuring out why we're loading that pytest plugin on Windows and not in the other os (rclpy doesn't depend on the package that provides the plugin).
- Change how we install python packages, to avoid doing weird imports while testing.

### DIFF
--- a/rclpy/test/__init__.py
+++ b/rclpy/test/__init__.py
@@ -12,18 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import importlib
+import os
 import sys
 
 assert 'rclpy' not in sys.modules, 'rclpy should not have been imported before running tests'
 
-# this will make the extensions load from the build folder
-import rclpy.impl  # noqa
-import test_rclpy  # noqa
+# this will load rclpy from the build folder
+sys.path.insert(0, os.environ.get('RCLPY_TEST_LIBRARY_DIR'))
 
-
-def _custom_import(name):
-    return importlib.import_module(name, package='test_rclpy')
-
-
-rclpy.impl._import = _custom_import
+import rclpy  # noqa
+assert rclpy  # silence pyflakes
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -28,16 +28,16 @@ if(WIN32 AND CMAKE_BUILD_TYPE STREQUAL "Debug")
   set(PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE_DEBUG}")
 endif()
 
-# enables using the Python extensions from the build space for testing
-file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/test_rclpy/__init__.py" "")
-
 ament_python_install_package(${PROJECT_NAME})
+
+# enables using the Python extensions from the build space for testing
+configure_file("resource/__init__.py.in" "${PROJECT_NAME}/__init__.py" @ONLY)
 
 function(set_properties _targetname _build_type)
   set_target_properties(${_targetname} PROPERTIES
     PREFIX ""
-    LIBRARY_OUTPUT_DIRECTORY${_build_type} "${CMAKE_CURRENT_BINARY_DIR}/test_${PROJECT_NAME}"
-    RUNTIME_OUTPUT_DIRECTORY${_build_type} "${CMAKE_CURRENT_BINARY_DIR}/test_${PROJECT_NAME}"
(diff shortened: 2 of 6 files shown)


dirk-thomas:
> There's some new nightly failures that are somewhat related with this problem, eg:
https://ci.ros2.org/view/nightly/job/nightly_win_deb/1366/testReport/junit/(root)/projectroot/test_client/

The referenced test passed?

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8072/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4079/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6578/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7944/)

ivanpauno:
> > There's some new nightly failures that are somewhat related with this problem, eg:
> > https://ci.ros2.org/view/nightly/job/nightly_win_deb/1366/testReport/junit/(root)/projectroot/test_client/
> 
> The referenced test passed?

Sorry, wrong link. Updated post.

dirk-thomas:
> https://ci.ros2.org/view/nightly/job/nightly_win_rel/1302/testReport/junit/(root)/projectroot/test_client/

Still referencing a passed test.

hidmic:
The original link shows the regression https://ci.ros2.org/view/nightly/job/nightly_win_deb/1366/testReport/junit/(root)/projectroot/test_client/, but it can also be seen everywhere in CI. There's https://github.com/ros2/build_cop/issues/235 tracking it.

ivanpauno:
There's something weird in the links, when I enter the first time, they show the regression and the error message.
When I enter again a time later, they show `Passed`.
If I enter to the job status, navigate through the failing tests and enter again, I see the correct regression message ... and the same link as before.

It happened again with @hidmic link ....

ivanpauno:
The grouping of the tests are looking well after this patch, see:
https://ci.ros2.org/job/ci_linux/8072/testReport/

rclpy/test/__init__.py
```diff
@@ -12,18 +12,13 @@
...
+import os
 import sys
 
 assert 'rclpy' not in sys.modules, 'rclpy should not have been imported before running tests'
 
-# this will make the extensions load from the build folder
-import rclpy.impl  # noqa
-import test_rclpy  # noqa
+# this will load rclpy from the build folder
+sys.path.insert(0, os.environ.get('RCLPY_TEST_LIBRARY_DIR'))
```
hidmic: @ivanpauno using that envvar is so shady... there has to be a natural way to do this. I wonder what `setuptools` does under the hood.

dirk-thomas:
> There's something weird in the links

If you look at the list of tests (https://ci.ros2.org/view/nightly/job/nightly_win_deb/1366/testReport/junit/(root)/projectroot/) you will find that there are three test results named `test_client`.

ivanpauno:
> If you look at the list of tests (https://ci.ros2.org/view/nightly/job/nightly_win_deb/1366/testReport/junit/(root)/projectroot/) you will find that there are three test results named `test_client`.

That's really inconvenient. This other doesn't have a repeated name:
https://ci.ros2.org/view/nightly/job/nightly_win_deb/1366/testReport/junit/(root)/projectroot/test_parameters_callback/

> @ivanpauno using that envvar is so shady... there has to be a natural way to do this. I wonder what setuptools does under the hood when installing a regular Python package.

The current solution is also shady ...

dirk-thomas:
> That's really inconvenient.

More like not smart naming of our tests. All three tests have the same name:

* rclcpp test client
* rclcpp_action test client
* rclpy test client

Giving them more precise and therefore unique names would resolve this.

ivanpauno:
Checking than this still have failures on windows when `launch_testing_ros` is present:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7968/)

ivanpauno:

PR #419, opened 2019-09-06
Timer uses ROS time by default

sloretz:
* Renamed WallTimer to Timer and made it accept a clock
* create_timer() uses the node's clock by default, which uses ROS time by default
* Executor stores and reuses clock to pass to Timer

### DIFF
--- a/rclpy/rclpy/timer.py
+++ b/rclpy/rclpy/timer.py
@@ -12,20 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from rclpy.clock import Clock
-from rclpy.clock import ClockType
 from rclpy.handle import Handle
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 from rclpy.utilities import get_default_context
 
 
-# TODO(mikaelarguedas) create a Timer or ROSTimer once we can specify custom time sources
-class WallTimer:
+class Timer:
 
-    def __init__(self, callback, callback_group, timer_period_ns, *, context=None):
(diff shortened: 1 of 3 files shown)


sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8086/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4090/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6590/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7964/)

rclpy/rclpy/timer.py
```diff
@@ -12,20 +12,16 @@
...
 
-from rclpy.clock import Clock
-from rclpy.clock import ClockType
 from rclpy.handle import Handle
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 from rclpy.utilities import get_default_context
 
 
-# TODO(mikaelarguedas) create a Timer or ROSTimer once we can specify custom time sources
-class WallTimer:
```
dirk-thomas: Doesn't this change break any code which uses `WallTimer`? If yes, I would rather preserve the `WallTimer` class with its current behavior and introduce the `Timer` class beside it.

sloretz: Outside of `rclpy` I'd expect it to only break type annotations. I wouldn't expect `WallTimer` to be used directly since it needs to be part of a `Node` to be served by an executor. It seems likely that all instances of `WallTimer` outside of `rclpy` are being created by `node.create_timer()`.

PR #420, opened 2019-09-06
Avoid the limitation of not being able to import rclpy C extensions at module level [alternative]

ivanpauno:
Alternative to https://github.com/ros2/rclpy/pull/417.

In this case, instead of installing the C extensions as a module of `rclpy` package, I installed them in a new package `rclpy_impl`.

In this way, we don't need to "hack" the imports while testing. 
This solve the same problem as #417, and also, it should solve the testing problems of `rclpy` on windows when `launch_testing_ros` is present.


ivanpauno:
CI up to rclpy (on windows, also built up to launch_testing_ros)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8088/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4092/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6592/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7966/)

ivanpauno:
So, this PR is solving all the test failures on Windows CI, plus allowing to import the C extensions at module level.
#417 just solves the second problem, but as proved in [this CI run](https://github.com/ros2/rclpy/pull/417#issuecomment-529011051), it doesn't solve the first one.

rclpy/CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ function(configure_python_c_extension_library _library_name)
   )
 
   install(TARGETS ${_library_name}
-    DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}"
+    DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}_impl"
```
dirk-thomas: This package doesn't own the namespace `rclpy_impl`. If another package with that name installs its files they will collide with this one.

hidmic:
> I thought the general goal was to delay that until any function from the C extension is actually being used?

Hmm, why is that a goal?

dirk-thomas:
> why is that a goal?

As the comments on the local imports state: "to avoid loading extensions on module import".

First of all if you import any Python module it doesn't need to load the C extensions yet - hence it is faster and lighter weight. Second it avoid the race condition between any customization you might want to do before trying to load the C extensions. Third any tool which needs to import the Python modules but doesn't want to ever call any of the API can do so without the need to have / load the C extensions.

hidmic:
> First of all if you import any Python module it doesn't need to load the C extensions yet - hence it is faster and lighter weight.

Is there an observable difference in speed? Do we have a measure of how bigger the memory footprint is? Correct me if I'm wrong but I'd expect `rmw` implementations to introduce the largest overhead in dynamic library loading and we're already deferring loading for those (at least on Linux, not sure about the other platforms?).

> Second it avoid the race condition between any customization you might want to do before trying to load the C extensions.

What kind of customization? Are we doing so? Could a downstream package achieve the same effect by delaying `rclpy` imports?

> Third any tool which needs to import the Python modules but doesn't want to ever call any of the API can do so without the need to have / load the C extensions.

Which goes back to the first statement, unless we apply that same logic to every single module import. I'd like to further understand why we need to give special treatment for an otherwise regular CPython module. It certainly makes it non-standard and creates issues from time to time -- which are easy to fix, sure, but only once you've understood all the implications of how we build and install `rclpy`.

ivanpauno:
@hidmic @dirk-thomas I've worked in a lazy import alternative, but I first want to finish with this discussion.

I understand @hidmic point, I think we're doing kind of a "premature optimization".
In general, `_rclpy` C extension will be loaded sooner or later, so I don't think is bad to load it in a global import.
The others, like `_rclpy_action`, should be only loaded when the module that uses it is loaded (`rclpy.action` in this case).
This is not currently happening, as `rclpy.impl.implementation_singleton` is doing all the imports in the same place, so if any C extension is loaded the others too.

----

There is currently two problems:
- The installed package folder structure is different than the package source directory structure, so pytest doesn't find the C extensions if we don't do something
  - A way to solve the problem, is to split each C extensions in its own package.
- Where to do the imports of the C extensions? Use lazy import?
  - I would move each import to the module where is used and I would avoid doing lazy import (I don't see much performance benefit).

dirk-thomas:
> I think we're doing kind of a "premature optimization".

The primary goal for doing lazy load is not performance improvement. It is to avoid the current race and in use cases which don't need the C extension not requiring it to be loaded.

> In general, `_rclpy` C extension will be loaded sooner or later, so I don't think is bad to load it in a global import.

I think you are missing some use cases:

* An application might just be interested in some constants or helper functions defined in `rclpy`, e.g. https://github.com/ros2/rclpy/blob/158fdef1f6745685b3556a1ff5c4282fee1f7492/rclpy/rclpy/node.py#L78 In that case it would never trigger API which needs the C extensions.
* A linter doing static code analysis needs to import all modules but doesn't execute any API.

> The others, like `_rclpy_action`, should be only loaded when the module that uses it is loaded (`rclpy.action` in this case).
> This is not currently happening, as `rclpy.impl.implementation_singleton` is doing all the imports in the same place, so if any C extension is loaded the others too.

With my proposal to convert the globally initialized handles into functions providing a lazily loaded singleton instance of the C extension that would not be the case anymore and each library would be loaded only when needed and on demand.

> The installed package folder structure is different than the package source directory structure, so pytest doesn't find the C extensions if we don't do something
> * A way to solve the problem, is to split each C extensions in its own package.

See my [above comment](https://github.com/ros2/rclpy/pull/420#discussion_r321918921) why I don't think that is feasible. Also the current approach would work just fine as is. In the cases it currently fails due to the race it would work once we do lazy loading.

ivanpauno:

dirk-thomas:
> We should still investigate ways of solving this issue.

@ivanpauno Is there any open ticket for this? If not, please create one and reference the previous approaches.

Also is the branch still needed?

ivanpauno:
> @ivanpauno Is there any open ticket for this? If not, please create one and reference the previous approaches.

See #560.

> Also is the branch still needed?

Deleted.

PR #422, opened 2019-09-11
Lazily import C extensions

ivanpauno:
Yet another alternative to #417 #420.
See https://github.com/ros2/rclpy/pull/420#issuecomment-529681952.

I ended up having a lot of problems with the `qos` module after applying lazy import, as all the default profiles are constructed at module level.
For working around the problem, I did a lot of changes that I don't like at all.

I would rather solve https://github.com/ros2/build_cop/issues/235 directly in `launch_testing_ros`, and figure out later how to avoid having to "hack" the imports while testing in `rclpy` (which is the root cause of https://github.com/ros2/build_cop/issues/235).

I would rather not merge this PR nor apply lazy import in general, as I believe is creating more problems than solving.

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -42,7 +42,6 @@ This will invalidate all entities derived from the context.
 
 import sys
 from typing import List
-from typing import TYPE_CHECKING
 
 from rclpy.context import Context
 from rclpy.executors import Executor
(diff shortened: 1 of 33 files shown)


ivanpauno:
This is the `launch_ros` patch that @hidmic sent:
https://github.com/ros2/launch_ros/pull/69

ivanpauno:

dirk-thomas:
> We should still investigate ways of solving this issue.

@ivanpauno Is there any open ticket for this? If not, please create one and reference the previous approaches.

Also is the branch still needed?

ivanpauno:
> @ivanpauno Is there any open ticket for this? If not, please create one and reference the previous approaches.

See https://github.com/ros2/rclpy/issues/560.

> Also is the branch still needed?

I deleted the branch on one of the PRs, and forgot to do ot in the others.

Issue #393, opened 2019-07-24
Support multiple `on parameter set` callbacks


ivanpauno:
#### Feature description

Similar to https://github.com/ros2/rclcpp/issues/722.

ivanpauno:
By the way, I think `set_parameters_callback` is not a good name.
The idea is to support both `on set` and `on change` callbacks.

The firsts are for rejecting or not the result, the later for updating things based on the new params.

P.S.: There is actually an implementation of the `on change` callback, but not local. You have to subscribe to a topic (at least, it's implemented in `rclcpp`, I guess that here too).

suab321321:
@ivanpauno  is this a beginner issue?

ivanpauno:
> @ivanpauno is this a beginner issue?

Yes. You have two implement two methods:

```python3
def add_on_set_parameters_callback(callback: Callable[[List[Parameter]], SetParametersResult]) -> None:
  ...
def remove_on_set_parameters_callback(callback: Callable[[List[Parameter]], SetParametersResult]) -> None:
  ...
```
They basically should add/remove a callback from a list.
All the callbacks on that list should be called in the same place where `self._parameters_callback` is called now.

suab321321:
@ivanpauno will start working today onwards.

suab321321:
xxx

suab321321:
okay

ivanpauno:
> @ivanpauno will start working today onwards.

Sure, I didn't read your reply before.
You can open a PR with a partial implementation if you want to ask anything.

suab321321:
hey @ivanpauno  I just open PR #427 could you please have a look at that confirm me if its okay..

suab321321:
hey @ivanpauno  sorry for my disturbance..but when I commit some changes in my branch that should automatically be able to update the PR which is related to that my branch..but here it is creating new pull request..
By mistake I also added another new PR trying to understand this..
And also I m unable to reopen my PR.

Sorry for my mistakes

jacobperron:
Closed by #457

PR #423, opened 2019-09-12
I just defined the add and remove callback methods

suab321321:
Issue #393 
@ivanpauno 

I cant seem to understand what @wjwwood did in  rclcpp.
So I have just defined these and have some questions

Questions:
1.Callback are there for specefic parameters or group of parameters right?So if I want to remove or add a callback I should also have the parameter(Parameter class) or groups of same kind parameters to either add or remove callbacks from list.
2.I dont know where the original list of callback is it the _paramters_callback(in rospy node.py line 144)?If it is then I think it should initalized as empty list but here it is initialized as None.
3.Also when a paramter is undeclared shouldnt its callback needs to be removed too?
3.(Silly One)Callbacks are triggered when we recevie some action (example when a subscriber recevies a message he will have a callback,which will manage what to do with recevied messaged),so what is point of having a callback in parameter setting and updating

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -756,6 +756,12 @@ class Node:
             self._parameter_event_publisher.publish(parameter_event)
 
         return result
+    
+    def add_on_set_paramters_callback(self,Callable:Callable([List[Parameter],SetParametersResult])):
+        """add the callback to list"""
+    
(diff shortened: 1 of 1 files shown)


suab321321:
Okay

suab321321:
@ivanpauno  this PR is already closed so my new push is not reflecting in here.

PR #443, opened 2019-10-10
Add Rate

### LINKED ISSUE #186: Implement Time Features in rclpy

sloretz:
Resolves #186

This adds a `Rate` class. It's a tool for running a loop at a constant rate.

Plus:
* Made context shutdown callbacks use `weakref.WeakMethod()`.
  * Should prevent the callback list from growing infinitely.
* Added `Handle.__bool__()`.
  * Used to avoid sleeping if the timer was destroyed.

Comparison with ROS 1:
* ROS 1: `rate = rospy.Rate(10)`; ROS 2 `rate = node.create_rate(10)`.
* Both raise `ROSInterruptException` if the client library has been shutdown
* ROS 1: may raise `ROSTimeMovedBackwardsException`; ROS 2: uses [timer time jump logic](https://github.com/ros2/rcl/blob/e6739884151c302838cfbe908fc9963ef4836f8f/rcl/src/rcl/timer.c#L60).
  * If it went back more than 1 period, `Rate` sleeps for 1 full period. If it went backwards less than a period, `Rate` sleeps for the time remaining. If the time source changed, sleeps for the time remaining.
* ROS 1: Almost always ok to call in a callback; ROS 2: An executor must be serving the timer, so may block if called in a callback in a `SingleThreadedExecutor`
* ROS 1: Had `Rate.remaining()`; ROS 2: not implemented. It should be easy to add if someone needs it.

### DIFF
--- a/rclpy/rclpy/timer.py
+++ b/rclpy/rclpy/timer.py
@@ -12,7 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import threading
+
+from rclpy.exceptions import ROSInterruptException
 from rclpy.handle import Handle
+from rclpy.handle import InvalidHandle
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 from rclpy.utilities import get_default_context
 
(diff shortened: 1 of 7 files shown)


rclpy/rclpy/timer.py
```diff
@@ -80,3 +84,65 @@ def time_since_last_call(self):
...
+        # This event is set to wake sleepers
+        self._event = threading.Event()
+        self._lock = threading.Lock()
+        self._num_sleepers = 0
+
+        # Set event when timer fires
+        self._timer.callback = self._event.set
+
+        # Set event when ROS is shutdown
+        context.on_shutdown(self._on_shutdown)
```
jacobperron: What if `context` is `None`? Use default context?

rclpy/rclpy/timer.py
```diff
@@ -80,3 +84,65 @@ def time_since_last_call(self):
     def time_until_next_call(self):
         with self.handle as capsule:
             return _rclpy.rclpy_time_until_next_call(capsule)
+
+
+class Rate:
```
jacobperron: Consider moving the implementation to it's own module `rate.py`.

rclpy/rclpy/timer.py
sloretz: Removed default value of `None` in 9ef2e97. Context can't be `None` because it is created by a `Node`, which handles using the default context.
https://github.com/ros2/rclpy/blob/b5b4e36c362be044f146df434e73328340c8843f/rclpy/rclpy/node.py#L137

rclpy/rclpy/timer.py
sloretz: I thought about it, but I think it fits in this module since it is a wrapper around a timer. It matches the [`Rate` class location in ROS 1](https://github.com/ros/ros_comm/blob/6b1ee1a43b5b0a970d85612435e269818c89a152/clients/rospy/src/rospy/timer.py#L47) too.

jacobperron: Fair enough.

sloretz:
CI (testing rclpy)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8350/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4283/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6806/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8252/)

charlotteheggem:
Hi, is it possible to make this feature be available for ROS2 dashing?

jacobperron:
We typically only backport bug fixes and not features. Though, I don't see much risk in backporting this particular feature and am open to it. @sloretz, what do you think?

PR #445, opened 2019-10-17
Drop rclpy test_remove_ros_args_empty test case.

hidmic:
Follow-up PR after https://github.com/ros2/rcl/pull/518.


hidmic:
Going in!

PR #448, opened 2019-10-23
Increase rclpy tests timeout

hidmic:
Precisely what the title says. 

I managed to reproduce recent `test_executor` failures on Linux when using RTI Connext (see https://ci.ros2.org/job/ci_linux/8479/testReport/junit/(root)/projectroot/test_executor/). It turns out it takes ~61 sec to complete... So I'll guess and say other recent failures are flakes as well (note how all fail exactly after 60s).

CI (up to `rclpy`, using RTI Connext only)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8539/)

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -220,7 +220,7 @@ if(BUILD_TESTING)
       PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE}"
       APPEND_ENV AMENT_PREFIX_PATH=${ament_index_build_path}
         PYTHONPATH=${CMAKE_CURRENT_BINARY_DIR}
-      TIMEOUT 60
+      TIMEOUT 90
       WERROR ON
     )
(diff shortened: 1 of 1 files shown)


hidmic:
@dirk-thomas I think `ament_cmake_add_pytest()` (and ament_cmake_add_gtest()` too) should generate a jUnit file in this case i.e. when timeouts are external to the testing tool. WDYT?

dirk-thomas:
That should already be the case: `ament_add_pytest_test` calls `ament_add_test` which invokes `run_test.py` which generates a result file in the [beginning](https://github.com/ament/ament_cmake/blob/cd05edad9948e84b189c2075a181d49aebd567cb/ament_cmake_test/ament_cmake_test/__init__.py#L102-L108).

mjcarroll:
Needed to be a tiny bit longer, I'm closing this in favor of #449

PR #449, opened 2019-10-23
Expand test timeout to deflake rmw_connext

mjcarroll:
I didn't see any duplication as in other tests, the following are just very substantial test files.

Based on a test local to my laptop, two test exceed the current 60 second timeout:

```
test_executor .....................   Passed   61.35 sec
test_node .........................   Passed  103.98 sec
```

And one is marginal:

```
test_time_source ..................   Passed   54.96 sec
```

Signed-off-by: Michael Carroll <michael@openrobotics.org>


mjcarroll:
[Build Status](https://ci.ros2.org/job/ci_linux/8541/)

hidmic:
A duplicate of #448 ? Either is fine, let's see which one is less flaky :)

mjcarroll:
Whoops, didn't see that you had opened that, I found that you're going to need > 100 seconds in some cases, so your threshold may not be enough.

PR #455, opened 2019-10-30
Get proper parameters with prefixes without dot separator

BMarchi:
For parameters without a "." prefix, the code wasn't working as it should. Connected to https://github.com/ros2/ros2cli/pull/389


BMarchi:
What do you mean by sanity linter?

hidmic:
> What do you mean by sanity linter?

Even if no tests are available, do a CI run on e.g. Linux to make sure at least linters are passing.

mjcarroll:
The `Epr` job runs linters.

hidmic:
Oh, that's right. Then go ahead if the ROSBoss approves.

PR #456, opened 2019-11-07
Expressly destroy a node's objects before the node.

nuclearsandwich:
This seems to reduce hangs during test runs described in
https://github.com/ros2/build_cop/issues/248.
Another few hours of testing will convince me to strengthen that statement.

The handles corresponding to the destroyed objects *should* be getting
destroyed explicitly when self.handle.destroy() is called below. It
seems however that when running with Fast-RTPS it's possible to get into
a state where multiple threads are waiting on futexes and none can move
forward. The rclpy context of this apparent deadlock is while clearing
a node's list of publishers or services (possibly others, although
publishers and services were the only ones observed).

I consider this patch to be a workaround rather than a fix as I'm not particularly proud of how little I understand why the existing destruction handling is insufficient.
I think there may either be a race condition between the rcl/rmw layer
and the rmw implementation layer which is being tripped by the
haphazardness of Python's garbage collector or there is a logical
problem with the handle destruction ordering in rclpy that only
Fast-RTPS is sensitive to. A further possibility is to play around with pausing the garbage collector and triggering it at explicit points to see if that changes results at all.

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1461,18 +1461,18 @@ class Node:
         # It will be destroyed with other publishers below.
         self._parameter_event_publisher = None
 
-        for p in self.__publishers:
-            self.destroy_publisher(p) 
-        for s in self.__subscriptions:
-            self.destroy_subscriber(s) 
-        for c in self.__clients:
-            self.destroy_client(c) 
-        for s in self.__services:
-            self.destroy_service(s) 
-        for t in self.__timers:
-            self.destroy_timer(t) 
-        for g in self.__guards:
-            self.destroy_guard(g) 
+        while self.__publishers:
+            self.destroy_publisher(self.__publishers.pop()) 
+        while self.__subscriptions:
+            self.destroy_subscriber(self.__subscriptions.pop()) 
+        while self.__clients:
+            self.destroy_client(self.__clients.pop()) 
+        while self.__services:
+            self.destroy_service(self.__services.pop()) 
+        while self.__timers:
+            self.destroy_timer(self.__timers.pop()) 
+        while self.__guards:
+            self.destroy_guard(self.__guards.pop()) 
         self.handle.destroy()
         self._wake_executor()
 
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1462,17 +1462,17 @@ class Node:
         self._parameter_event_publisher = None
 
         while self.__publishers:
-            self.destroy_publisher(self.__publishers.pop()) 
+            self.destroy_publisher(self.__publishers.pop())
         while self.__subscriptions:
-            self.destroy_subscriber(self.__subscriptions.pop()) 
+            self.destroy_subscription(self.__subscriptions.pop())
         while self.__clients:
-            self.destroy_client(self.__clients.pop()) 
+            self.destroy_client(self.__clients.pop())
         while self.__services:
-            self.destroy_service(self.__services.pop()) 
+            self.destroy_service(self.__services.pop())
         while self.__timers:
-            self.destroy_timer(self.__timers.pop()) 
+            self.destroy_timer(self.__timers.pop())
         while self.__guards:
-            self.destroy_guard(self.__guards.pop()) 
+            self.destroy_guard(self.__guards.pop())
         self.handle.destroy()
         self._wake_executor()
 
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1461,18 +1461,20 @@ class Node:
         # It will be destroyed with other publishers below.
         self._parameter_event_publisher = None
 
+        # Destroy dependent items eagerly to work around a possible hang
+        # https://github.com/ros2/build_cop/issues/248
         while self.__publishers:
-            self.destroy_publisher(self.__publishers.pop())
+            self.destroy_publisher(self.__publishers[0])
         while self.__subscriptions:
-            self.destroy_subscription(self.__subscriptions.pop())
+            self.destroy_subscription(self.__subscriptions[0])
         while self.__clients:
-            self.destroy_client(self.__clients.pop())
+            self.destroy_client(self.__clients[0])
         while self.__services:
-            self.destroy_service(self.__services.pop())
+            self.destroy_service(self.__services[0])
         while self.__timers:
-            self.destroy_timer(self.__timers.pop())
+            self.destroy_timer(self.__timers[0])
         while self.__guards:
-            self.destroy_guard_condition(self.__guards.pop())
+            self.destroy_guard_condition(self.__guards[0])
         self.handle.destroy()
         self._wake_executor()
 
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1461,12 +1461,18 @@ class Node:
         # It will be destroyed with other publishers below.
         self._parameter_event_publisher = None
 
-        self.__publishers.clear()
-        self.__subscriptions.clear()
-        self.__clients.clear()
-        self.__services.clear()
-        self.__timers.clear()
-        self.__guards.clear()
+        for p in self.__publishers:
+            self.destroy_publisher(p) 
+        for s in self.__subscriptions:
+            self.destroy_subscriber(s) 
+        for c in self.__clients:
+            self.destroy_client(c) 
+        for s in self.__services:
+            self.destroy_service(s) 
+        for t in self.__timers:
+            self.destroy_timer(t) 
+        for g in self.__guards:
+            self.destroy_guard(g) 
         self.handle.destroy()
         self._wake_executor()
 
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1472,7 +1472,7 @@ class Node:
         while self.__timers:
             self.destroy_timer(self.__timers.pop())
         while self.__guards:
-            self.destroy_guard(self.__guards.pop())
+            self.destroy_guard_condition(self.__guards.pop())
         self.handle.destroy()
         self._wake_executor()
 


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8620/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4471/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7036/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8540/)

mjcarroll:
This doesn't seem outrageous. It does seem like there may be a more elegant solution, but it probably requires some arcane knowledge of Python garbage collection's interactions with `C` extensions.  Unless @sloretz has any immediate thoughts, I'm good with merging this + adding a tracking issue for a longer-term solution.

dirk-thomas:
The new logic seems to remove items for the collections while iterating over them which shouldn't be done. The loops should be changed into `while self._[name]:` to avoid that.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8621/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4472/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7037/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8541/)

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8622/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4473/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7038/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8542/)

rclpy/rclpy/node.py
```diff
@@ -1461,12 +1461,18 @@ def destroy_node(self) -> bool:
...
         self._parameter_event_publisher = None
 
-        self.__publishers.clear()
-        self.__subscriptions.clear()
-        self.__clients.clear()
-        self.__services.clear()
-        self.__timers.clear()
-        self.__guards.clear()
+        while self.__publishers:
+            self.destroy_publisher(self.__publishers.pop())
```
sloretz: This is effectively the same as `self.__publishers.clear()`, but the items are removed one at a time. `pop()` removes the publisher from `self.__publishers`, which causes this logic to get skipped.

https://github.com/ros2/rclpy/blob/e566f3e15094dbb065fda72ee5969e826ff76e52/rclpy/rclpy/node.py#L1347-L1355

nuclearsandwich: Hah. So would the right thing to do be

```python3
self.destroy_publisher(self.__publishers[0])
```
?

Which would keep destroying the first publisher until there are none more?

sloretz: I think that would do it.

While you're there, I recommend adding a comment above this block that this is a workaround for an unknown bug that causes a deadlock.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8623/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4474/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7039/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8543/)

nuclearsandwich:
>  I don't know what causes the hang or how this works around it.

I've also been dissatisfied with this solution. After discussion with @sloretz I've started another experiment running with Python's GC disabled before 1461 and enabled after self.handle.destroy(). That has also seemingly stopped the hang from occurring over the last ~50 test runs which means we might need to look more closely at the handle destruction logic.

nuclearsandwich:
I ran experimental tests most of the day Saturday and I was able to reproduce a hang using this branch of rclpy with rmw_cyclonedds_cpp as well. Although it's in a slightly different place.

In [this gdb log](https://gist.github.com/nuclearsandwich/c845d073fa854ce020c8d42453d00faa) thread 2 is stuck during capsule destruction in a futex down in cyclonedds.

nuclearsandwich:
@hidmic and I dove into this today and we couldn't pinpoint a reason for the hang. It is still possible to hang with even with this PR although the incidence rate is lower.

We did a couple of experiments trying to reproduce the hang outside of launch tests using multiple threads and rclpy but were unable to do so.

I do have a bit more detail from the current reproduction since the destruction doesn't occur as part of a garbage collection. In python, we're inside [Handle.__destroy_self](https://github.com/ros2/rclpy/blob/e566f3e15094dbb065fda72ee5969e826ff76e52/rclpy/rclpy/handle.py#L192-L200)

```
 192        def __destroy_self(self):
 193            with self.__rlock:
 194                # Calls pycapsule destructor
>195                _rclpy_capsule.rclpy_pycapsule_destroy(self.__capsule)
 196                # Call post-destroy callbacks
 197                while self.__destroy_callbacks:
 198                    self.__destroy_callbacks.pop()(self)
 199                # get rid of references to other handles to break reference cycles
 200                del self.__required_handles
```

and in the C context we can see the pycapsule destructor calling rcl_publisher_fini which is carrying us into Fast-RTPS internals where we're stuck waiting for a lock:

```
#0  0x00007f40703f19f3 in futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x1f77a04) at ../sysdeps/unix/sysv/linux/futex-internal.h:88
#1  __pthread_cond_wait_common (abstime=0x0, mutex=0x1f779b0, cond=0x1f779d8) at pthread_cond_wait.c:502
#2  __pthread_cond_wait (cond=0x1f779d8, mutex=0x1f779b0) at pthread_cond_wait.c:655
#3  0x00007f40632a5951 in eprosima::fastrtps::TimedConditionVariable::wait<std::timed_mutex>(std::unique_lock<std::timed_mutex>&, std::function<bool ()>) (this=0x1f779d8, lock=..., predicate=...)
    at /home/osrf/debug_ws/src/eProsima/Fast-RTPS/include/fastrtps/rtps/resources/../../utils/TimedConditionVariable.hpp:79
#4  0x00007f406329fd0e in eprosima::fastrtps::rtps::ResourceEvent::unregister_timer (this=0x1f779a0, event=0x20deb20) at /home/osrf/debug_ws/src/eProsima/Fast-RTPS/src/cpp/rtps/resources/ResourceEvent.cpp:101
#5  0x00007f40632a8962 in eprosima::fastrtps::rtps::TimedEvent::~TimedEvent (this=0x1797100, __in_chrg=<optimized out>) at /home/osrf/debug_ws/src/eProsima/Fast-RTPS/src/cpp/rtps/resources/TimedEvent.cpp:42
#6  0x00007f40632a89a0 in eprosima::fastrtps::rtps::TimedEvent::~TimedEvent (this=0x1797100, __in_chrg=<optimized out>) at /home/osrf/debug_ws/src/eProsima/Fast-RTPS/src/cpp/rtps/resources/TimedEvent.cpp:44
#7  0x00007f4063367911 in eprosima::fastrtps::PublisherImpl::~PublisherImpl (this=0x20cef00, __in_chrg=<optimized out>) at /home/osrf/debug_ws/src/eProsima/Fast-RTPS/src/cpp/publisher/PublisherImpl.cpp:105
#8  0x00007f40633679ec in eprosima::fastrtps::PublisherImpl::~PublisherImpl (this=0x20cef00, __in_chrg=<optimized out>) at /home/osrf/debug_ws/src/eProsima/Fast-RTPS/src/cpp/publisher/PublisherImpl.cpp:115
#9  0x00007f406335a689 in eprosima::fastrtps::ParticipantImpl::removePublisher (this=0x1f16450, pub=0x18f46c0) at /home/osrf/debug_ws/src/eProsima/Fast-RTPS/src/cpp/participant/ParticipantImpl.cpp:91
#10 0x00007f406334e85b in eprosima::fastrtps::Domain::removePublisher (pub=0x18f46c0) at /home/osrf/debug_ws/src/eProsima/Fast-RTPS/src/cpp/Domain.cpp:116
#11 0x00007f4063bf99e3 in rmw_fastrtps_shared_cpp::__rmw_destroy_publisher (identifier=0x7f406427fc17 "rmw_fastrtps_cpp", node=0x191e3b0, publisher=0x162e0d0)
    at /home/osrf/debug_ws/src/ros2/rmw_fastrtps/rmw_fastrtps_shared_cpp/src/rmw_publisher.cpp:66
#12 0x00007f4064274f8f in rmw_destroy_publisher (node=0x191e3b0, publisher=0x162e0d0) at /home/osrf/debug_ws/src/ros2/rmw_fastrtps/rmw_fastrtps_cpp/src/rmw_publisher.cpp:295
#13 0x00007f406a533f34 in rmw_destroy_publisher (v2=0x191e3b0, v1=0x162e0d0) at /home/osrf/debug_ws/src/ros2/rmw_implementation/rmw_implementation/src/functions.cpp:285
#14 0x00007f406a758cc2 in rcl_publisher_fini (publisher=0x7f4064e68720, node=0x7f4064e686f0) at /home/osrf/debug_ws/src/ros2/rcl/rcl/src/rcl/publisher.c:231
#15 0x00007f406b231336 in _rclpy_destroy_publisher (pyentity=<PyCapsule at remote 0x7f40600a86f0>) at /home/osrf/debug_ws/src/ros2/rclpy/rclpy/src/rclpy/_rclpy.c:1259
#16 0x00007f4067599d78 in rclpy_pycapsule_destroy (_unused_self=<module at remote 0x7f406b4740e8>, args=(<PyCapsule at remote 0x7f40600a86f0>,))
    at /home/osrf/debug_ws/src/ros2/rclpy/rclpy/src/rclpy/_rclpy_pycapsule.c:100
```

This PR still improves matters, but it's not a full solution.

nuclearsandwich:
In discussion yesterday we've decided to merge this as an incremental improvement. I'll leave https://github.com/ros2/build_cop/issues/248 open or create a more focused issue to continue investigation.

PR #458, opened 2019-11-12
Provide logging severity for string

BMarchi:
Connected to https://github.com/ros2/ros2cli/pull/390

As the title says, this PR provides an API to get the logging severity from a name in string format

### DIFF
--- a/rclpy/src/rclpy/_rclpy_logging.c
+++ b/rclpy/src/rclpy/_rclpy_logging.c
@@ -14,6 +14,7 @@
 
 #include <Python.h>
 
+#include <rcutils/allocator.h>
 #include <rcutils/error_handling.h>
 #include <rcutils/logging.h>
 #include <rcutils/time.h>
@@ -169,6 +170,36 @@ rclpy_logging_rcutils_log(PyObject * Py_UNUSED(self), PyObject * args)
   Py_RETURN_NONE;
 }
 
+/// Get the log severity based on the log level string representation
+/**
+ * On failure, an exception is raised and NULL is returned if:
+ * Raises RuntimeError if invalid log level name is given.
+ * Raises ValueError if log_level is not a string
+ *
+ * \param[in] log_level Name of the log level.
+ * \return NULL on failure
+           Log level associated with the string representation
+ */
+static PyObject *
+rclpy_logging_severity_level_from_string(PyObject * Py_UNUSED(self), PyObject * args)
+{
+  int severity;
+  const char * log_level = NULL;
+  if (!PyArg_ParseTuple(args, "s", &log_level)) {
+    return NULL;
+  }
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+  rcutils_ret_t ret = rcutils_logging_severity_level_from_string(log_level, allocator, &severity);
+  if (ret != RCUTILS_RET_OK) {
+    PyErr_Format(PyExc_RuntimeError,
+      "Failed to get log severity from name \"%s\", return code: %d\n",
+      log_level, ret);
+    rcutils_reset_error();
+    return NULL;
+  }
+  return PyLong_FromLongLong(severity);
+}
+
 /// Define the public methods of this module
 static PyMethodDef rclpy_logging_methods[] = {
   {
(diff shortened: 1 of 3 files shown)


rclpy/test/test_logging.py
```diff
@@ -359,6 +359,22 @@ def test_clear_config(self):
...
 
+    def test_logging_severity_from_string(self):
+        log_levels = {
+            'UNSET': 0,
+            'DEBUG': 10,
+            'INFO': 20,
+            'WARN': 30,
+            'ERROR': 40,
+            'FATAL': 50
+        }
```
hidmic: @BMarchi hmm should we hard-code these integers here? Doesn't it defeat the purpose?

rclpy/src/rclpy/_rclpy_logging.c
```diff
@@ -169,6 +170,36 @@ rclpy_logging_rcutils_log(PyObject * Py_UNUSED(self), PyObject * args)
   Py_RETURN_NONE;
 }
 
+/// Get the log severity based on the log level string representation
+/**
+ * On failure, an exception is raised and NULL is returned if:
+ * Raises RuntimeError if invalid log level name is given.
+ * Raises ValueError if log_level is not a string
```
hidmic: @BMarchi nit: proof read this one, writing is awkward.

rclpy/test/test_logging.py
BMarchi: Why? Having them outside the test doesn't make any sense. I would even said that having an rcl function to make that won't have sense at all if that was the case.

hidmic: Actually, looking into the [`rclpy.logging` module](https://github.com/ros2/rclpy/blob/f293718b235ae3e86b2b4518ef0e9d2da3f2a2e9/rclpy/rclpy/logging.py#L22) I see severity levels are already duplicated w.r.t. [`rcutils`](https://github.com/ros2/rcutils/blob/fe215b4a3db580358f09bda6269e1356165fa920/include/rcutils/logging.h#L164).

I don't really like either, as it couples the implementation with that of a transitive dependency. But that's another discussion. Let's just use what's already defined in `rclpy.logging` i.e. let's keep the duplication in a single place.

dirk-thomas: > Actually, looking into the rclpy.logging module I see severity levels are already duplicated w.r.t. rcutils.

Pointing out one location where code was copies hardly justifies to repeat the same anti pattern. Instead it motivates to the remove the existing duplication instead.

hidmic: I'm not arguing for duplication, we should use the Python `enum` that's available and will likely stay available, regardless of how we manage to construct it. We can then deal with how we tunnel this information without duplication in a follow up PR. Do you agree @dirk-thomas ?

dirk-thomas: > We can then deal with how we tunnel this information without duplication in a follow up PR

Sure (as long as the follow up PR is happening in the near future).

rclpy/test/test_logging.py
```diff
@@ -359,6 +359,22 @@ def test_clear_config(self):
...
+        log_levels = {
+            'UNSET': rclpy.logging.LoggingSeverity.UNSET,
+            'DEBUG': rclpy.logging.LoggingSeverity.DEBUG,
+            'INFO': rclpy.logging.LoggingSeverity.INFO,
+            'WARN': rclpy.logging.LoggingSeverity.WARN,
+            'ERROR': rclpy.logging.LoggingSeverity.ERROR,
+            'FATAL': rclpy.logging.LoggingSeverity.FATAL
+        }
+        for level, severity in log_levels.items():
+            self.assertEqual(rclpy.logging.get_logging_severity_from_string(level), severity)
```
hidmic: @BMarchi `rclpy.logging.LoggingSeverity` is an [`IntEnum`](https://docs.python.org/3.6/library/enum.html#enum.IntEnum), you can do:

```python
for severity in rclpy.logging.LoggingSeverity:
    self.assertEqual(rclpy.logging.get_logging_severity_from_string(severity.name), severity)
```

BMarchi:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8732/)
- Arch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4563/)
- macOS [Build Status](http://ci.ros2.org/job/ci_osx/7134/)
- Windows [Build Status](http://ci.ros2.org/job/ci_windows/8647/)

PR #459, opened 2019-11-15
Support for ON_REQUESTED_INCOMPATIBLE_QOS and ON_OFFERED_INCOMPATIBLE_QOS events

jaisontj:
**_Depends on [ros2/rcl:535](https://github.com/ros2/rcl/pull/535)_**

Related to [this](https://github.com/ros2/ros2/issues/822) feature request. The design and implementation details can also be found there.

- Added `INCOMPATIBLE_QOS` event type to `QoSPublisherEventType` and `QoSSubscriptionEventType`
- Added support for the above events in the `QoSEventHandler`
- modified `test_qos_event.py` to include the new events.

Signed-off-by: Jaison Titus <jaisontj92@gmail.com>

### DIFF
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -27,6 +27,7 @@ class QoSPolicyKind(IntEnum):
     This enum matches the one defined in rmw/incompatible_qos_events_statuses.h
     """
 
+    # TODO(mm3188): obtain these enum values from the rmw layer, instead of hardcoding
     INVALID = 1 << 0
     DURABILITY = 1 << 1
     DEADLINE = 1 << 2
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -27,29 +27,28 @@ class QoSPolicyKind(IntEnum):
     This enum matches the one defined in rmw/incompatible_qos_events_statuses.h
     """
 
-    RMW_QOS_POLICY_INVALID = 1 << 0
-    RMW_QOS_POLICY_DURABILITY = 1 << 1
-    RMW_QOS_POLICY_DEADLINE = 1 << 2
-    RMW_QOS_POLICY_LIVELINESS = 1 << 3
-    RMW_QOS_POLICY_RELIABILITY = 1 << 4
-    RMW_QOS_POLICY_HISTORY = 1 << 5
-    RMW_QOS_POLICY_LIFESPAN = 1 << 6
+    INVALID = 1 << 0
+    DURABILITY = 1 << 1
+    DEADLINE = 1 << 2
+    LIVELINESS = 1 << 3
+    RELIABILITY = 1 << 4
+    HISTORY = 1 << 5
+    LIFESPAN = 1 << 6
 
 
 def qos_policy_name_from_kind(policy_kind: QoSPolicyKind):
     """Get QoS policy name from QoSPolicyKind enum."""
-
-    if policy_kind == RMW_QOS_POLICY_DURABILITY:
+    if policy_kind == QoSPolicyKind.DURABILITY:
         return 'DURABILITY_QOS_POLICY'
-    elif policy_kind == RMW_QOS_POLICY_DEADLINE:
+    elif policy_kind == QoSPolicyKind.DEADLINE:
         return 'DEADLINE_QOS_POLICY'
-    elif policy_kind == RMW_QOS_POLICY_LIVELINESS:
+    elif policy_kind == QoSPolicyKind.LIVELINESS:
         return 'LIVELINESS_QOS_POLICY'
-    elif policy_kind == RMW_QOS_POLICY_RELIABILITY:
+    elif policy_kind == QoSPolicyKind.RELIABILITY:
         return 'RELIABILITY_QOS_POLICY'
-    elif policy_kind == RMW_QOS_POLICY_HISTORY:
+    elif policy_kind == QoSPolicyKind.HISTORY:
         return 'HISTORY_QOS_POLICY'
-    elif policy_kind == RMW_QOS_POLICY_LIFESPAN:
+    elif policy_kind == QoSPolicyKind.LIFESPAN:
         return 'LIFESPAN_QOS_POLICY'
     else:
         return 'INVALID_QOS_POLICY'
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -20,8 +20,43 @@ from rclpy.impl.implementation_singleton import rclpy_action_implementation as _
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
 
+class QoSPolicyKind(IntEnum):
+    """
+    Enum for types of QoS policies that a Publisher or Subscription can set.
+
+    This enum matches the one defined in rmw/incompatible_qos_events_statuses.h
+    """
+
+    RMW_QOS_POLICY_INVALID = 1 << 0
+    RMW_QOS_POLICY_DURABILITY = 1 << 1
+    RMW_QOS_POLICY_DEADLINE = 1 << 2
+    RMW_QOS_POLICY_LIVELINESS = 1 << 3
+    RMW_QOS_POLICY_RELIABILITY = 1 << 4
+    RMW_QOS_POLICY_HISTORY = 1 << 5
+    RMW_QOS_POLICY_LIFESPAN = 1 << 6
+
+
+def qos_policy_name_from_kind(policy_kind: QoSPolicyKind):
+    """Get QoS policy name from QoSPolicyKind enum."""
+
+    if policy_kind == RMW_QOS_POLICY_DURABILITY:
+        return 'DURABILITY_QOS_POLICY'
+    elif policy_kind == RMW_QOS_POLICY_DEADLINE:
+        return 'DEADLINE_QOS_POLICY'
+    elif policy_kind == RMW_QOS_POLICY_LIVELINESS:
+        return 'LIVELINESS_QOS_POLICY'
+    elif policy_kind == RMW_QOS_POLICY_RELIABILITY:
+        return 'RELIABILITY_QOS_POLICY'
+    elif policy_kind == RMW_QOS_POLICY_HISTORY:
+        return 'HISTORY_QOS_POLICY'
+    elif policy_kind == RMW_QOS_POLICY_LIFESPAN:
+        return 'LIFESPAN_QOS_POLICY'
+    else:
+        return 'INVALID_QOS_POLICY'
+
+
 class InvalidQoSProfileException(Exception):
-    """Raised when concstructing a QoSProfile with invalid arguments."""
+    """Raised when constructing a QoSProfile with invalid arguments."""
 
     def __init__(self, *args):
         Exception.__init__(self, 'Invalid QoSProfile', *args)
--- a/rclpy/rclpy/qos_event.py
+++ b/rclpy/rclpy/qos_event.py
@@ -26,6 +26,22 @@ from rclpy.waitable import NumberOfEntities
 from rclpy.waitable import Waitable
 
 
+class QoSPolicyKind(IntEnum):
+    """
+    Enum for types of QoS policies that a Publisher or Subscription can set.
+
+    This enum matches the one defined in rmw/incompatible_qos_events_statuses.h
+    """
+
+    RMW_QOS_POLICY_INVALID = 1 << 0
+    RMW_QOS_POLICY_DURABILITY = 1 << 1
+    RMW_QOS_POLICY_DEADLINE = 1 << 2
+    RMW_QOS_POLICY_LIVELINESS = 1 << 3
+    RMW_QOS_POLICY_RELIABILITY = 1 << 4
+    RMW_QOS_POLICY_HISTORY = 1 << 5
+    RMW_QOS_POLICY_LIFESPAN = 1 << 6
+
+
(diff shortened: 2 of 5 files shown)


rclpy/rclpy/qos_event.py
```diff
@@ -72,6 +74,18 @@ class QoSSubscriptionEventType(IntEnum):
...
+"""
+Payload type for Subscription Incompatible QoS callback.
+
+Mirrors rmw_requested_incompatible_qos_status_t from rmw/types.h
+"""
+QoSRequestedIncompatibleQoSInfo = NamedTuple(
+    'QoSRequestedIncompatibleQoSInfo', [
+        ('total_count', 'int'),
+        ('total_count_change', 'int'),
+        ('last_policy_id', 'int'),
```
ivanpauno: An int is not really clear for this.
Maybe a python enum https://docs.python.org/3/library/enum.html?

wjwwood: 👍 for an enum for the policy id. Also, as I mentioned in the rmw pr, I think `policy_kind` makes more sense. The fact that it is an enum or is backed with an int versus a string versus a type or whatever is an implementation detail.

mm318: I've introduced a `QoSPolicyKind` `IntEnum`, but I'll figure out how to specify that in a `NamedTuple` later.

rclpy/rclpy/qos_event.py
```diff
@@ -94,6 +108,18 @@ class QoSSubscriptionEventType(IntEnum):
         ('total_count_change', 'int'),
     ])
 
+"""
+Payload type for Publisher Incompatible QoS callback.
+
+Mirrors rmw_offered_incompatible_qos_status_t from rmw/types.h
+"""
+QoSOfferedIncompatibleQoSInfo = NamedTuple(
```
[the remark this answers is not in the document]
mm318: Done.

rclpy/rclpy/qos.py
```diff
@@ -20,8 +20,43 @@
...
 
 
+class QoSPolicyKind(IntEnum):
+    """
+    Enum for types of QoS policies that a Publisher or Subscription can set.
+
+    This enum matches the one defined in rmw/incompatible_qos_events_statuses.h
+    """
+
+    RMW_QOS_POLICY_INVALID = 1 << 0
```
ivanpauno: why the RMW prefix? Considering the class name, I would just left `INVALID`. In that way, it can be used like: `QoSPolicyKind.INVALID`, which is really readable IMO.

ivanpauno: It would also be good to use the values defined in `rmw`, instead of manually write them here again.

I don't mind strongly, you can left a TODO.

rclpy/rclpy/qos_event.py
```diff
@@ -94,6 +108,20 @@ class QoSSubscriptionEventType(IntEnum):
...
 
+"""
+Payload type for Publisher Incompatible QoS callback.
+
+Mirrors rmw_offered_incompatible_qos_status_t from rmw/types.h
+"""
+QoSOfferedIncompatibleQoSInfo = QoSRequestedIncompatibleQoSInfo
+
+
+class UnsupportedEventTypeException(Exception):
```
ivanpauno: `RuntimeError` sounds like a better base class.
I would also rename the exception to `UnsupportedEventTypeError`, as `Error` is usually used as suffix in Python instead of `Exception`.

ivanpauno: Actually, `NotImplementedError` may be a better base class.

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -30,14 +33,36 @@ static
...
+  if (RCL_RET_OK == ret) {
+    return true;
+  }
+
+  PyObject * exception = PyExc_RuntimeError;
+  PyObject * pyqos_event_module = NULL;
+  PyObject * pyqos_event_exception = NULL;
+
+  if (RCL_RET_UNSUPPORTED == ret) {
+    PyObject * pyqos_event_module = PyImport_ImportModule("rclpy.qos_event");
```
ivanpauno: I think that it's better to define the new exception in `C`, like this:
https://github.com/ros2/rclpy/blob/3d1698a99cbde82322c82fde428c76dd2242e855/rclpy/src/rclpy/_rclpy.c#L5523.

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -30,14 +33,36 @@ static
...
-    PyErr_Format(
-      PyExc_RuntimeError,
-      "%s: %s", error_msg, rcl_get_error_string().str);
-    rcl_reset_error();
-    return false;
+  if (RCL_RET_OK == ret) {
+    return true;
+  }
+
+  PyObject * exception = PyExc_RuntimeError;
```
ivanpauno: This should be an `RCLError`, instead of `RuntimeError`.

rclpy/rclpy/qos.py
mm318: I am not sure how to do that. Do you have an example?

`QoSPublisherEventType` and `QoSSubscriptionEventType` were also done this way.

rclpy/src/rclpy/_rclpy_qos_event.c
mm318: If we define the new exception in C, then it looks like using it from Python module would be something like:
```
...
try:
    ...
except _rclpy.UnsupportedEventTypeError as e:
    ...
```
whereas if we define it in Python (as it is in this pull request currently), the usage would be like:
```
from rclpy.qos_event import UnsupportedEventTypeError
...
try:
    ...
except UnsupportedEventTypeError as e:
    ...
```
I think the latter is better, because you probably wouldn't want a user to be importing `rclpy.impl.*`

rclpy/rclpy/qos_event.py
mm318: The [Python documentation](https://docs.python.org/3/library/exceptions.html) says "All user-defined exceptions should also be derived from this class (Exception)", so I don't think `RuntimeError` or `NotImplementedError` are meant to be base classes.

rclpy/src/rclpy/_rclpy_qos_event.c
mm318: Done.

rclpy/rclpy/qos.py
mm318: Done.

mm318:
Are the `rclpy` tests only run with `rmw_fastrtps_cpp`? Should we make it run with additional `rmw` implementations? And if we do, how can we tell from within the test which `rmw` implementation is being used?

rclpy/rclpy/qos_event.py
ivanpauno: > The [Python documentation](https://docs.python.org/3/library/exceptions.html) says "All user-defined exceptions should also be derived from this class (Exception)", so I don't think `RuntimeError` or `NotImplementedError` are meant to be base classes.

IMHO, that means that `isinstance(CustomError(..), Exception)` should be `True`.
That's the case, because the base case of `RuntimeError` is `Exception`.
There's a nice hierarchy tree at the end of the page you linked.
AFAIK, it's a usual practice to subclass `RuntimeError` or any other python built-in exception.
e.g.: `RCLError` subclass from `RuntimeError`.

@wjwwood what do you think?

rclpy/src/rclpy/_rclpy_qos_event.c
ivanpauno: you can import it in `rclpy.qos_event` and then reexport it.

@wjwwood for a second opinion.

rclpy/rclpy/qos.py
ivanpauno: Not really nice, but here an example: https://github.com/ros2/rclpy/blob/3d1698a99cbde82322c82fde428c76dd2242e855/rclpy/src/rclpy/_rclpy_logging.c#L244

You could also create the constants in the [module init function](https://github.com/ros2/rclpy/blob/3d1698a99cbde82322c82fde428c76dd2242e855/rclpy/src/rclpy/_rclpy.c#L5516), but again, not super nice.

ivanpauno:
> Are the `rclpy` tests only run with `rmw_fastrtps_cpp`? Should we make it run with additional `rmw` implementations?

I'm ok with running them in only one implementation.
We also have jobs in `build.ros2.org` where only one implementation is available, so it will cover all off them there.

rclpy/rclpy/qos.py
mm318: I see. I'm not sure if this is worth the effort, because it won't automatically synchronize when new enums are added to `rmw_qos_policy_kind_t`.

I'll leave a `TODO` here.

ivanpauno: > I see. I'm not sure if this is worth the effort, because it won't automatically synchronize when new enums are added to rmw_qos_policy_kind_t.

Sounds good!

rclpy/src/rclpy/_rclpy_qos_event.c
wjwwood: I think both will work. It’s slightly less awkward to avoid importing code in the c extension and then reexport it as @ivanpauno said, but I don’t mind as long as it works.

rclpy/rclpy/qos_event.py
wjwwood: Correct, you can, and where appropriate, you should inherit from other exception types. They just mean you should raise a completely custom class that does not inherit from exception at all, directly or indirectly.

mm318: I will move the definition of this exception to `_rclpy.c` as suggested at https://github.com/ros2/rclpy/pull/459#discussion_r397849914 , in which case this exception will inherit from `RCLError` instead of `RuntimeError`.

rclpy/src/rclpy/_rclpy_qos_event.c
mm318: I have moved the definition of the new exception into C instead of Python.

mm318:
Thanks!

The following is the set of remaining pull requests for this feature:
- https://github.com/ros2/rclcpp/pull/924
- https://github.com/ros2/rclpy/pull/459
- https://github.com/ros2/ros2cli/pull/410
- https://github.com/ros2/demos/pull/416

After I add some more tests to `rclcpp` (to address https://github.com/ros2/rclcpp/pull/924#pullrequestreview-381177646), I'll request to kick of CI.

mm318:
@ros2/aws-oncall - please run this CI job
Gist: `https://gist.githubusercontent.com/mm318/e32e5d8c1e712f6098101fa6b32efc57/raw/48efc95ff1c41ef1cab6322f2b639221637c34a0/ros2_qos.repos`
BUILD args: `--packages-up-to ros2topic quality_of_service_demo_cpp quality_of_service_demo_py`
TEST args: `--packages-select rclcpp rclpy ros2topic quality_of_service_demo_cpp quality_of_service_demo_py`
Job: `ci_launcher`

dabonnie:
> @ros2/aws-oncall - please run this CI job
> Gist: `https://gist.githubusercontent.com/mm318/e32e5d8c1e712f6098101fa6b32efc57/raw/48efc95ff1c41ef1cab6322f2b639221637c34a0/ros2_qos.repos`
> BUILD args: `--packages-up-to ros2topic quality_of_service_demo_cpp quality_of_service_demo_py`
> TEST args: `--packages-select rclcpp rclpy ros2topic quality_of_service_demo_cpp quality_of_service_demo_py`
> Job: `ci_launcher`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9849/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5465/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8010/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9703/)

mm318:
I have pushed a fix to https://github.com/ros2/ros2cli/pull/410. @ros2/aws-oncall, can you please re-run CI (https://github.com/ros2/rclpy/pull/459#issuecomment-606451193)?

dabonnie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9856/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5472/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8016/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9711/)

mm318:
Hi @ivanpauno, can you take a look at the CI results at https://github.com/ros2/rclpy/pull/459#issuecomment-606820891? I think the macOS and Windows CI test failures are unrelated.

ivanpauno:
> Hi @ivanpauno, can you take a look at the CI results at [#459 (comment)](https://github.com/ros2/rclpy/pull/459#issuecomment-606820891)? I think the macOS and Windows CI test failures are unrelated.

Seems to be unrelated to me.

mm318:
Great! Then this set of pull requests are ready to merge:
- https://github.com/ros2/rclcpp/pull/924
- https://github.com/ros2/rclpy/pull/459
- https://github.com/ros2/ros2cli/pull/410
- https://github.com/ros2/demos/pull/416

Thanks, @ivanpauno!

ivanpauno:
> ros2/demos#416

Please, address the comment in that one. I will merge the others now.

PR #461, opened 2019-11-15
Future invokes done callbacks when done

### LINKED ISSUE #460: Future done callbacks sometimes discarded instead of called
## Bug report

**Required Info:**

- Operating System:
  - N/A
- Installation type:
  - N/A
- Version or commit hash:
  - master and earlier
- DDS implementation:
  - N/A
  - rclpy

#### Steps to reproduce issue
Create an `rclpy.task.Future`, and add a done callback to it. Set a result on the future so it becomes done. The done callback will never be called

```python3
import rclpy.task

fut = rclpy.task.Future()
fut.add_done_callback(lambda f: print('Future done!'))
fut.set_result(42)
```

#### Expected behavior

`Future done!` should be printed to the console.

#### Actual behavior

Nothing is printed to the console.

#### Additional information

The future isn't associated with an executor, so `Future._schedule_done_callbacks()` just discards them :upside_down_face: 

https://github.com/ros2/rclpy/blob/772115d45f23a3842660c79a1438f72b7610418d/rclpy/rclpy/task.py#L130-L136

The only future's who's done callbacks are called are ones that are known to the executor. This includes [futures used for service clients](https://github.com/ros2/rclpy/blob/772115d45f23a3842660c79a1438f72b7610418d/rclpy/rclpy/executors.py#L345) and [futures used for waitables](https://github.com/ros2/rclpy/blob/772115d45f23a3842660c79a1438f72b7610418d/rclpy/rclpy/executors.py#L369). I forgot when using [futures in `tf2_ros`](https://github.com/ros2/geometry2/blob/71f823e90b699c9535ce682db777f7bc88cdfbcd/tf2_ros/src/tf2_ros/buffer.py#L241), and as a result done callbacks on those futures are never called.

`asyncio` solves this problem by having a single global event loop that `asyncio.future.Future` [can always get a reference to](https://github.com/python/cpython/blob/abde52cd8e31830bfc06c5803221faae6172104a/Lib/asyncio/futures.py#L76). This won't work for rclpy since there may be multiple executors, and there is no global reference to them. Even if there was a global reference, the future can't know which executor it will be associated with.

`concurrent.futures` solves this problem by [calling the done callbacks at the point the future is completed](https://github.com/python/cpython/blob/25ce77dd2455abbb6e2c9e055bbc98954642fa7c/Lib/concurrent/futures/_base.py#L325). This means a done callback potentially block code completing the future, but the callbacks are always guaranteed to run.

sloretz:
Fixes #460 

If an executor is set, the future try to schedule done callback on it. If no executor is set, the future will invoke the done callbacks directly. This is sort of a hybrid of how `asyncio` and `concurrent.futures` handle this situation.

This should be backported to Dashing.

### DIFF
--- a/rclpy/rclpy/task.py
+++ b/rclpy/rclpy/task.py
@@ -15,6 +15,7 @@
 import inspect
 import sys
 import threading
+import warnings
 import weakref
 
 
@@ -61,7 +62,7 @@ class Future:
         with self._lock:
             if not self._done:
                 self._cancelled = True
-            self._schedule_done_callbacks()
+        self._schedule_or_invoke_done_callbacks()
 
     def cancelled(self):
         """
@@ -112,7 +113,7 @@ class Future:
             self._result = result
             self._done = True
             self._cancelled = False
-            self._schedule_done_callbacks()
+        self._schedule_or_invoke_done_callbacks()
 
     def set_exception(self, exception):
         """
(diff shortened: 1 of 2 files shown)


sloretz:
CI (Testing all packages above rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8649/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4492/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7056/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8568/)

sloretz:
CI looks ok, just flake8 warnings already fixed by https://github.com/ros2/ros2cli/pull/399 and https://github.com/ros2/launch_ros/pull/94

PR #463, opened 2019-11-15
make short key of a QoS policy accessible

dirk-thomas:
Allows to retrieve the name of the QoS value.

### DIFF
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -251,6 +251,17 @@ class QoSPolicyEnum(IntEnum):
         """Retrieve a policy type from a short name, case-insensitive."""
         return cls[name.upper()].value
 
+    @property
+    def short_key(self):
+        for k, v in self.__class__.__members__.items():
+            if k.startswith('RMW'):
+                continue
+            if self.value == v:
+                return k.lower()
+        raise AttributeError(
+            'failed to find value %s in %s' %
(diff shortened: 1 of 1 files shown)


rclpy/rclpy/qos.py
```diff
@@ -251,6 +251,17 @@ def get_from_short_key(cls, name):
         """Retrieve a policy type from a short name, case-insensitive."""
         return cls[name.upper()].value
 
+    @property
+    def short_key(self):
```
sloretz: This sort of exists in `name`, although it has the `RMW_` prefix. Would that work for your use case?

```python
>>> rclpy.qos.QoSProfile(depth=10).durability.name
'RMW_QOS_POLICY_DURABILITY_VOLATILE'
>>> rclpy.qos.DurabilityPolicy[rclpy.qos.QoSProfile(depth=10).durability.name]
<DurabilityPolicy.RMW_QOS_POLICY_DURABILITY_VOLATILE: 2>
```

emersonknapp: This is exposing the singular case of the plural already exposed on line 245 https://github.com/ros2/rclpy/pull/463/files#diff-8110981f890a0a14e8a04fb4d65d5832R245 - this removes more than just RMW_, it removes all redundant context for use as commandline arguments - e.g. `--durability-qos volatile` vs `--durability-qos qos_policy_durability_volatile`

PR #468, opened 2019-11-26
Make use of rcutils log severity defined enum instead of duplicating code

BMarchi:
Follow up of https://github.com/ros2/rclpy/pull/458. Remove log severity hard coded and copied values from rcutils

### DIFF
--- a/rclpy/src/rclpy/_rclpy_logging.c
+++ b/rclpy/src/rclpy/_rclpy_logging.c
@@ -199,6 +199,54 @@ rclpy_logging_severity_level_from_string(PyObject * Py_UNUSED(self), PyObject * 
   return PyLong_FromLongLong(severity);
 }
 
+/// Get log unset severity level as int from rcutils
+/// \return RCUTILS_LOG_SEVERITY_UNSET
+static PyObject *
+rclpy_get_unset_logging_severity(PyObject * Py_UNUSED(self), PyObject * Py_UNUSED(args))
+{
+  return PyLong_FromLongLong(RCUTILS_LOG_SEVERITY_UNSET);
+}
+
+/// Get log debug severity level as int from rcutils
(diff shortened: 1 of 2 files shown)


BMarchi:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8740/)
- Arch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4569/)
- macOS [Build Status](http://ci.ros2.org/job/ci_osx/7142/)
- Windows [Build Status](http://ci.ros2.org/job/ci_windows/8654/)

rclpy/src/rclpy/_rclpy_logging.c
```diff
@@ -199,6 +199,54 @@ rclpy_logging_severity_level_from_string(PyObject * Py_UNUSED(self), PyObject *
   return PyLong_FromLongLong(severity);
 }
 
+/// Get log unset severity level as int from rcutils
+/// \return RCUTILS_LOG_SEVERITY_UNSET
```
hidmic: @BMarchi nit^2: `\return RCUTILS_LOG_SEVERITY_UNSET as a PyLong`, here and everywhere else.

dirk-thomas:
The patch is fine for now.

I am still worried that this will get us further away from being able to import `rclpy` modules side effect free (see #420). But that shouldn't hold up this PR.

hidmic:
> I am still worried that this will get us further away from being able to import `rclpy` modules side effect free (see #420). But that shouldn't hold up this PR.

Yes, I remember the ticket and the discussions. Thanks for bumping that this one up. I also recall that most variations trying to circumvent the problem weren't great either. It'd be great if we could spend the time to really solve the problem i.e. C extensions need not be special, it's a trap to make them so.

PR #474, opened 2019-12-06
Guard against unexpected action responses

jacobperron:
Fixes https://github.com/ros2/demos/issues/417

If multiple action servers are running with the same action name, then it is possible that both
servers will reply to requests from the action client. In the event that this happens, we avoid
crashing the code by ignoring the unexpected response and logging a warning for the user.
We have similar behavior for the action client implementation in rclcpp.

---

Example output running the Python action tutorials client node when there are two action servers running:

```sh
$ ros2 run action_tutorials_py fibonacci_action_client
[INFO] [1575593387.895730727] [fibonacci_action_client]: Received feedback: array('i', [0, 1, 1])
[INFO] [1575593387.896286699] [fibonacci_action_client]: Goal accepted :)
[WARN] [1575593387.897040932] [fibonacci_action_client]: Ignoring unexpected goal response. There may be two or more action servers for the action 'fibonacci'
[INFO] [1575593387.897367530] [fibonacci_action_client]: Received feedback: array('i', [0, 1, 1])
[INFO] [1575593388.884332723] [fibonacci_action_client]: Received feedback: array('i', [0, 1, 1, 2])
[INFO] [1575593388.886578244] [fibonacci_action_client]: Received feedback: array('i', [0, 1, 1, 2])
[INFO] [1575593389.882818973] [fibonacci_action_client]: Received feedback: array('i', [0, 1, 1, 2, 3])
[INFO] [1575593389.884030116] [fibonacci_action_client]: Received feedback: array('i', [0, 1, 1, 2, 3])
[INFO] [1575593390.884331581] [fibonacci_action_client]: Received feedback: array('i', [0, 1, 1, 2, 3, 5])
[INFO] [1575593390.887038829] [fibonacci_action_client]: Received feedback: array('i', [0, 1, 1, 2, 3, 5])
[INFO] [1575593391.882985796] [fibonacci_action_client]: Received feedback: array('i', [0, 1, 1, 2, 3, 5, 8])
[INFO] [1575593391.884393489] [fibonacci_action_client]: Received feedback: array('i', [0, 1, 1, 2, 3, 5, 8])
[INFO] [1575593392.882920103] [fibonacci_action_client]: Received feedback: array('i', [0, 1, 1, 2, 3, 5, 8, 13])
[INFO] [1575593392.884465125] [fibonacci_action_client]: Received feedback: array('i', [0, 1, 1, 2, 3, 5, 8, 13])
[...]

PR #475, opened 2019-12-06
[dashing backport] Guard against unexpected action responses

jacobperron:
Backport #474 to Dashing.

If multiple action servers are running with the same action name, then it is possible that both
servers will reply to requests from the action client. In the event that this happens, we avoid
crashing the code by ignoring the unexpected response and logging a warning for the user.
We have similar behavior for the action client implementation in rclcpp.

### DIFF
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -148,6 +148,7 @@ class ActionClient(Waitable):
         check_for_type_support(action_type)
         self._node = node
         self._action_type = action_type
+        self._action_name = action_name
         with node.handle as node_capsule:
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
I ran CI for this and #477 together against Dashing. The failures are all related to linters which have been upgraded since Dashing and so are not blockers.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8859/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4653/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7254/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8772/)

PR #476, opened 2019-12-06
[eloquent backport] Guard against unexpected action responses

jacobperron:
Backport #474 to Eloquent.

If multiple action servers are running with the same action name, then it is possible that both
servers will reply to requests from the action client. In the event that this happens, we avoid
crashing the code by ignoring the unexpected response and logging a warning for the user.
We have similar behavior for the action client implementation in rclcpp.


jacobperron:

jacobperron:

nuclearsandwich:

PR #488, opened 2020-01-03
Avoid a reference cycle between Node and TimeSource

ivanpauno:
See https://github.com/ros2/rclpy/pull/470#issuecomment-570629990.

### DIFF
--- a/rclpy/rclpy/time_source.py
+++ b/rclpy/rclpy/time_source.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import weakref
+
 from rcl_interfaces.msg import SetParametersResult
 from rclpy.clock import ClockType
 from rclpy.clock import ROSClock
@@ -27,7 +29,7 @@ class TimeSource:
 
     def __init__(self, *, node=None):
         self._clock_sub = None
-        self._node = None
+        self._node_weak_ref = None
         self._associated_clocks = []
         # Zero time is a special value that means time is uninitialzied
         self._last_time_set = Time(clock_type=ClockType.ROS_TIME)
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/time_source.py
```diff
@@ -91,12 +97,14 @@ def attach_node(self, node):
 
     def detach_node(self):
         # Remove the subscription to the clock topic.
+        error = RuntimeError('Unable to destroy previously created clock subscription')
```
dirk-thomas: Since this temporary variable is only used in a single case I would rather leave the `RuntimeError` there and not at a variable for it.

rclpy/rclpy/time_source.py
```diff
@@ -91,12 +97,14 @@ def attach_node(self, node):
...
-            if self._node is None:
-                raise RuntimeError('Unable to destroy previously created clock subscription')
-            self._node.destroy_subscription(self._clock_sub)
+            node = self._get_node()
+            if node is None:
+                raise error
+            node.destroy_subscription(self._clock_sub)
         self._clock_sub = None
-        self._node = None
+        node = None
```
dirk-thomas: Why is this line necessary? The local var should go out of scope at the end of the function (which is right after this line) anyway.

rclpy/rclpy/time_source.py
```diff
@@ -119,9 +127,19 @@ def _on_parameter_event(self, parameter_list):
...
+                            .format(USE_SIM_TIME_NAME))
                 break
 
         return SetParametersResult(successful=True)
+
+    def _get_node(self):
+        if self._node_weak_ref is not None:
+            node = self._node_weak_ref()
+            if node is None:
+                raise RuntimeError('Attached node has gone out of scope before expected')
```
dirk-thomas: "before expected" sounds weird to me (I am not a native English speaker though). Maybe:

> Attached node has gone out of scope already

But actually in all the calls you check if the returned value is `None`. In which cases is the weak ref still set but the node is already out of scope and you actually want an exception here instead of just returning `None`?

rclpy/rclpy/time_source.py
sloretz: Is this meant to be `self._node_weak_ref = None`? It seems like the weak reference isn't needed after `detach_node()` is called.

rclpy/rclpy/time_source.py
```diff
@@ -119,9 +127,19 @@ def _on_parameter_event(self, parameter_list):
...
+            return node
+        return None
```
sloretz: I would think this case is worth raising for (trying to do an operation that requires a node, but the time source is not attached to one), and the other case of the weak reference being invalid is expected and not worth raising for.

rclpy/rclpy/time_source.py
ivanpauno: I was planning to use the `error` more than once, but then I didn't.

rclpy/rclpy/time_source.py
ivanpauno: Yes, @sloretz is correct.

rclpy/rclpy/time_source.py
ivanpauno: Yeah, returning just `None` is also an option. I updated the code.

rclpy/rclpy/time_source.py
ivanpauno: I changed the code to not raise an error in case the weak reference is invalid.
The code previous to this change was working when `self._node` was `None`, so I don't think that an error shold be raised in that case.

mjcarroll:
Let's make sure to include the `ros2*` test suite in this CI, to see if issues with nightly hangs are resolved (specifically `ros2topic` and `ros2action`)

ivanpauno:
> Let's make sure to include the `ros2*` test suite in this CI, to see if issues with nightly hangs are resolved (specifically `ros2topic` and `ros2action`)

The hangs observed in CI might be related with `rclpy` object destruction order problems.
This PR together with #490 might "fix" the problems.
But the value of this PR is completly unrelated to whatever those tests hang or not, so I'm just going to include `rclpy` in the CI run.

CI up to rclpy, only fastrtps:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8959/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4729/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7331/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8879/)

ivanpauno:
The windows fail is an unrelated flaky test, here this same branch passing:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8890/)

I'm merging this!

PR #490, opened 2020-01-06
Avoid reference cycle between Node and ParameterService

ivanpauno:
See https://github.com/ros2/rclpy/pull/470#issuecomment-570629990.

### DIFF
--- a/rclpy/rclpy/parameter_service.py
+++ b/rclpy/rclpy/parameter_service.py
@@ -159,5 +159,5 @@ class ParameterService:
     def _get_node(self):
         node = self._node_weak_ref()
         if node is None:
-            raise RuntimeError('Expected valid node weak reference')
+            raise ReferenceError('Expected valid node weak reference')
         return node
--- a/rclpy/rclpy/parameter_service.py
+++ b/rclpy/rclpy/parameter_service.py
@@ -25,125 +25,135 @@ from rclpy.validate_topic_name import TOPIC_SEPARATOR_STRING
 class ParameterService:
 
     def __init__(self, node):
-        self._node = node
+        def _describe_parameters_callback(self, request, response):
+            for name in request.names:
+                try:
+                    descriptor = node.describe_parameter(name)
+                except ParameterNotDeclaredException:
+                    descriptor = ParameterDescriptor()
+                response.descriptors.append(descriptor)
+            return response
+
+        def _get_parameters_callback(self, request, response):
+            for name in request.names:
+                p = node.get_parameter_or(name)
+                response.values.append(p.get_parameter_value())
+            return response
+
+        def _get_parameter_types_callback(self, request, response):
+            for name in request.names:
+                response.types.append(node.get_parameter_or(name).type_)
+            return response
+
+        def _list_parameters_callback(self, request, response):
+            names_with_prefixes = []
+            for name in node._parameters.keys():
+                if PARAMETER_SEPARATOR_STRING in name:
+                    names_with_prefixes.append(name)
+                    continue
+                elif request.prefixes:
+                    for prefix in request.prefixes:
+                        if name.startswith(prefix):
+                            response.result.names.append(name)
+                    continue
+                else:
+                    response.result.names.append(name)
+            if 1 == request.depth:
+                return response
+
+            if not request.DEPTH_RECURSIVE == request.depth:
+                names_with_prefixes = filter(
+                    lambda name:
+                        name.count(PARAMETER_SEPARATOR_STRING) < request.depth, names_with_prefixes
+                )
+            for name in names_with_prefixes:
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Why did you choose to refactor all methods to local function? This seems to make it much harder to write tests for the numerous functions since they are unaccessible from the outside. Wouldn't a `weakref` of the node be an option here too?

ivanpauno:
> Why did you choose to refactor all methods to local function? This seems to make it much harder to write tests for the numerous functions since they are unaccessible from the outside. Wouldn't a `weakref` of the node be an option here too?

Actually, it wasn't just only bad for unit testing, it wasn't breaking the reference cycle ([1](https://github.com/ros2/rclpy/blob/997703a0f5c841ea007b00719acc5a3076066747/rclpy/rclpy/node.py#L146)).

I now tested this together with #488, and #470 destruction order is ok.

rclpy/rclpy/parameter_service.py
```diff
@@ -64,28 +66,37 @@ def __init__(self, node):
...
 
     def _describe_parameters_callback(self, request, response):
-        for name in request.names:
-            try:
-                descriptor = self._node.describe_parameter(name)
-            except ParameterNotDeclaredException:
-                descriptor = ParameterDescriptor()
-            response.descriptors.append(descriptor)
+        node = self._node_weak_ref()
+        if node is not None:
```
dirk-thomas: If the node is not available anymore is it sufficient to return the unmodified `response` or should any kind of exception be raised instead?

Same below.

ivanpauno: I think it's ok to return the unmodified response (which is the current behavior).

Also, it's currently impossible that the callback would be called after the node gets destroyed (as just the node is storing a reference to the created services), so the weak reference should be never invalid.

dirk-thomas: > it's currently impossible that the callback would be called after the node gets destroyed ... so the weak reference should be never invalid.

If that is an invariant why not assert that explicitly?

ivanpauno: Fixed in https://github.com/ros2/rclpy/pull/490/commits/d41fc53d10d39eafdc943e839ddec2d1a9d883ac.

rclpy/rclpy/parameter_service.py
```diff
@@ -138,12 +145,19 @@ def _set_parameters_callback(self, request, response):
...
             response.result = SetParametersResult(
                     successful=False,
                     reason=str(e)
                 )
         return response
+
+    def _get_node(self):
+        node = self._node_weak_ref()
+        if node is None:
+            raise RuntimeError('Expected valid node weak reference')
```
dirk-thomas: Should this be a `ValueError` instead (or what is the reason to choose `RuntimeError`)?

ivanpauno: AFAIU, [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) is used when an argument passed by the user is not of the appropiate value. There's not "argument" involved here, just an attribute of the object being invalid.

`RuntimeError` is the choice when the error doesn't fall in any of the other categories ([link](https://docs.python.org/3/library/exceptions.html#RuntimeError)), and I don't think there is a more appropiate standard error.
I also can create a custom error if that sounds better.

dirk-thomas: I was mostly wondering because `RuntimeError` is treated differently by the command line tools - they suppress the stacktrace and only show the exception string. Not sure if that is intended in this case.

dirk-thomas: It sounds like [ReferenceError](https://docs.python.org/3/library/exceptions.html#ReferenceError) is what we want here.

ivanpauno: > It sounds like ReferenceError is what we want here.

Sounds like the right option! Updated the PR.

> I was mostly wondering because RuntimeError is treated differently by the command line tools - they suppress the stacktrace and only show the exception string. Not sure if that is intended in this case.

A little off-topic:
I have observed this, and find that behavior strange. Why do the command line tools do that?

dirk-thomas: Some kind of error messages don't warrant to show a stacktrace to the user. And in those cases we use `RuntimeError`.

ivanpauno:
CI up to rclpy, only fastrtps:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8967/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4737/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7339/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8901/)

PR #495, opened 2020-01-10
Add wrappers for RMW serialize and deserialize functions

jacobperron:
Related to https://github.com/ros2/rosbag2/issues/232

Added a serialization submodule with functions for serializing and deserializing ROS messages.

I noticed an issue related to the serialization of floating-point values. I believe it's a bug related to loss of precision for 32-bit floating point values as (de)serialization happens with underlying C-types. I'm not sure how to resolve it at the moment, but left a TODO and test that if uncommented
demonstrates the bug.

### DIFF
--- a/rclpy/rclpy/serialization.py
+++ b/rclpy/rclpy/serialization.py
@@ -1,0 +1,44 @@
+# Copyright 2020 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+from rclpy.type_support import check_for_type_support
+
+
+def deserialize_message(serialized_message: bytes, message_type):
+    """
+    Deserialize a ROS message.
+
+    :param serialized_message: The ROS message to deserialized.
+    :param message_types: The type of the serialized ROS message.
+    :return: The deserialized ROS message.
+    """
+
+    # this line imports the typesupport for the message module if not already done
+    check_for_type_support(message_type)
+    return _rclpy.rclpy_deserialize(serialized_message, message_type)
+
+
+def serialize_message(message) -> bytes:
+    """
+    Serialize a ROS message.
+
+    :param message: The ROS message to serialize.
+    :return: The serialized bytes.
+    """
+
+    # this line imports the typesupport for the message module if not already done
+    check_for_type_support(message)
+    return _rclpy.rclpy_serialize(message)
+
--- a/rclpy/rclpy/serialization.py
+++ b/rclpy/rclpy/serialization.py
@@ -16,6 +16,18 @@ from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 from rclpy.type_support import check_for_type_support
 
 
+def serialize_message(message) -> bytes:
+    """
+    Serialize a ROS message.
+
+    :param message: The ROS message to serialize.
+    :return: The serialized bytes.
+    """
+    # this line imports the typesupport for the message module if not already done
+    check_for_type_support(message)
+    return _rclpy.rclpy_serialize(message)
+
+
 def deserialize_message(serialized_message: bytes, message_type):
     """
     Deserialize a ROS message.
@@ -27,15 +39,3 @@ def deserialize_message(serialized_message: bytes, message_type):
     # this line imports the typesupport for the message module if not already done
     check_for_type_support(message_type)
     return _rclpy.rclpy_deserialize(serialized_message, message_type)
-
-
-def serialize_message(message) -> bytes:
-    """
-    Serialize a ROS message.
-
-    :param message: The ROS message to serialize.
-    :return: The serialized bytes.
-    """
-    # this line imports the typesupport for the message module if not already done
-    check_for_type_support(message)
-    return _rclpy.rclpy_serialize(message)
--- a/rclpy/rclpy/serialization.py
+++ b/rclpy/rclpy/serialization.py
@@ -23,9 +23,10 @@ def serialize_message(message) -> bytes:
     :param message: The ROS message to serialize.
     :return: The serialized bytes.
     """
+    message_type = type(message)
     # this line imports the typesupport for the message module if not already done
-    check_for_type_support(message)
-    return _rclpy.rclpy_serialize(message)
+    check_for_type_support(message_type)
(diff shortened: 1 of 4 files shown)


rclpy/test/test_serialization.py
```diff
@@ -0,0 +1,67 @@
...
+        msg_deserialized = deserialize_message(msg_serialized, msg_type)
+        assert msg == msg_deserialized
+
+
+# TODO(jacobperron): Fix this failing test.
+#                    I'm pretty sure that during (de)serialization we convert to a C float before
+#                    converting to a PyObject. This causes a loss of precision.
+# def test_set_float32():
+#     msg = BasicTypes()
+#     msg.float32_value = 3.14
```
dirk-thomas: Maybe try a float value which can be represented without rounding, e.g. 1.125?

jacobperron: I can confirm a value that doesn't require rounding, like 1.125, works fine. Do you think this TODO is still valid though?

dirk-thomas: The comment that the value will be stored in a C type still applies. I am just not sure what the TODO should be about?

jacobperron: Maybe there's nothing we can do about this. I could instead update the test to do the rounding in Python so it passes.

Karsten1987:
are you considering adding `rcl_take_serialized_message` and/or `rcl_publish_serialized_message` as well here?

I am asking because this would make it easy to check the interoperability between the serialization in python as well as c++.

jacobperron:
> are you considering adding rcl_take_serialized_message and/or rcl_publish_serialized_message as well here?

I hadn't, but we could expose these in a follow-up PR.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -4691,6 +4691,98 @@ rclpy_get_node_parameters(PyObject * Py_UNUSED(self), PyObject * args)
...
+  rosidl_message_type_support_t * ts =
+    (rosidl_message_type_support_t *)rclpy_common_get_type_support(pymsg_type);
+  if (!ts) {
+    return NULL;
+  }
+
+  // Create a serialized message object
+  rcl_serialized_message_t serialized_msg = rmw_get_zero_initialized_serialized_message();
+  // Just copy pointer to avoid extra allocation and copy
+  serialized_msg.buffer_capacity = serialized_buffer_size;
```
Karsten1987: can we guarantee that at this point the size and the capacity of the buffer are equivalent?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -4691,6 +4691,98 @@ rclpy_get_node_parameters(PyObject * Py_UNUSED(self), PyObject * args)
...
+    rcutils_ret = rmw_serialized_message_fini(&serialized_msg);
+    if (RCUTILS_RET_OK != rcutils_ret) {
+      PyErr_Format(RCLError,
+        "Failed to finalize serialized message: %s", rcutils_get_error_string().str);
+    }
+    return NULL;
+  }
+
+  // Bundle serialized message in a bytes object
+  return Py_BuildValue("y#", serialized_msg.buffer, serialized_msg.buffer_length);
```
Karsten1987: is it safe to ignore the capacity here? Does this also make the `bytes object` immutable?

jacobperron: The format string "y#" returns a [bytes](https://docs.python.org/3/library/stdtypes.html#bytes-objects) object which is immutable. So there's no capacity member AFAIK.

rclpy/src/rclpy/_rclpy.c
jacobperron: Yes, the input, declare as "y#", is immutable ([Reference](https://docs.python.org/3/c-api/arg.html#strings-and-buffers)).

rclpy/test/test_serialization.py
jacobperron: I've replaced the TODO with a test that illustrates what could happen with float32 types: 668c9b9

jacobperron:
CI up to rclpy:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9025/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4784/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7385/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8959/)

rclpy/rclpy/serialization.py
```diff
@@ -0,0 +1,41 @@
...
+def serialize_message(message) -> bytes:
+    """
+    Serialize a ROS message.
+
+    :param message: The ROS message to serialize.
+    :return: The serialized bytes.
+    """
+    # this line imports the typesupport for the message module if not already done
+    check_for_type_support(message)
+    return _rclpy.rclpy_serialize(message)
```
dirk-thomas: Nitpick: I would suggest swapping the order of these two functions. It feels just more natural to have 1) serialize and 2) deserialize (which is also the order they are used in the tests).

rclpy/rclpy/serialization.py
```diff
@@ -0,0 +1,41 @@
...
+
+def serialize_message(message) -> bytes:
+    """
+    Serialize a ROS message.
+
+    :param message: The ROS message to serialize.
+    :return: The serialized bytes.
+    """
+    # this line imports the typesupport for the message module if not already done
+    check_for_type_support(message)
```
dirk-thomas: According to the docblock of `check_for_type_support` the function expects a message type - not a message?

rclpy/rclpy/serialization.py
jacobperron: I purposefully added them alphabetically before, but swapping the order makes sense too.

rclpy/rclpy/serialization.py
jacobperron: I guess? It looks like both the message and the type have the same attributes that it expects, e.g.
  
        msg = BasicTypes()
        msg.__class__._TYPE_SUPPORT
        msg.__class__.__import_type_support__()
        BasicTypes.__class__._TYPE_SUPPORT
        BasicTypes.__class__.__import_type_support__()

I'm abusing this in the C-extension too:

https://github.com/ros2/rclpy/blob/668c9b9e3ff07ac78281d80a15394a38f896860a/rclpy/src/rclpy/_rclpy.c#L4703-L4704

To err on the side of caution, I've switched to using types ac346e4

rclpy/rclpy/serialization.py
jacobperron: Reordered 40782e7

rclpy/rclpy/serialization.py
dirk-thomas: I see. Maybe updating the docblock of that function would be good to state that it can be used with a message instance too. (Not necessary as part of this PR.)

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9044/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4796/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7394/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8974/)

PR #497, opened 2020-01-20
Fix object destruction order

ivanpauno:
Supersedes https://github.com/ros2/rclpy/pull/470.
See https://github.com/ros2/rclpy/pull/470#issuecomment-567058281.

Solves strange bugs related with destruction order.

Included in this PR:
- `rclpy_handle_t` struct, related methods, and usage of it in `Handle` class.
- Replaces usage of PyCapsule for `rclpy_handle_t` where needed.
- Some formatting changes (avoid unnecessary castings, etc).

Not included (subject of a follow up):
- `rmw_request_id_t` capsule is leaking.
- `rmw_qos_profile_t` capsule is leaking.
- Refactor `wait set` related code, so it can be wrapped by a `Handle`.
- Refactor `rclpy_action` objects, so they can be wrapped in a `Handle`.


ivanpauno:
CI above `rclpy`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9083/)

ivanpauno:
New CI, after fixing failures:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9087/)

rclpy/src/rclpy_common/include/rclpy_common/handle.h
```diff
@@ -0,0 +1,88 @@
...
+#endif
+
+#include <Python.h>
+#include <stddef.h>
+
+#include "rcutils/types/rcutils_ret.h"
+
+/// Wrapper that manages the lifetime of another object,
+/// allowing to establish dependencies between them.
+typedef struct rclpy_handle_t rclpy_handle_t;
```
hidmic: @ivanpauno nit: why not collapsing both `typedef` and `struct` definition below into a single statement?

rclpy/src/rclpy_common/include/rclpy_common/handle.h
```diff
@@ -0,0 +1,88 @@
...
+
+/// Signature of rclpy_handle_t destructors.
+typedef void (* rclpy_handle_destructor_t)(void *);
+
+/// Create a PyCapsule wrapping a rclpy_handle_t object.
+/**
+ * \sa: _rclpy_create_handle
+ * \param name: Name of the PyCapsule.
```
hidmic: @ivanpauno nit: here and below, colons after doxygen commands are not necessary.

rclpy/src/rclpy_common/src/handle.c
```diff
@@ -0,0 +1,127 @@
...
+#include "rclpy_common/handle.h"
+
+/// Creates a Handle object.
+rclpy_handle_t *
+_rclpy_create_handle(void * ptr, rclpy_handle_destructor_t destructor)
+{
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+  rclpy_handle_t * handle = allocator.zero_allocate(1, sizeof(rclpy_handle_t), allocator.state);
+  if (!handle) {
+    return NULL;
```
hidmic: @ivanpauno considering setting a `RuntimeError` or even a `MemoryError` exception before returning.

rclpy/src/rclpy_common/src/handle.c
```diff
@@ -0,0 +1,127 @@
...
+#include "rcutils/strdup.h"
+#include "rcutils/types/rcutils_ret.h"
+
+#include "rclpy_common/handle.h"
+
+/// Creates a Handle object.
+rclpy_handle_t *
+_rclpy_create_handle(void * ptr, rclpy_handle_destructor_t destructor)
+{
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
```
hidmic: @ivanpauno missing `assert`s on `ptr` and `destructor`?

rclpy/src/rclpy_common/src/handle.c
```diff
@@ -0,0 +1,127 @@
...
+ */
+rcutils_ret_t
+_rclpy_handle_add_dependency(rclpy_handle_t * dependent, rclpy_handle_t * dependency)
+{
+  assert(dependent);
+  assert(dependency);
+
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+
+  dependent->num_of_dependencies++;
```
hidmic: @ivanpauno I think this should not be incremented before we're certain the re-allocation succeeded.

rclpy/src/rclpy_common/src/handle.c
hidmic: Or do so in the caller scope.

rclpy/src/test/rclpy_common/test_handle.cpp
```diff
@@ -0,0 +1,95 @@
...
+// limitations under the License.
+
+#include <gtest/gtest.h>
+#include <memory>
+
+#include "rclpy_common/handle.h"
+
+void int_destructor(void * p)
+{
+  *reinterpret_cast<int *>(p) = 1;
```
hidmic: @ivanpauno nit: consider turning this into an increment, so that a destructor being called twice does not goes by unnoticed.

rclpy/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.5)
 
-project(rclpy C)
+project(rclpy)
```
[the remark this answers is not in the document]
ivanpauno: Because I added a test written in `CPP` (using `gtest`).

rclpy/src/rclpy/_rclpy.c
```diff
@@ -2766,7 +2744,7 @@ rclpy_wait(PyObject * Py_UNUSED(self), PyObject * args)
   if (!PyArg_ParseTuple(args, "O|K", &pywait_set, &timeout)) {
     return NULL;
   }
-  rcl_wait_set_t * wait_set = (rcl_wait_set_t *)PyCapsule_GetPointer(pywait_set, "rcl_wait_set_t");
+  rcl_wait_set_t * wait_set = PyCapsule_GetPointer(pywait_set, "rcl_wait_set_t");
```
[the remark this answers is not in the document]
ivanpauno: Good question. Currently, the `rcl_wait_set_t` capsule didn't have a deleter, and deletion was handled manually in `Python`.
To avoid doing this PR bigger, I didn't refactor how wait sets were used. It should be done in a follow up PR.

rclpy/CMakeLists.txt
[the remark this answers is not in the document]
ivanpauno: According to [cmake documentation](https://cmake.org/cmake/help/latest/command/project.html), the default is `C CXX`, so deleting `C` seems like the correct thing to do (in all the others `C/CXX` projects we don't specify a language).

rclpy/src/rclpy_common/include/rclpy_common/handle.h
ivanpauno: I moved the struct definition to the source file now, this is a forward declaration.

rclpy/src/rclpy/_rclpy.c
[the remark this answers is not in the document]
hidmic: Let's open an issue to track that work.

rclpy/CMakeLists.txt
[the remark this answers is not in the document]
hidmic: Fair enough.

rclpy/src/test/rclpy_common/test_handle.cpp
```diff
@@ -0,0 +1,95 @@
...
+  // Add dependency.
+  _rclpy_handle_add_dependency(dependent_handle, dependency_handle);
+
+  // Decrement reference of dependency, nothing is destructed.
+  _rclpy_handle_dec_ref(dependent_handle);
+  ASSERT_EQ(dependency_obj.get(), _rclpy_handle_get_pointer(dependency_handle));
+  EXPECT_EQ(1, *dependent_obj);
+  EXPECT_EQ(0, *dependency_obj);
+
+  // Decrement reference of dependent, both are destructed.
```
hidmic: @ivanpauno 
```suggestion
  // Decrement reference of dependency, dependency is destructed.
```

rclpy/src/test/rclpy_common/test_handle.cpp
```diff
@@ -0,0 +1,95 @@
...
+  ASSERT_EQ(dependent_obj.get(), _rclpy_handle_get_pointer(dependent_handle));
+  ASSERT_NE(nullptr, dependency_handle);
+  ASSERT_EQ(dependency_obj.get(), _rclpy_handle_get_pointer(dependency_handle));
+  EXPECT_EQ(0, *dependent_obj);
+  EXPECT_EQ(0, *dependency_obj);
+
+  // Add dependency.
+  _rclpy_handle_add_dependency(dependent_handle, dependency_handle);
+
+  // Decrement reference of dependency, nothing is destructed.
```
hidmic: @ivanpauno 
```suggestion
  // Decrement reference of dependent, dependent is destructed.
```

rclpy/rclpy/handle.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 Open Source Robotics Foundation, Inc.
+# Copyright 2019-2020 Open Source Robotics Foundation, Inc.
```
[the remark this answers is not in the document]
dirk-thomas: Adding the current year when making significant changes is totally fine (and even desired). We usually just don't do it...

rclpy/rclpy/handle.py
```diff
@@ -159,42 +153,9 @@ def __destroy(self):
...
-            while self.__destroy_callbacks:
-                self.__destroy_callbacks.pop()(self)
-            # get rid of references to other handles to break reference cycles
-            del self.__required_handles
+
+        # Calls pycapsule destructor
+        _rclpy_capsule.rclpy_pycapsule_destroy(self.__capsule)
+        # Call post-destroy callbacks
+        while self.__destroy_callbacks:
+            self.__destroy_callbacks.pop()(self)
```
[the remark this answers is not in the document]
ivanpauno: I think that deleting the feature makes sense.

rclpy/src/rclpy_common/src/handle.c
```diff
@@ -0,0 +1,146 @@
...
+  handle->ref_count++;
+  handle->destructor = destructor;
+
+  return handle;
+}
+
+/// Adds a dependency to a handle.
+/**
+ * The `dependency` handle reference count is incresead.
+ * The `dependent` handle stores `dependency` into its dependencies list.
```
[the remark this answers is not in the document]
ivanpauno: Do we want to check reference cycles?
I would rather not handle that case (like `std::shared_ptr`), and carefully check them when reviewing.

rclpy/rclpy/handle.py
```diff
@@ -68,30 +61,34 @@ def __hash__(self):
...
 
     def destroy(self, then=None):
         """
-        Destroy pycapsule as soon as possible without waiting for garbage collection.
+        Destroy the pycapsule as soon as possible without waiting for garbage collection.
+
+        The managed `rclpy_handle_t` object will not be destructed immediately if another handle
+        called `other.requires(this)`.
+        In that case, the managed object will be destroyed after all its
+        dependents are destroyed.
```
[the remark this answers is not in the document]
ivanpauno: > That duplicate knowledge is hard to do for Context. Without that knowledge duplicated it means rclpy.shutdown() doesn't actually shutdown the library; instead everything must wait for the garbage collector.

Up to now, shutting down rclpy wasn't destructing all the `rcl` created objects, as `Context` wasn't using a `Handle`. If we want that feature, I would rather do it in a follow-up PR, and limit this one to fix the order destruction problem (it can be done by keeping track of all the nodes created by a context).

There's also some inconsistency of what happens when a node is destructed:
- Publishers, Subscriptions, Clients, Services, Timers, Guards created by a node are destructed.
- Action clients and action servers aren't.

rclpy/src/rclpy_common/src/handle.c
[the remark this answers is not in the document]
hidmic: I agree with @ivanpauno. Preventing direct reference cycles only gives a false sense of safety, and growing this to become another `gc` is unnecessary. We just have to be careful. I will say though that I don't see the need for this API to be exposed. It could be handled entirely within the C extension, where we can be careful.

rclpy/rclpy/handle.py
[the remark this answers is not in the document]
hidmic: I will say though that you _could_ keep the list of Python callables within `rclpy_handle_t`. Agree on deleting it if nothing's using it.

rclpy/rclpy/handle.py
```diff
@@ -24,18 +24,15 @@ class InvalidHandle(Exception):
...
     ::
     with subscription.handle as pycapsule:
         ...
 
-    This class assumes the passed pycapsule has a destructor.
-    When this class destroys the capsule, it will call the destructor.
-    Then it will set the destructor to NULL so it is not called a second time when the capsule is
-    garbage collected.
+    :meth:`destroy` allows early destruction of the pycapsule.
+    The managed `rclpy_handle_t` will not be destructed until all its dependents are destructed.
```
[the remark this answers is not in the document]
ivanpauno: I've reworded the comment, hope it's clearer now.

rclpy/rclpy/handle.py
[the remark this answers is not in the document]
hidmic: We can also rely on things being explicitly destroyed in reversed order and fail if destruction of an object other objects still depend on is attempted.

---

If I may, I have to ask, where does this need of explicitly destroying objects come from? It's somewhat incompatible with Python, it creates invalid objects, it makes resource management more complex and it could be easily circumvented by calling `gc.collect` if need be. It almost seems to me like we're fighting the language to do things in a way it was not meant to.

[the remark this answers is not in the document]
sloretz: > If I may, I have to ask, where does this need of explicitly destroying objects come from? It's somewhat incompatible with Python, it creates invalid objects, it makes resource management more complex and it could be easily circumvented by calling gc.collect if need be. It almost seems to me like we're fighting the language to do things in a way it was not meant to.

I hear you. This is an old discussion. [I've been on that side before](https://github.com/ros2/rclpy/pull/127#discussion_r146351099), and I'm still mostly on that side. I agree we don't need to free memory right away. There's nothing special about our ram usage vs python's, so if the python garbage collector is good enough for some then it is good enough for all.

The reason to destroy these objects is not about their memory usage, it's about freeing system resources. Explicitly freeing system resources without waiting for the garbage collector is not fighting the language; it's pretty normal.

* https://docs.python.org/3/library/socket.html#socket.close
* https://docs.python.org/3/library/io.html#io.IOBase.close
* https://docs.python.org/3/library/subprocess.html#subprocess.Popen.kill

 Maybe it could be circumvented by calling `gc.collect()`, but that's not how python objects usually solve this situation: https://docs.python.org/3/tutorial/inputoutput.html .

> If you’re not using the with keyword, then you should call f.close() to close the file and immediately free up any system resources used by it. If you don’t explicitly close a file, Python’s garbage collector will eventually destroy the object and close the open file for you, but the file may stay open for a while. Another risk is that different Python implementations will do this clean-up at different times.

Yes `rcl` objects have to be destroyed in order, but that has nothing to do with why we want to destroy them. We want to destroy them because that's the only way the `rcl` API gives us to free system resources they use. Depending on the `rmw` implementation, nodes/pubs/subs/etc open sockets, they're using network bandwidth by continuing to participate in discovery, subscribers are receiving data from other publishers on the network, service servers may be accepting requests even if the intent is for another server to take their place, etc. That's where the need of explicitly destroying objects comes from.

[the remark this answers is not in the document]
ivanpauno: I understand the point of freeing resources. But, my point is that `rclpy.shutdown` is currently not doing so. I think that the responsability of doing that should be of the `Context` itself, and not of `Handle`. I can add that, but IMO it's subject of a follow up.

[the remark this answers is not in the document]
sloretz: > my point is that rclpy.shutdown is currently not doing so. I think that the responsability of doing that should be of the Context itself, and not of Handle. I can add that, but IMO it's subject of a follow up.

That's a fair point. I think I'm getting hung up on what I intended `Handle` to do vs the reality of that never being fully implemented. I do like the approach this PR takes, and I think it makes it pretty easy to get to a state of everything being cleaned up right away.

If `h1.requires(h0)` caused `h0` to store a weakref to `h1`, then `h0.destory()` could call `h1.destroy()` without worrying about destruction order of `h1`'s dependents because the reference counting in this PR would take care of that. I think doing that is important for the reasons above, but you're right it doesn't need to be done right away if the feature doesn't fully exist now. I'll dismiss my review since it seems to be blocking progress. Sorry about that.

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -281,15 +279,19 @@ rclpy_create_event(PyObject * Py_UNUSED(self), PyObject * args)
...
     return NULL;
   }
-
-  return pyevent;
+  _rclpy_handle_add_dependency(event_handle, parent_handle);
+  if (PyErr_Occurred()) {
+    _rclpy_handle_dec_ref(event_handle);
+  }
+  PyObject * event_capsule = _rclpy_create_handle_capsule(event_handle, "rcl_event_t");
+  return event_capsule;
```
hidmic: @ivanpauno nit: consider collapsing the last two statements.

rclpy/src/rclpy_common/include/rclpy_common/handle.h
```diff
@@ -0,0 +1,100 @@
...
+ * \param handle
+ */
+void *
+_rclpy_handle_get_pointer(rclpy_handle_t * handle);
+
+/// Adds a dependency to a handle.
+/**
+ * PyExc_RuntimeError is set, if `handle` is NULL, o if the managed pointer is NULL.
+ *
+ * \param dependency Handle object which its reference count will be incremented.
```
hidmic: @ivanpauno consider 
```suggestion
 * \param dependency Handle object whose reference count will be incremented.
```

ivanpauno: Solved in https://github.com/ros2/rclpy/pull/497/commits/034da37e83c05acab31b7efc4436aa55fb36e08b.

rclpy/rclpy/handle.py
```diff
@@ -24,18 +24,17 @@ class InvalidHandle(Exception):
...
     with subscription.handle as pycapsule:
         ...
 
-    This class assumes the passed pycapsule has a destructor.
-    When this class destroys the capsule, it will call the destructor.
-    Then it will set the destructor to NULL so it is not called a second time when the capsule is
-    garbage collected.
+    :meth:`destroy` allows early destruction of the pycapsule.
+
+    The managed `rclpy_handle_t` blocks destruction as long as another rcl object needs this one
```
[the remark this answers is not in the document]
ivanpauno: Solved in https://github.com/ros2/rclpy/pull/497/commits/034da37e83c05acab31b7efc4436aa55fb36e08b.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -719,26 +735,44 @@ rclpy_create_node(PyObject * Py_UNUSED(self), PyObject * args)
...
     PyErr_WarnFormat(
       PyExc_RuntimeWarning, stack_level, "Failed to fini arguments: %s",
       rcl_get_error_string().str);
+    rcl_reset_error();
   }
-  return PyCapsule_New(node, "rcl_node_t", _rclpy_destroy_node);
+
+  rclpy_handle_t * node_handle = _rclpy_create_handle(node, _rclpy_destroy_node);
+  if (!node_handle) {
+    rcl_ret_t ret = rcl_node_fini(node);
```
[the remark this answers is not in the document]
ivanpauno: Fixed in https://github.com/ros2/rclpy/pull/497/commits/3aee0703c52cbff8d1d81c394609724c95dacc15.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1390,7 +1423,25 @@ rclpy_create_publisher(PyObject * Py_UNUSED(self), PyObject * args)
     PyMem_Free(pub);
     return NULL;
   }
-  return PyCapsule_New(pub, "rclpy_publisher_t", _rclpy_destroy_publisher);
+
+  rclpy_handle_t * pub_handle = _rclpy_create_handle(pub, _rclpy_destroy_publisher);
+  if (!pub_handle) {
+    rcl_ret_t ret = rcl_publisher_fini(&(pub->publisher), pub->node);
```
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/rclpy/pull/497/commits/3aee0703c52cbff8d1d81c394609724c95dacc15.

rclpy/src/rclpy/_rclpy_qos_event.c
ivanpauno: Actually, error handling was wrong there. Fixed in https://github.com/ros2/rclpy/pull/497/commits/c0c56b9b4f24be37b4088154a5afea72821f5651.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1340,33 +1371,35 @@ rclpy_create_publisher(PyObject * Py_UNUSED(self), PyObject * args)
     return NULL;
   }
 
-  rcl_node_t * node = (rcl_node_t *)PyCapsule_GetPointer(pynode, "rcl_node_t");
+  rclpy_handle_t * node_handle = PyCapsule_GetPointer(pynode, "rcl_node_t");
+  if (!node_handle) {
+    return NULL;
+  }
+  rcl_node_t * node = _rclpy_handle_get_pointer(node_handle);
```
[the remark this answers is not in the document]
ivanpauno: Because, extracting the pointer from a `rclpy_handle_t` is less work than extracting it from a `PyCapsule`.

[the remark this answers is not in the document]
hidmic: That I understand, my point being that in most if not all use cases of `_rclpy_handle_get_pointer()` you've  previously pulled an `rclpy_handle_t` from a `PyCapsule` that later does not get used. So why not reusing `rclpy_handle_get_pointer_from_capsule()` where appropriate?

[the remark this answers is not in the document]
ivanpauno: I'm using `rclpy_handle_get_pointer_from_capsule` where I don't need the handle.
In the case you're asking, I need the handle later:
https://github.com/ros2/rclpy/blob/c0c56b9b4f24be37b4088154a5afea72821f5651/rclpy/src/rclpy/_rclpy.c#L1428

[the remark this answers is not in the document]
hidmic: Ohh, I missed that one. You're right, disregard.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9370/)

ivanpauno:
@hidmic About the regression test you asked, I gave it a try, but it's super hard to reproduce the problem.
The only example of the problem I found is https://github.com/ros2/rclpy/pull/470, which is not more reproducible after https://github.com/ros2/rclpy/pull/488 and https://github.com/ros2/rclpy/pull/490.

My main reason to move lifetime manangement to C is that destruction order is not documented in Python, nor how the gc works. Both of them are an implementation detail of a particular python interpreter, and users shouldn't make any supposition about how they work.

From [PEP 442](https://www.python.org/dev/peps/pep-0442/):

> Cyclic isolate (CI)
> A standalone subgraph of objects in which no object is referenced from the outside, containing one or several reference cycles, and whose objects are still in a usable, non-broken state: they can access each other from their respective finalizers.

> For CI objects, the order in which finalizers are called (step 2 above) is undefined.

hidmic:
I think listing that PEP to back up `rclpy_handler_t` rationale would be great for our future selves.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1441,19 +1471,16 @@ rclpy_remap_topic_name(PyObject * Py_UNUSED(self), PyObject * args)
   return result;
 }
 
-/// PyCapsule destructor for publisher
+/// Handles destructor for publisher
```
hidmic: @ivanpauno nit: 
```suggestion
/// Handle destructor for publisher
```

rclpy/src/rclpy_common/include/rclpy_common/handle.h
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * \param ptr Opaque pointer to the object being wrapped.
+ * \param destructor Function that will be called when the handle is destructed.
+ */
+rclpy_handle_t *
+_rclpy_create_handle(void * ptr, rclpy_handle_destructor_t destructor);
+
+/// Create a PyCapsule wrapping a rclpy_handle_t object.
+/**
+ * \param rclpy_handle_t Already constructed handle.
```
hidmic: @ivanpauno typo?
```suggestion
 * \param ptr Already constructed handle.
```

rclpy/test/test_handle.py
```diff
@@ -47,52 +47,3 @@ def test_handle_destroyed_when_not_used():
...
-            pass
-
-        node2.handle.destroy()
-        with pytest.raises(InvalidHandle):
-            with node2.handle:
-                pass
-        with node1.handle:
-            pass
-    finally:
-        rclpy.shutdown(context=context)
```
hidmic: @ivanpauno why were these tests dropped? they certainly have to be rewritten but still apply conceptually, no?

ivanpauno:
> I think listing that PEP to back up `rclpy_handler_t` rationale would be great for our future selves.

See https://github.com/ros2/rclpy/pull/497/commits/826e6ccadc5d2b14bc26ac8bff60a874e91cc094.

rclpy/src/rclpy/_rclpy.c
ivanpauno: See https://github.com/ros2/rclpy/pull/497/commits/a0f8f3d9a265e81ca446d0d761945fddb4ea88cb.

rclpy/src/rclpy_common/include/rclpy_common/handle.h
ivanpauno: See https://github.com/ros2/rclpy/pull/497/commits/a0f8f3d9a265e81ca446d0d761945fddb4ea88cb.

rclpy/test/test_handle.py
ivanpauno: Those tests were relaying in using `requires`, which is not exposed more.
I think that the tests I wrote in https://github.com/ros2/rclpy/blob/39a93ac5cde247b53137441e34f431d68ba211c4/rclpy/src/test/rclpy_common/test_handle.cpp are a good replacement.

hidmic: I see, fair enough.

rclpy/src/rclpy_common/include/rclpy_common/handle.h
```diff
@@ -0,0 +1,116 @@
...
+/// allowing to establish dependencies between them.
+typedef struct rclpy_handle_t rclpy_handle_t;
+
+/// Signature of rclpy_handle_t destructors.
+typedef void (* rclpy_handle_destructor_t)(void *);
+
+/// Create a PyCapsule wrapping a rclpy_handle_t object.
+/**
+ * Handles are used to manage lifetime of rcl objects.
+ * The main reason to use them, is that destruction order can't be controlled in Python.
```
hidmic: @ivanpauno nit:
```suggestion
 * The main reason to use them, is that destruction order cannot be guaranteed in Python.
```

rclpy/src/rclpy_common/include/rclpy_common/handle.h
```diff
@@ -0,0 +1,116 @@
...
+ * > containing one or several reference cycles, and whose objects are still in a usable,
+ * > non-broken state: they can access each other from their respective finalizers.
+ * >
+ * > For CI objects, the order in which finalizers are called (step 2 above) is undefined.
+ *
+ * Handles provide a basic way of establishing dependencies and keeping the reference count,
+ * with a mechanism completly decoupled of Python reference counting.
+ *
+ * If in the future this library is migrated to use `pybind11`, this class can be dropped and
+ * `std::shared_ptr` be used directly.
```
hidmic: @ivanpauno consider just saying it's akin to `std::shared_ptr`

rclpy/src/rclpy_common/include/rclpy_common/handle.h
```diff
@@ -0,0 +1,116 @@
...
+/// Wrapper that manages the lifetime of an object,
+/// allowing to establish dependencies between them.
+typedef struct rclpy_handle_t rclpy_handle_t;
+
+/// Signature of rclpy_handle_t destructors.
+typedef void (* rclpy_handle_destructor_t)(void *);
+
+/// Create a PyCapsule wrapping a rclpy_handle_t object.
+/**
+ * Handles are used to manage lifetime of rcl objects.
```
hidmic: @ivanpauno I'd move this rationale to the `typedef` above.

rclpy/src/rclpy_common/include/rclpy_common/handle.h
```diff
@@ -0,0 +1,116 @@
...
+ * The main reason to use them, is that destruction order can't be controlled in Python.
+ * From PEP 442:
+ * > Cyclic isolate (CI)
+ * > A standalone subgraph of objects in which no object is referenced from the outside,
+ * > containing one or several reference cycles, and whose objects are still in a usable,
+ * > non-broken state: they can access each other from their respective finalizers.
+ * >
+ * > For CI objects, the order in which finalizers are called (step 2 above) is undefined.
+ *
+ * Handles provide a basic way of establishing dependencies and keeping the reference count,
```
hidmic: @ivanpauno nit: 
```suggestion
 * Handles provide a way of establishing dependencies, keeping a reference count
```

rclpy/src/rclpy_common/include/rclpy_common/handle.h
```diff
@@ -0,0 +1,116 @@
...
+ * with a mechanism completly decoupled of Python reference counting.
```
hidmic: @ivanpauno nit: 
```suggestion
 * separately from Python's, ensuring proper destruction order. No protection against reference cycles is provided.
```

rclpy/src/rclpy_common/include/rclpy_common/handle.h
ivanpauno: Done in https://github.com/ros2/rclpy/pull/497/commits/269486e2444792a90bd0dc5c6e7a57c863272f6f.

rclpy/src/rclpy_common/include/rclpy_common/handle.h
ivanpauno: I simplified the comment in https://github.com/ros2/rclpy/pull/497/commits/269486e2444792a90bd0dc5c6e7a57c863272f6f, though still mentioned the migration to pybind11.

rclpy/src/rclpy_common/include/rclpy_common/handle.h
ivanpauno: I deleted this.
The above typedef already mentioned this rationale.

rclpy/src/rclpy_common/include/rclpy_common/handle.h
ivanpauno: Done in https://github.com/ros2/rclpy/pull/497/commits/269486e2444792a90bd0dc5c6e7a57c863272f6f.

rclpy/src/rclpy_common/include/rclpy_common/handle.h
ivanpauno: Done in https://github.com/ros2/rclpy/pull/497/commits/269486e2444792a90bd0dc5c6e7a57c863272f6f.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9492/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5216/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7764/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9431/)

ivanpauno:
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7770/)

PR #457, opened 2019-11-09
Support multiple parameters callback

suab321321:
@allenh1 now #427 is continued here as my previous repos were a lot messy

PR #504, opened 2020-02-06
Deprecate set_parameters_callback API

### LINKED ISSUE #499: Deprecate old set parameters callback API
#### Feature description

Similar to what was done in `rclcpp` (https://github.com/ros2/rclcpp/issues/791), once support for multiple parameter callbacks (https://github.com/ros2/rclpy/issues/393), we should deprecate the older API, which is no longer needed.

Specifically, the following function should be deprecated:

https://github.com/ros2/rclpy/blob/30eeae9b30ab15209798d62b501d1dec633e1db3/rclpy/rclpy/node.py#L1022-L1025

jacobperron:
Resolves #499 

Replaced by new API that supports multiple callbacks introduced in #457.
Replaced references to the old API with the new API.
Left tests for the old API that should be removed or updated when we remove
the deprecated API.

CI up to rclpy:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9254/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4939/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7561/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9174/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/51/)

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -24,6 +24,7 @@ from typing import Tuple
 from typing import TypeVar
 from typing import Union
 
+import warnings
 import weakref
 
 from rcl_interfaces.msg import FloatingPointRange
@@ -326,7 +327,7 @@ class Node:
         Declare and initialize a parameter.
 
         This method, if successful, will result in any callback registered with
-        :func:`set_parameters_callback` to be called.
+        :func:`add_on_set_parameters_callback` to be called.
 
         :param name: Fully-qualified name of the parameter, including its namespace.
         :param value: Value of the parameter to declare.
@@ -368,7 +369,7 @@ class Node:
         This allows you to declare several parameters at once without a namespace.
 
         This method, if successful, will result in any callback registered with
-        :func:`set_parameters_callback` to be called once for each parameter.
+        :func:`add_on_set_parameters_callback` to be called once for each parameter.
         If one of those calls fail, an exception will be raised and the remaining parameters will
         not be declared.
         Parameters declared up to that point will not be undeclared.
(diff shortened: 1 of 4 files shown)


rclpy/rclpy/node.py
```diff
@@ -1053,12 +1054,18 @@ def set_parameters_callback(
         callback: Callable[[List[Parameter]], SetParametersResult]
     ) -> None:
         """
-        Register a set parameters callback.
+        DEPRECATED. Register a set parameters callback.
```
hidmic: @jacobperron here and elsewhere, I think you may want to use the [`deprecated` directive](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated).

jacobperron: Using deprecated directive: 36743c7
Correct me if I'm wrong, but I think there is only this one place to use the directive (at least in the scope of this PR).

jacobperron: @hidmic Friendly ping.

jacobperron:
Rebuild with latest changes:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9557/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5259/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7803/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9478/)

PR #508, opened 2020-02-12
Call init and shutdown thread safely

ivanpauno:
The rcl functions aren't thread safe.
The upper layer has to lock.


ivanpauno:
> Out of curiosity, did this show up as errors somewhere?

No, I realized about the problem in another [discussion](https://github.com/ros2/rcl/pull/573#discussion_r377070661).
It's pretty hard to hit the problem, as `shutdown` should be called asynchronously while `init` is still running, which is an strange use-case.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9330/)

PR #512, opened 2020-02-22
Fixes max_jitter calculation

brawner:
The calculation of the max_jitter, which is a criterion for this test's success, was incorrect. This fixes the `abs()` calculation and adds min/max to the `__str__()` call.

### DIFF
--- a/rclpy/test/test_rate.py
+++ b/rclpy/test/test_rate.py
@@ -30,6 +30,8 @@ class RateRunner:
     def __init__(self, rate):
         self.avg_period = None
         self.max_jitter = None
+        self.min_period = None
+        self.max_period = None
         self.done = False
 
         self._num_measurements = 10
@@ -37,7 +39,8 @@ class RateRunner:
         self._thread.start()
 
     def __str__(self):
-        return 'avg period: {} max jitter: {}'.format(self.avg_period, self.max_jitter)
+        return 'avg period: {} max jitter: {} min: {} max: {} '.format(
+            self.avg_period, self.max_jitter, self.min_period, self.max_period)
 
     def _run(self, rate):
         try:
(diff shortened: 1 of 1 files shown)


brawner:
This is a flakey test on windows and mac. This was run with `--retest-until-fail 50` for this specific test.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9414/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5092/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7701/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9329/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/186/)

rclpy/test/test_rate.py
```diff
@@ -53,7 +56,9 @@ def _run(self, rate):
                 last_wake_time = now
 
             self.avg_period = sum(measurements) / len(measurements)
-            self.max_jitter = max([abs(m) - self.avg_period for m in measurements])  # noqa: C407
+            self.max_jitter = max([abs(m - self.avg_period) for m in measurements])  # noqa: C407
```
dirk-thomas: The change to the formula looks good to me. The test still seems to fail on macOS and Windows though.

brawner: It's a flakey test on those systems, see (https://github.com/ros2/rclpy/issues/513). I'm really curious why windows sees jitter with such very precise rate periods, but I can open a PR that doubles the max_jitter for windows/macos if we want to improve the reliability of this test.

clalancette: I can't really explain why, but its common that we have to relax the timing requirement on tests for them to pass on Windows.  Another recent example is in https://github.com/ros2/rclcpp/pull/993/files#diff-0ca2dd153932c387ce77cd330d3c70c6 .

brawner:
Is someone able to approve (and merge) this PR? @dirk-thomas @clalancette

PR #516, opened 2020-02-25
Embolden the warning in the docs about Client.call() potentially deadlocking

mm318:
The potential for the synchronous `call()` to cause a deadlock if called from inside a callback warrants a more serious warning than how it is currently being documented.


piraka9011:
@jacobperron @ivanpauno Hopefully a simple PR :)
Could any of you PTAL?

PR #518, opened 2020-03-03
Ensure logging is initialized only once

ivanpauno:
Connects to https://github.com/ros2/rcl/pull/579.
Connects to https://github.com/ros2/rclcpp/pull/998.

### DIFF
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -20,6 +20,10 @@ from typing import Optional
 import weakref
 
 
+g_logging_configure_lock = threading.Lock()
+g_logging_ref_count = 0
+
+
 class Context:
     """
     Encapsulates the lifecycle of init and shutdown.
@@ -36,12 +40,13 @@ class Context:
         self._lock = threading.Lock()
         self._callbacks = []
         self._callbacks_lock = threading.Lock()
+        self._logging_initialized = False
 
     @property
     def handle(self):
         return self._handle
 
-    def init(self, args: Optional[List[str]] = None):
+    def init(self, args: Optional[List[str]] = None, *, initialize_logging: bool = True):
         """
         Initialize ROS communications for a given context.
 
@@ -49,8 +54,15 @@ class Context:
         """
         # imported locally to avoid loading extensions on module import
         from rclpy.impl.implementation_singleton import rclpy_implementation
+        global g_logging_ref_count
         with self._handle as capsule, self._lock:
             rclpy_implementation.rclpy_init(args if args is not None else sys.argv, capsule)
+            if initialize_logging:
+                self._logging_initialized = True
+                with g_logging_configure_lock:
+                    g_logging_ref_count += 1
+                    if g_logging_ref_count == 1:
+                        rclpy_implementation.rclpy_logging_configure(self._handle)
 
     def ok(self):
         """Check if context hasn't been shut down."""
@@ -95,3 +107,12 @@ class Context:
                 callback()
             else:
                 self._callbacks.append(weakref.WeakMethod(callback, self._remove_callback))
+
+    def __del__(self):
+        from rclpy.impl.implementation_singleton import rclpy_implementation
+        global g_logging_ref_count
+        if self._logging_initialized:
+            with g_logging_configure_lock:
+                g_logging_ref_count -= 1
+                if g_logging_ref_count == 0:
+                    rclpy_implementation.rclpy_logging_fini()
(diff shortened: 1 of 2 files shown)


ivanpauno:
Rebased after https://github.com/ros2/rclpy/pull/497 went in.

ivanpauno:
CI jobs in https://github.com/ros2/rcl/pull/579#issuecomment-595304625.

rclpy/rclpy/context.py
```diff
@@ -36,21 +40,29 @@ def __init__(self):
...
 
         :param args: List of command line arguments.
         """
         # imported locally to avoid loading extensions on module import
         from rclpy.impl.implementation_singleton import rclpy_implementation
+        global g_logging_ref_count
         with self._handle as capsule, self._lock:
             rclpy_implementation.rclpy_init(args if args is not None else sys.argv, capsule)
+            if initialize_logging:
+                self._logging_initialized = True
```
hidmic, line 61: @ivanpauno should this be after the context manager, in case `rclpy_logging_configure` raises?

rclpy/rclpy/context.py
```diff
@@ -95,3 +108,12 @@ def on_shutdown(self, callback: Callable[[], None]):
...
                 self._callbacks.append(weakref.WeakMethod(callback, self._remove_callback))
+
+    def _logging_fini(self):
+        from rclpy.impl.implementation_singleton import rclpy_implementation
+        global g_logging_ref_count
+        if self._logging_initialized:
+            with g_logging_configure_lock:
+                g_logging_ref_count -= 1
+                if g_logging_ref_count == 0:
+                    rclpy_implementation.rclpy_logging_fini()
```
hidmic, line 119: @ivanpauno should `self._logging_initialized` go back to False?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -634,6 +635,56 @@ rclpy_init(PyObject * Py_UNUSED(self), PyObject * args)
...
+    return NULL;
+  }
+  rcl_context_t * context = rclpy_handle_get_pointer_from_capsule(pycontext, "rcl_context_t");
+  if (!context) {
+    return NULL;
+  }
+  rcl_allocator_t allocator = rcl_get_default_allocator();
+  rcl_ret_t ret = rcl_logging_configure(
+    &context->global_arguments,
+    &allocator);
```
hidmic, line 658: @ivanpauno nit: do we need to wrap this function call's arguments?

rclpy/rclpy/context.py
```diff
@@ -20,6 +20,10 @@
 import weakref
 
 
+g_logging_configure_lock = threading.Lock()
+g_logging_ref_count = 0
```
hidmic, line 24: @ivanpauno same comment as in https://github.com/ros2/rclcpp/pull/998 about where this reference count lives.

ivanpauno: See my reply in the `rclcpp` PR.

rclpy/rclpy/context.py
ivanpauno: It shouldn't be after the context manager because in that case in can be initialized twice.
But it definitely has to be after the `rclpy_logging_configure` function call.

ivanpauno: > It shouldn't be after the context manager because in that case it can be initialized twice.

I'm sorry, this is not correct as the local `lock` is protecting that.

ivanpauno: Fixed in https://github.com/ros2/rclpy/pull/518/commits/a5814ed5b7c5788ecd6e77aa11fa36580ed8efaa.

rclpy/rclpy/context.py
ivanpauno: See https://github.com/ros2/rclpy/pull/518/commits/a5814ed5b7c5788ecd6e77aa11fa36580ed8efaa.

rclpy/src/rclpy/_rclpy.c
ivanpauno: No, but I think both styles are accepted in our guidelines.
I fixed it in https://github.com/ros2/rclpy/pull/518/commits/a5814ed5b7c5788ecd6e77aa11fa36580ed8efaa.

rclpy/rclpy/context.py
```diff
@@ -73,6 +85,7 @@ def shutdown(self):
         with self._handle as capsule, self._lock:
             rclpy_implementation.rclpy_shutdown(capsule)
         self._call_on_shutdown_callbacks()
+        self._logging_fini()
```
wjwwood, line 88: @ivanpauno any idea why this is not called from `try_shutdown()` as well?

ivanpauno: no, I guess that was only a mistake.

ivanpauno: https://github.com/ros2/rclpy/pull/800 should fix it

PR #522, opened 2020-03-06
add env var to filter available RMW implementations

dirk-thomas:
Same as https://github.com/ros2/rmw/pull/199#pullrequestreview-370664368.

### DIFF
--- a/rclpy/rclpy/utilities.py
+++ b/rclpy/rclpy/utilities.py
@@ -96,7 +96,7 @@ def get_available_rmw_implementations():
         if missing_rmw_implementations:
             raise RuntimeError(
                 f'The RMW implementations {missing_rmw_implementations} '
-                "specified in 'RMW_IMPLEMENTATIONS' are not available ("
+                "specified in 'RMW_IMPLEMENTATIONS' are not available (" +
                 ', '.join(sorted(available_rmw_implementations)) + ')')
         available_rmw_implementations = {
             name for name in available_rmw_implementations
--- a/rclpy/rclpy/utilities.py
+++ b/rclpy/rclpy/utilities.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 import sys
 import threading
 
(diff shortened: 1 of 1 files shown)


rclpy/rclpy/utilities.py
```diff
@@ -71,9 +72,37 @@ def get_rmw_implementation_identifier():
...
+        'rmw_typesupport')
+    available_rmw_implementations = {
+        name for name in available_rmw_implementations
+        if name != 'rmw_implementation'}
+
+    # filter by implementations in environment variable if provided
+    rmw_implementations = os.environ.get('RMW_IMPLEMENTATIONS')
+    if rmw_implementations:
+        rmw_implementations = rmw_implementations.split(os.pathsep)
+        missing_rmw_implementations = set(rmw_implementations) - \
```
jacobperron: style nit: [pep8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) prefers wrapping line continuation in parentheses instead of using a backslash.

rclpy/rclpy/utilities.py
```diff
@@ -71,9 +72,37 @@ def get_rmw_implementation_identifier():
...
+            available_rmw_implementations
+        if missing_rmw_implementations:
+            raise RuntimeError(
+                f'The RMW implementations {missing_rmw_implementations} '
+                "specified in 'RMW_IMPLEMENTATIONS' are not available ("
+                ', '.join(sorted(available_rmw_implementations)) + ')')
```
jacobperron: I don't think this code works as expected. The following seems to do the trick:
```suggestion
                f"{', '.join(sorted(available_rmw_implementations))})")
```

rclpy/rclpy/utilities.py
```diff
@@ -71,9 +72,37 @@ def get_rmw_implementation_identifier():
...
+    Including an unavailable RMW implementation results in a RuntimeError.
+    """
+    available_rmw_implementations = ament_index_python.get_resources(
+        'rmw_typesupport')
+    available_rmw_implementations = {
+        name for name in available_rmw_implementations
+        if name != 'rmw_implementation'}
+
+    # filter by implementations in environment variable if provided
+    rmw_implementations = os.environ.get('RMW_IMPLEMENTATIONS')
```
jacobperron: Suggest renaming the variable to `filtered_rmw_implementations`.

The more that I think about it, the more I feel like we should rename the environment variable. `RMW_IMPLEMENTATIONS` is very similar to the existing environment variable `RMW_IMPLEMENTATION`.

What if we changed the environment variable name to `FILTER_AVAILABLE_RMM_IMPLEMENTATIONS`? I know it's a little verbose.

rclpy/rclpy/utilities.py
dirk-thomas: Since we use the backslash style throughout our code I am less inclined to diverge from that style here.

rclpy/rclpy/utilities.py
dirk-thomas: Imo both variables are similar since they do similar things. `RMW_IMPLEMENTATION` selects the one to be used, `RMW_IMPLEMENTATIONS` lists the ones which can be used. Adding `FILTER_` to one sounds arbitrary - the same argument could be made for prefixing the other with `SELECT_` / `CHOOSE_`.

rclpy/rclpy/utilities.py
dirk-thomas: I fixed it differently in 906fd4739e49c1cafe73d21e6f51d0860b63d8a1. With this I get an exception message like this:

> RuntimeError: The RMW implementations {'foo'} specified in 'RMW_IMPLEMENTATIONS' are not available (rmw_cyclonedds_cpp, rmw_fastrtps_cpp, rmw_fastrtps_dynamic_cpp)

dirk-thomas:
[Build Status](https://ci.ros2.org/job/ci_linux/9562/)

rclpy/rclpy/utilities.py
jacobperron: Thought I'd raise the concern; I'm not going to block on it.

rclpy/rclpy/utilities.py
jacobperron: I don't know if it's that prevalent; there's a dozen uses in rclpy that I'm inclined to go and change :grin:

dirk-thomas: `pep8` calls the parenthesis a preference only and the current code also passes `flake8` just fine.

rclpy/rclpy/utilities.py
dirk-thomas: If we want to we can still change it. As long as we haven't released the change a modification in a follow up PR should be straight forward.

PR #525, opened 2020-03-13
Set context when creating Timer.

DongheeYe:
rcl timer must create a guard condition in order to reset a canceled timer.
When calling rcl_create_timer, it will alls guard condition and it is need context.

Signed-off-by: Donghee Ye <donghee.ye@samsung.com>

It is related to : https://github.com/ros2/rcl/pull/589


ivanpauno:
CI here: https://github.com/ros2/rcl/pull/589#issuecomment-598679851.

PR #536, opened 2020-04-07
Create a default warning for qos incompatibility

emersonknapp:
Signed-off-by: Emerson Knapp <emerson.b.knapp@gmail.com>

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import copy
 import math
 
 from typing import Any
@@ -1151,14 +1150,9 @@ class Node:
 
         publisher_handle = Handle(publisher_capsule)
 
-        if event_callbacks:
-            event_callbacks_with_logger = copy.copy(event_callbacks)
-        else:
-            event_callbacks_with_logger = PublisherEventCallbacks()
-        event_callbacks_with_logger._logger = self.get_logger()
         publisher = Publisher(
             publisher_handle, msg_type, topic, qos_profile,
-            event_callbacks=event_callbacks_with_logger,
+            event_callbacks=event_callbacks or PublisherEventCallbacks(),
             callback_group=callback_group)
         self.__publishers.append(publisher)
         self._wake_executor()
@@ -1214,15 +1208,10 @@ class Node:
 
         subscription_handle = Handle(subscription_capsule)
 
-        if event_callbacks:
-            event_callbacks_with_logger = copy.copy(event_callbacks)
-        else:
-            event_callbacks_with_logger = PublisherEventCallbacks()
-        event_callbacks_with_logger._logger = self.get_logger()
         subscription = Subscription(
             subscription_handle, msg_type,
             topic, callback, callback_group, qos_profile, raw,
-            event_callbacks=event_callbacks_with_logger)
+            event_callbacks=event_callbacks or SubscriptionEventCallbacks())
         self.__subscriptions.append(subscription)
         callback_group.add_entity(subscription)
 
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import copy
 import math
 
 from typing import Any
@@ -1150,9 +1151,14 @@ class Node:
 
         publisher_handle = Handle(publisher_capsule)
 
+        if event_callbacks:
+            event_callbacks_with_logger = copy.copy(event_callbacks)
+        else:
+            event_callbacks_with_logger = PublisherEventCallbacks()
+        event_callbacks_with_logger._logger = self.get_logger()
         publisher = Publisher(
             publisher_handle, msg_type, topic, qos_profile,
-            event_callbacks=event_callbacks or PublisherEventCallbacks(),
+            event_callbacks=event_callbacks_with_logger,
             callback_group=callback_group)
         self.__publishers.append(publisher)
         self._wake_executor()
@@ -1208,10 +1214,15 @@ class Node:
 
         subscription_handle = Handle(subscription_capsule)
 
+        if event_callbacks:
+            event_callbacks_with_logger = copy.copy(event_callbacks)
+        else:
+            event_callbacks_with_logger = PublisherEventCallbacks()
+        event_callbacks_with_logger._logger = self.get_logger()
         subscription = Subscription(
             subscription_handle, msg_type,
             topic, callback, callback_group, qos_profile, raw,
-            event_callbacks=event_callbacks or SubscriptionEventCallbacks())
+            event_callbacks=event_callbacks_with_logger)
         self.__subscriptions.append(subscription)
         callback_group.add_entity(subscription)
 
--- a/rclpy/rclpy/qos_event.py
+++ b/rclpy/rclpy/qos_event.py
@@ -22,6 +22,7 @@ import rclpy
 from rclpy.callback_groups import CallbackGroup
 from rclpy.handle import Handle
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+from rclpy.logging import get_logger
 from rclpy.qos import qos_policy_name_from_kind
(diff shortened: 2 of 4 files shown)


rclpy/rclpy/qos_event.py
```diff
@@ -215,18 +225,36 @@ def create_event_handlers(
...
+                    'events, you will not be notified when Subscriptions request an incompatible '
+                    'QoS profile from Publishers on the same topic.',
+                    once=True)
         return event_handlers
 
 
 class PublisherEventCallbacks:
     """Container to provide middleware event callbacks for a Publisher."""
 
+    _logger = rclpy.logging.get_logger('SubscriptionEventCallbacks')
```
mm318, line 256: Is this the convention, where the logger is specific to the class instead of the class instance?

rclpy/rclpy/qos_event.py
```diff
@@ -262,10 +297,25 @@ def create_event_handlers(
...
                 callback=self.incompatible_qos,
                 event_type=QoSPublisherEventType.RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS,
                 parent_handle=publisher_handle))
+        else:
+            # Register default callback when not specified
+            try:
+                event_handlers.append(QoSEventHandler(
+                    callback_group=callback_group,
+                    callback=self._default_incompatible_qos_callback,
+                    event_type=QoSSubscriptionEventType.RCL_SUBSCRIPTION_OFFERED_INCOMPATIBLE_QOS,
```
mm318, line 313: `RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS`

rclpy/rclpy/qos_event.py
emersonknapp: It was the best way I could think of to warn once per process run, rather than for every publisher/subscription. I believe `once=True` applies to the logger instance, so I created one for the class rather than the instance.

emersonknapp: Oh but the name is definitely not correct. will fix

mm318: Should the logger used in the default callback function be the same as the one used during the registering of the default callback function be the same? We should get more guidance from OSRF.

mm318:
This pull request has been updated to behave similarly to the `rclcpp` version (in https://github.com/ros2/rclcpp/pull/1051).

rclpy/rclpy/node.py
```diff
@@ -1150,9 +1151,14 @@ def create_publisher(
 
         publisher_handle = Handle(publisher_capsule)
 
+        if event_callbacks:
+            event_callbacks_with_logger = copy.copy(event_callbacks)
```
ivanpauno, line 1155: why is this needed?

rclpy/rclpy/node.py
```diff
@@ -1150,9 +1151,14 @@ def create_publisher(
...
+        else:
+            event_callbacks_with_logger = PublisherEventCallbacks()
+        event_callbacks_with_logger._logger = self.get_logger()
```
ivanpauno, line 1158: This is extremely hacky. If somebody creates a custom `PublisherEventCallbacks` with the default subscription and they doesn't add `_logger` member, then the callback will fail.

Consider either moving the callbacks definitions from `PublisherEventCallbacks`/`SubscriptionEventCallbacks` to `Publisher`/`Subscription`, and adding the default subscriptions there when `use_default_callbacks` option is true, or any other solution that avoids this workaround.

rclpy/rclpy/node.py
mm318: I don't want to modify the `PublisherEventCallbacks`/`SubscriptionEventCallbacks` objects that the user pass in, in case they want to reuse them.

mm318: This change is no longer relevant.

rclpy/rclpy/node.py
mm318: I have taken a different approach. Please re-review the pull request.

mm318:
@ros2/aws-oncall - please run this CI job
Gist: `https://gist.githubusercontent.com/mm318/9ab95351a88ec18085660718f582912f/raw/779f53aa1e5ffcebc4c093d71f49bd5c7d2c97c4/ros2_incompatible_qos_callbacks.repos`
BUILD args: `--packages-up-to rclpy`
TEST args: `--packages-select rclpy`
Job: `ci_launcher`

rclpy/src/rclpy/_rclpy.c
```diff
@@ -866,6 +866,65 @@ rclpy_get_node_logger_name(PyObject * Py_UNUSED(self), PyObject * args)
...
+
+  rclpy_publisher_t * pub = rclpy_handle_get_pointer_from_capsule(pypublisher, "rclpy_publisher_t");
+  if (NULL == pub) {
+    return NULL;
+  }
+
+  const char * node_logger_name = rcl_node_get_logger_name(pub->node);
+  if (NULL == node_logger_name) {
+    Py_RETURN_NONE;
+  }
```
ivanpauno, line 893: This is fine, though I think that a bit unnecessary.

IMO, a nicer approach would have been the following:

```python3
callbacks = PublisherEventCallbacks()  # default callbacks is true, but they're not yet defined
# Default callbacks are added here, if `event_callbacks._use_default_callbacks` is true, and the user doesn't explicitly define that callback.
publisher = self.node.create_publisher(EmptyMsg, 'test_topic', 10, event_callbacks=callbacks)
```

In both cases the result the user sees is the same, so I won't block on the change.

rclpy/rclpy/node.py
ivanpauno: Thanks!

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10210/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5758/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8299/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10099/)

rclpy/src/rclpy/_rclpy.c
mm318: I thought about doing it this way, but with `event_callbacks.incompatible_qos` no longer `None`, it's harder to determine later on if it's the default one or a user provided one, which is needed to handle the `UnsupportedEventTypeError` exception differently.

mm318:
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/10099/)

Looks to me like the build test failures on Windows are unrelated, @ivanpauno.

rclpy/src/rclpy/_rclpy.c
ivanpauno: > I thought about doing the way you mentioned, but with event_callbacks.incompatible_qos no longer None, it's harder to determine later on if it's the default one or a user provided one, which is needed to handle the UnsupportedEventTypeError exception differently.

It could initially be `None`, except the user provided explicitly one.
If it's set to `None` and use `default_callbacks` is true, the event is registered with the default callback when the publisher is created.

I think the approach I'm mentioning is exactly the one used in `rclcpp`.

jacobperron:
FYI, I think this change is going to cause a string of new test failures in `ros2cli`, as the warning appears when running various CLI tools (e.g. ros2 topic echo).

dirk-thomas:
> FYI, I think this change is going to cause a string of new test failures in `ros2cli`, as the warning appears when running various CLI tools (e.g. ros2 topic echo).

@emersonknapp @mm318 @ros2/aws-oncall FYI

emersonknapp:
Thx for the heads up. Saw one such test failure in rosbag2, though that was resolvable by making the test less fragile. CLI is a bit different

dirk-thomas:
See https://ci.ros2.org/job/ci_linux/10244/testReport/ for a list of test failures in the `ros2cli` repo.

nuclearsandwich:
@emersonknapp since this PR will introduce test failures. I'm wondering if we should revert it before the nightlies and explore solutions in the morning?

emersonknapp:
I'm personally fine with that, I believe in a revert-friendly engineering culture. I'm unfortunately not at my computer for the remainder of the evening, if you or @mm318 would be able to do it?

PR #539, opened 2020-04-10
use f-string to fix flake8 warning

nuclearsandwich:
Resolves a flake8 error which would manifest as a KeyError at runtime
when this exception is created.

Spotted in [nightlies](https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1519/testReport/rclpy/flake8/F999____rclpy_exceptions_py_74_34_/) it took a minute to reproduce since the flake8 version we install claims to be incompatible with [pyflakes 2.2.0](https://pypi.org/project/pyflakes/#history) which was just released, which is likely why the error has the F999 code.

### DIFF
--- a/rclpy/rclpy/exceptions.py
+++ b/rclpy/rclpy/exceptions.py
@@ -71,8 +71,7 @@ class ParameterException(Exception):
     """Base exception for parameter-related errors."""
 
     def __init__(self, error_msg, parameters, *args):
-        Exception.__init__(self, '{error_msg}: {parameters}'.format(
-            error_msg=error_msg, parameters=parameters), *args)
+        Exception.__init__(self, f'{error_msg}: {parameters}')
 
 
 class ParameterNotDeclaredException(ParameterException):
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10067/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5657/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8187/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9970/)

nuclearsandwich:
The CMake warning is from https://github.com/ament/ament_cmake/pull/238 and is not related to this change.

rclpy/rclpy/exceptions.py
```diff
@@ -71,7 +71,8 @@ class ParameterException(Exception):
     """Base exception for parameter-related errors."""
 
     def __init__(self, error_msg, parameters, *args):
-        Exception.__init__(self, '{error_msg}: {parameters}'.format(error_msg, parameters), *args)
+        Exception.__init__(self, '{error_msg}: {parameters}'.format(
```
dirk-thomas, line 74: Please use an f-string instead.

nuclearsandwich:
Sanity rebuild on Linux alone [Build Status](https://ci.ros2.org/job/ci_linux/10083/)

dirk-thomas:
I updated the title and will go ahead and merge this since CI showed the flake8 warning is resolved.

PR #541, opened 2020-04-14
update includes to use non-entry point headers from detail subdir

dirk-thomas:
Related to ros2/rosidl#461.

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -25,7 +25,7 @@
 #include <rcl/remap.h>
 #include <rcl/time.h>
 #include <rcl/validate_topic_name.h>
-#include <rcl_interfaces/msg/parameter_type__struct.h>
+#include <rcl_interfaces/msg/detail/parameter_type__struct.h>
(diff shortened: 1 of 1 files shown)


rclpy/src/rclpy/_rclpy.c
```diff
@@ -25,7 +25,7 @@
 #include <rcl/remap.h>
 #include <rcl/time.h>
 #include <rcl/validate_topic_name.h>
-#include <rcl_interfaces/msg/parameter_type__struct.h>
+#include <rcl_interfaces/msg/detail/parameter_type__struct.h>
```
dirk-thomas, line 28: As an alternative this could also just include `rcl_interfaces/msg/parameter_type.h`?

jacobperron: I agree. I think the "detail" namespace should discourage users (including `rclpy`) from using those headers directly.

dirk-thomas: Changed in cea1736f4ede0610b16298d85e0b8fa6125efeaa.

jacobperron:
We should update the PR title.

PR #543, opened 2020-04-16
Revert "Create a default warning for qos incompatibility"

nuclearsandwich:
Reverts ros2/rclpy#536 until test failures in ros2cli can be addressed.

FYI @ivanpauno as the reviewer of #536.


ivanpauno:
Thanks for reverting @nuclearsandwich !

PR #544, opened 2020-04-16
Revert "Revert "Create a default warning for qos incompatibility""

mm318:
Reverts https://github.com/ros2/rclpy/pull/543

Let's work on solving the aforementioned test failures before merging this pull request.

### DIFF
--- a/rclpy/rclpy/qos_event.py
+++ b/rclpy/rclpy/qos_event.py
@@ -252,11 +252,7 @@ class SubscriptionEventCallbacks:
                     parent_handle=subscription_handle))
 
             except UnsupportedEventTypeError:
-                logger.warn(
-                    'This rmw implementation does not support ON_REQUESTED_INCOMPATIBLE_QOS '
-                    'events, you will not be notified when Subscriptions request an incompatible '
-                    'QoS profile from Publishers on the same topic.',
-                    once=True)
+                pass
 
         return event_handlers
 
@@ -333,10 +329,6 @@ class PublisherEventCallbacks:
                     parent_handle=publisher_handle))
 
             except UnsupportedEventTypeError:
-                logger.warn(
-                    'This rmw implementation does not support ON_OFFERED_INCOMPATIBLE_QOS '
-                    'events, you will not be notified when Publishers offer an incompatible '
-                    'QoS profile to Subscriptions on the same topic.',
-                    once=True)
+                pass
 
         return event_handlers
--- a/rclpy/rclpy/qos_event.py
+++ b/rclpy/rclpy/qos_event.py
@@ -22,6 +22,8 @@ import rclpy
 from rclpy.callback_groups import CallbackGroup
 from rclpy.handle import Handle
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+from rclpy.logging import get_logger
+from rclpy.qos import qos_policy_name_from_kind
 from rclpy.waitable import NumberOfEntities
 from rclpy.waitable import Waitable
 
@@ -186,6 +188,7 @@ class SubscriptionEventCallbacks:
         deadline: Optional[Callable[[QoSRequestedDeadlineMissedInfo], None]] = None,
         liveliness: Optional[Callable[[QoSLivelinessChangedInfo], None]] = None,
         incompatible_qos: Optional[Callable[[QoSRequestedIncompatibleQoSInfo], None]] = None,
+        use_default_callbacks: bool = True,
     ) -> None:
         """
         Create a SubscriptionEventCallbacks container.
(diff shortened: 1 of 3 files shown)


ivanpauno:
Continuing discussion from https://github.com/ros2/rclpy/pull/536:

> FYI, I think this change is going to cause a string of new test failures in ros2cli, as the warning appears when running various CLI tools (e.g. ros2 topic echo).

The warning that appeared is the one saying that `Fast-RTPS` doesn't support qos incompatibility callbacks, right?

I think the problem can be avoided in the following way:

- A warning is showed if the event type is not supported and the user explicitly required a callback.
- If the user doesn't specify a callback, the default one is registered. If the event isn't supported no warning is showed, and there's not default callback for that rmw vendor.

The approach commented here https://github.com/ros2/rclpy/pull/536#discussion_r408822436 will make this easier to implement IMO.

@mm318 let me know if this sounds reasonable to you.
In case that yes, I would delete the warning in `rclcpp` too: https://github.com/aws-ros-dev/rclcpp/blob/f6d0fc388fbf9baf3262aa158b32569228437de8/rclcpp/include/rclcpp/publisher.hpp#L100.

ivanpauno:
@mm318 The revert commit has to be singed too, to make the DCO check pass.

ivanpauno:
> A warning is showed if the event type is not supported and the user explicitly required a callback.

Sorry for the mistake here. Currently we're ricing an exception in this case, and that approach is correct. Ignore this item.

mm318:
Let's just remove the warning log message in the case of failing to register a default callback.

> The approach commented here [#536 (comment)](https://github.com/ros2/rclpy/pull/536#discussion_r408822436) will make this easier to implement IMO.

The structure shouldn't really change, because the logic is:
```
if (user provided callback) {  // the "use default callback" flag can still be true here
  register user provided callback
  if (failed to register) {
    throw exception
  }
} else if (use default callback) {  // the "user provided callback" was None
  register default callback
  if (failed to register) {
    log warning message  // will be removing this
    don't throw exception
  }
}
```
If we set the `incompatible_qos` field to a default callback beforehand (as suggested in https://github.com/ros2/rclpy/pull/536#discussion_r408822436), it would be more difficult to determine if it was a user provided callback or a default callback.

rclpy/rclpy/qos_event.py
```diff
@@ -194,33 +197,63 @@ def __init__(
...
+
+                event_type = QoSSubscriptionEventType.RCL_SUBSCRIPTION_REQUESTED_INCOMPATIBLE_QOS
+                event_handlers.append(QoSEventHandler(
+                    callback_group=callback_group,
+                    callback=_default_incompatible_qos_callback,
+                    event_type=event_type,
+                    parent_handle=subscription_handle))
+
+            except UnsupportedEventTypeError:
+                pass
```
mm318, line 255: I have removed the warning log message here.

rclpy/rclpy/qos_event.py
```diff
@@ -241,31 +275,60 @@ def __init__(
...
+                        'Last incompatible policy: {}'.format(policy_name))
+
+                event_handlers.append(QoSEventHandler(
+                    callback_group=callback_group,
+                    callback=_default_incompatible_qos_callback,
+                    event_type=QoSPublisherEventType.RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS,
+                    parent_handle=publisher_handle))
+
+            except UnsupportedEventTypeError:
+                pass
```
mm318, line 332: I have removed the warning log message here.

ivanpauno:
> The structure shouldn't really change, because the logic is:

Yes, sorry. I was convinced that the error was shown when creating the publisher.

mm318:
@ivanpauno, @nuclearsandwich, and @dirk-thomas, I am not sure which packages should go through CI for this pull request, but here is the relevant gist: `https://gist.githubusercontent.com/mm318/8bb18da56a1acff5cd6944f1e5ca2166/raw/c5d4d67220843da47b4591821a6fe4342baa9b00/ros2_incompatible_qos_callbacks.repos`

ivanpauno:
Full Linux CI here:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10252/)

If the others agree, I will just run CI covering `rclpy` and `rclcpp` in the other platforms (which are the repos being modified).

ivanpauno:
@mm318 There are a few more test failures than in the nightlies https://ci.ros2.org/view/nightly/job/nightly_linux_release/1514/.

Can you take a look locally to them?
Unfortunately, the test output that CI shows in `launch tests` isn't really useful.

mm318:
Sorry, there was something wrong with the `.repos` gist. Can you please re-run CI with this one: `https://gist.githubusercontent.com/mm318/8bb18da56a1acff5cd6944f1e5ca2166/raw/5d2aa0956ef303b1583d39dbd0367668c8f39796/ros2_incompatible_qos_callbacks.repos`? Thanks!

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10272/)

mm318:
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/10272/)

Hi @ivanpauno, I just rebased. Can you please re-run CI?

ivanpauno:
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5820/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8365/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10180/)

ivanpauno:
Ok, CI looks good, going finally in!

mm318:
Great! Thanks!

PR #548, opened 2020-04-26
Fix bug that not to get expected data because use less timeout

Barry-Xu-2018:
Signed-off-by: Barry Xu <Barry.Xu@sony.com>

### DIFF
--- a/rclpy/test/test_logging_rosout.py
+++ b/rclpy/test/test_logging_rosout.py
@@ -19,6 +19,7 @@ import rclpy
 from rclpy.executors import SingleThreadedExecutor
 
 TEST_PARAMETERS = [
+    # name, enable_global_rosout_logs, enable_node_rosout, expected_data
     ('enable_global_rosout_enable_node_rosout', True, True, True),
     ('enable_global_rosout_disable_node_rosout', True, False, False),
     ('disable_global_rosout_enable_node_rosout', False, True, False),
@@ -71,8 +72,11 @@ def test_enable_rosout(
         raw=True
     )
 
-    node.get_logger().info('SOMETHING')
-    executor.spin_once(timeout_sec=1)
+    cycle_count = 0
+    while cycle_count < 5 and raw_subscription_msg is None:
+        node.get_logger().info('SOMETHING')
+        cycle_count += 1
+        executor.spin_once(timeout_sec=1)
 
     if expected_data:
         assert (raw_subscription_msg is not None)
--- a/rclpy/test/test_logging_rosout.py
+++ b/rclpy/test/test_logging_rosout.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import pytest
+import time
 
 from rcl_interfaces.msg import Log
 import rclpy
@@ -27,13 +28,12 @@ TEST_PARAMETERS = [
     ('disable_global_rosout_disable_node_rosout', False, False, False),
 ]
 
-receive_expected_rosout_msg = False  # True if expected rosout msg is received
+rosout_subscription_msg = None  # None=No result yet
 
 
 def rosout_subscription_callback(msg):
-    global receive_expected_rosout_msg
-    if (LoggingSeverity(msg.level) == LoggingSeverity.INFO and msg.msg == 'SOMETHING'):
-        receive_expected_rosout_msg = True
+    global rosout_subscription_msg
+    rosout_subscription_msg = msg
 
 
 @pytest.mark.parametrize(
@@ -63,8 +63,8 @@ def test_enable_rosout(
     )
     executor.add_node(node)
 
-    global receive_expected_rosout_msg
-    receive_expected_rosout_msg = False
+    global rosout_subscription_msg
+    rosout_subscription_msg = None
     # create subscriber of 'rosout' topic
     node.create_subscription(
         Log,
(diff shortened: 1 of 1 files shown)


fujitatomoya:
i made up pseudo code https://github.com/fujitatomoya/rclpy/commit/9b33ab26834b892066a49a2a695acce877000b54,

i believe that this is more sustainable, what do you think?

Barry-Xu-2018:
> @Barry-Xu-2018 @wjwwood @ivanpauno
> 
> not only adding the iteration loop, but also checking the contents of rosout if expected message which contains INFO and 'SOMETHING'?

Absolutely. Update codes.

Barry-Xu-2018:
> i made up pseudo code [fujitatomoya@9b33ab2](https://github.com/fujitatomoya/rclpy/commit/9b33ab26834b892066a49a2a695acce877000b54),
> 
> i believe that this is more sustainable, what do you think?

Thanks for your example.  I update codes like this.

rclpy/test/test_logging_rosout.py
```diff
@@ -17,20 +17,23 @@
...
-raw_subscription_msg = None  # None=No result yet
+receive_expected_rosout_msg = False  # True if expected rosout msg is received
 
 
-def raw_subscription_callback(msg):
-    global raw_subscription_msg
-    raw_subscription_msg = msg
+def rosout_subscription_callback(msg):
+    global receive_expected_rosout_msg
+    if (LoggingSeverity(msg.level) == LoggingSeverity.INFO and msg.msg == 'SOMETHING'):
```
fujitatomoya, line 35: more precisely checking node name also? I do not think that is necessary.

Barry-Xu-2018: > more precisely checking node name also? I do not think that is necessary.

I think so too.

rclpy/test/test_logging_rosout.py
```diff
@@ -17,20 +17,23 @@
...
 TEST_PARAMETERS = [
+    # name, enable_global_rosout_logs, enable_node_rosout, expected_data
     ('enable_global_rosout_enable_node_rosout', True, True, True),
     ('enable_global_rosout_disable_node_rosout', True, False, False),
     ('disable_global_rosout_enable_node_rosout', False, True, False),
     ('disable_global_rosout_disable_node_rosout', False, False, False),
 ]
 
-raw_subscription_msg = None  # None=No result yet
+receive_expected_rosout_msg = False  # True if expected rosout msg is received
```
clalancette, line 30: Why did you end up changing from storing the message to storing a boolean?  It doesn't matter a *huge* amount here, but storing the message meant we could do more detailed checks on it (like it contained bytes, and the length wasn't 0).  Unless there is a good reason, I'd suggest to go back to storing the message.

fujitatomoya: i guess this is because of my suggested code, but i do agree with you.

rclpy/test/test_logging_rosout.py
```diff
@@ -60,26 +63,26 @@ def test_enable_rosout(
...
+        1
     )
 
-    node.get_logger().info('SOMETHING')
-    executor.spin_once(timeout_sec=1)
+    cycle_count = 0
+    while cycle_count < 5 and receive_expected_rosout_msg is False:
+        node.get_logger().info('SOMETHING')
+        cycle_count += 1
+        executor.spin_once(timeout_sec=1)
```
wjwwood, line 80: Looping like this looks a lot more robust to me, thanks.

I am curious though about the structure and the decision to try logging multiple times. I think it's ok for the test to assert that a log message is received (it is supposed to be delivered reliably), though. Why not just keep spinning, rather than resending the logging each time.

One reason I could see for this is that this doesn't wait for the subscription to be "matched" to the rosout publisher, so it might "miss" the initial logging message.

There are two ways you could try to improve this I think. You could either:

- loop until you receive a log on the subscription, resending each loop as you do now
  - this ensures communication has been established
- then send one log message, and loop until the subscription receives that exact message or a time elapses
  - assert that the exact message was received

Or you could do something similar except you use the matched status as a precondition for checking for the message:

- wait for the subscription (and/or the publisher) to be matched
- send on log message, and loop until it is received

----

Separately, I'd recommend looping based purely on a time limit, rather than a cycle count, because you cannot ensure that there are not < 5 events to be handled when you enter this loop. The implementation may queue up as many events as is needed prior to this...

rclpy/test/test_logging_rosout.py
Barry-Xu-2018: Thanks for comments.

OK. I want to store the message but not raw message. Check the message outside the callback.

rclpy/test/test_logging_rosout.py
Barry-Xu-2018: ** not ensure that there are not < 5 events to be handled **, this is a little strange. Maybe in the future, these will be some build-in events which need to be handled if user add a node into executor.  

Based on understand your comments, implementing the codes likes below

```python
    max_difference_time = 5
    begin_time = time.time()
    while rosout_subscription_msg is None:
        node.get_logger().info('SOMETHING')
        executor.spin_once(timeout_sec=1)
        if int(time.time() - begin_time) > max_difference_time:
            break
```

wjwwood: Yeah, that looks right, but I'd combine the while loop and break condition, to something like:

```python
while rosout_subscription_msg is None and time_not_expired:
    # ...
```

Just because it is easier to read.

Barry-Xu-2018: Agree. Update codes.

hidmic:
CI up to `rclpy`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10653/) (canary build)

hidmic:
@Barry-Xu-2018 `flake8` isn't happy, please check locally.

Barry-Xu-2018:
@hidmic 

>  `flake8` isn't happy, please check locally.

Thanks for notification.  
I have checked in my local environment and didn't find these flake8 errors.  
After checking, I find CI use Ubuntu 20.04 environment (But I use Ubuntu 18.04).   
Anyway, I have fixed flake8 errors.

PR #551, opened 2020-04-28
Fix warning about pytaken_msg maybe being uninitialized

sloretz:
Seen in https://ci.ros2.org/job/nightly_linux_release/1525/gcc/new/

This initializes `pytaken_msg` with `NULL`, and changes the check that `pytaken_msg` is not `NULL` to cover both raw and non-raw cases.

PR #552, opened 2020-04-28
Fix flaky test expecting wrong return type of rclpy_take

sloretz:
#542 made `rclpy_take()` return a 2-tuple of `(message, message_info)`, but wrote a test that expected `rclpy_take()` to return `None` when there is no message to take. However, it seems it actually crashed in that case. #550 fixed the crash by making `rclpy_take()` return `(None, None)` when there was no message to take from the middleware, but the test still expected `None`. This causes the test added in #542 to be flaky: https://github.com/ros2/rclpy/pull/551#issuecomment-620327015 . This PR fixes the issue by making the test expect `(None, None)` and documenting this return code in `rclpy_take()`.

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -322,8 +322,10 @@ class Executor:
 
     def _take_subscription(self, sub):
         with sub.handle as capsule:
-            msg, _ = _rclpy.rclpy_take(capsule, sub.msg_type, sub.raw)
-        return msg
+            msg_info = _rclpy.rclpy_take(capsule, sub.msg_type, sub.raw)
+            if msg_info is not None:
+                return msg_info[0]
+        return None
 
     async def _execute_subscription(self, sub, msg):
         if msg:
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -3301,7 +3301,8 @@ rclpy_message_info_to_dict(rmw_message_info_t * message_info)
 /**
  * \param[in] pysubscription Capsule pointing to the subscription to process the message
  * \param[in] pymsg_type Instance of the message type to take
- * \return Tuple of (Python message with all fields populated with received message, message_info)
+ * \return Tuple of (Python message with all fields populated with received message, message_info),
+ *         or (None, None) if ther was no message to take.
  */
 static PyObject *
 rclpy_take(PyObject * Py_UNUSED(self), PyObject * args)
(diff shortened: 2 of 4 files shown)


sloretz:
CI (testing just rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10544/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6000/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8590/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10417/)

wjwwood:
Yeah, I wonder if returning `(None, None)` was the right move. Like is that more valuable as a return type than just `None`? AFAIK, it would never be `(something, None)` or `(None, something)`, so it seems useless to return a tuple of none to me.

I think @iluetkeb should comment on this to see what was originally intended.

iluetkeb:
The function already returned (None, None) when I modified it. The only thing I did was change the type of the first return value.

I agree with @wjwwood that there is not much value in returning a tuple of None and it is rather unusual. However, there are a few places in the executor code where a tuple is expected, which would need to be modified if we change this to return just a single None.

hidmic:
> However, there are a few places in the executor code where a tuple is expected, which would need to be modified if we change this to return just a single None

And that's why https://github.com/ros2/rclpy/pull/550 keeps its return value a tuple, to keep unpacking simple. I don't mind changing it to `None`.

sloretz:
Well, since no one approved the PR f808d21ff36d8c986443f2bc09a0deaccdf1362e restores `rclpy_take()` returning `None` if there's no message to take.

sloretz:
CI (testing just rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10551/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6006/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8593/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10433/)

iluetkeb:
@hidmic I don't have approval rights...

How is this: If you open this PR against our rclpy fork on boschresearch/rclpy, I will accept it and complete the necessary changes to the executor for a PR against ros2/rclpy.

hidmic:
Test failures seem unrelated. I think we're good to merge.

@iluetkeb not sure I follow you here, why not building on top of this branch? You can always rebase afterwards.

iluetkeb:
> not sure I follow you here, why not building on top of this branch?

To minimize the number of PRs, but anyway, since this I merged now, it's easy anyway ;-)

PR #555, opened 2020-05-05
Fix bad rmw_time_t to nanoseconds conversion.

hidmic:
Fixes https://github.com/ros2/ros2cli/issues/503 (i.e. the failure, not the hang). `rmw_time_t` instances were being forced into an `rcutils_time_duration_t` value (actually, first into a single `uint64_t` total nanosecond count, then into an unsigned `PY_LONG_LONG`, and then into an `rcutils_time_duration_t` i.e. an `int64_t`), causing overflows.

---

This happened in the first place because, RMW graph APIs are not currently bound to return anything reasonable for QoS related durations that do not make sense in a given context. That is the case for data reader (subscribers' backbone) lifespans in `rmw_connext_cpp` (see [here](https://github.com/ros2/rmw_connext/blob/7781a5e5982032be9460840e33646028af21129e/rmw_connext_shared_cpp/src/qos.cpp#L259) and [here](https://github.com/ros2/rmw_connext/blob/ae7117ec9942953d159b6a25feb2bb0a0e17c857/rmw_connext_shared_cpp/src/types/custom_subscriber_listener.cpp#L55-L56)), which come with whatever value was in the stack at the moment that `rmw_qos_profile_t` variable enters scope.

I see there's some discussion in https://github.com/ros2/rmw/issues/210 about introducing some consistency in `rmw_time_t` values. I'll defer to them on what to do about RMWs in general.

### DIFF
--- a/rclpy/src/rclpy_common/src/common.c
+++ b/rclpy/src/rclpy_common/src/common.c
@@ -140,11 +140,29 @@ rclpy_convert_to_py_names_and_types(rcl_names_and_types_t * names_and_types)
   return pynames_and_types;
 }
 
+// TODO(hidmic): revisit after discussions in
+// https://github.com/ros2/rmw/issues/210 and
+// https://github.com/ros2/rmw/issues/215 have
+// settled.
+static
+int64_t
+_convert_rmw_time_to_ns(const rmw_time_t * duration)
+{
+  int64_t partial = RCUTILS_S_TO_NS(duration->sec);
+  if (partial < 0LL || (uint64_t)partial < duration->sec) {
+    return INT64_MAX;
+  }
+  int64_t total = partial + duration->nsec;
+  if (total < 0LL || total < partial || (uint64_t)total < duration->nsec) {
+    return INT64_MAX;
+  }
+  return total;
+}
+
 static
 PyObject *
 _convert_rmw_time_to_py_duration(const rmw_time_t * duration)
 {
-  uint64_t total_nanoseconds = RCUTILS_S_TO_NS(duration->sec) + duration->nsec;
   PyObject * pyduration_module = NULL;
   PyObject * pyduration_class = NULL;
   PyObject * args = NULL;
@@ -163,7 +181,8 @@ _convert_rmw_time_to_py_duration(const rmw_time_t * duration)
   if (!args) {
     goto cleanup;
   }
-  kwargs = Py_BuildValue("{sK}", "nanoseconds", total_nanoseconds);
+  int64_t total_nanoseconds = _convert_rmw_time_to_ns(duration);
+  kwargs = Py_BuildValue("{sL}", "nanoseconds", total_nanoseconds);
   if (!kwargs) {
     goto cleanup;
   }
--- a/rclpy/src/rclpy_common/src/common.c
+++ b/rclpy/src/rclpy_common/src/common.c
@@ -140,34 +140,6 @@ rclpy_convert_to_py_names_and_types(rcl_names_and_types_t * names_and_types)
   return pynames_and_types;
 }
 
-// TODO(hidmic): revisit after discussions in
-// https://github.com/ros2/rmw/issues/210 and
-// https://github.com/ros2/rmw/issues/215 have
-// settled.
-static
-bool
-_convert_rmw_time_to_ns(const rmw_time_t * duration, int64_t * nanoseconds)
-{
-  assert(duration != NULL);
-  assert(nanoseconds != NULL);
-  int64_t partial = RCUTILS_S_TO_NS(duration->sec);
-  if (partial < 0LL || (uint64_t)partial < duration->sec) {
-    goto fail_convert_rmw_time_to_ns;
-  }
-  int64_t total = partial + duration->nsec;
-  if (total < 0LL || total < partial || (uint64_t)total < duration->nsec) {
-    goto fail_convert_rmw_time_to_ns;
-  }
-  *nanoseconds = total;
-  return true;
-fail_convert_rmw_time_to_ns:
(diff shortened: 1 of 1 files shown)


hidmic:
FYI @emersonknapp.

dirk-thomas:
I don't think I have the necessary context for a meaningful review. It seems wrong to me to just map "invalid" input to some value without any warning / error.

> This happened in the first place because, RMW graph APIs are not currently bound to return anything reasonable for QoS related durations that do not make sense in a given context.

Why can't the rmw API not be forced to set reasonable values in these cases?

hidmic:
> It seems wrong to me to just map "invalid" input to some value without any warning / error.

We can certainly add a warning if overflow would ensue, though that'll add noise for RMWs that can potentially return garbage.

> Why can't the rmw API not be forced to set reasonable values in these cases?

I believe there's no consensus as to what would be a reasonable value in some cases. There're some who even think that `rmw_time_t` doesn't need to exist (see https://github.com/ros2/rmw/issues/215). Zeroing values or defining an INFINITY would make sense to me, but I have not been actively involved in the design to make that call.

In any case, this patch is still necessary.

jacobperron:
> In any case, this patch is still necessary.

Makes sense to guard against overflow to me, since we're casting from uint64_t to int64_t.

dirk-thomas:
> RMWs that can potentially return garbage.

Imo that is simply unacceptable API. Do RMW implementations actually return uninitialized data? If yes, this might open us up to undefined which we should really avoid.

emersonknapp:
I don't think that they return unitialized values, but we don't currently have a defined constant. fastrtps and connext return one value for "unspecified", and cyclone has some whole other constant that it seems to use for that value. https://github.com/ros2/rmw/issues/210 is an attempt to discuss some ROS2-global constants that we would dictate in the API and expect rmw implementations to use.

hidmic:
> I don't think that they return unitialized values

That's not exactly true. Graph cache in `rmw_connext_cpp` gets populated with QoS data [here](https://github.com/ros2/rmw_connext/blob/ae7117ec9942953d159b6a25feb2bb0a0e17c857/rmw_connext_shared_cpp/src/types/custom_subscriber_listener.cpp#L55-L56), for which it uses [`dds_remote_qos_to_rmw_qos`](https://github.com/ros2/rmw_connext/blob/7781a5e5982032be9460840e33646028af21129e/rmw_connext_shared_cpp/src/qos.cpp#L309) which in turn may use an empty [`dds_qos_lifespan_to_rmw_qos_lifespan`]( https://github.com/ros2/rmw_connext/blob/7781a5e5982032be9460840e33646028af21129e/rmw_connext_shared_cpp/src/qos.cpp#L259) implementation. As the first `rmw_qos_profile_t` instance isn't initialized, nowhere is. The same goes for any QoS getter that doesn't initialize that duration prior to function call. 

> fastrtps and connext return one value for "unspecified"

Which value?

> Instead of working around the problem here the RMW implementations should be updated to return a defined value.

This patch is orthogonal to that change.

> The only logic which should be present here is to check valid boundaries and show an error / raise an exception when that is violated.

We can raise an exception if the `uint64_t` to `int64_t` conversion is not feasible, sure. But bear in mind that unreasonable values are not invalid values, type wise. We've set ourselves up for a problem by introducing `rmw_time_t` in the first place IMHO.

hidmic:
See https://github.com/ros2/rmw_connext/pull/422 and https://github.com/ros2/rclpy/pull/555/commits/0ef805b387961fad319abb2264f16ecdccc38196. It's not a silver bullet for the `rmw_time_t` <-> `rcutils_time_point_t` problem, but it's an improvement.

emersonknapp:
> Which value? I don't have a strong opinion against zero initialization though.

I don't have it up and running in front of me right now, but I was seeing (from https://github.com/ros2/rosbag2/pull/335#issuecomment-606306677)

FastRTPS seems to always return the following when a QoS policy duration was unspecified (e.g. {0, 0} passed in)

```
sec: 0x7FFFFFFF 
nsec: 0xFFFFFFFF
```

Cyclone seems to always return the following when a QoS policy duration was unspecified (e.g. {0, 0} passed in)

```
sec = 9223372036 == 0x225C17D04 
nsec = 854775807 == 0x32F2D7FF
```

hidmic:
CI up to `ros2cli`, `rclpy`, and `rmw_connext` (when possible):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10632/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6064/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8657/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10538/)

dirk-thomas:
How about an alternative: construct the Python `Duration` with two arguments - seconds and nanoseconds - as provided by `rmw_time_t` and instead perform the check if the value matches the bounds of the underlying C type in `rclpy_create_duration()`?

hidmic:
CI up to `ros2cli`, `rclpy`, and `rmw_connext` (when possible):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10638/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6068/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8661/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10546/)

hidmic:
@dirk-thomas PTAL

dirk-thomas:
> instead perform the check if the value matches the bounds of the underlying C type in `rclpy_create_duration()`?

Does something need to be changed in this function?

PR #556, opened 2020-05-07
Remove MANUAL_BY_NODE liveliness API

ivanpauno:
See https://github.com/ros2/rmw/pull/227.


rclpy/src/rclpy/_rclpy.c
```diff
@@ -4300,16 +4300,7 @@ rclpy_assert_liveliness(PyObject * Py_UNUSED(self), PyObject * args)
...
-    rcl_node_t * node = rclpy_handle_get_pointer_from_capsule(pyentity, "rcl_node_t");
-    if (RCL_RET_OK != rcl_node_assert_liveliness(node)) {
-      PyErr_Format(
-        RCLError,
-        "Failed assert liveliness on the Node: %s", rcl_get_error_string().str);
-      rcl_reset_error();
-      return NULL;
-    }
-  } else if (PyCapsule_IsValid(pyentity, "rclpy_publisher_t")) {
+  if (PyCapsule_IsValid(pyentity, "rclpy_publisher_t")) {
```
jacobperron, line 4303: Please update the docblock and error message in the `else-block` to indicate that this call is only valid for Publishers.

ivanpauno: Great catch! See https://github.com/ros2/rclpy/pull/556/commits/5cde700ef29c65f8996e66fa092288d013092a03.

jacobperron: nit: the exception message below can also be updated.

ivanpauno: See https://github.com/ros2/rclpy/pull/556/commits/069fcba130896d1aa109281a2449d754094f5945

PR #558, opened 2020-05-14
Explicitly add DLL directories for Windows before importing

jacobperron:
New in Python 3.8, we should call os.add_dll_directory for directories
containing the DLLs we intend to import as well as their recursive
dependencies.

See https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew

Unblocks https://github.com/ros2/ci/pull/432 🤞

### DIFF
--- a/rclpy/rclpy/impl/__init__.py
+++ b/rclpy/rclpy/impl/__init__.py
@@ -17,8 +17,20 @@ import os
 
 
 def _import(name):
+    dll_dir_handles = []
     try:
-        return importlib.import_module(name, package='rclpy')
+        # New in Python 3.8: on Windows we should call 'add_dll_directory()' for directories
+        # containing DLLs we depend on.
+        # https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew
+        if os.name == 'nt':
+            path_env = os.environ['PATH'].split(';')
+            for prefix_path in path_env:
+                if os.path.exists(prefix_path):
+                    dll_dir_handles.append(os.add_dll_directory(prefix_path))
+
+        imported_module = importlib.import_module(name, package='rclpy')
+
+        return imported_module
     except ImportError as e:
         if e.path is not None and os.path.isfile(e.path):
             e.msg += \
@@ -27,3 +39,6 @@ def _import(name):
                 (e.path, 'https://index.ros.org/doc/ros2/Troubleshooting/'
                          '#import-failing-even-with-library-present-on-the-system')
         raise
+    finally:
+        for handle in dll_dir_handles:
+            handle.close()
--- a/rclpy/rclpy/impl/__init__.py
+++ b/rclpy/rclpy/impl/__init__.py
@@ -15,19 +15,16 @@
 import importlib
 import os
 
+from rpyutils import add_dll_directories_from_env
+
 
 def _import(name):
-    # New in Python 3.8: on Windows we should call 'add_dll_directory()' for directories
-    # containing DLLs we depend on.
-    # https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew
-    dll_dir_handles = []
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/impl/__init__.py
```diff
@@ -17,8 +17,20 @@
...
 
 def _import(name):
+    dll_dir_handles = []
     try:
-        return importlib.import_module(name, package='rclpy')
+        # New in Python 3.8: on Windows we should call 'add_dll_directory()' for directories
+        # containing DLLs we depend on.
+        # https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew
+        if os.name == 'nt':
+            path_env = os.environ['PATH'].split(';')
```
dirk-thomas, line 26: There is a reason why Python chose to *not* use all paths in `PATH`. Maybe it would be better to not bypass this mechanism entirely but select specific paths only to be added, e.g. `rclpy`, message packages.

rclpy/rclpy/impl/__init__.py
```diff
@@ -17,8 +17,20 @@
 
 
 def _import(name):
+    dll_dir_handles = []
     try:
-        return importlib.import_module(name, package='rclpy')
+        # New in Python 3.8: on Windows we should call 'add_dll_directory()' for directories
+        # containing DLLs we depend on.
+        # https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew
+        if os.name == 'nt':
```
nuclearsandwich, line 25:
```suggestion
        if os.name == 'nt' and hasattr(os, 'add_dll_directory'):
```

rclpy/rclpy/impl/__init__.py
dirk-thomas: Message packages should probably contain their own logic to ensure their path is added too.

jacobperron: > There is a reason why Python chose to not use all paths in PATH

What I gather from the issue[1] is that *where* DLLs are loaded from was not consistent before Python 3.8 (e.g. in general, you couldn't assume that PATH was ever searched for DLLs; it was using the process default ordering[2]), which has some security implications[3]. The change in Python 3.8 forces a consistent search order.

The reason to not include `PATH` in the search order is a little above my head, but I think it boils down to a minimizing the risk of a so-called *DLL preloading attack*. I agree we could try could try down-selecting to just what we need, though I'm not sure of a proper way to achieve this. Maybe using CMake variables that are exported by packages producing DLLs? Also, I guess there will be other DLLs installed on the system we rely on that should also be added (e.g. OpenSSL).

---

[1] https://bugs.python.org/issue36085
[2] https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order
[3] https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-security

jacobperron: Yeah, you're right. `rosidl_generator_py` needs a similar patch. Pretty much anywhere we use `importlib.import_module`.

rclpy/rclpy/impl/__init__.py
jacobperron: Added as part of 241a02c

jacobperron:
I also had to make the same patch to the test module 4f1676c

jacobperron:
I've moved the logic to a new common package `rpyutils` (a795304).

This PR now depends on https://github.com/ros2/ros2/pull/920 and https://github.com/ros2/rpyutils/pull/2

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10797/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6182/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8790/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/10691/)

PR #559, opened 2020-05-16
PARAM_REL_TOL comment was incorrect in the api doc.

dawonn-haval:
See: http://docs.ros2.org/latest/api/rclpy/api/node.html

Signed-off-by: Dereck Wonnacott <dereck.wonnacott@havalus.com>

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -92,14 +92,17 @@ NodeNameNonExistentError = _rclpy.NodeNameNonExistentError
 
 
 class Node:
-
-    PARAM_REL_TOL = 1e-6
-
     """
     A Node in the ROS graph.
 
     A Node is the primary entrypoint in a ROS system for communication.
     It can be used to create ROS entities such as publishers, subscribers, services, etc.
+    """
+
+    PARAM_REL_TOL = 1e-6
+    """
+    Relative tolerance for floating point parameter values' comparison.
+    See `math.isclose` documentation.
     """
 
     def __init__(


rclpy/rclpy/node.py
```diff
@@ -92,16 +92,20 @@
...
 
     A Node is the primary entrypoint in a ROS system for communication.
     It can be used to create ROS entities such as publishers, subscribers, services, etc.
     """
 
+    PARAM_REL_TOL = 1e-6
+    """
+    Minimum change in parameter value to register a change.
+
+    Changes smaller than this value will be ignored.
```
hidmic, line 106: @dawonn-haval this isn't exactly true. I see `PARAM_REL_TOL` being used to tell if a parameter value to which a floating point range constraint applies meets the constraint or not.

dirk-thomas: @dawonn-haval Friendly ping.

dawonn-haval:
This totally dropped of my radar, sorry about the delay. He's totally right, I misunderstood the logic.

Updated the comment and rebased.

rclpy/rclpy/node.py
dawonn-haval: This totally dropped of my radar, sorry about the delay. He's totally right, I misunderstood the logic.

Updated the comment and rebased.

rclpy/rclpy/node.py
```diff
@@ -92,16 +92,21 @@
...
     """
     A Node in the ROS graph.
 
     A Node is the primary entrypoint in a ROS system for communication.
     It can be used to create ROS entities such as publishers, subscribers, services, etc.
     """
 
+    PARAM_REL_TOL = 1e-6
+    """
+    Maximum differnance between two floating point values to determine that they are 'close'.
```
hidmic, line 104: @dawonn-haval consider: 
```suggestion
    Relative tolerance for floating point parameter values' comparison. See `math.isclose` documentation.
```
to avoid repeating documentation.

rclpy/rclpy/node.py
```diff
@@ -92,16 +92,21 @@
...
+
+    This value is used to validate float parameter changes that are 'close' to defined min,
```
hidmic, line 106: @dawonn-haval consider:
```suggestion
    This tolerance is used to ensure floating point parameter values comply with range constraints, if any. See `ParameterDescriptor` and `FloatingPointRange` messages documentation.
```

rclpy/rclpy/node.py
```diff
@@ -92,16 +92,21 @@
...
     A Node is the primary entrypoint in a ROS system for communication.
     It can be used to create ROS entities such as publishers, subscribers, services, etc.
     """
 
+    PARAM_REL_TOL = 1e-6
+    """
+    Relative tolerance for floating point parameter values' comparison. See `math.isclose` documentation.
+
+    This tolerance is used to ensure floating point parameter values comply with range constraints, if any. See `ParameterDescriptor` and `FloatingPointRange` messages documentation.
+    max, and step values.
```
hidmic, line 107: @dawonn-haval leftover?

hidmic:
@dawonn-haval DCO and linters are not happy.

dirk-thomas:
@dawonn-haval Friendly ping.

dirk-thomas:
@dawonn-haval Friendly ping.

dawonn-haval:
> @dawonn-haval the fix itself looks good to me, but what's with all these style changes?

my editor is not configured for the ROS2 style guide, so every time I save it auto-formats incorrectly. So for ROS2 contribs I need to fix linting errors manually and 'save without formatting'... I try to check each commit after I push it to make sure it's clean but clearly not an expert...

hidmic:
@dawonn-haval it's no silver bullet, better to have your editor configured, but I've found out that `autopep8` goes a long way.

dawonn-haval:
yeah, my editor is configured for work, and every time I save I need to use 'save without formatting.' 

I clearly just hit ctrl+s mindlessly for the dozen'th time...

PR #562, opened 2020-05-29
protect access to global logging calls with a mutex

wjwwood:
Counter part to https://github.com/ros2/rclcpp/pull/1125

Note this pull request introduces C++ code to the rclpy python library (though it keeps the main files as pure C). I started by trying to implement it without std mutex from C++, but that was taking a lot more time, so I decided to create new C functions to do the mutex locking and call those from the main file. In the future we could reimplement these few functions to be C only if we wanted to do so.


wjwwood:
CI (full because I couldn't come up with a good way to cover what might be affected without potentially missing something):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10935/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6281/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8903/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10836/)

rclpy/src/rclpy/detail/logging_mutex.cpp
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+
+#include "./logging_mutex.hpp"
+
+#include <memory>
+#include <mutex>
+
+#include "rcutils/macros.h"
+
+std::shared_ptr<std::recursive_mutex>
```
ivanpauno, line 22: nit: we could make use of a `std::recursive_mutex` directly, I don't see an advantage of using `std::shared_ptr` here (all Python calls will happen after "library construction" and before "library destruction").

rclpy/src/rclpy/detail/logging_mutex.hpp
```diff
@@ -0,0 +1,47 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLPY__DETAIL__LOGGING_MUTEX_HPP_
+#define RCLPY__DETAIL__LOGGING_MUTEX_HPP_
+
+#include <memory>
+#include <mutex>
+
+#include "rcutils/visibility_control_macros.h"
```
ivanpauno, line 21: do we set rcutils visibility options when compiling rclpy?

rclpy/src/rclpy/detail/logging_mutex.hpp
```diff
@@ -0,0 +1,47 @@
...
+// it at destruction time can keep it alive.
+// In that way, a destruction ordering problem between static objects is avoided.
+RCUTILS_LOCAL
+std::shared_ptr<std::recursive_mutex>
+get_global_logging_mutex();
+
+}  // namespace detail
+}  // namespace rclpy
+
+#endif  // RCLPY__DETAIL__LOGGING_MUTEX_HPP_
```
ivanpauno, line 47: nit: missing new line before EOF

rclpy/src/rclpy/detail/logging_mutex.cpp
wjwwood: Yeah I thought about this and considered changing it, but I also don't think it is really a problem to use a shared pointer here.

I can change it though.

wjwwood:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10947/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6292/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8914/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10853/)

rclpy/src/rclpy/detail/logging_mutex.hpp
ivanpauno: > I don't understand the question, do you mean like a compiler definition or something?

Yes, I thought it was a project dependent macro, but it isn't.
It's ok as-is.

wjwwood:
The CI failures are, in my opinion, unrelated. Just waiting on second review from @jacobperron, but I think this is gtg.

PR #568, opened 2020-06-09
Add resolved_name() method to publisher

crystaldust:
The method internally call the c function [`rcl_publisher_get_topic_name`](https://github.com/ros2/rcl/blob/master/rcl/include/rcl/publisher.h#L456) in rcl library.

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -5439,17 +5439,27 @@ rclpy_service_info_get_received_timestamp(PyObject * Py_UNUSED(self), PyObject *
 static PyObject *
 rclpy_publisher_get_topic_name(PyObject * Py_UNUSED(self), PyObject * args)
 {
-    PyObject * pyentity;
-    if (!PyArg_ParseTuple(args, "O", &pyentity)) {
-        return NULL;
-    }
+  PyObject * pyentity;
+  if (!PyArg_ParseTuple(args, "O", &pyentity)) {
+    return NULL;
+  }
 
-    rclpy_publisher_t * publisher = rclpy_handle_get_pointer_from_capsule(
-        pyentity, "rclpy_publisher_t");
-    if (!publisher) {
-        return NULL;
-    }
-    return PyUnicode_FromString(rcl_publisher_get_topic_name(&publisher->publisher));
+  rclpy_publisher_t * publisher = rclpy_handle_get_pointer_from_capsule(
+    pyentity, "rclpy_publisher_t");
+  if (!publisher) {
+    return NULL;
+  }
+
+  const char * topic_name = rcl_publisher_get_topic_name(&publisher->publisher);
+  if (!topic_name) {
+    PyErr_Format(
+      RCLError,
+      "Failed to get topic name: %s",
+      rcl_get_error_string().str);
+    rcl_reset_error();
+    return NULL;
+  }
+  return PyUnicode_FromString(topic_name);
 }
 
 /// Define the public methods of this module
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -5418,7 +5418,6 @@ rclpy_service_info_get_source_timestamp(PyObject * Py_UNUSED(self), PyObject * a
   return PyLong_FromLongLong(service_info->source_timestamp);
 }
 
-
 /// Retrieves the received timestsamp number from a rmw_service_info_t capsule
 /**
  * Raises RuntimeError on failure.
@@ -5436,24 +5435,21 @@ rclpy_service_info_get_received_timestamp(PyObject * Py_UNUSED(self), PyObject *
   return PyLong_FromLongLong(service_info->received_timestamp);
 }
 
+/// Retrieve the topic name from a rclpy_publisher_t
 static PyObject *
-rclpy_publisher_get_resolved_name(PyObject * Py_UNUSED(self), PyObject * args)
+rclpy_publisher_get_topic_name(PyObject * Py_UNUSED(self), PyObject * args)
 {
     PyObject * pyentity;
     if (!PyArg_ParseTuple(args, "O", &pyentity)) {
         return NULL;
     }
 
-    if (PyCapsule_IsValid(pyentity, "rclpy_publisher_t")) {
-        rclpy_publisher_t *publisher = rclpy_handle_get_pointer_from_capsule(
-            pyentity, "rclpy_publisher_t");
-        const char * topic_name = rcl_publisher_get_topic_name(&publisher->publisher);
-        return Py_BuildValue("s", topic_name);
-    } else {
-        PyErr_Format(
-            PyExc_TypeError, "Passed capsule is not a valid Publisher.");
+    rclpy_publisher_t * publisher = rclpy_handle_get_pointer_from_capsule(
+        pyentity, "rclpy_publisher_t");
+    if (!publisher) {
         return NULL;
     }
+    return PyUnicode_FromString(rcl_publisher_get_topic_name(&publisher->publisher));
 }
 
 /// Define the public methods of this module
@@ -5863,7 +5859,7 @@ static PyMethodDef rclpy_methods[] = {
     "Retrieve received timestamp from service_info"
   },
   {
-    "rclpy_publisher_get_resolved_name", rclpy_publisher_get_resolved_name,
+    "rclpy_publisher_get_topic_name", rclpy_publisher_get_topic_name,
     METH_VARARGS,
     "Get the resolved name(topic) of publisher"
   },
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -5436,6 +5436,26 @@ rclpy_service_info_get_received_timestamp(PyObject * Py_UNUSED(self), PyObject *
   return PyLong_FromLongLong(service_info->received_timestamp);
 }
 
+static PyObject *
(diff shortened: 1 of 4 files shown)


crystaldust:
@sloretz Let's talke about the feature here.
I see the publisher class in rclcpp has a method [`get_topic_name`](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/publisher_base.cpp#L122) which calls the same c function rcl_publisher_get_topic_name, should we keep the method name the same in rclpy? Since at first sight `resolved_name` seems to be something like the name of the host machine the node is running on.

sloretz:
> I see the publisher class in rclcpp has a method get_topic_name which calls the same c function rcl_publisher_get_topic_name, should we keep the method name the same in rclpy? Since at first sight resolved_name seems to be something like the name of the host machine which the node is running on.

That's a good point! `topic` seems clearer to me than `resolved` when on the publisher. I don't think the naming between rclcpp and rclpy has to be *exactly* the same since there are different norms for naming in C++ and Python. I wouldn't expect to see a `get_...()` method on a Python class. Now that you mention it, I think an `@property` called `topic_name` would be a good way to access this, while still being similar enough to `rclcpp` that C++ users can find it.

rclpy/test/test_publisher.py
```diff
@@ -0,0 +1,82 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
```
sloretz, line 1: New file should use year 2020

rclpy/test/test_publisher.py
```diff
@@ -0,0 +1,82 @@
...
+        self.context = rclpy.context.Context()
+        rclpy.init(context=self.context)
+        self.node = rclpy.create_node(
+            TEST_NODE,
+            namespace=TEST_NAMESPACE,
+            context=self.context,
+            parameter_overrides=[
+                Parameter('initial_foo', Parameter.Type.INTEGER, 4321),
+                Parameter('initial_bar', Parameter.Type.STRING, 'init_param'),
+                Parameter('initial_baz', Parameter.Type.DOUBLE, 3.14)
```
sloretz, line 57: I think this test could be minimized. These `Parameter`'s can be removed since they're not needed by the test, same with `TEST_NODE` and `TEST_NAMESPACE`.

rclpy/test/test_publisher.py
```diff
@@ -0,0 +1,82 @@
...
+            ],
+            cli_args=[
+                '--ros-args', '-p', 'initial_fizz:=buzz',
+                '--params-file', str(TEST_RESOURCES_DIR / 'test_parameters.yaml'),
+                '-p', 'initial_buzz:=1.'
+            ],
+            automatically_declare_parameters_from_overrides=False
```
sloretz, line 64: Can remove this argument

rclpy/test/test_publisher.py
```diff
@@ -0,0 +1,82 @@
...
+            context=self.context,
+            parameter_overrides=[
+                Parameter('initial_foo', Parameter.Type.INTEGER, 4321),
+                Parameter('initial_bar', Parameter.Type.STRING, 'init_param'),
+                Parameter('initial_baz', Parameter.Type.DOUBLE, 3.14)
+            ],
+            cli_args=[
+                '--ros-args', '-p', 'initial_fizz:=buzz',
+                '--params-file', str(TEST_RESOURCES_DIR / 'test_parameters.yaml'),
+                '-p', 'initial_buzz:=1.'
```
sloretz, line 62: I think the only args needed for this test would be remapping arguments: `['--ros-args', '-r', '/original/topic/name:=/remapped/topic/name']`. Then the test can create a publisher with the name `/original/topic/name` and make sure `resolved_name()` (or `publisher.topic_name` if you go with a `@property`) returns `/remapped/topic/name`.

rclpy/test/test_publisher.py
```diff
@@ -0,0 +1,82 @@
...
+            ],
+            automatically_declare_parameters_from_overrides=False
+        )
+        self.base_publisher = self.node.create_publisher(BasicTypes, TEST_TOPIC, 0)
+        self.position_publisher = self.node.create_publisher(BasicTypes, TEST_TOPIC_POSITION, 0)
+        self.speed_publisher = self.node.create_publisher(BasicTypes, TEST_TOPIC_SPEED, 0)
```
sloretz, line 68: Only need one publisher for this test, and I would create it in the test function itself.

rclpy/test/test_publisher.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @classmethod
+    def tearDown(self):
+        self.node.destroy_node()
+        rclpy.shutdown(context=self.context)
+
+    def test_resolved_name(self):
+        self.assertEqual(self.base_publisher.resolved_name(), full_topic(TEST_TOPIC))
+        self.assertEqual(self.position_publisher.resolved_name(), full_topic(TEST_TOPIC_POSITION))
+        self.assertEqual(self.speed_publisher.resolved_name(), full_topic(TEST_TOPIC_SPEED))
```
sloretz, line 78: Since `rclpy` uses `pytest` as the test runner, let's use `pytest` style assertions instead of `self.assert...()`. It's simpler and `pytest` is great about printing info when tests fail:

```
def test_resolved_name(self):
   # ... create publisher code here ...

   assert '/remapped/topic/name' == publisher.topic_name

   # ... destroy publisher code here ...

**Comment by crystaldust** [comment]
I've made some updates following the comments. The test cases cover the node namespace effects on topic_name, and the remapping rules.

I used to think we should have full coverage on naming and remapping, while found the rules are pretty complicated, but they are well covered in rcl [test_remap.cpp](https://github.com/ros2/rcl/blob/master/rcl/test/rcl/test_remap.cpp) and [test_remap_integration.cpp](https://github.com/ros2/rcl/blob/master/rcl/test/rcl/test_remap_integration.cpp). Since rclpy is just a binding on rcl, we may not need to test every aspect on remapping on rclpy.

File: rclpy/src/rclpy/_rclpy.c [hunk 7]
```diff
@@ -5436,6 +5435,23 @@ rclpy_service_info_get_received_timestamp(PyObject * Py_UNUSED(self), PyObject *
...
+    if (!PyArg_ParseTuple(args, "O", &pyentity)) {
+        return NULL;
+    }
+
+    rclpy_publisher_t * publisher = rclpy_handle_get_pointer_from_capsule(
+        pyentity, "rclpy_publisher_t");
+    if (!publisher) {
+        return NULL;
+    }
+    return PyUnicode_FromString(rcl_publisher_get_topic_name(&publisher->publisher));
```
sloretz, line 5452: One more thing to handle. It's possible `rcl_publisher_get_topic_name()` could fail, in which case the return value is `NULL` and an `rcl` error message is set. If this happens, I think we should raise an exception - and there's a custom exception for errors in `rcl` stored in the variable `RCLError` in this file.


Here's an example:

https://github.com/ros2/rclpy/blob/8f601c69ce9775ee3764c2bcfa2cd22cd9c31a1d/rclpy/src/rclpy/_rclpy.c#L184-L188

`PyErr_Format()` creates the exception of type `RCLError`, `rcl_get_error_string()` gets the error message set by `rcl_publisher_get_topic_name()`, and `rcl_reset_error()` clears the error message so it doesn't get in the way of the next RCL call. Mind adding that here?

rclpy/test/test_publisher.py
```diff
@@ -0,0 +1,103 @@
...
+
+        Args:
+            test_topics: A list of binary tuple in the form (topic, expected topic), the
+            former will be passed to node.create_publisher and the latter will be compared
+            with publisher.topic_name
+            node: The node used to create the publisher. The node's namespace will have
+            an effect on the publisher's topic_name.
+        """
+        for topic_tuple in test_topics:
+            topic, target_topic = topic_tuple
```
sloretz, line 68: small nitpick :upside_down_face: , this could be combined into `for topic, target_topic in test_topics:`

sloretz:
Looks like the PR is really close to completion. Running linux CI testing just `rclpy` to see if new tests pass and if linters pass

[Build Status](https://ci.ros2.org/job/ci_linux/11115/)

rclpy/test/test_publisher.py
crystaldust: Aha, my stupid! How could I forget Python's sweet sugar

rclpy/src/rclpy/_rclpy.c
crystaldust: That makes sense, I'll add the handle here.

crystaldust:
Fixed the two snippets, should I do a rebase before the PR is merged?

crystaldust:
Opps, I'll take a look at the pylint errors in the CI build

crystaldust:
`test/test_publisher.py` added to CMakeLists.txt :)

sloretz:
CI (Testing just `rclpy`)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11154/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6441/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9097/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11069/)

sloretz:
Approved and CI is green; merging :tada:

Thank you @crystaldust for contributing and iterating!

sloretz:
I think it would be safe to backport this change to Foxy, though there appears to be a merge conflict with Eloquent. I'll add this to the Foxy backport board.

PR #569, opened 2020-06-09
Doc+fix rclpy_handle_get_pointer_from_capsule()

sloretz:
This documents `rclpy_handle_get_pointer_from_capsule()` and fixes a case where `RuntimeError` is set when an exception was already set by [PyCapsule_GetPointer](https://docs.python.org/3.7/c-api/capsule.html#c.PyCapsule_GetPointer).

### DIFF
--- a/rclpy/src/rclpy_common/include/rclpy_common/handle.h
+++ b/rclpy/src/rclpy_common/include/rclpy_common/handle.h
@@ -61,7 +61,7 @@ rclpy_create_handle_capsule(void * ptr, const char * name, rclpy_handle_destruct
 
 /// Returns the object managed by the rclpy_handle_t wrapped in a PyCapsule.
 /**
- *  PyExec_RuntimeError is set if the PyCapsule is valid but the `rclpy_handle_t` is not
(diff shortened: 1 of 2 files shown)


sloretz:
CI (testing just rclpy)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11076/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6390/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9036/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10992/)

sloretz:
Ignoring windows test failure from cppcheck as unrelated to this PR.

```
There is an unknown macro here somewhere. Configuration is required. If RCUTILS_STRINGIFY is a macro then please configure it.
```

> lgtm, should be backported?'

Yup, but only to Foxy. The file changed doesn't appear to exist in Eloquent.

sloretz:
This does not need to be released immediately, it can wait for the next `rclpy` release. The bug is unlikely to be encountered, and if it is the symptom is just extra stderr console spam about an exception already being set.

PR #571, opened 2020-06-09
Add `topic_name` property to Subscription

audrow:


rclpy/src/rclpy/_rclpy.c
```diff
@@ -1161,6 +1161,27 @@ rclpy_get_subscriptions_info_by_topic(PyObject * Py_UNUSED(self), PyObject * arg
...
+
+  rclpy_subscription_t * sub =
+    rclpy_handle_get_pointer_from_capsule(pysubscription, "rclpy_subscription_t");
+  if (NULL == sub) {
+    return NULL;
+  }
+
+  const char * subscription_name = rcl_subscription_get_topic_name(&(sub->subscription));
+  if (NULL == subscription_name) {
+    Py_RETURN_NONE;
```
sloretz, line 1180: I think in this case `rcl_subscription_get_topic_name()` woul have set an error message. I think this should raise an exception in that case

example:
https://github.com/ros2/rclpy/blob/78f8a9e3d10964b2c7f347980ebd0e1eb570b526/rclpy/src/rclpy/_rclpy.c#L1973-L1975

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1161,6 +1161,27 @@ rclpy_get_subscriptions_info_by_topic(PyObject * Py_UNUSED(self), PyObject * arg
   return _get_info_by_topic(args, "subscriptions", rcl_get_subscriptions_info_by_topic);
 }
 
+static PyObject *
+rclpy_get_resolved_subscription_name(PyObject * Py_UNUSED(self), PyObject * args)
```
sloretz, line 1165: nit, I would name this `rclpy_get_subscription_topic_name`, like `rclpy_get_node_name()` is the `rclpy` wrapper around `rcl_node_get_name()`.

rclpy/test/test_subscription.py
```diff
@@ -0,0 +1,56 @@
...
+
+import rclpy
+from rclpy.node import Node
+from test_msgs.msg import Empty
+
+
+class TestSubscriber(Node):
+
+    def __init__(self, topic):
+        super().__init__('minimal_subscriber')
```
sloretz, line 25: I recommend adding a few remap rules that affect a couple of the test cases to confirm the topic name returned is actually the one being used, and not the original unremapped topic name.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -1161,6 +1161,27 @@ rclpy_get_subscriptions_info_by_topic(PyObject * Py_UNUSED(self), PyObject * arg
   return _get_info_by_topic(args, "subscriptions", rcl_get_subscriptions_info_by_topic);
 }
 
+static PyObject *
```
fujitatomoya, line 1164: could you add description what this method is capable along with others?

rclpy/test/test_subscription.py
```diff
@@ -0,0 +1,56 @@
...
+    rclpy.init()
+
+
+@pytest.mark.parametrize('topic_name, expected', [
+    ('topic', '/topic'),
+    ('/topic', '/topic'),
+    ('sample/topic', '/sample/topic'),
+    ('/sample/topic', '/sample/topic'),
+    ('sample/topic/name', '/sample/topic/name'),
+    ('/sample/topic/name', '/sample/topic/name'),
```
fujitatomoya, line 51: I would add false cases.

fujitatomoya: i was thinking corner cases as unit test,
e.g) ('/sample/topics', '/sample/topic'), false expected.

fujitatomoya:
@audrow 

thanks, i will look into it.

rclpy/src/rclpy/_rclpy.c
sloretz: Whoops, looks like the `RCLError` code got put in the wrong place :upside_down_face: It should be here after the `rcl_subscription_get_topic_name()` call.

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11123/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6424/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9075/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11047/)

rclpy/src/rclpy/_rclpy.c
audrow: Oops.  I believe it should be good now.

rclpy/test/test_subscription.py
audrow: If I understand correctly, you're suggestion of negative tests would help make the tests more thorough, where as in the commit you've reviewed they were pretty one dimensional (make sure there is a leading slash, effectively).  

Towards this end, I added cases for remappings and namespaces, which test that the topic name is being resolved in more complex and varied circumstances.  In this case, I think that testing for direct equality is more informative than checking that the function does not give me something that I don't expect, as it is far less likely to be what I expect than anything else if things are implemented incorrectly. 

Thanks for this feedback.

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11169/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6447/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9104/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11079/)

PR #577, opened 2020-06-17
Pass rcutils_include_dirs to cppcheck 

claireyywang:
This fixes the rclpy's `unknown macro` error on osx and windows, but it may not be the best strategy, cuz it will require manually pass in `include_dirs` to cppcheck in all other repos listed on this test report https://ci.ros2.org/view/nightly/job/nightly_win_rel/1576/testReport/

We've encountered the same error before and managed to fix it in this ticket https://github.com/ament/ament_lint/issues/116/ I'm not sure why this is happening again. 

Original issue https://github.com/ros2/ros2/issues/942

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -187,7 +187,7 @@ endif()
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   # Give cppcheck hints about macro definitions coming from outside this package
-  set(ament_cmake_cppcheck_ADDITIONAL_INCLUDE_DIRS ${rcutils_INCLUDE_DIRS})
+  get_target_property(ament_cmake_cppcheck_ADDITIONAL_INCLUDE_DIRS rcutils::rcutils INTERFACE_INCLUDE_DIRECTORIES)
   ament_lint_auto_find_test_dependencies()
 
   find_package(ament_cmake_pytest REQUIRED)
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -185,6 +185,9 @@ if(NOT WIN32)
 endif()
 
 if(BUILD_TESTING)
+  # Give cppcheck hints about macro definitions coming from outside this package
+  set(ament_cmake_cppcheck_ADDITIONAL_INCLUDE_DIRS ${rcutils_INCLUDE_DIRS})
+
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
 
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -185,9 +185,6 @@ if(NOT WIN32)
 endif()
 
 if(BUILD_TESTING)
-  # Give cppcheck hints about macro definitions coming from outside this package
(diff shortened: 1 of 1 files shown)


claireyywang:
windows CI [Build Status](https://ci.ros2.org/job/ci_windows/11106/)

dirk-thomas:
This approach doesn't seem like the right solution. Maybe the original decision to not include dirs from other packages needs to be reconsidered (see https://github.com/ament/ament_lint/pull/117#discussion_r441854363).

jacobperron:
> This approach doesn't seem like the right solution. Maybe the original decision to not include dirs from other packages needs to be reconsidered (see ament/ament_lint#117 (comment)).

IIRC, the reason I opted to *not* include headers from all dependencies was that the running time of cppcheck increase by a lot (e.g. the test timeout was easily exceeded). It still might be worth trying again; perhaps there have been performance improvements to cppcheck since then.

claireyywang:
> It used to be rare that we needed to give these hints, but if it starting to become a common occurrence maybe we should investigate alternatives.

`rclpy` is one of the packages having this error. The bulk of errors from [test report](https://ci.ros2.org/view/nightly/job/nightly_win_rel/1576/testReport/) seem to be from a bunch of `rosidl_typesupport` and `rcl` packages. It's hard to estimate from the report how many are affected, but if this approach is the right way to go, I will need to start passing macro `include_dirs` to all those packages.

claireyywang:
I guess another way is to pass all macro `include_dirs`  used by the package to cppcheck as well, instead of just `include_dirs` of the package being tested, but I don't know enough `cmake` or build process to see the down side of this approach.

rclpy/CMakeLists.txt
```diff
@@ -186,6 +186,8 @@ endif()
 
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
+  # Give cppcheck hints about macro definitions coming from outside this package
+  set(ament_cmake_cppcheck_ADDITIONAL_INCLUDE_DIRS ${rcutils_INCLUDE_DIRS})
```
dirk-thomas, line 190: Instead of the classic CMake variable this should use the modern CMake target:

```
get_target_property(ament_cmake_cppcheck_ADDITIONAL_INCLUDE_DIRS rcutils::rcutils INTERFACE_INCLUDE_DIRECTORIES)
```

dirk-thomas:
I am mostly concerned about the locality of a change. If a package starts using a macro it might need to update the `cppcheck` arguments (not great but not terrible either).

But that change shouldn't be exposed to downstream packages and if upstream packages change their implementation that shouldn't effect this package either.

The exception would be if an upstream package is already explicitly listed for additional include directories - at that point any future change in the upstream package might cause a regression in this package. That would be case which might bubble through the stack which would be really unfortunate.

Beside my minor inline comment this patch addresses a specific issue we are having (when unpinning) so I am fine to merge it (with the comment addressed). Maybe a ticket to follow up this to see if there is any way to automate this while not killing performance would be good to create.

claireyywang:
CI again 
linux [Build Status](https://ci.ros2.org/job/ci_linux/11265/)
linux-aaaarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/6524/)
osx [Build Status](https://ci.ros2.org/job/ci_osx/9224/)
windows [Build Status](https://ci.ros2.org/job/ci_windows/11173/)

PR #580, opened 2020-06-18
improve error message if rclpy C extensions are not found

dirk-thomas:
There have been numerous tickets / questions regarding problems loading the `rclpy` C extensions (just the latest example https://answers.ros.org/question/355064/ros2-modulenotfounderror-no-module-named-rclpy_rclpy/):

> ModuleNotFoundError: No module named ‘rclpy._rclpy’

This PR adds more context to this specific failure (the expected library not being found / existing). See the referenced PR to add the linked section to the docs.

### DIFF
--- a/rclpy/rclpy/impl/__init__.py
+++ b/rclpy/rclpy/impl/__init__.py
@@ -14,6 +14,7 @@
 
 import importlib
 import os
+from pathlib import Path
 
 from rpyutils import add_dll_directories_from_env
 
@@ -28,10 +29,9 @@ def _import(name):
     except ImportError as e:
         if e.path is None:
             import sysconfig
-            expected_path = os.path.join(
-                os.path.dirname(os.path.dirname(__file__)),
+            expected_path = Path(__file__).parents[1] / (
                 name[1:] + sysconfig.get_config_var('EXT_SUFFIX'))
-            assert not os.path.isfile(expected_path)
+            assert not expected_path.is_file()
             e.msg += \
                 f"\nThe C extension '{expected_path}' isn't present on the " \
                 "system. Please refer to 'https://index.ros.org/doc/ros2/" \
--- a/rclpy/rclpy/impl/__init__.py
+++ b/rclpy/rclpy/impl/__init__.py
@@ -26,6 +26,18 @@ def _import(name):
         with add_dll_directories_from_env('PATH'):
             return importlib.import_module(name, package='rclpy')
     except ImportError as e:
+        if e.path is None:
+            import sysconfig
+            expected_path = os.path.join(
+                os.path.dirname(os.path.dirname(__file__)),
+                name[1:] + sysconfig.get_config_var('EXT_SUFFIX'))
+            assert not os.path.isfile(expected_path)
+            e.msg += \
+                "\nThe C extension '%s' isn't present on the system." \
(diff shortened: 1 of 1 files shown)


rclpy/rclpy/impl/__init__.py
```diff
@@ -26,6 +26,18 @@ def _import(name):
         with add_dll_directories_from_env('PATH'):
             return importlib.import_module(name, package='rclpy')
     except ImportError as e:
+        if e.path is None:
+            import sysconfig
+            expected_path = os.path.join(
+                os.path.dirname(os.path.dirname(__file__)),
```
ivanpauno, line 32: nit: Prefer using `pathlib` over `os.path`

rclpy/rclpy/impl/__init__.py
```diff
@@ -26,6 +26,18 @@ def _import(name):
...
+                name[1:] + sysconfig.get_config_var('EXT_SUFFIX'))
+            assert not os.path.isfile(expected_path)
+            e.msg += \
+                "\nThe C extension '%s' isn't present on the system." \
+                " Please refer to '%s' for possible solutions" % (
+                    expected_path,
+                    'https://index.ros.org/doc/ros2/Troubleshooting/Installation-Troubleshooting/'
+                    '#import-failing-without-library-present-on-the-system')
```
ivanpauno, line 40: nit: prefer using f-strings

rclpy/rclpy/impl/__init__.py
```diff
@@ -26,6 +26,18 @@ def _import(name):
         with add_dll_directories_from_env('PATH'):
             return importlib.import_module(name, package='rclpy')
     except ImportError as e:
+        if e.path is None:
+            import sysconfig
+            expected_path = os.path.join(
+                os.path.dirname(os.path.dirname(__file__)),
+                name[1:] + sysconfig.get_config_var('EXT_SUFFIX'))
+            assert not os.path.isfile(expected_path)
```
ivanpauno, line 34: can you explain what this assert is checking?
I don't get its purpose, neither what `expected_path` is.

ivanpauno: After rechecking, I got what `expected_path` is.

I think it would be better something like:
```python3
if os.path.is_file(expected_path):
    e.msg += ...
```

So if the file exist you get a normal import error, instead of an assert error.

dirk-thomas: The assertion is only ensuring the logic of the code. If the file does exist this branch of the code should never be taken since `e.path` would not be `None`. Therefore it should stay an `assert`.

ivanpauno: sounds reasonable

rclpy/rclpy/impl/__init__.py
dirk-thomas: Done in 8a1623b6be331821bfc461a54702bd0a12e179e1.

rclpy/rclpy/impl/__init__.py
dirk-thomas: Done in e76dc55bbc3de0b0201aea4994ee1bf0fb791be1.

dirk-thomas:

dirk-thomas:

PR #592, opened 2020-06-30
[backport dashing] improve error message if rclpy C extensions are not found

dirk-thomas:
Backport of #580.

### DIFF
--- a/rclpy/rclpy/impl/__init__.py
+++ b/rclpy/rclpy/impl/__init__.py
@@ -27,11 +27,11 @@ def _import(name):
                 name[1:] + sysconfig.get_config_var('EXT_SUFFIX'))
             assert not expected_path.is_file()
             e.msg += \
-                f"\nThe C extension '{expected_path}' isn't present on the " \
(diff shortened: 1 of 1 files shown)


rclpy/rclpy/impl/__init__.py
```diff
@@ -14,12 +14,24 @@
...
     try:
         return importlib.import_module(name, package='rclpy')
     except ImportError as e:
+        if e.path is None:
+            import sysconfig
+            expected_path = Path(__file__).parents[1] / (
+                name[1:] + sysconfig.get_config_var('EXT_SUFFIX'))
+            assert not expected_path.is_file()
+            e.msg += \
+                f"\nThe C extension '{expected_path}' isn't present on the " \
```
ivanpauno, line 30: does dashing minimum python version support f-strings?

dirk-thomas: Good catch. Dashing targets Python 3.5 which doesn't support f-strings. Fixed in 0262fc488bbcf24686af2eafe525b3e271f903b3.

Issue #484, opened 2019-12-17
Ability to configure domain ID in rclpy


fujitatomoya:
#### Feature description

ability to set the Domain ID on a per-node basis.

the same ability with https://github.com/ros2/rclcpp/issues/910

#### Implementation considerations

1. add domain id with rclpy Node Class option. (default RCL_NODE_OPTIONS_DEFAULT_DOMAIN_ID)
2. if it is RCL_NODE_OPTIONS_DEFAULT_DOMAIN_ID, the rcl overrides domain id with ROS_DOMAIN_ID environmental variable.

fujitatomoya:
related to https://github.com/ros2/rclcpp/issues/910

ivanpauno:
Closing, as I think it has been clarified in https://github.com/ros2/rclcpp/issues/910#issuecomment-566520259.
Please, reopen if not.

fujitatomoya:
@ivanpauno 

this is for the same ability with https://github.com/ros2/rclcpp/issues/910 for rclpy. i believe that this is needed for rclpy also not only for rclcpp.

ivanpauno:
> @ivanpauno
> 
> this is for the same ability with [ros2/rclcpp#910](https://github.com/ros2/rclcpp/issues/910) for rclpy. i believe that this is needed for rclpy also not only for rclcpp.

Sorry, I didn't understand that.
The title sounds like rclpy is not using `ROS_DOMAIN_ID` environment variable.
Can you update the description and reopen? Thanks!

fujitatomoya:
@ivanpauno 

> The title sounds like rclpy is not using ROS_DOMAIN_ID environment variable.

my bad, that confuses you.

> Can you update the description and reopen? Thanks!

will do it right now.

suab321321:
@ivanpauno sir I would like to contribute to this.

ivanpauno:
For the same reason than here https://github.com/ros2/rclcpp/issues/910#issuecomment-555000410, it's better to wait now. The other set of PRs will be done soon, after that, this will be unblocked.

fujitatomoya:
@ivanpauno 

this one too, you can leave it to us.

Barry-Xu-2018:
@ivanpauno @fujitatomoya 

After checking current implementation of rclpy, it only supports use default rcl_opitions.   

[...]

PR #594, opened 2020-07-02
Add ability to configure domain ID

Barry-Xu-2018:
Related to #484

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -60,7 +60,12 @@ if TYPE_CHECKING:
     from rclpy.node import Node  # noqa: F401
 
 
-def init(*, args: Optional[List[str]] = None, context: Context = None, domain_id: int = get_default_domain_id()) -> None:
+def init(
+    *,
+    args: Optional[List[str]] = None,
+    context: Context = None,
+    domain_id: int = get_default_domain_id()
+) -> None:
     """
     Initialize ROS communications for a given context.
 
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -52,6 +52,7 @@ from rclpy.utilities import get_rmw_implementation_identifier  # noqa: F401
 from rclpy.utilities import ok  # noqa: F401 forwarding to this module
 from rclpy.utilities import shutdown as _shutdown
 from rclpy.utilities import try_shutdown  # noqa: F401
+from rclpy.utilities import get_default_domain_id
 
 # Avoid loading extensions on module import
 if TYPE_CHECKING:
@@ -59,7 +60,7 @@ if TYPE_CHECKING:
     from rclpy.node import Node  # noqa: F401
 
 
-def init(*, args: Optional[List[str]] = None, context: Context = None) -> None:
+def init(*, args: Optional[List[str]] = None, context: Context = None, domain_id: int = get_default_domain_id()) -> None:
     """
     Initialize ROS communications for a given context.
 
@@ -68,7 +69,7 @@ def init(*, args: Optional[List[str]] = None, context: Context = None) -> None:
         (see :func:`.get_default_context`).
     """
     context = get_default_context() if context is None else context
-    return context.init(args)
+    return context.init(args, domain_id=domain_id)
 
 
 # The global spin functions need an executor to do the work
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -61,7 +61,10 @@ class Context:
         from rclpy.impl.implementation_singleton import rclpy_implementation
         global g_logging_ref_count
         with self._handle as capsule, self._lock:
-            rclpy_implementation.rclpy_init(args if args is not None else sys.argv, capsule, domain_id)
+            rclpy_implementation.rclpy_init(
+                args if args is not None else sys.argv,
+                capsule,
+                domain_id)
             if initialize_logging and not self._logging_initialized:
                 with g_logging_configure_lock:
                     g_logging_ref_count += 1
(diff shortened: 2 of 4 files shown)


Barry-Xu-2018:
After checking failure log for  `Rpr__rclpy__ubuntu_focal_amd64 — Build finished.`, I find the error is the same as below.  
It seems that test environment has problem.

```
Traceback:
../../src/rclpy/rclpy/test/__init__.py:23: in <module>
    import rclpy.impl  # noqa
../../src/rclpy/rclpy/rclpy/__init__.py:47: in <module>
    from rclpy.context import Context
../../src/rclpy/rclpy/rclpy/context.py:21: in <module>
../../src/rclpy/rclpy/rclpy/impl/implementation_singleton.py:31: in <module>
    rclpy_implementation = _import('._rclpy')
../../src/rclpy/rclpy/rclpy/impl/__init__.py:28: in _import
/usr/lib/python3.8/importlib/__init__.py:127: in import_module
E   ModuleNotFoundError: No module named 'rclpy._rclpy'
E   The C extension '/tmp/ws/src/rclpy/rclpy/rclpy/_rclpy.cpython-38-x86_64-linux-gnu.so' isn't present on the system. Please refer to 'https://index.ros.org/doc/ros2/Troubleshooting/Installation-Troubleshooting/#import-failing-without-library-present-on-the-system' for possible solutions
```

rclpy/src/rclpy/_rclpy.c
```diff
@@ -5494,6 +5506,24 @@ rclpy_publisher_get_topic_name(PyObject * Py_UNUSED(self), PyObject * args)
...
 }
 
+
+/// Retrieves default domain id defined in rcl layer
+/**
+ *
+ * \return the defined domain id
+ */
+static PyObject *
+rclpy_get_default_domain_id(PyObject * Py_UNUSED(self), PyObject * args)
```
fujitatomoya, line 5516: i was expecting this would call https://github.com/ros2/rcl/blob/6f3c3fb10391edaacb2676e2345a754bef1b76f2/rcl/include/rcl/domain_id.h#L40-L42 to get default domain id from rcl.

rclpy/rclpy/context.py
```diff
@@ -46,7 +47,11 @@ def __init__(self):
     def handle(self):
         return self._handle
 
-    def init(self, args: Optional[List[str]] = None, *, initialize_logging: bool = True):
+    def init(self,
+             args: Optional[List[str]] = None,
+             *,
+             initialize_logging: bool = True,
+             domain_id: int = rclpy_implementation.rclpy_get_default_domain_id()):
```
fujitatomoya, line 54: why call implementation instead of calling utilities.get_default_domain_id?

rclpy/src/rclpy/_rclpy.c
Barry-Xu-2018: This function is called at  
https://github.com/Barry-Xu-2018/rclpy/blob/02d0bb6aa33b8e44a97ecac6cae4bbb8fce20f34/rclpy/rclpy/__init__.py#L63-L68. 
rcl_get_default_domain_id() may return error.  
rclpy_get_default_domain_id() directly return RCL_DEFAULT_DOMAIN_ID defined in rcl layer.  

Besides, this domain id (RCL_DEFAULT_DOMAIN_ID) will be set to init_options in rclpy_init() (_rclpy.c)  
In rcl_init(), if domain id equal to RCL_DEFAULT_DOMAIN_ID, rcl_get_default_domain_id() will be called.  
Refer to https://github.com/ros2/rcl/blob/6f3c3fb10391edaacb2676e2345a754bef1b76f2/rcl/src/rcl/init.c#L146-L154.   

So not directly call rcl_get_default_domain_id() here.

rclpy/rclpy/context.py
Barry-Xu-2018: It is because utilities.py has imported context.py.
https://github.com/Barry-Xu-2018/rclpy/blob/02d0bb6aa33b8e44a97ecac6cae4bbb8fce20f34/rclpy/rclpy/utilities.py#L22  

In order to avoid interdependence, have to call rclpy_implementation.

rclpy/src/rclpy/_rclpy.c
fujitatomoya: that makes sense.

rclpy/rclpy/utilities.py
```diff
@@ -39,6 +39,12 @@ def get_default_context(*, shutting_down=False):
         return g_default_context
 
 
+def get_default_domain_id():
```
fujitatomoya, line 42: maybe we could add test in https://github.com/ros2/rclpy/blob/master/rclpy/test/test_utilities.py.

rclpy/rclpy/__init__.py
```diff
@@ -52,14 +52,20 @@
...
+from rclpy.utilities import get_default_domain_id
 
 # Avoid loading extensions on module import
 if TYPE_CHECKING:
     from rclpy.executors import Executor  # noqa: F401
     from rclpy.node import Node  # noqa: F401
 
 
-def init(*, args: Optional[List[str]] = None, context: Context = None) -> None:
+def init(
```
fujitatomoya, line 63: Can we add test with specific domain_id in https://github.com/ros2/rclpy/blob/master/rclpy/test/test_init_shutdown.py?

Barry-Xu-2018: OK. I will supplement test codes.

rclpy/rclpy/utilities.py
Barry-Xu-2018: OK. I will supplement test codes.

Barry-Xu-2018:
> https://github.com/ros2/rclpy/pull/594#issuecomment-652859803

Sorry. It is my mistake.  
The failure is caused from my modification.  
I will fix it.

Barry-Xu-2018:
@fujitatomoya 

Sorry.  
Since the change is a bit big, I will close this PR and create new PR to review.

PR #596, opened 2020-07-08
Add ability to configure domain ID

Barry-Xu-2018:
Signed-off-by: Barry Xu <barry.xu@sony.com>

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -59,7 +59,12 @@ if TYPE_CHECKING:
     from rclpy.node import Node  # noqa: F401
 
 
-def init(*, args: Optional[List[str]] = None, context: Context = None) -> None:
+def init(
+    *,
+    args: Optional[List[str]] = None,
+    context: Context = None,
+    domain_id: Optional[int] = None
+) -> None:
     """
     Initialize ROS communications for a given context.
 
@@ -68,7 +73,7 @@ def init(*, args: Optional[List[str]] = None, context: Context = None) -> None:
         (see :func:`.get_default_context`).
     """
     context = get_default_context() if context is None else context
-    return context.init(args)
+    return context.init(args, domain_id=domain_id)
 
 
 # The global spin functions need an executor to do the work
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -64,7 +64,7 @@ class Context:
         else:
             if domain_id < 0:
                 raise RuntimeError(
-                    'Unexpected error: domain id ({}) should never be lower than zero.'
+                    'Domain id ({}) should never be lower than zero.'
                     .format(domain_id))
 
         global g_logging_ref_count
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -46,7 +46,11 @@ class Context:
     def handle(self):
         return self._handle
 
-    def init(self, args: Optional[List[str]] = None, *, initialize_logging: bool = True):
+    def init(self,
+             args: Optional[List[str]] = None,
+             *,
+             initialize_logging: bool = True,
+             domain_id: Optional[int] = None):
         """
         Initialize ROS communications for a given context.
 
@@ -54,9 +58,16 @@ class Context:
         """
         # imported locally to avoid loading extensions on module import
         from rclpy.impl.implementation_singleton import rclpy_implementation
+
+        if domain_id is None:
+            domain_id = rclpy_implementation.rclpy_get_default_domain_id()
+
         global g_logging_ref_count
         with self._handle as capsule, self._lock:
-            rclpy_implementation.rclpy_init(args if args is not None else sys.argv, capsule)
+            rclpy_implementation.rclpy_init(
+                args if args is not None else sys.argv,
+                capsule,
+                domain_id)
             if initialize_logging and not self._logging_initialized:
                 with g_logging_configure_lock:
                     g_logging_ref_count += 1
@@ -122,3 +133,9 @@ class Context:
                         raise RuntimeError(
                             'Unexpected error: logger ref count should never be lower that zero')
                 self._logging_initialized = False
+
+    def get_domain_id(self):
+        """Get domain id of context."""
+        from rclpy.impl.implementation_singleton import rclpy_implementation
+        with self._handle as capsule, self._lock:
+            return rclpy_implementation.rclpy_get_domain_id_in_context(capsule)
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -61,6 +61,11 @@ class Context:
 
         if domain_id is None:
             domain_id = rclpy_implementation.rclpy_get_default_domain_id()
+        else:
+            if domain_id < 0:
+                raise RuntimeError(
+                    'Unexpected error: domain id ({}) should never be lower than zero.'
+                    .format(domain_id))
 
         global g_logging_ref_count
         with self._handle as capsule, self._lock:
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -64,7 +64,7 @@ class Context:
         else:
             if domain_id < 0:
                 raise RuntimeError(
-                    'Domain id ({}) should never be lower than zero.'
+                    'Domain id ({}) should not be lower than zero.'
                     .format(domain_id))
 
         global g_logging_ref_count
@@ -143,4 +143,4 @@ class Context:
         """Get domain id of context."""
         from rclpy.impl.implementation_singleton import rclpy_implementation
         with self._handle as capsule, self._lock:
-            return rclpy_implementation.rclpy_get_domain_id_in_context(capsule)
+            return rclpy_implementation.rclpy_context_get_domain_id(capsule)
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -61,19 +61,14 @@ class Context:
 
         global g_logging_ref_count
         with self._handle as capsule, self._lock:
-            if domain_id is not None:
-                if domain_id < 0:
-                    raise RuntimeError(
-                        'Domain id ({}) should not be lower than zero.'
-                        .format(domain_id))
-                rclpy_implementation.rclpy_init(
-                    args if args is not None else sys.argv,
-                    capsule,
-                    domain_id)
-            else:
-                rclpy_implementation.rclpy_init(
-                    args if args is not None else sys.argv,
-                    capsule)
+            if domain_id is not None and domain_id < 0:
+                raise RuntimeError(
+                    'Domain id ({}) should not be lower than zero.'
+                    .format(domain_id))
+            rclpy_implementation.rclpy_init(
+                args if args is not None else sys.argv,
+                capsule,
+                domain_id)
             if initialize_logging and not self._logging_initialized:
                 with g_logging_configure_lock:
                     g_logging_ref_count += 1
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -59,20 +59,21 @@ class Context:
         # imported locally to avoid loading extensions on module import
         from rclpy.impl.implementation_singleton import rclpy_implementation
 
-        if domain_id is None:
-            domain_id = rclpy_implementation.rclpy_get_default_domain_id()
-        else:
-            if domain_id < 0:
-                raise RuntimeError(
-                    'Domain id ({}) should not be lower than zero.'
-                    .format(domain_id))
-
         global g_logging_ref_count
         with self._handle as capsule, self._lock:
-            rclpy_implementation.rclpy_init(
-                args if args is not None else sys.argv,
-                capsule,
-                domain_id)
+            if domain_id is not None:
+                if domain_id < 0:
+                    raise RuntimeError(
+                        'Domain id ({}) should not be lower than zero.'
+                        .format(domain_id))
+                rclpy_implementation.rclpy_init(
+                    args if args is not None else sys.argv,
+                    capsule,
+                    domain_id)
+            else:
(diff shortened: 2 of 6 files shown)


Barry-Xu-2018:
After checking error, I find below 2 functions doesn't exist  
- rcl_init_options_get_domain_id
- rcl_init_options_set_domain_id

`Rpr__rclpy__ubuntu_focal_amd64` environment use rcl package from http://repo.ros2.org/ubuntu/testing/pool/main/r/ros-foxy-rcl/

||||
|:--|:--|:--|
|ros-foxy-rcl-dbgsym_1.1.5-1focal.20200624.145908_amd64.deb | 2020-06-24 15:26 | 180K |  
| ros-foxy-rcl-dbgsym_1.1.5-1focal.20200625.081610_arm64.deb | 2020-06-25 08:28 | 177K |  
| ros-foxy-rcl_1.1.5-1focal.20200624.145908_amd64.deb | 2020-06-24 15:26 | 118K |  
| ros-foxy-rcl_1.1.5-1focal.20200625.081610_arm64.deb | 2020-06-25 08:28 | 113K |  
| ros-foxy-rcl_1.1.5-1focal.debian.tar.xz | 2020-06-04 03:48 | 8.2K |  
| ros-foxy-rcl_1.1.5-1focal.dsc | 2020-06-04 03:48 | 1.4K |  
| ros-foxy-rcl_1.1.5.orig.tar.gz | 2020-06-04 03:48 | 181K |  

So have to wait for rcl package updated

Barry-Xu-2018:
@ros-pull-request-builder

I find rcl package has been updated in http://repo.ros2.org/ubuntu/testing/pool/main/r/ros-foxy-rcl/.  
This change depends on rcl commit (6/30).   
I think current update rcl package has included dependent interfaces.   
Could you help to do retest ?

rclpy/rclpy/__init__.py
```diff
@@ -59,7 +59,12 @@
     from rclpy.node import Node  # noqa: F401
 
 
-def init(*, args: Optional[List[str]] = None, context: Context = None) -> None:
+def init(
+    *,
+    args: Optional[List[str]] = None,
+    context: Context = None,
+    domain_id: Optional[int] = None
```
fujitatomoya, line 66: domain_id should be checked if it is greater than or equal to zero?

Barry-Xu-2018: Yes. thanks.  
I supplement checking codes.

Barry-Xu-2018:
@fujitatomoya 

> domain_id should be checked if it is greater than or equal to zero?

Updated codes.  
Please review again.

rclpy/rclpy/context.py
```diff
@@ -61,6 +61,11 @@ def init(self,
 
         if domain_id is None:
             domain_id = rclpy_implementation.rclpy_get_default_domain_id()
+        else:
+            if domain_id < 0:
+                raise RuntimeError(
+                    'Unexpected error: domain id ({}) should never be lower than zero.'
```
fujitatomoya, line 67:
```suggestion
                    'domain id ({}) should never be lower than zero.'
```

I think that this is an `expected error`.

Barry-Xu-2018: I refer to template of below existing codes.  

https://github.com/ros2/rclpy/blob/9b26f5647d2ab7206e66c85de2dc9776b1f608b7/rclpy/rclpy/context.py#L121-L123

fujitatomoya: this case you pointed is actually `unexpected`, it expects the counter never goes less than zero. but with domain id, user application actually can specify the negative domain id.

Barry-Xu-2018: OK. Update it.

Barry-Xu-2018:
@ivanpauno 

Please help to review this commit.

rclpy/rclpy/context.py
```diff
@@ -46,17 +46,33 @@ def __init__(self):
...
         """
         # imported locally to avoid loading extensions on module import
         from rclpy.impl.implementation_singleton import rclpy_implementation
+
+        if domain_id is None:
+            domain_id = rclpy_implementation.rclpy_get_default_domain_id()
+        else:
+            if domain_id < 0:
+                raise RuntimeError(
+                    'Domain id ({}) should never be lower than zero.'
```
ivanpauno, line 67:
```suggestion
                    'Domain id ({}) should not be lower than zero.'
```

rclpy/src/rclpy/_rclpy.c
```diff
@@ -587,6 +590,15 @@ rclpy_init(PyObject * Py_UNUSED(self), PyObject * args)
     rcl_init_options_t init_options = rcl_get_zero_initialized_init_options();
     rcl_ret_t ret = rcl_init_options_init(&init_options, allocator);
     if (RCL_RET_OK == ret) {
+      // Set domain id
+      ret = rcl_init_options_set_domain_id(&init_options, (size_t)domain_id);
```
ivanpauno, line 594: Adding a check for `domain_id > 0` here too doesn't sound like a bad idea.
I always prefer doing the check in the closest place to the unsafe cast.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -5494,6 +5506,62 @@ rclpy_publisher_get_topic_name(PyObject * Py_UNUSED(self), PyObject * args)
...
+  return PyLong_FromSize_t(RCL_DEFAULT_DOMAIN_ID);
+}
+
+/// Retrieves domain id from init_options of context
+/**
+ * \param[in] pyargs context Python object
+ * \return domain id
+ */
+static PyObject *
+rclpy_get_domain_id_in_context(PyObject * Py_UNUSED(self), PyObject * args)
```
ivanpauno, line 5527: It would be more accurate to use here https://github.com/ros2/rcl/blob/70d13257a97f1278fd8e8a38f380e33b011f32e7/rcl/include/rcl/context.h#L267, which is the actual domain id used by the context.

rclpy/src/rclpy/_rclpy.c
Barry-Xu-2018: @ivanpauno 

Agree. But there is a limitation here.   
Please look at below codes

In context.py,  
```python
class Context:
  ...
  def init(...): 
     ...
        if domain_id is None:
            domain_id = rclpy_implementation.rclpy_get_default_domain_id() # return SIZE_MAX
        else:
            if domain_id < 0:
                raise RuntimeError(
                    'Domain id ({}) should never be lower than zero.'
                    .format(domain_id))
       ...
            rclpy_implementation.rclpy_init(
                args if args is not None else sys.argv,
                capsule,
                domain_id)
```

If user doesn't specify domain id, we will get default domain id which is defined as `SIZE_MAX`.   
In _rclpy.c, use `PyArg_ParseTuple()` to get parameter from python. For domain_id, if use 'L' (convert C long long), while domain_id is default value,  rclpy_implementation.rclpy_init() report error "OverflowError: int too big to convert".   
So we have to think domain_id  passing to `rclpy_init()` is unsigned long long.

ivanpauno: @Barry-Xu-2018 argh, good point.

I would prefer if `SIZE_MAX` in the C side is converted to `None` in the `Python` side.
`None` expresses intent better in this case.

- `rclpy_get_default_domain_id` won't be needed.
- `rclpy_init` will have to parse an optional long long ("OO|K" is the required format string). `domain_id` should be initialized to `SIZE_MAX` (`PyArg_ParseTuple` doesn't modify optional arguments when no specified).

@Barry-Xu-2018 does that make sense to you?

Barry-Xu-2018: @ivanpauno 

Thanks for sharing your thoughts.   
Agree. I will update codes.

Barry-Xu-2018:
@ivanpauno   

Please review again.

rclpy/rclpy/context.py
```diff
@@ -46,17 +46,34 @@ def __init__(self):
...
+                        'Domain id ({}) should not be lower than zero.'
+                        .format(domain_id))
+                rclpy_implementation.rclpy_init(
+                    args if args is not None else sys.argv,
+                    capsule,
+                    domain_id)
+            else:
+                rclpy_implementation.rclpy_init(
+                    args if args is not None else sys.argv,
+                    capsule)
```
ivanpauno, line 76: I think that the logic can be simplified a bit:
```suggestion
            if domain_id is not None and domain_id < 0:
                raise RuntimeError(
                    'Domain id ({}) should not be lower than zero.'
                    .format(domain_id))
            rclpy_implementation.rclpy_init(
                args if args is not None else sys.argv,
                capsule,
                domain_id)
```

rclpy/rclpy/utilities.py
```diff
@@ -39,6 +39,13 @@ def get_default_context(*, shutting_down=False):
         return g_default_context
 
 
+def get_domain_id_in_context(context=None):
```
ivanpauno, line 42: This utility function doesn't sound necessary.

Doing `get_default_context().get_domain_id()` or `my_context.get_domain_id()` look better than passing the context to a function.

rclpy/src/rclpy/_rclpy.c
ivanpauno: @Barry-Xu-2018 this comment still seems to be valid.

rclpy/rclpy/context.py
Barry-Xu-2018: @ivanpauno 

I tried this way. But while `domain_id` is None, code reports error as below.  
python thinks there is third argument. 

```shell
2: self = <rclpy.context.Context object at 0x7f7a2060eeb0>, args = None
2: 
2:     def init(self,
2:              args: Optional[List[str]] = None,
2:              *,
2:              initialize_logging: bool = True,
2:              domain_id: Optional[int] = None):
2:         """
2:         Initialize ROS communications for a given context.
2:     
2:         :param args: List of command line arguments.
2:         """
2:         # imported locally to avoid loading extensions on module import
2:         from rclpy.impl.implementation_singleton import rclpy_implementation
2:     
2:         global g_logging_ref_count
2:         with self._handle as capsule, self._lock:
2:             if domain_id is not None and domain_id < 0:
2:                 raise RuntimeError(
2:                     'Domain id ({}) should not be lower than zero.'
2:                     .format(domain_id))
2: >           rclpy_implementation.rclpy_init(
2:                 args if args is not None else sys.argv,
2:                 capsule,
2:                 domain_id)
2: E           TypeError: argument 3 must be int, not None
```

rclpy/rclpy/utilities.py
Barry-Xu-2018: @ivanpauno 

This utility function is only for context' get_domain_id(). 
```python
class Context:
    ...
    def get_domain_id(self):
    """Get domain id of context."""
         return rclpy_implementation.rclpy_context_get_domain_id(capsule)
```

This way likes context' ok().  
```python
class Context:
    ...
    def ok(self):
        """Check if context hasn't been shut down."""
            return rclpy_implementation.rclpy_ok(capsule)
```

rclpy/rclpy/context.py
Barry-Xu-2018: After google, I find below solution.  

https://stackoverflow.com/questions/42333354/how-to-permit-none-or-a-specific-type-as-an-argument-to-a-python-c-extension-fun.  

Do you want to implement like above way?

rclpy/rclpy/utilities.py
ivanpauno: Yeah, but we don't want to have all the methods of `Context` as free functions.

In the case of `ok`, we use that in every spinning loop. So having a free function was a good idea:

```python3
try:
    while rclpy.ok():
        ...
        rclpy.spin_once()
        ...
except KeyboardInterrupt:
    ...
```

In this case, a free function sounds unnecessary.

Barry-Xu-2018:
@ivanpauno 

> With #596 (comment) addressed and get_domain_id_in_context deleted I think this is ready to go.

Got it. Please look at commit `5a8c761`.

rclpy/test/test_init_shutdown.py
```diff
@@ -79,3 +79,16 @@ def test_create_node_without_init():
     context = rclpy.context.Context()
     with pytest.raises(NotInitializedException):
         rclpy.create_node('foo', context=context)
+
+
+def test_init_with_domain_id():
+    rclpy.init(domain_id=123)
+    rclpy.shutdown()
+    context = rclpy.context.Context()
+    rclpy.init(context=context, domain_id=123)
```
ivanpauno, line 88:
```suggestion
    rclpy.init(context=context, domain_id=123)
    assert context.get_domain_id() == 123
```

rclpy/test/test_init_shutdown.py
```diff
@@ -79,3 +79,16 @@ def test_create_node_without_init():
     context = rclpy.context.Context()
     with pytest.raises(NotInitializedException):
         rclpy.create_node('foo', context=context)
+
+
+def test_init_with_domain_id():
+    rclpy.init(domain_id=123)
```
ivanpauno, line 85:
```suggestion
    rclpy.init(domain_id=123)
    assert rclpy.get_default_context().get_domain_id() == 123
```

Barry-Xu-2018:
@ivanpauno  

Update test codes at commit `43c42c3`.

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11742/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6822/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9579/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11680/)

ivanpauno:
Thanks @Barry-Xu-2018 !

PR #590, opened 2020-06-30
[backport dashing] fix moved troubleshooting url

dirk-thomas:
Backport of #579.

PR #597, opened 2020-07-10
[Dashing] Wrap lines to shorten line length.

nuclearsandwich:
Backports #586 but does not include the f-string change since they are not available in Dashing's minimum supported Python version, 3.5.

### DIFF
--- a/rclpy/rclpy/impl/__init__.py
+++ b/rclpy/rclpy/impl/__init__.py
@@ -34,8 +34,9 @@ def _import(name):
                 'possible solutions' % expected_path
         if e.path is not None and os.path.isfile(e.path):
             e.msg += \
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The second commit in #590 already made sure that the code passed the linter?

nuclearsandwich:
> The second commit in #590 already made sure that the code passed the linter?

Does that mean this PR is spurious and I can drop 586 from the board of things that need backporting?

dirk-thomas:
Assuming that CI for Dashing already passes I would think this backport is not necessary anymore.

nuclearsandwich:
http://build.ros2.org/view/Ddev/job/Ddev__rclpy__ubuntu_bionic_amd64/70/ is happy so I am happy.

Issue #588, opened 2020-06-23
Regression on expected number of event handlers in test_qos_event


azeey:
Likely a regression from https://github.com/ros2/rclpy/pull/572.

Failing build: http://build.ros2.org/view/Fci/job/Fci__nightly-cyclonedds_ubuntu_focal_amd64/145/testReport/junit/rclpy.src.ros2.rclpy.rclpy.test.test_qos_event/TestQoSEvent/test_subscription_constructor/

Output:
```
self = <test.test_qos_event.TestQoSEvent testMethod=test_subscription_constructor>

    def test_subscription_constructor(self):
        callbacks = SubscriptionEventCallbacks()
        liveliness_callback = Mock()
        deadline_callback = Mock()
        message_callback = Mock()
        incompatible_qos_callback = Mock()
        expected_num_event_handlers = 0 if self.is_fastrtps else 1
    
        # No arg
        subscription = self.node.create_subscription(
            EmptyMsg, self.topic_name, message_callback, 10)
        self.assertEqual(len(subscription.event_handlers), expected_num_event_handlers)
        self.node.destroy_subscription(subscription)
    
        # Arg with no callbacks
        subscription = self.node.create_subscription(
            EmptyMsg, self.topic_name, message_callback, 10, event_callbacks=callbacks)
        self.assertEqual(len(subscription.event_handlers), expected_num_event_handlers)
        self.node.destroy_subscription(subscription)
    
        # Arg with one of the callbacks
        callbacks.deadline = deadline_callback
        expected_num_event_handlers += 1
        subscription = self.node.create_subscription(
            EmptyMsg, self.topic_name, message_callback, 10, event_callbacks=callbacks)
        self.assertEqual(len(subscription.event_handlers), expected_num_event_handlers)
        self.node.destroy_subscription(subscription)
    
        # Arg with two callbacks
        callbacks.liveliness = liveliness_callback
        expected_num_event_handlers += 1
        subscription = self.node.create_subscription(
            EmptyMsg, self.topic_name, message_callback, 10, event_callbacks=callbacks)
        self.assertEqual(len(subscription.event_handlers), expected_num_event_handlers)
        self.node.destroy_subscription(subscription)
    
        # Arg with three callbacks
        callbacks.incompatible_qos = incompatible_qos_callback
        try:
            subscription = self.node.create_subscription(
                EmptyMsg, self.topic_name, message_callback, 10, event_callbacks=callbacks)
[...]

PR #601, opened 2020-07-15
Default incompatible qos callback should be set when there's no user specified callback

ivanpauno:
It's currently being set when there's no user specified "message lost" callback, which doesn't make sense.
The regression was introduced in https://github.com/ros2/rclpy/pull/572.

Partially addresses https://github.com/ros2/rclpy/issues/588.
There's another failure in the same test file, but not related to this one.

### DIFF
--- a/rclpy/rclpy/qos_event.py
+++ b/rclpy/rclpy/qos_event.py
@@ -242,20 +242,6 @@ class SubscriptionEventCallbacks:
                 callback=self.incompatible_qos,
                 event_type=QoSSubscriptionEventType.RCL_SUBSCRIPTION_REQUESTED_INCOMPATIBLE_QOS,
                 parent_handle=subscription_handle))
-
-        if self.liveliness:
-            event_handlers.append(QoSEventHandler(
-                callback_group=callback_group,
-                callback=self.liveliness,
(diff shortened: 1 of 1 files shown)


rclpy/rclpy/qos_event.py
```diff
@@ -242,20 +242,6 @@ def create_event_handlers(
...
-                event_type=QoSSubscriptionEventType.RCL_SUBSCRIPTION_LIVELINESS_CHANGED,
-                parent_handle=subscription_handle))
-
-        if self.message_lost:
-            event_handlers.append(QoSEventHandler(
-                callback_group=callback_group,
-                callback=self.message_lost,
-                event_type=QoSSubscriptionEventType.RCL_SUBSCRIPTION_MESSAGE_LOST,
-                parent_handle=subscription_handle))
         elif self.use_default_callbacks:
```
ivanpauno, line 245: This diff is hard to read.
The `elif` block is applying to a different `if` now.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11478/)

ivanpauno:
Failures are matching the ones in http://build.ros2.org/view/Fci/job/Fci__nightly-cyclonedds_ubuntu_focal_amd64/162, and unrelated to this PR.
https://github.com/ros2/rclpy/pull/603 will complete the fix.

PR #603, opened 2020-07-15
Destroy event handlers owned by publishers/subscriptions when calling publisher.destroy()/subscription.destroy()

ivanpauno:
The other half of https://github.com/ros2/rclpy/pull/601 to completely fix https://github.com/ros2/rclpy/issues/588.

The assumption in comment https://github.com/ros2/rclpy/blob/be1deb6746ae260a88c47aa2399f3fd6f5bfddc9/rclpy/test/test_qos_event.py#L271-L272 was wrong because some of the objects created in the previous test cases were still around.

### DIFF
--- a/rclpy/test/test_qos_event.py
+++ b/rclpy/test/test_qos_event.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import gc
 import unittest
 from unittest.mock import Mock
 
@@ -53,6 +54,9 @@ class TestQoSEvent(unittest.TestCase):
     def tearDown(self):
         self.node.destroy_node()
         rclpy.shutdown(context=self.context)
+        del self.context
+        del self.node
+        gc.collect()
 
     def test_publisher_constructor(self):
         callbacks = PublisherEventCallbacks()
--- a/rclpy/test/test_qos_event.py
+++ b/rclpy/test/test_qos_event.py
@@ -54,9 +54,6 @@ class TestQoSEvent(unittest.TestCase):
     def tearDown(self):
         self.node.destroy_node()
         rclpy.shutdown(context=self.context)
-        del self.node
-        del self.context
-        gc.collect()
 
     def test_publisher_constructor(self):
         callbacks = PublisherEventCallbacks()
--- a/rclpy/test/test_qos_event.py
+++ b/rclpy/test/test_qos_event.py
@@ -54,8 +54,8 @@ class TestQoSEvent(unittest.TestCase):
     def tearDown(self):
         self.node.destroy_node()
         rclpy.shutdown(context=self.context)
-        del self.context
         del self.node
+        del self.context
         gc.collect()
 
     def test_publisher_constructor(self):
--- a/rclpy/test/test_qos_event.py
+++ b/rclpy/test/test_qos_event.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import gc
 import unittest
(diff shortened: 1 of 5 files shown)


ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11480/)

rclpy/test/test_qos_event.py
```diff
@@ -53,6 +54,9 @@ def setUp(self):
     def tearDown(self):
         self.node.destroy_node()
         rclpy.shutdown(context=self.context)
+        del self.context
+        del self.node
+        gc.collect()
```
clalancette, line 59: I see what you are doing here, but I'm not entirely sure why.  You are hand-destroying both objects, and then making sure they are collected so that the destructors run (and presumably release the underlying `rcl` resources).

But in the case of the node, shouldn't call `self.node.destroy_node()` above be enough?  If not, why not?  As for the context, maybe we should add a `destroy_context()` method as well?  Neither is very pythonic, but it at least gives users a deterministic way to destroy the objects, rather than forcing gc runs (which may have other consequences).

ivanpauno: > But in the case of the node, shouldn't calling self.node.destroy_node() above be enough? If not, why not?

I think it should be enough, but if I delete the `del self.node` line it stops working.
I will investigate a bit more to find the reason.

ivanpauno: See https://github.com/ros2/rclpy/pull/603/commits/ef3b7851ded2eea6f8f41711af2e385c78a7bd2d

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11486/)

ivanpauno:
> Change seems reasonable to me, but the test failures seem odd. At least, I don't see test_waitable_with_timer failing in the nightly Linux. @ivanpauno any thoughts?

That test is extremely unrelated, so it's maybe flaky.

I'm testing here only `rmw_cyclonedds` which is not the usual nightly conditions (the reported bug #588 didn't happen in nightlies).
I cannot find a similar failure in the only cyclonedds job in buld.ros2.org, so I will investigate a bit more before merging.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11496/) (including fastrtps)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11495/)  (only cyclone)

ivanpauno:

ivanpauno:
> That test is extremely unrelated, so it's maybe flaky.

It seems to be related actually :confused:

ivanpauno:
Ok, I think I've got it right now:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11497/)  (only cyclonedds)

clalancette:
This is looking good.  However, there is enough change in here that I'll suggest a more thorough test before merging.  I think we should run CI on all platforms and test with `--packages-above rclpy` just to make sure there isn't something else lurking out there.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11499/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6689/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9423/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11382/)

ivanpauno:
@clalancette the failing windows build is due to a bug in rosbag (nightlies did fail too).
I think this is safe to merge now, it's highly unlikely that this behavior is platform dependent.

Issue #585, opened 2020-06-19
`MultiThreadedExecutor:spin_until_future_complete` can block when the future is ready


craigh92:
## Bug report

**Required Info:**

- Operating System:
  -  Ubuntu 20.04
- Installation type:
  - From source
- Version or commit hash:
  - Foxy
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclpy

#### Steps to reproduce the issue

 - Publish a message from one node (with latching_qos)
 - In another node, subscribe to the topic with a callback that sets the result of the future
 - Start both nodes in a `MuliThreadedExecutor`
 - Spin until future complete

```python
import rclpy
from rclpy.executors import MultiThreadedExecutor, SingleThreadedExecutor
from rclpy.node import Node
from std_msgs.msg import String
from rclpy.qos import QoSProfile, QoSDurabilityPolicy
from rclpy.task import Future
import os

latching_qos = QoSProfile(depth=1,
    durability=QoSDurabilityPolicy.RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL)

def main():

    rclpy.init()
    
    # Set up publisher
    pubnode = Node('pubnode_' + str(os.getpid()))
    pub1 = pubnode.create_publisher(String, 'topic1', latching_qos)
    msg1 = String()
    msg1.data = "hello1"
    pubnode.get_logger().info("Publishing hello1")
    pub1.publish(msg1)

    # Set up listener
    future_msgs = Future()
    subnode = Node('subnode_' + str(os.getpid()))
    subnode.create_subscription(String, 'topic1', lambda msg : ([
            subnode.get_logger().info("Received message on topic1"),
            future_msgs.set_result(msg)
    ]), latching_qos)

    # Start nodes
    exe = MultiThreadedExecutor()
    exe.add_node(pubnode)
    exe.add_node(subnode)

    future_msgs.add_done_callback(lambda fut : print("Future is done"))
    exe.spin_until_future_complete(future_msgs)

if __name__ == '__main__':
    main()
```

#### Expected behavior
The subnode should receive the message, set the future as complete, and then the program should exit.

#### Actual behavior
The subnode receives the message, sets the future as complete, but the `exe.spin_until_future_complete(future_msgs)` never returns.

#### Additional information

This only happens with the `MultiThreadedExector`. If I swap this out for the `SingleThreadedExecutor` then it works as expected.

If the pub node is running in a different process, it also works as expected.

[...]

PR #605, opened 2020-07-20
MultiThreadedExecutor spin_until_future complete should not continue waiting when the future is done

ivanpauno:
Fixes https://github.com/ros2/rclpy/issues/585.

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -324,7 +324,7 @@ class Executor:
         Wait for and execute a single callback.
 
         This should behave in the same way as :meth:`spin_once`.
-        If needed by the implementation, it should be awake other threads waiting.
+        If needed by the implementation, it should awake other threads waiting.
 
         :param future: The executor will wait until this future is done.
         :param timeout_sec: Maximum seconds to wait. Block forever if ``None`` or negative.
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -735,7 +735,7 @@ class MultiThreadedExecutor(Executor):
     def _spin_once_impl(
         self,
         timeout_sec: float = None,
-        wait_condition: Callable[[], bool] = lambda: True
+        wait_condition: Callable[[], bool] = lambda: False
     ) -> None:
         try:
             handler, entity, node = self.wait_for_ready_callbacks(
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -287,14 +287,14 @@ class Executor:
         """Execute callbacks until a given future is done or a timeout occurs."""
         if timeout_sec is None or timeout_sec < 0:
             while self._context.ok() and not future.done() and not self._is_shutdown:
-                self.spin_once(timeout_sec=timeout_sec)
+                self.spin_once_until_future_complete(future, timeout_sec)
         else:
             start = time.monotonic()
             end = start + timeout_sec
             timeout_left = timeout_sec
 
             while self._context.ok() and not future.done() and not self._is_shutdown:
-                self.spin_once(timeout_sec=timeout_left)
+                self.spin_once_until_future_complete(future, timeout_left)
                 now = time.monotonic()
 
                 if now >= end:
@@ -309,6 +309,19 @@ class Executor:
         A custom executor should use :meth:`wait_for_ready_callbacks` to get work.
 
         :param timeout_sec: Seconds to wait. Block forever if ``None`` or negative.
+            Don't wait if 0.
+        """
+        raise NotImplementedError()
+
+    def spin_once_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
+        """
+        Wait for and execute a single callback.
+
+        This should behave in the same way as :meth:`spin_once`.
+        If needed by the implementation, it should be awake other threads waiting.
+
+        :param future: The executor will wait until this future is done.
+        :param timeout_sec: Maximum seconds to wait. Block forever if ``None`` or negative.
             Don't wait if 0.
         """
         raise NotImplementedError()
@@ -686,6 +699,9 @@ class SingleThreadedExecutor(Executor):
             if handler.exception() is not None:
                 raise handler.exception()
 
+    def spin_once_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
+        self.spin_once(timeout_sec)
+
 
 class MultiThreadedExecutor(Executor):
     """
@@ -717,3 +733,8 @@ class MultiThreadedExecutor(Executor):
             pass
         else:
             self._executor.submit(handler)
+
+    def spin_once_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
+        self.spin_once(timeout_sec)
+        if future.done():
+            self.wake()
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -755,4 +755,5 @@ class MultiThreadedExecutor(Executor):
         self._spin_once_impl(timeout_sec)
 
     def spin_once_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
+        future.add_done_callback(lambda x: self.wake())
         self._spin_once_impl(timeout_sec, future.done)
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -460,7 +460,8 @@ class Executor:
         :param timeout_sec: Seconds to wait. Block forever if ``None`` or negative.
             Don't wait if 0.
         :param nodes: A list of nodes to wait on. Wait on all nodes if ``None``.
-        :param condition: A callable that makes the function return immediately when it evaluates to True.
+        :param condition: A callable that makes the function return immediately when it evaluates
+            to True.
         """
         timeout_timer = None
         timeout_nsec = timeout_sec_to_nsec(timeout_sec)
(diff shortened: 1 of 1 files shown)


rclpy/rclpy/executors.py
```diff
@@ -717,3 +733,8 @@ def spin_once(self, timeout_sec: float = None) -> None:
             pass
         else:
             self._executor.submit(handler)
+
+    def spin_once_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
+        self.spin_once(timeout_sec)
+        if future.done():
```
fujitatomoya, line 739: future might not be done yet in child thread, then main thread will call in this function, wait on `rcl_wait`.

ivanpauno:
My original approach was incorrect as `executor.wake` should be called when the future is done.

I was going to simply add a done callback to the future that would call `executor.wake()`, but that didn't work because the executor isn't waken when a guard condition is triggered.
@sloretz was that intentional? IMO, `wait_for_ready_callbacks` should return when a guard condition was triggered, and not continue waiting for work.

For the moment, I have fixed the issue by adding an artificial "condition" argument to `wait_for_ready_callbacks`.

hidmic:
> I was going to simply add a done callback to the future that would call executor.wake(), but that didn't work because the executor isn't waken when a guard condition is triggered.
@sloretz was that intentional? IMO, wait_for_ready_callbacks should return when a guard condition was triggered, and not continue waiting for work.

I second this question. To trigger a guard condition seems like the "natural" solution for this problem.

rclpy/rclpy/executors.py
```diff
@@ -429,7 +448,8 @@ def can_execute(self, entity: WaitableEntityType) -> bool:
     def _wait_for_ready_callbacks(
         self,
         timeout_sec: float = None,
-        nodes: List['Node'] = None
+        nodes: List['Node'] = None,
+        condition: Callable[[], bool] = lambda: True,
```
fujitatomoya, line 452:
```suggestion
        condition: Callable[[], bool] = lambda: False,
```

unless the following error will happen with `SingleThreadedExecutor`

```
ros2 run my_package my_node --exe=S
[INFO] [1595381337.338499864] [pubnode_22677]: Publishing hello1: Thread MainThread
Using <class 'rclpy.executors.SingleThreadedExecutor'>
  File "/root/ros2_ws/tmp_ws/install/my_package/lib/my_package/my_node", line 11, in <module>
    load_entry_point('my-package', 'console_scripts', 'my_node')()
  File "/root/ros2_ws/tmp_ws/build/my_package/my_package/my_node.py", line 70, in main
    exe.spin_until_future_complete(future_msgs)
  File "/root/ros2_ws/tmp_ws/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 296, in spin_until_future_complete
    self.spin_once_until_future_complete(future, timeout_sec)
  File "/root/ros2_ws/tmp_ws/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 713, in spin_once_until_future_complete
    self.spin_once(timeout_sec)
  File "/root/ros2_ws/tmp_ws/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 702, in spin_once
  File "/root/ros2_ws/tmp_ws/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 688, in wait_for_ready_callbacks
    return next(self._cb_iter)
  File "/root/ros2_ws/tmp_ws/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 668, in _wait_for_ready_callbacks
    raise ConditionReachedException()
rclpy.executors.ConditionReachedException
```

rclpy/rclpy/executors.py
```diff
@@ -706,14 +732,27 @@ def __init__(self, num_threads: int = None, *, context: Context = None) -> None:
                 num_threads = 1
         self._executor = ThreadPoolExecutor(num_threads)
 
-    def spin_once(self, timeout_sec: float = None) -> None:
+    def _spin_once_impl(
+        self,
+        timeout_sec: float = None,
+        wait_condition: Callable[[], bool] = lambda: True
```
fujitatomoya, line 738:
```suggestion
        wait_condition: Callable[[], bool] = lambda: False
```

rclpy/rclpy/executors.py
ivanpauno: Thanks!
I inverted the logic before committing, and I didn't realize that I forgot to change the `condition` argument default.

sloretz:
> I was going to simply add a done callback to the future that would call executor.wake(), but that didn't work because the executor isn't waken when a guard condition is triggered.
@sloretz was that intentional? IMO, wait_for_ready_callbacks should return when a guard condition was triggered, and not continue waiting for work.

It's intentional. [User created guard conditions count as work](https://github.com/ros2/rclpy/blob/75c51c945d6029bea61571ced43c1d1b9c3c6703/rclpy/rclpy/executors.py#L613-L620), while the internal guard condition does not count as work. `Executor.wake()` tells the executor to wake up so it can reexamine the entities it's waiting for work on. For example, [this tells the executor to wake once when the data was taken from the rmw layer](https://github.com/ros2/rclpy/blob/75c51c945d6029bea61571ced43c1d1b9c3c6703/rclpy/rclpy/executors.py#L404) so the same entity can be added back to the wait list, but no actual work has been done.

rclpy/rclpy/executors.py
```diff
@@ -440,14 +460,15 @@ def _wait_for_ready_callbacks(
...
+        :param condition: A callable that makes the function return immedaitely when it's True.
         """
         timeout_timer = None
         timeout_nsec = timeout_sec_to_nsec(timeout_sec)
         if timeout_nsec > 0:
             timeout_timer = Timer(None, None, timeout_nsec, self._clock, context=self._context)
 
         yielded_work = False
-        while not yielded_work and not self._is_shutdown:
+        while not yielded_work and not self._is_shutdown and not condition():
```
sloretz, line 471: I guess this solves the problem. If the subscription callback were being executed in another thread, and that thread calls `future.done()` just after this loop is entered, it would still be ok because the callback would trigger the executor's guard condition after completing and that would cause this loop to re-run.

ivanpauno:
> It's intentional. [User created guard conditions count as work](https://github.com/ros2/rclpy/blob/75c51c945d6029bea61571ced43c1d1b9c3c6703/rclpy/rclpy/executors.py#L613-L620), while the internal guard condition does not count as work.

Thanks for clarifying, I didn't realize of the difference.

>  `Executor.wake()` tells the executor to wake up so it can reexamine the entities it's waiting for work on. For example, [this tells the executor to wake once when the data was taken from the rmw layer](https://github.com/ros2/rclpy/blob/75c51c945d6029bea61571ced43c1d1b9c3c6703/rclpy/rclpy/executors.py#L404) so the same entity can be added back to the wait list, but no actual work has been done.

That makes sense, but I wouldn't call the method `wake`.
`wake` sounds like the executor is actually waking up (stopping to spin).
I would use something like `refresh_entities`, or similar.

Should we have both alternatives?
I think that with one internal guard condition and an extra flag we can achieve both behaviors.

rclpy/rclpy/executors.py
ivanpauno: > I guess this solves the problem. If the subscription callback were being executed in another thread, and that thread calls `future.done()` just after this loop is entered, it would still be ok because the callback would trigger the executor's guard condition after completing and that would cause this loop to re-run.

That was the idea.
At some point I deleted the callback, I've re-added it in https://github.com/ros2/rclpy/pull/605/commits/7864a7b9fd1f8f63ff5d2c623a774246bbfb54b7.

ivanpauno:
@sloretz friendly ping

rclpy/rclpy/executors.py
```diff
@@ -313,6 +319,19 @@ def spin_once(self, timeout_sec: float = None) -> None:
         """
         raise NotImplementedError()
 
+    def spin_once_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
+        """
+        Wait for and execute a single callback.
+
+        This should behave in the same way as :meth:`spin_once`.
+        If needed by the implementation, it should be awake other threads waiting.
```
hidmic, line 327: @ivanpauno nit: 
```suggestion
        If needed by the implementation, it should awake other threads waiting.
```

rclpy/rclpy/executors.py
```diff
@@ -440,14 +460,15 @@ def _wait_for_ready_callbacks(
         :param timeout_sec: Seconds to wait. Block forever if ``None`` or negative.
             Don't wait if 0.
         :param nodes: A list of nodes to wait on. Wait on all nodes if ``None``.
+        :param condition: A callable that makes the function return immedaitely when it's True.
```
hidmic, line 463: @ivanpauno consider: 
```suggestion
        :param condition: A callable that makes the function return immediately when it evaluates to True.
```

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11844/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6905/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9659/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11767/)

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11882/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6927/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9679/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11818/)

ivanpauno:
@sloretz let me know if you have any comments about this https://github.com/ros2/rclpy/pull/605#issuecomment-664392934, maybe we can simplify this in a follow up.

nuclearsandwich:
I don't believe any backports have been opened for this PR yet so I am bumping it from patch release 8 for Dashing.

PR #606, opened 2020-07-21
MultiThreadedExecutor spin_until_future complete should return once future is done

fujitatomoya:
proposal fix for https://github.com/ros2/rclpy/issues/585


fujitatomoya:
@ivanpauno 

what about using single thread for `spin_until_future_complete()`? I think that this fix aligned with [rclcpp](https://github.com/ros2/rclcpp/blob/a0376768f7d077d7543bb9a2cdd9d2a72b495538/rclcpp/include/rclcpp/executor.hpp#L239). at least this can fix the problem.

ivanpauno:
I think that trying the multithreaded approach is better, it's definitely possible to fix the current error.

ivanpauno:
Closing, as the idea is to make actual usage of the multithreaded capability in `spin_until_future_complete`.

PR #612, opened 2020-08-03
Use absolute parameter events topic name

ivanpauno:
Follow up of https://github.com/ros2/rclcpp/pull/929.

PR #613, opened 2020-08-03
Use global namespace for parameter events topic

jacobperron:
Similar to https://github.com/ros2/rclcpp/pull/929.

As a side note, it appears we've implemented a parameter service in rclpy, but not a parameter client. I couldn't find any ticket tracking the addition of a parameter client, but perhaps it's something to consider adding. It might be a good ticket for "help wanted". @sloretz what do you think?


jacobperron:
Duplicates https://github.com/ros2/rclpy/pull/612

PR #545, opened 2020-04-23
Feature/services timestamps

iluetkeb:
Connects to https://github.com/ros2/rmw/pull/217

PR #615, opened 2020-08-10
skip test relying on source timestamps with Connext

dirk-thomas:
Skip test with Connext which have been failing since they have been added:
* ros2/rclpy#545: http://build.ros2.org/view/Rci/job/Rci__nightly-connext_ubuntu_focal_amd64/42/testReport/rclpy.rclpy.test.test_client/TestClient/test_service_timestamps/
* ros2/rclpy#542: http://build.ros2.org/view/Rci/job/Rci__nightly-connext_ubuntu_focal_amd64/42/testReport/rclpy.rclpy.test.test_node/TestNodeAllowUndeclaredParameters/test_take/

The missing feature is ticketed in ros2/rmw_connext#405.

Before: [Build Status](https://ci.ros2.org/job/ci_linux/11858/)
After: [Build Status](https://ci.ros2.org/job/ci_linux/11861/)

### DIFF
--- a/rclpy/test/test_client.py
+++ b/rclpy/test/test_client.py
@@ -19,6 +19,7 @@ from rcl_interfaces.srv import GetParameters
 import rclpy
 import rclpy.executors
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+from rclpy.utilities import get_rmw_implementation_identifier
 
 
 # TODO(sloretz) Reduce fudge once wait_for_service uses node graph events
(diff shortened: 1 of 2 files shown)


gbiggs:
Is there an alternative test for `_rclpy.rclpy_take()` available?

dirk-thomas:
> Is there an alternative test for _rclpy.rclpy_take() available?

I don't understand the question - what for?

gbiggs:
I'm probably misunderstanding something, but it looks like you're skipping the only tests here that call `rclpy_take()` for Connext. I was asking if that function is tested with Connext somewhere else or if it's just left untested with Connext. If the function relies on source timestamps and is completely unusable with Connext then that's fine, but it's probably worth being more specific about that in the skip comment.

iluetkeb:
One test is timestamp-specific (`get_service_timestamps`), but the other (`test_take`) isn't. In the C++ tests I've #ifdef'd out the check blocks only. Not sure if sth like that is possible for pyunit?

dirk-thomas:
> I'm probably misunderstanding something, but it looks like you're skipping the only tests here that call `rclpy_take()` for Connext. I was asking if that function is tested with Connext somewhere else or if it's just left untested with Connext.

The specific test was added in the above referenced PR and has failed for Connext since then. This patch only skips the always failing test.

Adding a separate test to check `rclpy_take()` without considering the source timestamp would be an option as well as modifying the test to skip part of the logic for Connext.

Since this change doesn't reduce the previously covered (and passed) code I don't think either should happen in this PR.

PR #620, opened 2020-09-13
fix _get_parameter_types_callback Enum to ParameterType conversion

### LINKED ISSUE #618: Bug: _get_parameter_types_callback returns Enum type rather than required int type
In `parameter_services.py:88` (https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/parameter_service.py#L88) there needs to be an additional `.value` at the end of `node.get_parameter_or(name).type_`. This is because `type_` returns an Enum value, but rather what we want is the integer value type from https://github.com/ros2/rcl_interfaces/blob/master/rcl_interfaces/msg/ParameterType.msg.

See https://stackoverflow.com/questions/24487405/enum-getting-value-of-enum-on-string-conversion.

nyxaria:
As discussed in #618, this fixes the conversion from Python Enum type to the `int` type in `ParameterType.msg`.

Before:
```
...
"parameter_service.py", line 90, in _get_parameter_types_callback
response.types.append(node.get_parameter_or(name).type_)
TypeError: an integer is required (got type Type)
```

Fixes #618

### DIFF
--- a/rclpy/rclpy/parameter_service.py
+++ b/rclpy/rclpy/parameter_service.py
@@ -85,7 +85,7 @@ class ParameterService:
     def _get_parameter_types_callback(self, request, response):
         node = self._get_node()
         for name in request.names:
-            response.types.append(node.get_parameter_or(name).type_)
+            response.types.append(node.get_parameter_or(name).type_.value)
         return response
 
     def _list_parameters_callback(self, request, response):


clalancette (changes requested):
Can you please add a test (either to [test_parameter.py](https://github.com/ros2/rclpy/blob/master/rclpy/test/test_parameter.py) or [test_parameters_callback.py](https://github.com/ros2/rclpy/blob/master/rclpy/test/test_parameters_callback.py)) to ensure that this is doing what we expect?  Thanks.

nyxaria:
I will try to do this in the following week, moving houses at the moment and swamped with work so I have no time to fix this now. Apologies for the delays.

clalancette:
@nyxaria No problem, thanks for the heads up.

clalancette:
@nyxaria Hope your house moving went well.  Friendly ping to add a test here, thanks.

sloretz:
Oops, it looks like someone opened a duplicate PR and it got merged (#641). Thanks for the PR!  Closing.

PR #621, opened 2020-09-18
Add in the topic name when QoS events are fired.

clalancette:
This makes debugging for the user a little easier.

This is the rclpy counterpart to https://github.com/ros2/rclcpp/pull/1286 .  @codebot FYI.

### DIFF
--- a/rclpy/rclpy/qos_event.py
+++ b/rclpy/rclpy/qos_event.py
@@ -225,7 +225,7 @@ class SubscriptionEventCallbacks:
         self.use_default_callbacks = use_default_callbacks
 
     def create_event_handlers(
-        self, callback_group: CallbackGroup, subscription_handle: Handle,
+        self, callback_group: CallbackGroup, subscription_handle: Handle, topic_name: str,
     ) -> List[QoSEventHandler]:
         with subscription_handle as subscription_capsule:
             logger = get_logger(_rclpy.rclpy_get_subscription_logger_name(subscription_capsule))
@@ -250,9 +250,9 @@ class SubscriptionEventCallbacks:
                 def _default_incompatible_qos_callback(event):
                     policy_name = qos_policy_name_from_kind(event.last_policy_kind)
                     logger.warn(
-                        'New publisher discovered on this topic, offering incompatible QoS. '
+                        "New publisher discovered on topic '{}', offering incompatible QoS. "
                         'No messages will be received from it. '
-                        'Last incompatible policy: {}'.format(policy_name))
+                        'Last incompatible policy: {}'.format(topic_name, policy_name))
 
                 event_type = QoSSubscriptionEventType.RCL_SUBSCRIPTION_REQUESTED_INCOMPATIBLE_QOS
                 event_handlers.append(QoSEventHandler(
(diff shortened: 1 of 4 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12355/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7324/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10067/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12268/)

fujitatomoya (approved):
lgtm

audrow (approved):
Looks good to me!

rclpy/rclpy/qos_event.py
```diff
@@ -225,7 +225,7 @@ def __init__(
         self.use_default_callbacks = use_default_callbacks
 
     def create_event_handlers(
-        self, callback_group: CallbackGroup, subscription_handle: Handle,
+        self, callback_group: CallbackGroup, subscription_handle: Handle, topic_name: str,
```
ivanpauno, line 228: I think it's fine to break API, because this is mostly used internally.
It seems weird to add the `topic_name` as an argument only because one of the "default generated" callbacks need it.

Maybe `create_event_handlers` could take a `rclpy.subscription.Subscription` instead of both the `topic_name` and the `subscription_handle`. That seems a bit more flexible if we want to query something else from the `Subscription` in the future.

ivanpauno (commented):
I have a minor comment, otherwise lgtm

rclpy/rclpy/qos_event.py
clalancette: > I think it's fine to break API, because this is mostly used internally.

Yeah, that was my thinking.

> Maybe `create_event_handlers` could take a `rclpy.subscription.Subscription` instead of both the `topic_name` and the `subscription_handle`. That seems a bit more flexible if we want to query something else from the `Subscription` in the future.

The issue there is that we create a circular dependency between subscription.py and qos_event.py.  subscription.py needs to import qos_event.py to do things, and then passing a `subscription.Subscription` into a `SubscriptionEventHandler` means that qos_event.py needs a dependency on subscription.py to get the subscription type.  I think Python will figure this out, and it may even be possible to only have this dependency at type-checking time, but I'd rather not have the circular dependency at all.  Other thoughts welcome, though.

ivanpauno: > The issue there is that we create a circular dependency between subscription.py and qos_event.py. subscription.py needs to import qos_event.py to do things, and then passing a subscription.Subscription into a SubscriptionEventHandler means that qos_event.py needs a dependency on subscription.py to get the subscription type. I think Python will figure this out, and it may even be possible to only have this dependency at type-checking time, but I'd rather not have the circular dependency at all. Other thoughts welcome, though.

Circular dependencies at type checking time aren't an issue, they can be solved using `typing.TYPE_CHECKING` and quoting the type annotation.
I find the circular dependency "normal" in this case, but feel free to ignore the comment.

ivanpauno (approved):

rclpy/rclpy/qos_event.py
clalancette: Thanks for the feedback.  I think I'm going to go ahead with this patch as-is, just to get this in.  If we need to have more information from the Subscription object in the future, we can revisit it then.

Issue #640, opened 2020-11-27
ros2 service call 'get_parameter_types' leads to coredump.


fujitatomoya:
## Bug report

- Operating System:
  - ubuntu20.04
- Installation type:
  - source build
- Version or commit hash:
  - https://github.com/ros2/ros2/commit/c059de9aadbf3d9ad189ca17d62f0110683c0a65
- DDS implementation:
  - Fast-DDS
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

- service server

```
root@5c7a784cba7f:~/ros2_ws/colcon_ws# ros2 run demo_nodes_py talker
[INFO] [1606456496.047371800] [talker]: Publishing: "Hello World: 0"
[INFO] [1606456497.032341427] [talker]: Publishing: "Hello World: 1"
[INFO] [1606456498.032360895] [talker]: Publishing: "Hello World: 2"
Traceback (most recent call last):
  File "/root/ros2_ws/colcon_ws/install/demo_nodes_py/lib/demo_nodes_py/talker", line 11, in <module>
    load_entry_point('demo-nodes-py', 'console_scripts', 'talker')()
  File "/root/ros2_ws/colcon_ws/build/demo_nodes_py/demo_nodes_py/topics/talker.py", line 44, in main
    rclpy.spin(node)
  File "/root/ros2_ws/colcon_ws/install/rclpy/lib/python3.8/site-packages/rclpy/__init__.py", line 196, in spin
    executor.spin_once()
  File "/root/ros2_ws/colcon_ws/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 711, in spin_once
    raise handler.exception()
  File "/root/ros2_ws/colcon_ws/install/rclpy/lib/python3.8/site-packages/rclpy/task.py", line 239, in __call__
    self._handler.send(None)
  File "/root/ros2_ws/colcon_ws/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 426, in handler
    await call_coroutine(entity, arg)
  File "/root/ros2_ws/colcon_ws/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 380, in _execute_service
    response = await await_or_execute(srv.callback, request, srv.srv_type.Response())
  File "/root/ros2_ws/colcon_ws/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 118, in await_or_execute
    return callback(*args)
[...]

PR #641, opened 2020-11-27
Fix Enum not being comparable with ints in get_parameter_types service

fujitatomoya:
fix https://github.com/ros2/rclpy/issues/640

### DIFF
--- a/rclpy/rclpy/parameter_service.py
+++ b/rclpy/rclpy/parameter_service.py
@@ -85,7 +85,7 @@ class ParameterService:
     def _get_parameter_types_callback(self, request, response):
         node = self._get_node()
         for name in request.names:
-            response.types.append(node.get_parameter_or(name).type_)
+            response.types.append(node.get_parameter_or(name).type_.value)
         return response
 
     def _list_parameters_callback(self, request, response):


fujitatomoya:
@ivanpauno @clalancette 

would you check this minor fix when you get back? thanks in advance.

rclpy/rclpy/parameter.py
```diff
@@ -23,7 +23,7 @@
 
 class Parameter:
 
-    class Type(Enum):
+    class Type(IntEnum):
```
iuhilnehc-ynos, line 26: Using IntEnum as a base class will change the behavior, which will make compareing a variable of Type with int value return True. Maybe the author wants the variable explicitly compare with the Type value, such as Type.BOOL, not int(1).

What about updating from
```
response.types.append(param.type_)
```
to
```
response.types.append(param.value_)
```

fujitatomoya: i see your point, thanks. let's keep this open to get more feedback.

iuhilnehc-ynos: Maybe we can just update "_type"  by appending ".value"

ivanpauno (approved):
This seems reasonable to me.

When was https://github.com/ros2/rclpy/issues/640 introduced?
It seems something a Python update broke, but maybe I'm wrong.

rclpy/rclpy/parameter.py
ivanpauno: @iuhilnehc-ynos alternative also sounds reasonable.

fujitatomoya:
> When was https://github.com/ros2/rclpy/issues/640 introduced?

i'm not sure, i was checking on https://github.com/ros2/rclcpp/issues/1482, and accidentally found that.

rclpy/rclpy/parameter.py
ivanpauno: @fujitatomoya can you apply @iuhilnehc-ynos suggestion, instead of modifying `Enum` to `IntEnum`.
I think that not modifying the type is a better idea.

fujitatomoya: yeah, okay i will do that in today.

fujitatomoya: > Maybe we can just update "type_" by appending ".value" for

thanks for the suggestion, it's been addressed with this one.

fujitatomoya:
@iuhilnehc-ynos @ivanpauno 

friendly ping, could you have a look again? (I confirmed that this fixes problem, and rclpy passes all test in local environment.)

fujitatomoya:
sorry there was error for `DCO`, i just did `git push -f`, now it is ready to review.

iuhilnehc-ynos (approved):
LGTM

ivanpauno (approved):

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13208/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8139/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10924/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13225/)

ivanpauno:
Windows build warning is unrelated, merging!

Thanks @fujitatomoya !

Issue #643, opened 2020-12-07
Memory leak on rclpy subscriber


yfedi:
During the work, we noticed that memory consumption of our docker container(with ROS2 Foxy)  has been increased with time. We did research and found that this memory is allocated for the node, written in python.

We tried to reproduce the issue with a simple application and we reproduced it on clean foxy docker with a demo application.

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04(doesn't matter in this case, as in docker other OS is used)
- Installation type:
  - binaries 
- Version or commit hash:
  - Foxy
- DDS implementation:
  - default
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue
1. Download docker file from the repository [dockerfile](https://github.com/osrf/docker_images/blob/master/ros/foxy/ubuntu/focal/desktop/Dockerfile)
2. Go to the folder where you've saved dockerfile in the previous step Build docker with a command like
`docker build -t ros2-foxy .`
3. Run docker with the listener from demo py package
`docker run -it --rm ros2-foxy ros2 run demo_nodes_py listener`
4. Run rqt tool on your environment or in the other docker
5. Check if /chatter topic is available. It should be created by the listener node
6. In rqt, configure message publisher to send messages to the chatter with high frequency, e.g. 5000Hz. The frequency doesn't matter, but with such frequency, we will see issue sooner 
7. Check docker(where the listener is run) memory consumption. In my case it was **~60Mb**
8. Wait for some time, in my case about 30 min

#### Expected behavior
Memory consumption is on the same level
#### Actual behavior
Memory consumption is increased to **330mB**
#### Additional information
Memory consumption is not decreased when we stop message publishing.

clalancette:
Out of curiousity, do you know if it also leaks in Rolling?  We've fixed a number of leaks there.  You should be able to `apt-get install ros-rolling-desktop` and re-run your test.

Knowing whether it also leaks on Rolling would give us some additional information on where to look.  Thanks.

fujitatomoya:
based on ros2 https://github.com/ros2/ros2/commit/32b29971204aee3b10afdfd4bc9dd4efa708439b

-  [demo_nodes_py](https://github.com/ros2/demos/tree/master/demo_nodes_py) with 1kHz publication

as we can see below, RSS increases from 40448KB to 215236KB. huge memory is allocated on heap(anon) area.

```
[...]

PR #645, opened 2020-12-09
memory leak (#643)

fujitatomoya:
address https://github.com/ros2/rclpy/issues/643

Co-authored-by: Chen Lihui <lihui.chen@sony.com>

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -3446,10 +3446,12 @@ rclpy_message_info_to_dict(rmw_message_info_t * message_info)
   PyObject * source_timestamp = PyLong_FromLongLong(message_info->source_timestamp);
   if (source_timestamp != NULL) {
     PyDict_SetItemString(dict, "source_timestamp", source_timestamp);
+    Py_DecRef(source_timestamp);
   }
-  PyObject * received_timestamp = PyLong_FromLongLong(message_info->source_timestamp);
+  PyObject * received_timestamp = PyLong_FromLongLong(message_info->received_timestamp);
   if (received_timestamp != NULL) {
-    PyDict_SetItemString(dict, "received_timestamp", source_timestamp);
+    PyDict_SetItemString(dict, "received_timestamp", received_timestamp);
+    Py_DecRef(received_timestamp);
   }
 
   // check for errors
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -3436,30 +3436,14 @@ rclpy_take_raw_with_info(rcl_subscription_t * subscription, rmw_message_info_t *
 static PyObject *
 rclpy_message_info_to_dict(rmw_message_info_t * message_info)
 {
-  PyObject * dict = PyDict_New();
+  PyObject * dict = Py_BuildValue(
+    "{s:L,s:L}",
+    "source_timestamp", message_info->source_timestamp,
+    "received_timestamp", message_info->received_timestamp);
   if (dict == NULL) {
-    PyErr_Format(PyExc_RuntimeError, "Failed to create dictionary object");
+    PyErr_Format(PyExc_RuntimeError, "Failed to create dictionary object for message info");
     return NULL;
   }
-
-  // we bail out at the end in case of errors
-  PyObject * source_timestamp = PyLong_FromLongLong(message_info->source_timestamp);
-  if (source_timestamp != NULL) {
-    PyDict_SetItemString(dict, "source_timestamp", source_timestamp);
-    Py_DecRef(source_timestamp);
-  }
-  PyObject * received_timestamp = PyLong_FromLongLong(message_info->received_timestamp);
-  if (received_timestamp != NULL) {
-    PyDict_SetItemString(dict, "received_timestamp", received_timestamp);
-    Py_DecRef(received_timestamp);
-  }
-
-  // check for errors
-  if (source_timestamp == NULL || received_timestamp == NULL) {
-    Py_DECREF(dict);
-    dict = NULL;
-  }
-
   return dict;
 }
 


fujitatomoya:
@iuhilnehc-ynos 

i include you as `co-author`, could you review this just in case?

iuhilnehc-ynos (approved):

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3446,10 +3446,12 @@ rclpy_message_info_to_dict(rmw_message_info_t * message_info)
   PyObject * source_timestamp = PyLong_FromLongLong(message_info->source_timestamp);
   if (source_timestamp != NULL) {
     PyDict_SetItemString(dict, "source_timestamp", source_timestamp);
+    Py_DecRef(source_timestamp);
```
clalancette, line 3449: I'll preface this comment by saying that I don't fully understand the reference counting mechanism in Python.  So this probably requires someone else to look at it.

But I'm not convinced this is the right thing to do.  The [documentation](https://docs.python.org/3/c-api/dict.html#c.PyDict_SetItemString) for `PyDict_SetItemString` says this:

```
This function does not steal a reference to val.
```

If `PyDict_SetItemString` doesn't steal a reference, then by doing a decref here, aren't there now *zero* references to the data?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3446,10 +3446,12 @@ rclpy_message_info_to_dict(rmw_message_info_t * message_info)
...
   }
-  PyObject * received_timestamp = PyLong_FromLongLong(message_info->source_timestamp);
+  PyObject * received_timestamp = PyLong_FromLongLong(message_info->received_timestamp);
```
clalancette, line 3451: Yikes, this is clearly buggy.

clalancette (commented):
I have a comment/question inline.  Maybe @sloretz could leave his thoughts as well.

Separately, from my brief perusal of the Python docs, it seems like using `Py_BuildValue` is the preferred way to create dictionaries like this.  I'll suggest that we just rewrite `rclpy_message_info_to_dict` to use that instead; I think it will be slightly easier to do all of the necessary error checking that way.

rclpy/src/rclpy/_rclpy.c
sloretz: IIUC "stealing" in the CPython API means the function does not increase the reference count of the object passed to it. If you bake a cake you own it :cake: . If I steal the cake from you, you don't get to eat it, I do :yum: . When a function "steals" your reference, you don't get to decref it because it's not yours anymore.

`PyLong_FromLongLong()` creates a new py object with a reference count of = 1. `PyDict_SetItemString()` does not steal the reference, so it must increase the reference count to 2. This `Py_DecRef` brings the reference count back to 1; meaning the dictionary is the only one referencing the object.

clalancette: > `PyLong_FromLongLong()` creates a new py object with a reference count of = 1.

Right.

> `PyDict_SetItemString()` does not steal the reference, so it must increase the reference count to 2.

This is the part I wasn't clear on (and the documentation didn't clarify).  In the CPython API, is it just assumed that whenever you pass a `PyObject` into a function, a new reference is taken?

iuhilnehc-ynos:
> it seems like using Py_BuildValue is the preferred way to create dictionaries 

:+1:

iuhilnehc-ynos:
@fujitatomoya 

It seems that I can't update your repository or add a code update suggestion in this PR after approval.
Could you update `rclpy_message_info_to_dict` by following @clalancette 's suggestion?

```c
static PyObject *
rclpy_message_info_to_dict(rmw_message_info_t * message_info)
{
  PyObject * dict = Py_BuildValue(
    "{s:L,s:L}",
    "source_timestamp", message_info->source_timestamp,
    "received_timestamp", message_info->received_timestamp);
  if (dict == NULL) {
    PyErr_Format(PyExc_RuntimeError, "Failed to create dictionary object");
    return NULL;
  }
  return dict;
}

```

fujitatomoya:
yeah, i will. thanks for comments to all 👍

rclpy/src/rclpy/_rclpy.c
fujitatomoya: > whenever you pass a PyObject into a function, a new reference is taken?

if it does not steal, i think yes.

fujitatomoya:
@clalancette @sloretz 
CC: @iuhilnehc-ynos 

requesting review once more time, thanks!

fujitatomoya:
@clalancette 

i don't think https://build.ros2.org/job/Rpr__rclpy__ubuntu_focal_amd64/137/ is related to the fix.

clalancette:

rclpy/src/rclpy/_rclpy.c
```diff
@@ -3436,28 +3436,14 @@ rclpy_take_raw_with_info(rcl_subscription_t * subscription, rmw_message_info_t *
...
 rclpy_message_info_to_dict(rmw_message_info_t * message_info)
 {
-  PyObject * dict = PyDict_New();
+  PyObject * dict = Py_BuildValue(
+    "{s:L,s:L}",
+    "source_timestamp", message_info->source_timestamp,
+    "received_timestamp", message_info->received_timestamp);
   if (dict == NULL) {
-    PyErr_Format(PyExc_RuntimeError, "Failed to create dictionary object");
+    PyErr_Format(PyExc_RuntimeError, "Failed to create dictionary object for message info");
```
clalancette, line 3444: Do we still need this?  The [documentation](https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue) for `Py_BuildValue` says:

```
 Returns the value or NULL in the case of an error; an exception will be raised if NULL is returned.
```

clalancette (commented):
Overall, this looks good to me.  I have one more question inline, then I'm happy to approve and run CI on this.

rclpy/src/rclpy/_rclpy.c
iuhilnehc-ynos: You are right, there will be a SystemError exception set internal if NULL is returned.
`If there is an error in the format string, the SystemError exception is set and NULL returned.`

> Do we still need this?
No, we don't need it. But I think using detailed information raised by `RuntimeError` for this operation is more useful.

clalancette: All right, that's fair.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13261/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8188/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10975/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13290/)

fujitatomoya:
@clalancette 
CC: @iuhilnehc-ynos 

I do not think any CI unstable reason and https://build.ros2.org/job/Rpr__rclpy__ubuntu_focal_amd64/144/ is related to this fix. but it is flaky and AFAIK it should use `try & except` statement to catch the exception in test code.

fujitatomoya:
> it is flaky and AFAIK it should use try & except statement to catch the exception in test code.

follow-up issue is created https://github.com/ros2/rclpy/issues/648

clalancette:
I'm going to go ahead and merge this; all of the failing tests are unrelated.  Thanks for the fix and the iterations.

PR #649, opened 2020-12-14
Don't throw an exception if timer canceled while sleeping.

clalancette:
It is a valid situation that may happen sometimes in threading,
so just quietly clean ourselves up and don't throw an
exception.

This resolves the failures we are seeing in the latest CI builds, e.g. https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/1447/testReport/junit/rclpy.rclpy.test/test_rate/test_shutdown_wakes_rate/ .

Here's a build where I ran the tests 10 times with this patch in place: https://ci.ros2.org/job/ci_linux-aarch64/8191/

@sloretz I'd appreciate a look by you, since you put this code in to start with.  I'm not sure if what I'm doing here is OK.

### DIFF
--- a/rclpy/rclpy/timer.py
+++ b/rclpy/rclpy/timer.py
@@ -131,6 +131,7 @@ class Rate:
                 self._event.clear()
         if self._is_shutdown:
             self.destroy()
+            raise ROSInterruptException()
 
     def sleep(self):
         """
--- a/rclpy/rclpy/timer.py
+++ b/rclpy/rclpy/timer.py
@@ -131,7 +131,6 @@ class Rate:
                 self._event.clear()
         if self._is_shutdown:
             self.destroy()
-            raise ROSInterruptException()
 
     def sleep(self):
         """
--- a/rclpy/test/test_rate.py
+++ b/rclpy/test/test_rate.py
@@ -17,6 +17,7 @@ import time
 
 import pytest
 import rclpy
+from rclpy.exceptions import ROSInterruptException
 from rclpy.executors import SingleThreadedExecutor
 
 # Hz
@@ -129,7 +130,7 @@ def test_shutdown_wakes_rate():
 
     rate = node.create_rate(0.0000001)
 
-    _thread = threading.Thread(target=sleep_check_exception, args=rate, daemon=True)
+    _thread = threading.Thread(target=sleep_check_exception, args=(rate,), daemon=True)
     _thread.start()
     executor.shutdown()
     node.destroy_node()
--- a/rclpy/test/test_rate.py
+++ b/rclpy/test/test_rate.py
@@ -111,6 +111,15 @@ class TestRate:
         self._thread.join()
 
 
+def sleep_check_exception(rate):
+    try:
+        rate.sleep()
+    except ROSInterruptException:
+        # rate.sleep() can raise ROSInterruptException if the context is
+        # shutdown while it is sleeping.  Just ignore it here.
+        pass
+
+
 def test_shutdown_wakes_rate():
     context = rclpy.context.Context()
     rclpy.init(context=context)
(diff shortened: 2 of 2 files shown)


rclpy/rclpy/timer.py
```diff
@@ -131,7 +131,6 @@ def _postsleep(self):
                 self._event.clear()
         if self._is_shutdown:
             self.destroy()
-            raise ROSInterruptException()
```
sloretz, line 134: I don't think this code path is part of a timer being canceled. This exception is raised when `self._is_shutdown` is set, which happens when the context is shutdown.

sloretz (changes requested):
I don't think this is the right fix. I think the test should expect an exception from `rate.sleep()` when ROS is shutdown. I don't know why this wasn't caught in the CI for #443. Were exceptions from daemon threads ignored before?

sloretz:
Likely fix: Catch `ROSInteruptException` in thread created by this test:

https://github.com/ros2/rclpy/blob/master/rclpy/test/test_rate.py#L114-L128

It looks like within the last month pytest started paying attention to exceptions in these daemon threads. https://github.com/pytest-dev/pytest/commit/d50df85e26b28e28ab5fb9a50df791068a051192

clalancette:
> It looks like within the last month pytest started paying attention to exceptions in these daemon threads. [pytest-dev/pytest@d50df85](https://github.com/pytest-dev/pytest/commit/d50df85e26b28e28ab5fb9a50df791068a051192)

Ah!  Good catch, that is likely what caused us to start failing the test.

> Likely fix: Catch `ROSInteruptException` in thread created by this test:
> 
> https://github.com/ros2/rclpy/blob/master/rclpy/test/test_rate.py#L114-L128

I considered that before doing this, and that would also fix the current problem.  But it also means that all downstream consumers of `rate.sleep` also have to do this to be ready for this situation.  Is that the API we are going for?

sloretz:
> But it also means that all downstream consumers of rate.sleep also have to do this to be ready for this situation. Is that the API we are going for?

I think it's a good idea. It lets downstream code know that `rate.sleep()` has woken early, and why. [ROS 1 Rate raises the same exception](http://wiki.ros.org/ROS/Tutorials/WritingPublisherSubscriber%28python%29#rospy_tutorials.2FTutorials.2FWritingPublisherSubscriber.CA-03913aa01eb4dc084a8831e8982cf5d7c49e193f_17).

clalancette:
> > 
> 
> I think it's a good idea. It lets downstream code know that `rate.sleep()` has woken early, and why. [ROS 1 Rate raises the same exception](http://wiki.ros.org/ROS/Tutorials/WritingPublisherSubscriber%28python%29#rospy_tutorials.2FTutorials.2FWritingPublisherSubscriber.CA-03913aa01eb4dc084a8831e8982cf5d7c49e193f_17).

All right, sounds good.  I'll change this to catch the exception in the test instead.  Thanks for the feedback.

clalancette:
I've changed the test to catch the exception now.  @sloretz please take another look.  In the meantime, here is CI (`--packages-select rclpy --retest-until-fail 2`):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13268/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8199/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10983/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13297/)

rclpy/test/test_rate.py
```diff
@@ -120,7 +129,7 @@ def test_shutdown_wakes_rate():
 
     rate = node.create_rate(0.0000001)
 
-    _thread = threading.Thread(target=rate.sleep, daemon=True)
+    _thread = threading.Thread(target=sleep_check_exception, args=rate, daemon=True)
```
sloretz, line 132: `args=rate` -> `args=(rate,)`

clalancette: Fixed in e2ba3ba.

clalancette:
New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13269/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8200/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10984/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13298/)

sloretz (approved):

PR #651, opened 2020-12-22
Use Pybind11's CMake functions for pycapsule lib

sloretz:
This is a small check to see if Pybind11's CMake function works on all
platforms. It works fine locally on ~~Linux~~ Ubuntu Focal using `CMAKE_BUILD_TYPE=Debug`.

@cottsay @IanTheEngineer FYI

PR #652, opened 2020-12-22
Update just pycapsule lib to use pybind11

sloretz:
Follow up test from #651 

This updates the `_rclpy_pycapsule` lib to use Pybind11 instead of raw CPython APIs where possible. Pybind11 would enable `rclpy` to not use `pycapsule` at all if we made some thin C++ classes, but that would take a bit of refactoring. This PR looks at translating the APIs to pybind11 without any changes. This might be a good intermediate step in moving rclpy to pybind11.

@cottsay @IanTheEngineer FYI


sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13320/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8250/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11051/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13364/)

sloretz:
Code size increase - the size of the pycapsule shared library is almost 50 times bigger than the same shared library on the master branch.

master

```console
$ ls -l _rclpy_pycapsule.cpython-38-x86_64-linux-gnu.so
-rw-r--r-- 1 sloretz sloretz 36608 Dec 22 14:54 _rclpy_pycapsule.cpython-38-x86_64-linux-gnu.so
```

this branch

```console
$ ls -l _rclpy_pycapsule.cpython-38-x86_64-linux-gnu.so 
-rw-r--r-- 1 sloretz sloretz 1810720 Dec 22 14:52 _rclpy_pycapsule.cpython-38-x86_64-linux-gnu.so
```

sloretz:
@azeey FYI

clalancette:
A 5 times code size increase (at best) seems like a lot for what is essentially a lateral move for this code.  If you are using the Python bindings at all, code size and performance are probably not your top concern, but it is still very large.  The question I have to ask is: given that we already have a working version of the hand-coded version of this library, does the move to pybind11 still make sense?  That is, does the move have enough benefits that we are willing to take a 5-50 times increase in compiled code size?

I don't have a good answer for that, and I'll leave it to the maintainers of rclpy to answer.  But I do think it should be considered.

ivanpauno:
> does the move to pybind11 still make sense?

IMO, one of the biggest pain points of the current implementation is object lifetimes.
RCL objects may keep a reference of other rcl objects and then you need to destroy them in order, and when creating bindings to a garbage collected language that's a nightmare.

If we move to pybind11 and don't use capsules at all, pybind11 [smart pointers support](https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html) might make things a lot easier.

emersonknapp:
I'll also note that we could do a C++ implementation, and manage object lifetimes via C++ destructors - there's nothing forcing us to do a pure-c implementation as far as I can tell. If we created simple C++ wrapper objects for the CPython API objects - it would probably remove a lot of the lifetime management code. These might be easier to expose with pybind, as well.

Has it been considered to just selectively expose `rclcpp` to Python to implement rclpy functionality? This may have drawbacks I haven't considered.

sloretz:
Rebased on #667

> I'll also note that we could do a C++ implementation, and manage object lifetimes via C++ destructors

@emersonknapp Definitely. Converting 1-to-1 is a hopefully quick process that will uncover any spots that are sensitive to pybind11's overhead. Using C++ and RAII comes after and adds a lot of benefit. You might be interested in the doc in #665.

cottsay (approved):

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13726/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8597/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11438/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13800/)

PR #653, opened 2020-12-29
include qos_profile_* constants in documentation

goekce:
This PR makes `qos_profile_*` constants in [QoS documentation](http://docs.ros2.org/foxy/api/rclpy/api/qos.html#rclpy.qos.QoSProfile) visible. Their usage is recommended on https://index.ros.org/doc/ros2/Releases/Release-Dashing-Diademata/#rclpy but hiding them in the documentation may confuse users.

The problem is that sphinx-autodoc currently does not include module-level variables, unless these variables include `#:` style comments. See https://github.com/sphinx-doc/sphinx/issues/1063

I did not look into source files other than `qos.py`, because including variables using this workaround is tedious. It is better to wait for https://github.com/sphinx-doc/sphinx/issues/1063

Note that this PR is only for foxy. 

fixes the issue: https://github.com/ros2/docs.ros2.org/issues/49

### DIFF
--- a/rclpy/docs/source/conf.py
+++ b/rclpy/docs/source/conf.py
@@ -197,6 +197,6 @@ epub_exclude_files = ['search.html']
 autoclass_content = 'both'
 
 autodoc_default_options = {
-    'members': None,
-    'undoc-members': True,
+    'members': True,  # document members
+    'undoc-members': True,  # also document members without documentation
 }
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -381,18 +381,37 @@ class LivelinessPolicy(QoSPolicyEnum):
 # Alias with the old name, for retrocompatibility
 QoSLivelinessPolicy = LivelinessPolicy
 
+# The details of the following profiles can be found at
+# 1. ROS QoS principles:
+#    https://design.ros2.org/articles/qos.html
+# 2. ros2/rmw : rmw/include/rmw/qos_profiles.h
+
+#: Used for initialization. Should not be used as the actual QoS profile.
 qos_profile_unknown = QoSProfile(**_rclpy.rclpy_get_rmw_qos_profile(
     'qos_profile_unknown'))
+#: Uses the default QoS settings defined in the DDS vendor tool
 qos_profile_system_default = QoSProfile(**_rclpy.rclpy_get_rmw_qos_profile(
     'qos_profile_system_default'))
+#: For sensor data, using best effort reliability and small
+#: queue depth
 qos_profile_sensor_data = QoSProfile(**_rclpy.rclpy_get_rmw_qos_profile(
     'qos_profile_sensor_data'))
+#: For services, using reliable reliability and volatile durability
 qos_profile_services_default = QoSProfile(**_rclpy.rclpy_get_rmw_qos_profile(
     'qos_profile_services_default'))
+#: For parameter communication. Similar to service QoS profile but with larger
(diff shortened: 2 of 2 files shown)


rclpy/docs/source/conf.py
```diff
@@ -194,9 +194,9 @@
 
 # -- Extension configuration -------------------------------------------------
 
-autoclass_content = 'both'
+#autoclass_content = 'both'
```
jacobperron, line 197: Why was this setting commented out? Seems like we could leave it how it was.

rclpy/rclpy/qos.py
```diff
@@ -381,20 +381,23 @@ class LivelinessPolicy(QoSPolicyEnum):
 # Alias with the old name, for retrocompatibility
 QoSLivelinessPolicy = LivelinessPolicy
 
+# module-level data members are not automatically documented by Sphinx. Use an
+# empty ``#:`` as a workaround. See
+# https://github.com/sphinx-doc/sphinx/issues/1063
```
jacobperron, line 386: Rather than appending an empty `#:` as a workaround, I think it would be better to add some content to the doc comment.

It could be something minimal, for example,

```python
#: A good QoS profile for sensor data, using best effort reliability.
qos_profile_sensor_data = QoSProfile(**_rclpy.rclpy_get_rmw_qos_profile(
    'qos_profile_sensor_data'))
```

Note, it looks like Sphinx is happy to have the comment appearing the line before the variable.

rclpy/docs/source/conf.py
```diff
@@ -194,9 +194,9 @@
...
 
 autodoc_default_options = {
-    'members': None,
-    'undoc-members': True,
+    'members': True,  # document members
```
jacobperron, line 200: I've tried reverting this change and the `qos_profile_*` variables still appear in the docs. I think it is the addition of the `#:` comments that is making them appear, and this line has no affect on them. I think we can undo this change.

jacobperron (changes requested):
Thanks for looking into it! I've left feedback below.

jacobperron:
@goekce Note, we enforce that contributors "sign" the Developer Certificate of Origin (DCO)](https://developercertificate.org/) (hence the failing DCO check). Please take a look at our [developer guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#change-control-process) and "sign-off" on all of your commits when contributing to the ROS 2 project.

rclpy/docs/source/conf.py
goekce: sorry, it seems that I forgot to revert it back while experimenting!

rclpy/docs/source/conf.py
goekce: >  I think it is the addition of the `#:` comments that is making them appear, and this line has no affect on them.

True, still: `None` and `True` keywords have the same meaning. I would stick to one of them to avoid confusion. What do you think?

rclpy/rclpy/qos.py
goekce: This is a good idea! I never used these profiles, but can try to get some info from rmw docs.

rclpy/docs/source/conf.py
goekce: I think it makes more sense to include this minor change in another commit.

jacobperron: > I think it makes more sense to include this minor change in another commit.

Sounds good to me.

jacobperron (approved):
Thanks for the improvements!

PR #656, opened 2021-01-08
Use CFFI to generate _rclpy_logging

sloretz:
This is a prototype using the `python-cffi` to generate the CPython extension instead of writing it ourselves. This pull request replaces _rclpy_logging with it. This is an alternative to pybind11 and #652.

CFFI works by giving it a definition file and a bit of extra C code with all the includes for those definitions, and asking it to generate a CPython extension. That extension is then compiled by CMake, and used in the Python code. In the case of _rclpy_logging it gets rid of all the manual CPython extension writing, and in the process moving all the logic that was in C into Python instead.

@cottsay @iantheengineer @azeey FYI

PR #659, opened 2021-01-13
Use Pybind11 to generate _rclpy_logging

sloretz:
~~This is a prototype PR meant to be used for an apples-to-apples comparison with #656.~~

Updated to convert _rclpy_logging to use pybind11, part of #665


sloretz:
Rebased on #667

rclpy/src/rclpy/_rclpy_logging.cpp
```diff
@@ -0,0 +1,193 @@
...
+void
+rclpy_logging_rcutils_log(
+  int severity,
+  const char * name,
+  const char * message,
+  const char * function_name,
+  const char * file_name,
+  uint64_t line_number)
+{
+  RCUTILS_LOGGING_AUTOINIT;
```
cottsay, line 130: Does this require special consideration since it initializes a global variable? Really, the entire logging subsystem for rcutils seems to rely on globals. Does that jive with the Pybind11 model?

sloretz: That's a good question. The only time I know of that global state matters is with subinterpreters. If a program used two python sub-interpreters and imported this module in both, then logging being initialized in one sub-interpreter would mean it's initialized in the other too. The same issue would come up with changes to the other globals like [the output format](https://github.com/ros2/rcutils/blob/d413e135d9d2a520f84b70ce3b0fd9625592ec50/src/logging.c#L88), and [the output handler](https://github.com/ros2/rcutils/blob/d413e135d9d2a520f84b70ce3b0fd9625592ec50/src/logging.c#L94).

It doesn't seem to be a problem because Pybind11 [hardcodes the module size to -1](https://github.com/pybind/pybind11/blob/c0fbb02c9fc1fe969f2740253469499385e6dc20/include/pybind11/pybind11.h#L1041), meaning Pybind11 modules [can't support subinterpreters](https://docs.python.org/3/c-api/module.html#c.PyModuleDef.m_size) anyways.

sloretz:
Rebased + two commits related to #681

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13743/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8612/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11454/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13817/)

rclpy/src/rclpy/_rclpy_logging.cpp
cottsay: Thanks for explaining. Seems like a non-issue, then.

cottsay (approved):

PR #660, opened 2021-01-13
Use Cython to generate _rclpy_logging

sloretz:
Uses Cython for the _rclpy_logging module. This is meant to be an apples-to-apples comparison with #656 and #659


sloretz:
Closing; future work in #665

PR #662, opened 2021-01-14
Use SIP to generate _rclpy_logging

sloretz:
This goes along with #656, #659, and #660 to give an apples-to-apples comparison of different Python binding generators


sloretz:
Closing; future work in #665

PR #664, opened 2021-01-20
Fix two clang static analysis warnings.

clalancette:
The first one in pycapsule is truly a bug; if destructor
was ever NULL, we would have reported an error and then
immediately dereferenced the NULL pointer, leading to a
crash.

The second one is a false positive.  The documentation for
PyErr_NoMemory() says that it always returns NULL, but
clang static analysis can't tell that just from the function
declaration.  Thus, call PyErr_NoMemory() to set the error,
and return NULL ourselves.

### DIFF
--- a/rclpy/src/rclpy/_rclpy_pycapsule.c
+++ b/rclpy/src/rclpy/_rclpy_pycapsule.c
@@ -95,6 +95,7 @@ rclpy_pycapsule_destroy(PyObject * Py_UNUSED(self), PyObject * args)
 
   if (!destructor) {
     PyErr_Format(PyExc_ValueError, "PyCapsule does not have a destructor.");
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13462/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8373/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11186/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13526/)

fujitatomoya (approved):
lgtm

clalancette:
Thanks for the reviews, merging.

PR #673, opened 2021-01-27
Use Py_XDECREF for pynode_names_and_namespaces

clalancette:
As pointed out by clang static analysis, it is possible for
pynode_names_and_namespaces to be NULL during cleanup.  Thus
use Py_XDECREF everywhere to do that checking for us.

### DIFF
--- a/rclpy/src/rclpy/_rclpy.c
+++ b/rclpy/src/rclpy/_rclpy.c
@@ -3825,7 +3825,7 @@ cleanup:
     PyErr_Format(
       RCLError,
       "Failed to destroy node_names: %s", rcl_get_error_string().str);
-    Py_DECREF(pynode_names_and_namespaces);
+    Py_XDECREF(pynode_names_and_namespaces);
     rcl_reset_error();
     return NULL;
   }
@@ -3833,7 +3833,7 @@ cleanup:
     PyErr_Format(
       RCLError,
       "Failed to destroy node_namespaces: %s", rcl_get_error_string().str);
-    Py_DECREF(pynode_names_and_namespaces);
+    Py_XDECREF(pynode_names_and_namespaces);
     rcl_reset_error();
     return NULL;
   }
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13520/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8423/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11246/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13587/)

emersonknapp (approved):
Nice - is there a clang static analysis build going on on the buildfarm for this?

clalancette:
> Nice - is there a clang static analysis build going on on the buildfarm for this?

Not currently, no.  I've been running this locally and just fixing the bugs as I see them.  The goal is indeed to have jobs available to do this either on https://ci.ros2.org, or as part of the PR jobs.

clalancette:
The test failures on macOS are the same as in the nightlies: https://ci.ros2.org/view/nightly/job/nightly_osx_repeated/2233/#showFailuresLink .  So I'm going to go ahead and merge this.  Thanks for the review.

j-rivero:
Might be related to aarch64 regressions today: https://ci.ros2.org/job/nightly_linux-aarch64_repeated/1488/testReport/junit/(root)/projectroot/test_client/ ?

clalancette:
> Might be related to aarch64 regressions today: https://ci.ros2.org/job/nightly_linux-aarch64_repeated/1488/testReport/junit/(root)/projectroot/test_client/ ?

I don't think so; I just merged it this morning, so it wouldn't have been used in the nightlies.

j-rivero:
> I don't think so; I just merged it this morning, so it wouldn't have been used in the nightlies.

Sorry, I meant that the problem in the nightly might have been resolved by this or do you think that the root cause is different.

clalancette:
> Sorry, I meant that the problem in the nightly might have been resolved by this or do you think that the root cause is different.

I think it is probably different; this only would be a problem during unlikely error paths.  I'll suggest to open a new issue with that failure, since it is likely something different.

Issue #665, opened 2021-01-21
Use Pybind11 for rclpy Python Bindings


sloretz:
## Feature request

@azeey @iantheengineer @cottsay FYI

#### Feature description

This is a ticket to use Pybind11 for the CPython extension instead of manually writing code to interact with the CPython API.

#### Implementation considerations

Plan: https://docs.google.com/document/d/1BsTlxgzAkpX-5Q7FW6_Hyw1fszDEQef_OEvYmdnrBrE/edit?usp=sharing

- [x] Replace custom CMake with Pybind11's pybind11_add_module (#667)
    - [x] Investigate CMake issues in Windows Debug https://github.com/ros2/rclpy/pull/667#issuecomment-778687182
    - [x] Fix windows warning reported in #687, (#688)
    - [ ] Remove code unsetting Py_DEBUG definition once all modules use Pybind11 (https://github.com/ros2/rclpy/pull/681#issuecomment-784579212) #814
- [ ] Convert each module to use Pybind11 with the existing C to Python API
    - [x] _rclpy (#675 :heavy_check_mark:, #691 :heavy_check_mark:, #692 :heavy_check_mark:, #693 :heavy_check_mark:, #694 :heavy_check_mark:, #695 :heavy_check_mark:, #696 :heavy_check_mark:, #699 :heavy_check_mark:, #701 :heavy_check_mark:, #703 :heavy_check_mark:, #706 :heavy_check_mark: , #709 :heavy_check_mark: , #712 :heavy_check_mark: , #715 :heavy_check_mark: , #717 :heavy_check_mark: , #718 :heavy_check_mark: , #719 :heavy_check_mark:, #721 :heavy_check_mark: , #723 :heavy_check_mark:, #730 :heavy_check_mark: , #732 :heavy_check_mark: , #733 :heavy_check_mark:, #735 :heavy_check_mark:, #736 :heavy_check_mark:, #738 :heavy_check_mark:)
    - [x] _rclpy_action (#678)
    - [x] _rclpy_handle (#668)
    - [x] _rclpy_logging (#659)
    - [x] _rclpy_pycapsule (#652)
    - [ ] _rclpy_signal_handler (#728)
        - skipping - uses C atomic types that can't be used with C++ due to [p0943r1](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0943r1.html) #814
        - Actually, couldn't skip because it uses rclpy_common, so rewrote in Python.
[...]

PR #675, opened 2021-01-29
Split rclpy module for easier porting to pybind11

sloretz:
@cottsay specifically, here's a way to split the giant `_rclpy` module in pieces that can be ported a little at a time.  This PR has split out exceptions (first commit) and `rclpy_context_get_domain_id` (second commit).

There are two modules: `_rclpy` and `_rclpy_pybind11`. Inside `rclpy.impl.implementation_singleton` the functions and exceptions on `_rclpy_pybind11` gets set on `_rclpy`. `_rclpy` itself needs the exception objects to raise them in C code, so it also imports `_rclpy_pybind11`. To do that, I had to convert `_rclpy` to use [Multi phase Initialization](https://docs.python.org/3/c-api/module.html#multi-phase-initialization).

@azeey @IanTheEngineer @cottsay FYI

Part of #665


rclpy/src/rclpy/exceptions.hpp
```diff
@@ -0,0 +1,62 @@
...
+namespace rclpy
+{
+
+std::string append_rcl_error(std::string prepend)
+{
+  prepend += ": ";
+  prepend += rcl_get_error_string().str;
+  rcl_reset_error();
+  return prepend;
+}
```
cottsay, line 32: Won't including this header in multiple `*.cpp` files that get compiled into the same binary cause a symbol collision?

rclpy/src/rclpy/context.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include <rcl/context.h>
+#include <rcl/types.h>
+
+#include <stdexcept>
+
+extern "C" {
+#include "rclpy_common/handle.h"
+}
+
+
```
cottsay, line 29: Is this double-newline meaningful?

rclpy/src/rclpy/_rclpy.c
```diff
@@ -6068,98 +6033,94 @@ static PyMethodDef rclpy_methods[] = {
...
+  {Py_mod_exec, rclpy_exec},
+  {0, NULL}
+};
+
+/// Define the Python module
+static struct PyModuleDef _rclpymodule = {
+  PyModuleDef_HEAD_INIT,
+  "_rclpy",
+  rclpy__doc__,
+  0,
```
cottsay, line 6114: Could you elaborate on why this changed from `-1` to `0`?

sloretz: It's the size of the [module's global state](https://docs.python.org/3/c-api/module.html#c.PyModuleDef.m_size). The reason for the change to `0` is `Non-negative m_size is required for multi-phase initialization`, and multi-phase initialization is needed to allow this module to import the new exceptions module.
 
The docs say -1 means there is state in global variables, and so it would be unsafe to use multiple sub-interpretters. Looking at this again, there are global variables storing references to the exceptions. IIUC this change means another change is needed to store those in the module's state instead.
https://github.com/ros2/rclpy/blob/812ea5082d008dc5c94a7d49e69df2d55415eec5/rclpy/src/rclpy/_rclpy.c#L53-L57

cottsay: Great explanation, thanks. Makes sense to me.

> ...another change is needed to store those in the module's state instead.

Are you planning to amend this PR or follow-up with another?

rclpy/src/rclpy/exceptions.hpp
sloretz: I suppose so. It looks like `inline` would solve the issue ([Explanation 2. 1.](https://en.cppreference.com/w/cpp/language/inline))

cottsay: Even if this gets inlined, I would assume the `RCLError` constructor would still have the issue. Any reason not to create a `.cpp` file for the implementation?

sloretz: No reason other than laziness :upside_down_face:

rclpy/src/rclpy/_rclpy.c
sloretz: Amend this one

rclpy/src/rclpy/exceptions.hpp
sloretz: b802e7e33bdd72e2bf5226ae06f3d68dfe14db1b + 49d5ec0a5f7659749d5a2b67db332f0713d3ab8c + d8182ad put made a cpp file for these and added them to `rclpy_common` - which does double duty of also introducing `pybind11` to `rclpy_common`.

rclpy/src/rclpy/context.hpp
sloretz: Habit from Python PEP 8 requiring double line breaks between definitions in module scope. Removed extra newline in da5edcc

rclpy/src/rclpy/_rclpy.c
sloretz: Uses module state instead of global variables in 30621b6

rclpy/src/rclpy/_rclpy_qos_event.c
```diff
@@ -32,13 +32,18 @@ typedef PyObject * (* _qos_event_data_filler_function)(_qos_event_callback_data_
 
 static
 bool
-_check_rcl_return(rcl_ret_t ret, const char * error_msg)
+_check_rcl_return(rclpy_module_state_t * module_state, rcl_ret_t ret, const char * error_msg)
 {
+  if (!module_state) {
+    PyErr_Format(PyExc_RuntimeError, "_check_rcl_return got NULL module state");
+    return NULL;
```
cottsay, line 39: Should this be `return false`?

rclpy/src/rclpy/context.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <Python.h>
+#include <pybind11/pybind11.h>
+#include <rcl/context.h>
+#include <rcl/types.h>
+
+#include <stdexcept>
+
+extern "C" {
+#include "rclpy_common/handle.h"
+}
```
cottsay, line 27: This is out-of-scope for this review, but is there any reason not to incorporate these linkage directives in the header itself? That seems to be the more common practice, but I could be missing something.

rclpy/src/rclpy/context.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace rclpy
+{
+/// Retrieves domain id from init_options of context
+/**
+ * \param[in] pycontext Capsule containing rcl_context_t
+ * \return domain id
+ */
+size_t
+context_get_domain_id(py::capsule pycontext)
```
cottsay, line 41: Is the idea to only ever include this file once? I'm worried about symbol collisions again.

rclpy/src/rclpy/_rclpy.c
```diff
@@ -6068,98 +6400,132 @@ static PyMethodDef rclpy_methods[] = {
...
-    Py_DECREF(m);
-    return NULL;
+  module_state->UnsupportedEventTypeError =
+    PyObject_GetAttrString(exceptions_module, "UnsupportedEventTypeError");
+  if (NULL == module_state->UnsupportedEventTypeError) {
+    Py_DECREF(module_state->NodeNameNonExistentError);
+    Py_DECREF(module_state->UnknownROSArgsError);
+    Py_DECREF(module_state->RCLInvalidROSArgsError);
+    Py_DECREF(module_state->RCLError);
+    Py_DECREF(exceptions_module);
```
cottsay, line 6471: It's worth pointing out that these pointers aren't getting reset after the decrefs. Presumably Python will free the memory at its leisure, but I'm not sure what becomes of the module state, or if it sticks around for a subsequent attempt or is also freed.

Might be an opportunity to centralize this cleanup in a cleanup block at the end of the function, as is the pattern used throughout this file.

cottsay (approved):
A few comments that you should take a look at, but nothing blocking.

The stateful module change looks like it was a lot of work - well done.

rclpy/src/rclpy/_rclpy.c
sloretz: IIUC an exception raised here becomes an exception raised [here](https://github.com/python/cpython/blob/6f8c8320e9eac9bc7a7f653b43506e75916ce8e8/Lib/importlib/util.py#L245), which means it will get thrown when something tries to access an attribute of the module.

The state itself exists on [the module PyObject](https://github.com/python/cpython/blob/6f8c8320e9eac9bc7a7f653b43506e75916ce8e8/Objects/moduleobject.c#L14) (the thing in `sys.modules["module.name"]`). [The module state is `free`'d when the module PyObject gets garbage collected](https://github.com/python/cpython/blob/6f8c8320e9eac9bc7a7f653b43506e75916ce8e8/Objects/moduleobject.c#L693). The module PyObject is going to stick around as long as there is a reference to it, which is usually forever unless someone calls [`importlib.reload()`](https://github.com/python/cpython/blob/6f8c8320e9eac9bc7a7f653b43506e75916ce8e8/Lib/importlib/__init__.py#L174). The state isn't reused, so no need to reset the pointers.

rclpy/src/rclpy/_rclpy_qos_event.c
sloretz: `return false` in 02a795e2f9222524f8637d81087741a0d58f6150

rclpy/src/rclpy/context.hpp
sloretz: Laziness - I didn't need to type the `#ifndef __cplusplus` this way. It looks like `handle.h` already has it though, so removed this in 83b9368bc3ceaec955f1fe2c7635da41e09ff2d4

rclpy/src/rclpy/context.hpp
sloretz: Split `context.hpp` into `.cpp` and `.hpp` in 4f5c39a7fa87071cd52c2418f5bb520a78be1529.

I would expect this to only be included once in the first pass, but once we convert to a `Context` class then I would expect it to be included multiple times.

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13741/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8610/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11452/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13815/)

sloretz:
CI again after fixing warning on OSX

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13745/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8614/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11456/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13819/)

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13748/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8617/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11459/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13822/)

cottsay (approved):
:tada:

PR #677, opened 2021-01-31
Fix the NULL check for destroy_ros_message.

clalancette:
clang static analysis pointed out that the check for NULL here
was wrong.  We were checking the double pointer for NULL, not
the derefenced one that we just attempted to fill in.  Fix the
check to look at the dereferenced pointer.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13564/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8455/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11285/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13626/)

fujitatomoya (approved):

Issue #680, opened 2021-02-16
👨‍🌾 rclpy fails during cmake configuration on windows CI


azeey:
## Bug report

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - source
- Version or commit hash:
  - 0ebae0cea78ee63864aea8b70b2962e24f42cb69
- DDS implementation:
  - default
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue
rclpy is failing on CI (nightly_win_deb since build [1899](https://ci.ros2.org/view/nightly/job/nightly_win_deb/1899/)) during cmake configuration. The reported error comes from `ament_cmake_python` while it's trying to set `PYTHON_EXECUTABLE`.

```
CMake Error at C:/ci/ws/install/share/ament_cmake_python/cmake/ament_cmake_python-extras.cmake:61 (message):
  execute_process( -c 'from distutils.sysconfig import get_python_lib;import
  os;print(os.path.relpath(get_python_lib(prefix='C:/ci/ws/install'),
  start='C:/ci/ws/install').replace(os.sep, '/'))') returned error code The
  parameter is incorrect
Call Stack (most recent call first):
  C:/ci/ws/install/share/ament_cmake_python/cmake/ament_cmake_python-extras.cmake:22 (_ament_cmake_python_get_python_install_dir)
  C:/ci/ws/install/share/ament_cmake_python/cmake/ament_python_install_package.cmake:27 (_ament_cmake_python_register_environment_hook)
  CMakeLists.txt:39 (ament_python_install_package)
```

#### Expected behavior
Builds successfully

#### Actual behavior

Build fails

sloretz:
Seems likely that it was caused by #665 

There's prior experience with Pybind11 issues on Windows Debug in rosbag2

PR #681, opened 2021-02-16
Copy windows debug fixes for pybind11

sloretz:
Hopefully fixes https://github.com/ros2/rclpy/issues/680

Copies windows debug fixes from rosbag2_py for using Pybind11 on Windows
debug and RelWithDebInfo.

For more info see the original PRs

https://github.com/ros2/rosbag2/pull/538
https://github.com/ros2/rosbag2/pull/531

Signed-off-by: Shane Loretz <sloretz@openrobotics.org>
Signed-off-by: Shane Loretz <sloretz@osrfoundation.org>


sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13679/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8558/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11396/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13746/)
* Windows debug: [Build Status](https://ci.ros2.org/job/ci_windows/13744/)

clalancette (commented):
This is clearly an improvement, as Windows Debug no longer fails.  However, there are still warnings present in the Windows Debug build.  @sloretz any thoughts on what's causing that?

sloretz:
> This is clearly an improvement, as Windows Debug no longer fails. However, there are still warnings present in the Windows Debug build. @sloretz any thoughts on what's causing that?

Implemented workaround to make sure Py_DEBUG on `rclpy_common` matches setting on the Python Modules. This workaround should not be necessary on Pybind11 2.6 and beyond, because that [adds `Py_DEBUG` as an interface compile definition to the `pybind11::pybind11` target](https://github.com/pybind/pybind11/blob/8de7772cc72daca8e947b79b83fea46214931604/tools/pybind11Tools.cmake#L116).

Anyways, if that's correct then this build should have no warnings:

Windows Debug [Build Status](https://ci.ros2.org/job/ci_windows/13752/)

sloretz:
Whoops, Windows debug again: [Build Status](https://ci.ros2.org/job/ci_windows/13754/)

clalancette:
Heh, it got worse: 2 warnings -> 4 warnings :).

sloretz:
Windows Debug again [Build Status](https://ci.ros2.org/job/ci_windows/13762/)

Made `Py_DEBUG` definition on `rclpy_common` `PUBLIC` so it's transitive - not sure why that would be needed.

clalancette:
So looking at the compile line for one of the files showing the problem, I see this (I split the lines for clarify):

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.28.29333\bin\HostX64\x64\CL.exe
 /c /IC:\dev\ros2_ws\src\ros2\rclpy\rclpy\src\rclpy_common\include
 /IC:\dev\ros2_ws\install\include /IC:\Python38\include /Zi /nologo /W3 /WX- /diagnostics:column
 /Od /Ob0 /D WIN32 /D _WINDOWS /D RCLPY_COMMON_BUILDING_DLL /D Py_DEBUG /D RCUTILS_ENABLE_FAULT_INJECTION
 /D _CRT_SECURE_NO_WARNINGS /D DEFAULT_RMW_IMPLEMENTATION=rmw_cyclonedds_cpp /D "CMAKE_INTDIR=\"Debug\""
 /D rclpy_common_EXPORTS /D _WINDLL /D _MBCS /Gm- /RTC1 /MDd /GS /fp:precise /Zc:wchar_t /Zc:forScope
 /Zc:inline /Fo"rclpy_common.dir\Debug\\" /Fd"rclpy_common.dir\Debug\vc142.pdb" /Gd /TC
 /errorReport:queue C:\dev\ros2_ws\src\ros2\rclpy\rclpy\src\rclpy_common\src\common.c
 C:\dev\ros2_ws\src\ros2\rclpy\rclpy\src\rclpy_common\src\handle.c
```

It is clear from that command-line that Py_DEBUG is being set to an empty macro.  The compile warning is this:

```
C:\Python38\include\pyconfig.h(321,1): warning C4005: 'Py_DEBUG': macro redefinition (compiling source file C:\dev\ros2_ws\src\ros2\rclpy\rclpy\src\rclpy_common\src\handle.c) [C:\dev\ros2_ws\build\rclpy\rclpy_common.vcxproj]
C:\dev\ros2_ws\src\ros2\rclpy\rclpy\src\rclpy_common\src\handle.c : message : see previous definition of 'Py_DEBUG' [C:\dev\ros2_ws\build\rclpy\rclpy_common.vcxproj]
C:\Python38\include\pyconfig.h(321,1): warning C4005: 'Py_DEBUG': macro redefinition (compiling source file C:\dev\ros2_ws\src\ros2\rclpy\rclpy\src\rclpy_common\src\common.c) [C:\dev\ros2_ws\build\rclpy\rclpy_common.vcxproj]
C:\dev\ros2_ws\src\ros2\rclpy\rclpy\src\rclpy_common\src\common.c : message : see previous definition of 'Py_DEBUG' [C:\dev\ros2_ws\build\rclpy\rclpy_common.vcxproj]
```

That warnings comes from the `pyconfig.h` file is included, which has this bit in it:

```
#ifdef _DEBUG
#          define Py_DEBUG
#endif
```

A few questions/observations then:
* Does Windows not allow redefining a macro, even if it is the same definition twice (as it appears to be in this case)?
* What part of the system is setting `_DEBUG` on?
* One way around this looks like to *not* define `Py_DEBUG` on the compile command-line.  But that would seem to have been what we had earlier in this PR (before 35dfed7).

sloretz:
How'd you get the compile commands on Windows?

> What part of the system is setting _DEBUG on?

It looks like the compiler sets `_DEBUG` because `/MDd` is set: https://docs.microsoft.com/en-us/cpp/c-runtime-library/debug?view=msvc-160

clalancette:
> How'd you get the compile commands on Windows?

I did:

```
set VERBOSE=1
rd /Q /S build\rclpy
python_d C:\Python38\Scripts\colcon.exe build --merge-install --cmake-args -DCMAKE_BUILD_TYPE=Debug --packages-select rclpy --event-handlers console_direct+
```

sloretz:
> Does Windows not allow redefining a macro, even if it is the same definition twice (as it appears to be in this case)?

Yeah, can confirm with an MRE that windows complains when a compiler definition is set both on the compiler command line and in code.

rclpy/CMakeLists.txt
```diff
@@ -146,13 +194,24 @@ target_link_libraries(_rclpy_handle PRIVATE
...
 
+# Workaround to ensure rclpy_common has same definition for Py_DEBUG
+# TODO(sloretz) remove this workaround when using pybind11 2.6+
+get_target_property(_rclpy_compile_definitions _rclpy COMPILE_DEFINITIONS)
+foreach(_rclpy_definition ${_rclpy_compile_definitions})
+  if("Py_DEBUG" STREQUAL _rclpy_definition)
+    target_compile_definitions(rclpy_common PUBLIC Py_DEBUG)
+    break()
+  endif()
+endforeach()
```
clalancette, line 213: Given what we now know, it seems like this isn't the correct workaround.  That is, we specifically *don't* want to add `Py_DEBUG` to the command-line, since that ends up with the warning.  It kind of looks to me like we need to remove this, and then see if there is something else that is also adding `Py_DEBUG` to the command-line.  If it is pybind11, maybe the thing to do here is to augment `clean_windows_flags` to also remove that flag?

sloretz:
Windows debug is green :tada: [Build Status](https://ci.ros2.org/job/ci_windows/13802/)

> That is, we specifically don't want to add Py_DEBUG to the command-line, since that ends up with the warning. It kind of looks to me like we need to remove this, and then see if there is something else that is also adding Py_DEBUG to the command-line.

Putting everything together, the Python interpreter has a special debug mode, and it [uses the macro `Py_DEBUG` to enable or disable this mode](https://docs.python.org/3/c-api/intro.html#debugging-builds). The source of the warning is [`pyconfig.h` sets `Py_DEBUG`](https://github.com/python/cpython/blob/6f8c8320e9eac9bc7a7f653b43506e75916ce8e8/PC/pyconfig.h#L320-L322) when `_DEBUG` is set, and `pybind11_add_module` [sets `Py_DEBUG` via the command line](https://github.com/pybind/pybind11/blob/3b1dbebabc801c9cf6f0953a4c20b904d444f879/tools/pybind11Tools.cmake#L145-L147) when a debug Python interpreter is used. This definition being set in two places is why the warning comes up.

One might think, just don't let `pybind11_add_module` set `Py_DEBUG` and all would be fine, but Pybind11 has some extra logic to allow using the Python release interpreter when building in Debug mode. That logic also uses `Py_DEBUG`. If `_DEBUG` is set but `Py_DEBUG` is not, then it will [undefine `_DEBUG`](https://github.com/pybind/pybind11/blob/3b1dbebabc801c9cf6f0953a4c20b904d444f879/include/pybind11/detail/common.h#L106-L108), [include `Python.h`](https://github.com/pybind/pybind11/blob/3b1dbebabc801c9cf6f0953a4c20b904d444f879/include/pybind11/detail/common.h#L112), then [restore `_DEBUG`](https://github.com/pybind/pybind11/blob/3b1dbebabc801c9cf6f0953a4c20b904d444f879/include/pybind11/detail/common.h#L134-L136). If we remove `Py_DEBUG` then `_DEBUG` will get unset and [CPython will link against release Python libs instead of the debug Python libs](https://github.com/python/cpython/blob/6f8c8320e9eac9bc7a7f653b43506e75916ce8e8/PC/pyconfig.h#L276-L281).

As long as `Py_DEBUG` is set and `pybind11/pybind11.h` was included instead of or before `<Python.h>`, then everything is happy. If code uses `pybind11_add_module` but includes `Python.h`, then there's the duplicate definition warning.

One way to fix this is to include `pybind11/pybind11.h` instead of `Python.h`, but that requires changing the language to C, and not everything is happy about that yet. Once all the PRs converting modules to use Pybind11 are in then this will be solved.

> If it is pybind11, maybe the thing to do here is to augment `clean_windows_flags` to also remove that flag?

This seems like the best fix, but it needs to be undone once the modules are using Pybind11. I made it in a separate macro in 1932bee so I could apply it to only the modules that have not yet been converted.

clalancette:
Woohoo!  I also tested this fix locally, and it seems to do the trick for me.  I'll do a brief review here, and then we can merge.  Thanks for the work.

clalancette (approved):
The changes look OK to me, so approving.  One question: should we open up an issue to track reverting these changes once everything is converted to PyBind11?  Or are you going to track it in #665?

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13730/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8600/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11441/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13804/)

sloretz:
>  One question: should we open up an issue to track reverting these changes once everything is converted to PyBind11? Or are you going to track it in #665?

I'll track it in #665. It should be removed in each PR as it's merged.

PR #686, opened 2021-02-26
Add history and depth to topic endpoint string

jacobperron:
I'm not sure why history and depth were left out originally. Maybe @jaisontj or @mm318 can answer?

AFAICT, we seem to be able to get this info from the endpoints just fine.

### DIFF
--- a/rclpy/rclpy/topic_endpoint_info.py
+++ b/rclpy/rclpy/topic_endpoint_info.py
@@ -164,6 +164,8 @@ class TopicEndpointInfo:
         result += 'Endpoint type: %s\n' % self.endpoint_type.name
         result += 'GID: %s\n' % '.'.join(format(x, '02x') for x in self.endpoint_gid)
         result += 'QoS profile:\n'
+        result += '  History: %s\n' % self.qos_profile.history.name
+        result += '  Depth: %s\n' % self.qos_profile.depth
         result += '  Reliability: %s\n' % self.qos_profile.reliability.name
         result += '  Durability: %s\n' % self.qos_profile.durability.name
         result += '  Lifespan: %d nanoseconds\n' % self.qos_profile.lifespan.nanoseconds
--- a/rclpy/test/test_topic_endpoint_info.py
+++ b/rclpy/test/test_topic_endpoint_info.py
@@ -95,6 +95,8 @@ class TestQosProfile(unittest.TestCase):
             'Endpoint type: INVALID\n' \
             'GID: \n' \
             'QoS profile:\n' \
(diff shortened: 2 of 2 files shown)


mm318:
I think it was left out because the history and depth QoS settings didn't contribute to QoS incompatibility, and this feature was implemented with debugging QoS incompatibilities in mind.

Overall, it was an oversight. Thanks for making the enhancement!

mm318 (approved):

fujitatomoya (approved):
lgtm

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13759/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8626/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11469/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13834/)

jacobperron:
So, I guess I got lucky with the default rmw. It looks like the history and depth values are only reported by rmw_cyclonedds_cpp, and not rmw_fastrtps_cpp or rmw_connext_cpp. Instead, we always get

```
History: UNKNOWN
Depth: 0
```

Knowing that this data isn't provided by 2/3 of the officially supported RMWs, I'm not sure if it makes sense to move forward with this PR. What do others think?

cc/ @ivanpauno

ivanpauno:
> Knowing that this data isn't provided by 2/3 of the officially supported RMWs, I'm not sure if it makes sense to move forward with this PR. What do others think?

I would say no, except that it can also be corrected in the other rmw implementations.
AFAIR not all qos settings are shared through discovery in DDS, only the ones that are used for qos matching.

I'm not sure what cyclonedds is doing, I will double check.

emersonknapp:
> Knowing that this data isn't provided by 2/3 of the officially supported RMWs, I'm not sure if it makes sense to move forward with this PR. What do others think?

Ah yes, I remember this now. I think there is no requirement for the participants to report these values because they're "participant-internal" values. It's worth leaving a comment in this code to inform the next person to wonder about it.

ivanpauno:
> I'm not sure what cyclonedds is doing, I will double check.

I tried, when you use a cyclonedds node and then call `ros2 topic info -v <topic_name>` also using cyclone it works correctly.
But if you create a node with fastrtps and then `ros2 topic info -v <topic_name>` with cyclonedds the information is not accurate.

i.e.: cyclone is sharing more metadata than what DDS requires (or that's my guess)

jacobperron:
Closing, since history and depth info isn't available for most RMWs.

PR #706, opened 2021-03-10
Convert wait_set functions to pybind11

sloretz:
Part of #665

This also replaces `rclpy_destroy_wait_set` and `rclpy.executors._WaitSet` with a pycapsule destructor.


rclpy/src/rclpy/wait_set.hpp
```diff
@@ -0,0 +1,115 @@
...
+ * Raises RuntimeError if the entity type is unknown
+ * Raises RCLError if any lower level error occurs
+ *
+ * \param[in] entity_type string defining the entity ["subscription, client, service"]
+ * \param[in] pywait_set Capsule pointing to the wait set structure
+ * \param[in] pyentity Capsule pointing to the entity to add
+ * \return Index in waitset entity was added at
+ */
+size_t
+wait_set_add_entity(const std::string entity_type, py::capsule pywait_set, py::capsule pyentity);
```
cottsay, line 73: Is there any value in the `const` qualifier for the string here? If it were passed by reference, I could see a performance gain, but I think it's just clutter in the header on a parameter that's passed by value, right?

rclpy/src/rclpy/wait_set.cpp
```diff
@@ -0,0 +1,296 @@
...
+    number_of_events,
+    context,
+    rcl_get_default_allocator());
+  if (ret != RCL_RET_OK) {
+    throw RCLError("failed to initialize wait set");
+  }
+}
+
+void
+wait_set_clear_entities(py::capsule pywait_set)
```
cottsay, line 107: Note that this returned `True` on success before. It doesn't seem to be used, though, so this seems OK to change.

rclpy/src/rclpy/wait_set.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+py::list
+get_ready_entities(const std::string entity_type, py::capsule pywait_set)
+{
+  if (0 != std::strcmp("rcl_wait_set_t", pywait_set.name())) {
+    throw py::value_error("capsule is not an rcl_wait_set_t");
+  }
+  auto wait_set = static_cast<rcl_wait_set_t *>(pywait_set);
+
+  py::list entity_ready_list;
```
cottsay, line 257: I think this can be dropped.
```suggestion
```

rclpy/src/rclpy/wait_set.cpp
```diff
@@ -0,0 +1,296 @@
...
+  }
+  if (index >= num_entities) {
+    throw std::out_of_range("wait set index too big");
+  }
+  return nullptr != entities[index];
+}
+
+template<typename EntityArray>
+py::list
+_get_ready_entities(const EntityArray ** entities, const size_t num_entities)
```
cottsay, line 237: Wow, this is so much better than before.

rclpy/rclpy/executors.py
```diff
@@ -521,7 +510,8 @@ def _wait_for_ready_callbacks(
                 entity_count += waitable.get_num_entities()
 
             # Construct a wait set
-            with _WaitSet() as wait_set, ExitStack() as context_stack:
+            wait_set = _rclpy.rclpy_get_zero_initialized_wait_set()
+            with ExitStack() as context_stack:
```
cottsay, line 514: Now that the GC can destroy the wait set when it pleases, are there any dependencies between the wait set and the entities that we need to handle?

rclpy/src/rclpy/wait_set.hpp
sloretz: I added it to ask the compiler to double check that I wasn't assigning to entity_type instead of doing comparisons. I can remove it.

rclpy/rclpy/executors.py
sloretz: None that needs to be handled by the wait set. `context_stack` makes sure the entities in the wait set stay alive until `rcl_wait()` finishes. After that it's totally fine to fini the wait set even if entities no longer exist because it doesn't do anything with those pointers.

rclpy/src/rclpy/wait_set.cpp
sloretz: Removed unused variable in 3121085 (didn't use gh suggestions to make DCO bot happy)

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13866/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8721/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11576/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13956/)

rclpy/src/rclpy/wait_set.cpp
```diff
@@ -0,0 +1,295 @@
...
+  PyMem_Free(wait_set);
+}
+
+py::capsule
+get_zero_initialized_wait_set()
+{
+  auto deleter = [](rcl_wait_set_t * ptr) {PyMem_FREE(ptr);};
+  auto wait_set = std::unique_ptr<rcl_wait_set_t, decltype(deleter)>(
+    static_cast<rcl_wait_set_t *>(PyMem_Malloc(sizeof(rcl_wait_set_t))),
+    deleter);
```
azeey, line 63: In the other pybind11 conversions, we've been throwing `bad_malloc` if `PyMem_Malloc` failed. Should we do the same here?

rclpy/src/rclpy/wait_set.hpp
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Raises RCLError if the wait set could not be initialized
+ *
+ * \param[in] pywait_set Capsule pointing to the wait set structure
+ * \param[in] node_name string name of the node to be created
+ * \param[in] number_of_subscriptions int
+ * \param[in] number_of_guard_conditions int
+ * \param[in] number_of_timers int
+ * \param[in] number_of_clients int
+ * \param[in] number_of_services int
```
azeey, line 40: Can you add docs for `pycontext`?

sloretz: Documented pycontext in dbc0c2335ef4151d9e346e5915428ee0539a5942

rclpy/src/rclpy/wait_set.cpp
sloretz: throws `std::bad_alloc` if `PyMem_Malloc` fails in bb10dab5683597429488e5ba85c193edee66f998

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13885/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8738/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11593/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13975/)

azeey (approved):

PR #708, opened 2021-03-10
Add API for checking QoS profile compatibility

audrow:
To do:
- [ ] Add Galactic release note

### DIFF
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -473,13 +473,21 @@ class QoSPresetProfiles(Enum):
         return cls[name.upper()].value
 
 
-class QoSCompatibility(IntEnum):
-    OK = 0
-    WARNING = 1
-    ERROR = 2
+QoSCompatibility = _rclpy.QoSCompatibility
 
 
 def qos_check_compatible(publisher_qos: QoSProfile, subscription_qos: QoSProfile):
+    """
+    Check if two QoS profiles are compatible.
+
+    Two QoS profiles are compatible if a publisher and subscription
+    using the QoS policies can communicate with each other.
+
+    If any policies have value "system default" or "unknown" then it is possible that
+    compatibility cannot be determined.
+    In this case, the value QoSCompatility.WARNING is set as part of
+    the returned structure.
+    """
     result = _rclpy.rclpy_qos_check_compatible(
         publisher_qos.get_c_qos_profile(),
         subscription_qos.get_c_qos_profile()
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -485,7 +485,7 @@ def qos_check_compatible(publisher_qos: QoSProfile, subscription_qos: QoSProfile
         subscription_qos.get_c_qos_profile()
     )
     compatibility = QoSCompatibility(
-        result.getCompatibility()
+        result.compatibility
     )
-    reason = result.getReason()
+    reason = result.reason
     return compatibility, reason
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -266,9 +266,6 @@ class QoSProfile:
             ', '.join(f'{slot[1:]}=%s' % getattr(self, slot) for slot in self.__slots__)
         )
 
-    def is_compatible(self, qos_profile):
-        return qos_check_compatible(self, qos_profile)
-
 
 class QoSPolicyEnum(IntEnum):
     """
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -477,9 +477,9 @@ class QoSPresetProfiles(Enum):
 
 
 class QoSCompatibility(IntEnum):
-    Ok = 0
-    Warning = 1
-    Error = 2
+    OK = 0
+    WARNING = 1
+    ERROR = 2
 
 
 def qos_check_compatible(publisher_qos: QoSProfile, subscription_qos: QoSProfile):
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -266,6 +266,9 @@ class QoSProfile:
             ', '.join(f'{slot[1:]}=%s' % getattr(self, slot) for slot in self.__slots__)
         )
 
+    def is_compatible(self, qos_profile):
+        return qos_check_compatible(self, qos_profile)
+
 
 class QoSPolicyEnum(IntEnum):
     """
@@ -471,3 +474,21 @@ class QoSPresetProfiles(Enum):
     def get_from_short_key(cls, name):
         """Retrieve a policy type from a short name, case-insensitive."""
         return cls[name.upper()].value
+
+
+class QoSCompatibility(IntEnum):
+    Ok = 0
+    Warning = 1
+    Error = 2
+
+
+def qos_check_compatible(publisher_qos: QoSProfile, subscription_qos: QoSProfile):
+    result = _rclpy.rclpy_qos_check_compatible(
+        publisher_qos.get_c_qos_profile(),
+        subscription_qos.get_c_qos_profile()
+    )
+    compatibility = QoSCompatibility(
+        result.getCompatibility()
+    )
+    reason = result.getReason()
+    return compatibility, reason
--- a/rclpy/src/rclpy/_rclpy_pybind11.cpp
+++ b/rclpy/src/rclpy/_rclpy_pybind11.cpp
@@ -75,6 +75,24 @@ PYBIND11_MODULE(_rclpy_pybind11, m) {
   .value("RCL_PUBLISHER_LIVELINESS_LOST", RCL_PUBLISHER_LIVELINESS_LOST)
   .value("RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS", RCL_PUBLISHER_OFFERED_INCOMPATIBLE_QOS);
 
+  py::enum_<rmw_qos_compatibility_type_t>(m, "QoSCompatibility")
+  .value("OK", RMW_QOS_COMPATIBILITY_OK)
+  .value("WARNING", RMW_QOS_COMPATIBILITY_WARNING)
+  .value("ERROR", RMW_QOS_COMPATIBILITY_ERROR);
+
+  py::class_<rclpy::QoSCheckCompatibleResult>(
+    m, "QoSCheckCompatibleResult",
+    "Result type for checking QoS compatibility with result")
+  .def(py::init<>())
+  .def_property_readonly(
+    "compatibility", [](const rclpy::QoSCheckCompatibleResult & result) {
+      return static_cast<int>(result.compatibility);
+    })
+  .def_property_readonly(
+    "reason", [](const rclpy::QoSCheckCompatibleResult & result) {
+      return result.reason;
+    });
+
   py::register_exception<rclpy::RCUtilsError>(m, "RCUtilsError", PyExc_RuntimeError);
   py::register_exception<rclpy::RMWError>(m, "RMWError", PyExc_RuntimeError);
   auto rclerror = py::register_exception<rclpy::RCLError>(m, "RCLError", PyExc_RuntimeError);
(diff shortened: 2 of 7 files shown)


rclpy/rclpy/qos.py
```diff
@@ -471,3 +471,21 @@ def short_keys(cls):
     def get_from_short_key(cls, name):
         """Retrieve a policy type from a short name, case-insensitive."""
         return cls[name.upper()].value
+
+
+class QoSCompatibility(IntEnum):
+    OK = 0
+    WARNING = 1
+    ERROR = 2
```
sloretz, line 479: I think it would be a good idea to create the enum with pybind11 so that the RCL constants aren't duplicated in python

See `ClockType` as an example

https://github.com/ros2/rclpy/blob/4d76c3a522d1ffdffcb2a5e7d925781f01fbfa07/rclpy/src/rclpy/_rclpy_pybind11.cpp#L38-L42
https://github.com/ros2/rclpy/blob/4d76c3a522d1ffdffcb2a5e7d925781f01fbfa07/rclpy/rclpy/clock.py#L21

rclpy/src/rclpy/_rclpy_pybind11.cpp
```diff
@@ -62,6 +63,22 @@ PYBIND11_MODULE(_rclpy_pybind11, m) {
...
+      return static_cast<int>(result.compatibility);
+    })
+  .def_property_readonly(
+    "reason", [](const rclpy::QoSCheckCompatibleResult & result) {
+      return result.reason;
+    })
+  .def(
+    "__repr__", [](const rclpy::QoSCheckCompatibleResult & result) {
+      (void) result;
+      return "Result type for checking QoS compatibility with result";
```
sloretz, line 79: I think this would fit better as a class doc string than the output of `__repr__`. I would ditch the `def` of `__repr__` completely and add this string as a third parameter to the constructor.

```c++
  py::class_<rclpy::QoSCheckCompatibleResult>(
    m, "QoSCheckCompatibleResult",
    "Result type for checking QoS compatibility with result")
...
```

The result is when some one does `help(QoSCheckCompatibleResult)` they see this string describing the type.

```
class QoSCheckCompatibleResult(pybind11_builtins.pybind11_object)
 |  Result type for checking QoS compatibility with result
...
```

rclpy/src/rclpy/qos.cpp
```diff
@@ -0,0 +1,87 @@
...
+    *publisher_qos_profile_,
+    *subscription_qos_profile_,
+    &compatible,
+    reason_c_str,
+    reason_size);
+
+  if (RMW_RET_OK != ret) {
+    auto error_str = rmw_get_error_string().str;
+    rmw_reset_error();
+    throw RCLError(error_str);
```
sloretz, line 63: Throw `RMWError` instead. It will call [`rmw_get_error_string()` and `rmw_reset_error()` for you](https://github.com/ros2/rclpy/blob/4d76c3a522d1ffdffcb2a5e7d925781f01fbfa07/rclpy/src/rclpy_common/src/exceptions.cpp#L61). The text given to the constructor is prepended and should be more specific to where it was thrown, like `throw RMWError("failed to compare QoS profiles")`

rclpy/src/rclpy/qos.hpp
```diff
@@ -0,0 +1,81 @@
...
+};
+
+/// Result type for checking QoS compatibility
+/**
+ * \see rclcpp::qos_check_compatible()
+ */
+struct QoSCheckCompatibleResult
+{
+  /// Compatibility result.
+  QoSCompatibility compatibility;
```
sloretz, line 43: Rather than defining a `QoSCompatibility` enum in this header, I would make this type [`rmw_qos_compatibility_type_t`](https://github.com/ros2/rmw/blob/ace271650ad39a656d81239050c9639c4190f077/rmw/include/rmw/qos_profiles.h#L126). Then there's no need for the enum above or the switch statement in `qos_check_compatible`

rclpy/src/rclpy/qos.cpp
```diff
@@ -0,0 +1,87 @@
...
+    case RMW_QOS_COMPATIBILITY_WARNING:
+      result.compatibility = QoSCompatibility::WARNING;
+      break;
+    case RMW_QOS_COMPATIBILITY_ERROR:
+      result.compatibility = QoSCompatibility::ERROR;
+      break;
+    default:
+      auto error_str = "Unexpected compatibility value returned by rmw '" +
+        std::to_string(compatible) + "'";
+      throw RCLError(error_str);
```
sloretz, line 82: Same comment about `RMWError`

rclpy/rclpy/qos.py
```diff
@@ -471,3 +471,21 @@ def short_keys(cls):
...
+
+
+def qos_check_compatible(publisher_qos: QoSProfile, subscription_qos: QoSProfile):
```
jacobperron, line 482: It would be good to add a docstring for this function, similar to the C++ equivalent.

rclpy/test/test_qos.py
```diff
@@ -124,3 +126,52 @@ def test_preset_profiles(self):
...
             QoSPresetProfiles.SYSTEM_DEFAULT.value ==
             QoSPresetProfiles.get_from_short_key('system_default'))
+
+
+class TestCheckQosCompatibility(unittest.TestCase):
+
+    def test_compatible(self):
+        qos = QoSProfile(
+            depth=1,
+            durability=QoSDurabilityPolicy.VOLATILE,
```
jacobperron, line 136: nit: we've explicitly set every policy except for `reliability`. Consider adding it here for completeness.

rclpy/test/test_qos.py
```diff
@@ -124,3 +126,52 @@ def test_preset_profiles(self):
...
+            liveliness_lease_duration=Duration(seconds=1),
+        )
+        compatibility, reason = qos_check_compatible(
+            qos, qos
+        )
+
+        assert compatibility == QoSCompatibility.OK
+        assert reason == ''
+
+    def test_incompatible(self):
```
jacobperron, line 149: It might be worth adding a comment that this test is assuming a DDS implementation. It's possible that a "best effort" publisher and "reliable" subscription is a valid match in a non-DDS implementation.

rclpy/test/test_qos.py
```diff
@@ -124,3 +126,52 @@ def test_preset_profiles(self):
...
+        )
+
+        compatibility, reason = qos_check_compatible(
+            pub_qos, sub_qos
+        )
+
+        assert compatibility == QoSCompatibility.ERROR
+        assert reason != ''
+
+    def test_warn_of_possible_incompatibility(self):
```
jacobperron, line 166: Same here. This is assuming behavior we expect from a DDS implementation.

rclpy/test/test_qos.py
```diff
@@ -124,3 +126,65 @@ def test_preset_profiles(self):
...
+    def test_compatible(self):
+        for durability in [QoSDurabilityPolicy.VOLATILE, QoSReliabilityPolicy.RELIABLE]:
+            qos = QoSProfile(
+                depth=1,
+                durability=durability,
+                lifespan=Duration(seconds=1),
+                deadline=Duration(seconds=1),
+                liveliness=QoSLivelinessPolicy.AUTOMATIC,
+                liveliness_lease_duration=Duration(seconds=1),
+            )
```
audrow, line 142: @jacobperron, is this what you meant in https://github.com/ros2/rclpy/pull/708#discussion_r597353191?

jacobperron: Sorry, I was suggesting we add a `reliability` keyword arg like this:
```diff
qos = QoSProfile(
                depth=1,
+               reliability=QoSReliabilityPolicy.RELIABLE,
+               durability=QoSDurabilityPolicy.VOLATILE,
-               durability=durability,
                lifespan=Duration(seconds=1),
                deadline=Duration(seconds=1),
                liveliness=QoSLivelinessPolicy.AUTOMATIC,
                liveliness_lease_duration=Duration(seconds=1),
            )
```

Note *durability* and *reliability* are difference policy kinds. The current code as written is technically wrong (assigning `QoSReliabilityPolicy.RELIABLE` as the durability).

audrow: > Note *durability* and *reliability* are different policy kinds. The current code as written is technically wrong (assigning `QoSReliabilityPolicy.RELIABLE` as the durability).

Ah, I get it. I was thinking of them as different types of policies. Thanks for the correction.

audrow:
@ros-pull-request-builder retest this please

jacobperron (approved):
LGTM

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14069/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8889/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11761/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14161/)

rclpy/src/rclpy/qos.cpp
```diff
@@ -0,0 +1,70 @@
...
+  char reason_c_str[reason_size] = "";
+  rmw_ret_t ret = rmw_qos_profile_check_compatible(
+    *publisher_qos_profile_,
+    *subscription_qos_profile_,
+    &result.compatibility,
+    reason_c_str,
+    reason_size);
+
+  if (RMW_RET_OK != ret) {
+    auto error_str = rmw_get_error_string().str;
```
jacobperron, line 61: There's a clang warning about this line. I fixed a similar case else where by making a copy:
```suggestion
    std::string error_str(rmw_get_error_string().str);
```

audrow: Thanks, @jacobperron!

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14078/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8897/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11771/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14170/)

rclpy/src/rclpy/qos.hpp
```diff
@@ -0,0 +1,75 @@
...
+{
+  /// Compatibility result.
+  rmw_qos_compatibility_type_t compatibility;
+
+  /// Reason for a (possible) incompatibility.
+  /**
+   * Set if compatiblity is RMW_QOS_COMPATIBILITY_WARNING or RMW_QOS_COMPATIBILITY_ERROR.
+   * Not set if the QoS profiles are compatible.
+   */
+  std::string reason;
```
sloretz, line 44: Nit, the above could be simplify this to `char reason[2048];` and ditch `reason_c+str` in the function.

sloretz: I think you can ditch `reason_size` too with `sizeof(QoSCheckCompatibleResult::reason)` to avoid duplicating the number  `2048`.

rclpy/src/rclpy/qos.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <rmw/qos_profiles.h>
+#include <rmw/types.h>
+
+#include <pybind11/pybind11.h>
+
+#include <string>
+
+#include "qos.hpp"
+#include "rclpy_common/exceptions.hpp"
+#include "rclpy_common/handle.h"
```
sloretz, line 29: No need for this header, no handle's are being used.

rclpy/src/rclpy/qos.cpp
```diff
@@ -0,0 +1,62 @@
...
+  QoSCheckCompatibleResult result;
+  rmw_ret_t ret = rmw_qos_profile_check_compatible(
+    *publisher_qos_profile_,
+    *subscription_qos_profile_,
+    &result.compatibility,
+    result.reason,
+    sizeof(QoSCheckCompatibleResult::reason));
+
+  if (RMW_RET_OK != ret) {
+    rmw_reset_error();
```
sloretz, line 55: No need to call `rmw_reset_error()`, `RMWError` does that too.

sloretz (approved):
LGTM with green CI

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14084/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8902/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11777/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14178/)

audrow:
Thanks, @sloretz and @jacobperron for the reviews!

PR #709, opened 2021-03-10
Use Pybind11 for name functions

sloretz:
Part of #665 

This uses pybind11 for name-related functions. I also added `RCUtilsError` since it was useful for one of the functions, and relaxed test expectations on the `TypeError` human readable text since the text output by pybind11 is a little bit different when an argument is the wrong type.


sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13886/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8739/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11594/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13976/)

sloretz:
Rebased on master

rclpy/src/rclpy/names.cpp
```diff
@@ -0,0 +1,289 @@
...
+  rcutils_allocator_t rcutils_allocator = rcutils_get_default_allocator();
+  rcutils_string_map_t substitutions_map = rcutils_get_zero_initialized_string_map();
+
+  rcutils_ret_t rcutils_ret = rcutils_string_map_init(&substitutions_map, 0, rcutils_allocator);
+  if (rcutils_ret == RCUTILS_RET_BAD_ALLOC) {
+    rcl_reset_error();
+    throw std::bad_alloc();
+  } else if (rcutils_ret != RCUTILS_RET_OK) {
+    throw RCLError("failed to initialize string map");
+  }
```
cottsay, line 141: This should be `rcutils` error handling, not `rcl`.

rclpy/src/rclpy/names.hpp
```diff
@@ -0,0 +1,116 @@
...
+ * \return tuple of error message and invalid index if invalid, or
+ * \return None if valid
+ */
+py::object
+get_validation_error_for_namespace(const char * namespace_);
+
+/// Validate a node name and return error message and index of invalidation.
+/**
+ * Raises MemoryError if memory could not be allocated
+ * Raises RuntimeError if an unexpected error happened while validating the node name
```
cottsay, line 70:
```suggestion
 * Raises RCLError if an unexpected error happened while validating the node name
```

rclpy/src/rclpy/names.hpp
```diff
@@ -0,0 +1,116 @@
...
+ * \param[in] node_name name of the node to be validated
+ * \return tuple of error message and invalid index if invalid, or
+ * \return None if valid
+ */
+py::object
+get_validation_error_for_node_name(const char * node_name);
+
+/// Expand a topic name
+/**
+ * Raises ValueError if the topic name, node name, or namespace are not valid.
```
cottsay, line 81: These last three functions can raise RCLError. Is that omitted intentionally?

cottsay (changes requested):
A few quick comments to address.

It's worth noting that several of the return types are changing from List to Tuple in this PR, though it's probably the right thing to do since it matches the documentation and is more performant.

rclpy/src/rclpy/names.cpp
cottsay: The docs should be updated to reflect the other exception type, too.

rclpy/src/rclpy/names.cpp
```diff
@@ -0,0 +1,289 @@
...
+  return py::make_tuple(validation_message, invalid_index);
+}
+
+py::object
+get_validation_error_for_namespace(const char * namespace_)
+{
+  int validation_result;
+  size_t invalid_index;
+  rmw_ret_t ret = rmw_validate_namespace(namespace_, &validation_result, &invalid_index);
+  if (ret == RCL_RET_BAD_ALLOC) {
```
hidmic, line 91: @sloretz in the same vein as @cottsay comments about `rcutils` error handling vs. `rcl` error handling, it'd be good to deal with `rmw` error handling using `rmw` API.

rclpy/src/rclpy/names.cpp
```diff
@@ -0,0 +1,289 @@
...
+  rcl_ret_t ret = rcl_get_default_topic_name_substitutions(&substitutions_map);
+  if (ret == RCL_RET_BAD_ALLOC) {
+    rcl_reset_error();
+    throw std::bad_alloc();
+  } else if (ret != RCL_RET_OK) {
+    auto exception = RCLError("failed to get default substitutions");
+    // finalize the string map before returning
+    rcutils_ret = rcutils_string_map_fini(&substitutions_map);
+    if (rcutils_ret != RCUTILS_RET_OK) {
+      fprintf(
```
hidmic, line 152: @sloretz FYI there's `RCUTILS_SAFE_FWRITE_TO_STDERR` too.

rclpy/src/rclpy/names.cpp
```diff
@@ -0,0 +1,289 @@
...
+  if (!node) {
+    throw py::error_already_set();
+  }
+
+  const rcl_node_options_t * node_options = rcl_node_get_options(node);
+  if (nullptr == node_options) {
+    throw RCLError("failed to get node options");
+  }
+
+  char * output_cstr = NULL;
```
hidmic, line 268: @sloretz nit: 
```suggestion
  char * output_cstr = nullptr;
```

rclpy/src/rclpy/names.cpp
```diff
@@ -0,0 +1,289 @@
...
+    };
+  auto resolved_name = std::unique_ptr<char, decltype(name_deleter)>(
+    output_cstr, name_deleter);
+
+  if (ret != RCL_RET_OK) {
+    throw RCLError("failed to resolve name");
+  }
+
+  return std::string{resolved_name.get()};
+}
```
hidmic, line 288: @sloretz if I'm not mistaken, `output_cstr` will remain a `nullptr` on `rcl_node_resolve_name()` error. In which case I think you can get away with simpler code e.g.:
```suggestion
  if (ret != RCL_RET_OK) {
    throw RCLError("failed to resolve name");
  }

  std::string output{output_cstr};
  node_options->allocator.deallocate(output_cstr, node_options->allocator.state);
  return output;
}
```

hidmic (commented):
Overall looks great!

rclpy/src/rclpy/names.cpp
gbalke: I think there a similar issue in `get_validation_error_for_full_topic_name`, `get_validation_error_for_node_name`.

On further investigation, there seems to be some handling for this in rcl. They seem to be translated to rcl error codes though?
https://github.com/ros2/rcl/blob/0ca8c6162136594e490ba69fbf288de1b5f7073d/rcl/src/rcl/expand_topic_name.c#L79

hidmic: Both `rmw` and `rcl` error handling APIs are aliases of `rcutils`'. That's why people tend to use them interchangeably. Which means we'll have a huge body of code to migrate if that stops being true at some point.

clalancette: Regardless, I think it is worthwhile to use the correct types (even if they are aliases of each other).  Besides the fact that it makes migration possible (though that is unlikely), there is no question for future readers of the code on whether this is correct or not.

rclpy/src/rclpy/names.cpp
hidmic: Hmm, `std::string` could throw on construction. Nevermind.

rclpy/src/rclpy/names.cpp
hidmic: Oops, I skimmed over the fact you are formatting. Either disregard or drop the integer formatting and use two safe writes.

rclpy/src/rclpy/names.cpp
sloretz: I added `RMWError` in fc32a0a99ebafc5e28efa343d0e6c868c977d9f7 and used rcl, rcutuils, or rmw constants and functions for each case in c269b6ac92dd3f15aafebffd3117055cfc34cad2

rclpy/src/rclpy/names.cpp
sloretz: NULL -> nullptr in dcfc7e28d67657fcb329704b7e78e3a88ef38d6c

rclpy/src/rclpy/names.hpp
sloretz: Added note about raising RCLError in c269b6ac92dd3f15aafebffd3117055cfc34cad2

rclpy/src/rclpy/names.cpp
sloretz: Updated docs with RMWError, RCUtilsError, or RCLError in c269b6ac92dd3f15aafebffd3117055cfc34cad2

rclpy/src/rclpy/names.hpp
sloretz: I think this doc got fixed in c269b6ac92dd3f15aafebffd3117055cfc34cad2 too

rclpy/src/rclpy/names.cpp
sloretz: Uses two calls to `RCUTILS_SAFE_FWRITE_TO_STDERR` in 90f1052

rclpy/src/rclpy/names.cpp
sloretz: Oops, actually fixed this one in d7fd3b0d20c4e8602ccdaae09a75b926c5d2d11d

rclpy/src/rclpy/names.cpp
sloretz: Good call on std::string constructor being able to throw. I fixed another spot where I overlooked that in f2d242ab6e3ce2c7cd5230c17bbd99e9f5efec92

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13959/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8792/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11659/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14042/)

rclpy/src/rclpy/names.cpp
```diff
@@ -0,0 +1,293 @@
...
+    only_expand,
+    &output_cstr);
+
+  auto name_deleter = [&](char * name) {
+      node_options->allocator.deallocate(name, node_options->allocator.state);
+    };
+  auto resolved_name = std::unique_ptr<char, decltype(name_deleter)>(
+    output_cstr, name_deleter);
+
+  if (ret != RCL_RET_OK) {
```
hidmic, line 287: @sloretz nit^N: flip operands, here and elsewhere.

rclpy/src/rclpy/names.cpp
```diff
@@ -0,0 +1,293 @@
...
+    node_options->allocator,
+    is_service,
+    only_expand,
+    &output_cstr);
+
+  auto name_deleter = [&](char * name) {
+      node_options->allocator.deallocate(name, node_options->allocator.state);
+    };
+  auto resolved_name = std::unique_ptr<char, decltype(name_deleter)>(
+    output_cstr, name_deleter);
```
hidmic, line 285: @sloretz FYI, nit^N: [`RCPPUTILS_SCOPE_EXIT`](https://github.com/ros2/rcpputils/blob/71ceb6b0597eee71f91dfeffe77fca7cc1a9cdb4/include/rcpputils/scope_exit.hpp#L65) may be handy in these cases.

hidmic (approved):
LGTM but for a few minor nits I won't block on.

rclpy/src/rclpy/names.cpp
sloretz: Flipped conditionals in 193a026

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13961/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8794/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11661/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14044/)

rclpy/src/rclpy/names.cpp
sloretz: Noted, I'll use this on the next PR once the rcpputils dependency is added.

hidmic (approved):

PR #719, opened 2021-03-17
[pybind11] Node Accessors

gbalke:
Part of #665.


rclpy/src/rclpy/2
```diff
@@ -0,0 +1,49 @@
+// Copyright 2021 Open Source Robotics Foundation, Inc.
```
sloretz, line 1: This file has a strange name `2`. Was it accidentally committed?

rclpy/src/rclpy/node.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+  auto node = static_cast<rcl_node_t *>(
+    rclpy_handle_get_pointer_from_capsule(pynode.ptr(), "rcl_node_t"));
+  if (!node) {
+    throw py::error_already_set();
+  }
+
+  const char * node_name = rcl_node_get_name(node);
+  if (!node_name) {
+    return py::none();
```
sloretz, line 55: Huh, I see this is what the code was doing already, but it seems incorrect. It looks like [when `rcl_node_get_name` fails it sets the error state](https://github.com/ros2/rcl/blob/fc3f93bd792ae467d26d6581da54158328dcd111/rcl/src/rcl/node.c#L435-L436), so returning `None` means the error state is still set. That's not a big problem, AFAIK it means it will print extra error messages the next time something tries to set the error state again.

I would recommend changing this API to throw `RCLError` instead of returning None. That means the return type could be `const char *` too (IIUC pybind11 will handle the conversion to a python string from that return type).

rclpy/src/rclpy/node.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+  auto node = static_cast<rcl_node_t *>(
+    rclpy_handle_get_pointer_from_capsule(pynode.ptr(), "rcl_node_t"));
+  if (!node) {
+    throw py::error_already_set();
+  }
+
+  const char * node_namespace = rcl_node_get_namespace(node);
+  if (!node_namespace) {
+    return py::none();
```
sloretz, line 72: Same comment  about making this API throw RCLError here.

rclpy/src/rclpy/node.hpp
```diff
@@ -0,0 +1,49 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLPY__NODE_HPP_
+#define RCLPY__NODE_HPP_
+
+#include <pybind11/pybind11.h>
+
+#include <string>
```
sloretz, line 20: Unused header?

rclpy/src/rclpy/node.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+#include <memory>
+#include <stdexcept>
+#include <string>
+
+#include "rclpy_common/handle.h"
+
+#include "rclpy_common/exceptions.hpp"
+
+#include "names.hpp"
```
sloretz, line 39: I think this should be including `"node.hpp"`

rclpy/src/rclpy/node.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <rcl/expand_topic_name.h>
+#include <rcl/rcl.h>
+#include <rcl/remap.h>
+#include <rcl/types.h>
+#include <rcl/validate_topic_name.h>
+#include <rcutils/error_handling.h>
+#include <rmw/error_handling.h>
+#include <rmw/validate_full_topic_name.h>
+#include <rmw/validate_namespace.h>
+#include <rmw/validate_node_name.h>
```
sloretz, line 29: It seems like there are a lot of unused headers here. I see reason for:

* pybind11/pybind11.h for all the `py::` stuff
* rcl/node.h for `rcl_node_t`, `rcl_node_get_name()`, `rcl_node_get_namespace()`
* rclpy_common/handle.h for the `rclpy_handle_get_pointer_from_capsule

And if you change the code to throw `RCLError` instead: "rclpy_common/exceptions.hpp"

But if you don't change it and instead just want to clear the error state: "rcl/error_handling.h"

I'm pretty sure the rest are unused.

sloretz (changes requested):

rclpy/src/rclpy/2
gbalke: I'm guessing I accidentally typed `:w 2` at some point. I'll fix it.

rclpy/src/rclpy/node.hpp
gbalke: I forgot to clean up headers after the copy/pasting. Sorry about that!

hidmic: Still unused?

hidmic (approved):
Overall LGTM pending green CI

gbalke:
CI up to `rclpy`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13996/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8819/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11689/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14073/)

sloretz (approved):

PR #717, opened 2021-03-17
Convert names_and_types graph APIs to pybind11

hidmic:
Precisely what the title says. Part of #665. This patch also brings `rcpputils` as a dependency.

CI up to `rclpy`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13960/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8793/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11660/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14043/)

PR #724, opened 2021-03-19
Make sure to list all C++ sources last.

hidmic:
To make sure `pybind11` pragmas override CPython ones. See https://github.com/ros2/rclpy/pull/717#issuecomment-802856792 for further reference. 

Debug CI up to `rclpy`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13995/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8818/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11688/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14072/)

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -129,9 +129,12 @@ endfunction()
 
 add_library(rclpy_common SHARED
   src/rclpy_common/src/common.c
+  src/rclpy_common/src/handle.c
+  # List C++ sources last to ensure pybind11 pragmas
+  # involving the Py_DEBUG macro override those from
+  # CPython
   src/rclpy_common/src/common.cpp
   src/rclpy_common/src/exceptions.cpp
-  src/rclpy_common/src/handle.c
 )
 target_link_libraries(rclpy_common PUBLIC
   pybind11::pybind11


rclpy/CMakeLists.txt
```diff
@@ -129,9 +129,12 @@ endfunction()
 
 add_library(rclpy_common SHARED
   src/rclpy_common/src/common.c
+  src/rclpy_common/src/handle.c
+  # List C++ sources last to ensure pybind11 pragmas
+  # involving the Py_DEBUG macro override those from
+  # CPython
```
clalancette, line 135: Oy, this is fragile.  Will this particular issue go away once rclpy is "pure" pybind11?

sloretz:
It looks like `handle.c` was last in the order before, so explicitly ordering here may not fix it. I was looking at another path where the C code is put into C++ files, and hopefully will open a PR after I figure out a small snag with C++ `void *` to function pointer casting.

rclpy/CMakeLists.txt
sloretz: Yeah, this issue is only a consequence of using both C and C++ files in the `rclpy_common` library. Once it's all C++ then we can rely on pybind11.h to fix the linking pragmas for us.

hidmic:
> It looks like handle.c was last in the order before,

That's right, but it wasn't until I introduced `common.cpp` that linkage to Python libraries became necessary.

sloretz:
> That's right, but it wasn't until I introduced common.cpp that linkage to Python libraries became necessary.

Oh, because exceptions.cpp doesn't use Python! I get it :)

hidmic:
Hmm, it's not an ordering issue :/

hidmic:
The issue may be elsewhere, see https://github.com/ros2/rclpy/pull/726.

hidmic:
Alright, this wasn't the right fix.

PR #725, opened 2021-03-19
Make all rclpy_common compilation units use C++

sloretz:
By making all compilation units use C++, it enables including `<pybind1/pybind11.h>` first in all compilation units. I think this should make the linker pragmas correct when building Windows debug, regardless the order the compilation units are linked.


sloretz:
Windows debug CI: [Build Status](https://ci.ros2.org/job/ci_windows/14074/)

sloretz:
Didn't fix the issue, closing

PR #739, opened 2021-03-26
Convert Client to use C++ Class

sloretz:
This converts the Client functions to a C++ class. It includes a couple infrastructure things needed for the conversion

## `class Handle`
This is an RAII wrapper around the `rclpy_handle_t` type. This class can be deleted as soon as all types are no longer using `rclpy_handle_t`. When constructed it increments the handle's reference to block destruction, and when destructed it decrements the reference. This allows the `Client` to keep a node alive, so a `Client` won't be destructed after a node.

## `class Destroyable`
While refactoring I found while `std::shared_ptr<>` solves the destruction order issue, there's still a need to block destruction of an object while it's in use. This class implements that very similarly to how it's done for `rclpy_handle_t` types - though I did choose to make a design change.

When destruction is requested on an `rclpy_handle_t`, new threads are allowed to use it as long as there is another thread using it. I didn't like that it meant two very fast threads using it back and forth (with overlap in their use) could block destruction forever. I chose to make `Destroyable` block new use of itself after destruction is requested, so it gets destroyed as soon as the current uses of it finish.

## `class Client`

This is the conversion to a `Client` class. It inherits from Destroyable and overrides `destroy()` to allow early destruction. It keeps a `Handle` instance to the node to prevent it from being destructed before the client. Once a `Node` class is created, that type will become `std::shared_ptr<rcl_node_t>` instead.

When `~Client()` is called, the destruction order between `rcl_node_t` and `rclpy_client_t` is guaranteed by the `std::shared_ptr<rcl_client_t>` being after `node_handle_` in the class definition. When `Client::destroy()` is called, the destruction order is guaranteed by releasing the `std::shared_ptr<rcl_client_t>` before releasing the `std::shared_ptr<Handle> node_handle_`.

### DIFF
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -33,7 +33,7 @@ class Client:
     def __init__(
         self,
         context: Context,
-        client_handle,
+        client_impl: _rclpy.Client,
         srv_type: SrvType,
         srv_name: str,
         qos_profile: QoSProfile,
@@ -46,7 +46,7 @@ class Client:
            should call :meth:`.Node.create_client`.
 
         :param context: The context associated with the service client.
-        :param client_handle: :class:`Handle` wrapping the underlying ``rcl_client_t`` object.
+        :param client_impl: :class:`_rclpy.Client` wrapping the underlying ``rcl_client_t`` object.
         :param srv_type: The service type.
         :param srv_name: The name of the service.
         :param qos_profile: The quality of service profile to apply the service client.
@@ -54,7 +54,7 @@ class Client:
             nodes default callback group is used.
         """
         self.context = context
-        self.__handle = client_handle
+        self.__client = client_impl
         self.srv_type = srv_type
         self.srv_name = srv_name
         self.qos_profile = qos_profile
@@ -122,8 +122,8 @@ class Client:
         if not isinstance(request, self.srv_type.Request):
             raise TypeError()
 
-        with self.handle as capsule:
-            sequence_number = _rclpy.rclpy_send_request(capsule, request)
+        with self.handle:
+            sequence_number = self.__client.send_request(request)
         if sequence_number in self._pending_requests:
             raise RuntimeError('Sequence (%r) conflicts with pending request' % sequence_number)
 
@@ -140,8 +140,8 @@ class Client:
 
         :return: ``True`` if a server is ready, ``False`` otherwise.
         """
-        with self.handle as capsule:
-            return _rclpy.rclpy_service_server_is_available(capsule)
+        with self.handle:
+            return self.__client.service_server_is_available()
 
     def wait_for_service(self, timeout_sec: float = None) -> bool:
         """
@@ -165,7 +165,7 @@ class Client:
 
     @property
     def handle(self):
-        return self.__handle
+        return self.__client
 
     def destroy(self):
-        self.handle.destroy()
+        self.__client.destroy_when_not_in_use()
--- a/rclpy/src/rclpy/client.cpp
+++ b/rclpy/src/rclpy/client.cpp
@@ -21,48 +21,26 @@
 #include <string>
 
 #include "rclpy_common/common.h"
-#include "rclpy_common/handle.h"
-
-#include "rclpy_common/exceptions.hpp"
 
 #include "client.hpp"
+#include "rclpy_common/exceptions.hpp"
+#include "utils.hpp"
 
 namespace rclpy
 {
-static void
-_rclpy_destroy_client(void * p)
-{
-  auto cli = static_cast<rclpy_client_t *>(p);
-  if (!cli) {
-    // Warning should use line number of the current stack frame
-    int stack_level = 1;
-    PyErr_WarnFormat(
-      PyExc_RuntimeWarning, stack_level, "_rclpy_destroy_client got NULL pointer");
-    return;
-  }
 
-  rcl_ret_t ret = rcl_client_fini(&(cli->client), cli->node);
-  if (RCL_RET_OK != ret) {
-    // Warning should use line number of the current stack frame
-    int stack_level = 1;
-    PyErr_WarnFormat(
-      PyExc_RuntimeWarning, stack_level, "Failed to fini client: %s",
-      rcl_get_error_string().str);
-    rcl_reset_error();
-  }
-  PyMem_Free(cli);
+void
+Client::destroy()
+{
+  rcl_client_.reset();
+  node_handle_.reset();
 }
 
-py::capsule
-client_create(
-  py::capsule pynode, py::object pysrv_type, std::string service_name,
-  py::capsule pyqos_profile)
+Client::Client(
+  py::capsule pynode, py::object pysrv_type, const char * service_name, py::capsule pyqos_profile)
+: node_handle_(std::make_shared<Handle>(pynode))
 {
-  auto node = static_cast<rcl_node_t *>(
-    rclpy_handle_get_pointer_from_capsule(pynode.ptr(), "rcl_node_t"));
-  if (!node) {
-    throw py::error_already_set();
-  }
+  auto node = node_handle_->cast<rcl_node_t *>("rcl_node_t");
 
   auto srv_type = static_cast<rosidl_service_type_support_t *>(
     rclpy_common_get_type_support(pysrv_type.ptr()));
@@ -80,22 +58,31 @@ client_create(
     client_ops.qos = *qos_profile;
   }
 
-  // Use smart pointer to make sure memory is free'd on error
-  auto deleter = [](rclpy_client_t * ptr) {_rclpy_destroy_client(ptr);};
-  auto cli = std::unique_ptr<rclpy_client_t, decltype(deleter)>(
-    static_cast<rclpy_client_t *>(PyMem_Malloc(sizeof(rclpy_client_t))),
-    deleter);
-  if (!cli) {
-    throw std::bad_alloc();
-  }
-  cli->client = rcl_get_zero_initialized_client();
-  cli->node = node;
+  // Create a client
+  rcl_client_ = std::shared_ptr<rcl_client_t>(
+    new rcl_client_t,
+    [this](rcl_client_t * client)
+    {
+      auto node = node_handle_->cast_or_warn<rcl_node_t *>("rcl_node_t");
+
+      rcl_ret_t ret = rcl_client_fini(client, node);
+      if (RCL_RET_OK != ret) {
+        // Warning should use line number of the current stack frame
+        int stack_level = 1;
+        PyErr_WarnFormat(
+          PyExc_RuntimeWarning, stack_level, "Failed to fini client: %s",
+          rcl_get_error_string().str);
+        rcl_reset_error();
+      }
+      delete client;
+    });
+
+  *rcl_client_ = rcl_get_zero_initialized_client();
 
   rcl_ret_t ret = rcl_client_init(
-    &(cli->client), node, srv_type,
-    service_name.c_str(), &client_ops);
-  if (ret != RCL_RET_OK) {
-    if (ret == RCL_RET_SERVICE_NAME_INVALID) {
+    rcl_client_.get(), node, srv_type, service_name, &client_ops);
+  if (RCL_RET_OK != ret) {
+    if (RCL_RET_SERVICE_NAME_INVALID == ret) {
       std::string error_text{"failed to create client due to invalid service name '"};
       error_text += service_name;
       error_text += "': ";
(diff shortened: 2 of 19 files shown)


rclpy/src/rclpy/handle.cpp
```diff
@@ -0,0 +1,59 @@
...
+    auto c_handle = static_cast<rclpy_handle_t *>(pyhandle_);
+    _rclpy_handle_dec_ref(c_handle);
+  }
+}
+
+Handle &
+Handle::operator=(const Handle & other)
+{
+  pyhandle_ = other.pyhandle_;
+
```
hidmic, line 46: @sloretz hmm, should we decrement the ref count before assignment?

rclpy/rclpy/client.py
```diff
@@ -140,8 +138,7 @@ def service_is_ready(self) -> bool:
 
         :return: ``True`` if a server is ready, ``False`` otherwise.
         """
-        with self.handle as capsule:
-            return _rclpy.rclpy_service_server_is_available(capsule)
+        return self.__client.service_server_is_available()
```
hidmic, line 141: @sloretz hmm, why no `with self.__client` context manager here? What if it concurs with `Client.destroy()`?

rclpy/src/rclpy/destroyable.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+void
+Destroyable::exit(py::object, py::object, py::object)
+{
+  if (0u == use_count) {
+    throw std::runtime_error("Internal error: Destroyable use_count would be negative");
+  }
+
+  --use_count;
+  if (please_destroy_ && use_count == 0u) {
```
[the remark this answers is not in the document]
sloretz: Flipped equality operands in 835142350614f1f2a960cfe4e1adef8ebe2a801f

rclpy/src/rclpy/destroyable.cpp
```diff
@@ -0,0 +1,73 @@
...
+    destroy();
+  }
+}
+
+void
+Destroyable::destroy()
+{
+  // Normally would be pure virtual, but then pybind11 can't create bindings for this class
+  throw std::runtime_error("Internal error: Destroyable subclass didn't override destroy()");
```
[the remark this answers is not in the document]
sloretz: Throwing `NotImplementedError` in e345c7a4b8a039abe9450b8fee81bc93b749d4e1

rclpy/src/rclpy/client.hpp
```diff
@@ -17,67 +17,91 @@
...
+   */
+  bool
+  service_server_is_available();
+
+  /// Take a response from a given client
+  /**
+   * Raises ValueError if pyclient is not a client capsule
+   *
+   * \param[in] pyresponse_type Instance of the message type to take
+   * \return 2-tuple sequence number and received response or None, None if there is no response
```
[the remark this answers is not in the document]
sloretz: Fixed sentence in 52db86ddda834a52ccf6c402b5ecb5039fe20264

rclpy/src/rclpy/client.cpp
```diff
@@ -175,14 +133,8 @@ _rclpy_destroy_service_info(PyObject * pycapsule)
...
+Client::take_response(py::object pyresponse_type)
 {
-  auto client = static_cast<rclpy_client_t *>(
-    rclpy_handle_get_pointer_from_capsule(pyclient.ptr(), "rclpy_client_t"));
-  if (!client) {
-    throw py::error_already_set();
-  }
-
   destroy_ros_message_signature * destroy_ros_message = nullptr;
   void * taken_response_ptr = rclpy_create_from_py(pyresponse_type.ptr(), &destroy_ros_message);
```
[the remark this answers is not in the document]
sloretz: Using `create_from_py` in 554e3c012c7c43b01b9c392c5aa3704e7ec63f5d

rclpy/src/rclpy/client.cpp
```diff
@@ -105,63 +92,34 @@ client_create(
...
+Client::send_request(py::object pyrequest)
 {
-  auto client = static_cast<rclpy_client_t *>(
-    rclpy_handle_get_pointer_from_capsule(pyclient.ptr(), "rclpy_client_t"));
-  if (!client) {
-    throw py::error_already_set();
-  }
-
   destroy_ros_message_signature * destroy_ros_message = nullptr;
   void * raw_ros_request = rclpy_convert_from_py(pyrequest.ptr(), &destroy_ros_message);
```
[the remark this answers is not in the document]
sloretz: I didn't see `convert_from_py` on master yet, but I did see `convert_to_py` and used it in 554e3c012c7c43b01b9c392c5aa3704e7ec63f5d

rclpy/rclpy/client.py
sloretz: Oops, I started out thinking I'd replace the context manager with a check in C++, but never implemented it. I added back the `with` statements in 5ba08d2e15879313170d7304f285088680ac4ccb

rclpy/src/rclpy/handle.cpp
sloretz: `other` will decrement the reference count in its destructor, so I think this function needs to increment the counter to keep it alive.

rclpy/rclpy/client.py
```diff
@@ -33,7 +32,7 @@ class Client:
     def __init__(
         self,
         context: Context,
-        client_handle,
+        client_impl,
```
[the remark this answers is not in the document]
sloretz: Added type annotation in 5ba08d2e15879313170d7304f285088680ac4ccb

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14102/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8915/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11791/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14199/)

rclpy/src/rclpy/handle.cpp
sloretz: Ohhhh, you mean decrement the ref count of `this->pyhandle_`! Definitely.

sloretz: Decref the old `pyhandle_` before assigning the new one in 1bf8f71

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14103/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8916/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11793/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14200/)

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14118/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8926/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11807/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14215/)

rclpy/rclpy/client.py
```diff
@@ -122,8 +122,8 @@ def call_async(self, request: SrvTypeRequest) -> Future:
         if not isinstance(request, self.srv_type.Request):
             raise TypeError()
 
-        with self.handle as capsule:
-            sequence_number = _rclpy.rclpy_send_request(capsule, request)
+        with self.handle:
+            sequence_number = self.__client.send_request(request)
```
hidmic, line 126: @sloretz nit^N: perhaps using the outcome of `self.handle.__enter__` is easier to read (though equivalent)?

```suggestion
        with self.handle as client:
            sequence_number = client.send_request(request)
```
Same elsewhere.

rclpy/src/rclpy/client.cpp
```diff
@@ -80,22 +58,31 @@ client_create(
...
-    static_cast<rclpy_client_t *>(PyMem_Malloc(sizeof(rclpy_client_t))),
-    deleter);
-  if (!cli) {
-    throw std::bad_alloc();
-  }
-  cli->client = rcl_get_zero_initialized_client();
-  cli->node = node;
+  // Create a client
+  rcl_client_ = std::shared_ptr<rcl_client_t>(
+    new rcl_client_t,
```
hidmic, line 63: @sloretz meta: why the mixure of allocation sources? I see `PyMem_Alloc`, `new`, and `allocator.allocate` across the code base.

rclpy/src/rclpy/client.hpp
```diff
@@ -17,67 +17,91 @@
...
-  py::capsule pynode, py::object pysrv_type, std::string service_name,
-  py::capsule pyqos_profile);
+class Client : public Destroyable
+{
+public:
+  /// Create a client
+  /**
+   * This function will create a client for the given service name.
+   * This client will use the typesupport defined in the service module
+   * provided as pysrv_type to send messages over the wire.
```
hidmic, line 38: @sloretz nit^N: I'd drop `over the wire`:
```suggestion
   * provided as pysrv_type to send messages.
```
We're careful elsewhere not to give details about the transport. Perhaps we should be everywhere.

rclpy/src/rclpy/client.hpp
```diff
@@ -17,67 +17,91 @@
...
+   *
+   * Raises ValueError if the capsules are not the correct types
```
hidmic, line 40: @sloretz nit:
```suggestion
   * Raises ValueError if the capsules are not of the correct types
```

rclpy/src/rclpy/handle.cpp
hidmic: Sorry for the imprecise comment, glad you were able to decode me :sweat_smile:

hidmic (approved):
LGTM

rclpy/src/rclpy/client.cpp
gbalke: Why not use the smart pointer library to allocate the object? http://www.cplusplus.com/reference/memory/make_shared/

gbalke (approved):
Looks good. I am enjoying the scoping benefits of classes :smile:.

I am a little curious about the "not-virtual" virtual `destroy` function but I see that it's simply a quirk of pybind11. https://pybind11.readthedocs.io/en/stable/advanced/classes.html

I guess since this is intended to be bound immediately, there's no good reason to make a `PyDestroyable` wrapper.

rclpy/rclpy/client.py
sloretz: I went the opposite direction in a6149d10cc9120bf582381698c31f1d137efc296 and made the context manager return `None` when it's entered. I'm a little cautions because I don't know if pybind11 has a way to figure out the `Destroyable` type returned at `__enter__` is an instance of `rclpy::Client`. If `__enter__` returns `Destroyable d`, then is `client.send_request`  effectively `reinterpret_cast<Client *>(d)->send_request()`? I think I recall @mxgrey showing me a case (IIRC involving multiple inheritance) where that's unsafe to do.

rclpy/src/rclpy/client.hpp
sloretz: This reads fine to me with or without `of`.

rclpy/src/rclpy/client.hpp
sloretz: Removed mention of transport in 8e20909efb0dcd2e31bf892c0ccdf236018fa250

rclpy/src/rclpy/client.cpp
sloretz: >  why the mixure of allocation sources?

Laziness. `new` throws if allocation fails, so the code doesn't have to check PyMem_Malloc's return or cleanup the `shared_ptr` to prevent attempting to free it twice. Maybe a PyMem_Malloc wrapper that throws on failure would be convenient here.

> Why not use the smart pointer library to allocate the object? http://www.cplusplus.com/reference/memory/make_shared/

I didn't use `make_shared` because I don't think it allows a custom deleter.

rclpy/rclpy/client.py
hidmic: Aha! Interesting. Good to know.

rclpy/src/rclpy/client.hpp
hidmic: Uh, didn't know you could phrase it like that. Nevermind then.

sloretz:
I'll take a second pass at the new/delete use in a follow up PR.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14136/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8942/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11823/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14235/)

sloretz:
Fixed Clang warning, running CI just to `rclpy` since it already passed with `--packages-above-and-dependencies` and the change seems small enough to not need a re-run of everything downstream

(build: `--packages-up-to rclpy` test: `--packages-select rclpy`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14149/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8953/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11834/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14248/)

sloretz:
CI LGTM, WIndows warning is in cyclonedds

hidmic (approved):

PR #745, opened 2021-03-31
Convert Timer to use a C++ Class

ahcorde:

This converts the Timer functions to a C++ class. 

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/rclpy/rclpy/timer.py
+++ b/rclpy/rclpy/timer.py
@@ -47,33 +47,41 @@ class Timer:
 
     @property
     def timer_period_ns(self):
-        val = self.__timer.get_timer_period()
+        with self.__timer:
+            val = self.__timer.get_timer_period()
         self.__timer_period_ns = val
         return val
 
     @timer_period_ns.setter
     def timer_period_ns(self, value):
         val = int(value)
-        self.__timer.change_timer_period(val)
+        with self.__timer:
+            self.__timer.change_timer_period(val)
         self.__timer_period_ns = val
 
     def is_ready(self):
-        return self.__timer.is_timer_ready()
+        with self.__timer:
+            return self.__timer.is_timer_ready()
 
     def is_canceled(self):
-        return self.__timer.is_timer_canceled()
+        with self.__timer:
+            return self.__timer.is_timer_canceled()
 
     def cancel(self):
-        self.__timer.cancel_timer()
+        with self.__timer:
+            self.__timer.cancel_timer()
 
     def reset(self):
-        self.__timer.reset_timer()
+        with self.__timer:
+            self.__timer.reset_timer()
 
     def time_since_last_call(self):
-        return self.__timer.time_since_last_call()
+        with self.__timer:
+            return self.__timer.time_since_last_call()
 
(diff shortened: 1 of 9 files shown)


rclpy/rclpy/timer.py
```diff
@@ -37,52 +36,44 @@ def __init__(self, callback, callback_group, timer_period_ns, clock, *, context=
...
     @property
     def clock(self):
         return self._clock
 
     @property
     def timer_period_ns(self):
-        with self.handle as capsule:
-            val = _rclpy.rclpy_get_timer_period(capsule)
-        self._timer_period_ns = val
+        val = self.__timer.get_timer_period()
```
sloretz, line 50: Please keep the `context` manager so that the timer can't be destroyed while this method is called.

```python
with self.__timer:
    val = self.__timer.get_timer_period()
```

sloretz: Same comment for the rest of the functions in this file

rclpy/src/rclpy/timer.cpp
```diff
@@ -33,223 +34,173 @@
...
+    rclpy_handle_get_pointer_from_capsule(pycontext.ptr(), "rcl_context_t"));
+  if (!context) {
     throw py::error_already_set();
   }
-  rcl_ret_t ret = rcl_timer_reset(timer);
+
+  // Create a client
+  rcl_timer_ = std::shared_ptr<rcl_timer_t>(
+    new rcl_timer_t,
+    [this](rcl_timer_t * timer)
```
sloretz, line 58: no need to capture `this` pointer in this lambda because it only uses the `timer` argument.

rclpy/src/rclpy/timer.hpp
```diff
@@ -17,127 +17,141 @@
...
- * Raises RCLError if the timer period could not be changed
- *
- * \param[in] pytimer Capsule pointing to the timer
- * \param[in] period_nsec the new period in nanoseconds
- */
-void
-change_timer_period(py::capsule pytimer, int64_t period_nsec);
+  /// Reset the timer
+  /**
+   * Raise ValueError if capsule is not a timer
```
sloretz, line 54: I think you can delete this line since `timer` is no longer an argument

sloretz: Same comment about deleting the `ValueError` comment on most of the functions below

sloretz (commented):
Looking good! Just a few more comments

ahcorde:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14204/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9002/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11896/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14307/)

sloretz (approved):

ahcorde:
The test failure on MacOS is unrelated and the test failure on Windows @sloretz is working on a PR

PR #743, opened 2021-03-30
Combine pybind11 modules into one

sloretz:
Part of #665 

This is a step towards getting rid of the C code in `rclpy_common`. There was trouble on Windows debug before when trying to put code that used pybind11 into `rclpy_common` that led to removing it in #727. We don't need to figure that out if all the code is in the same shared library, so this PR moves all the pybind11 code into a single shared library.

This has an added benefit of reducing the code size from 4.8 MB to 2.8 MB.

PR #749, opened 2021-04-01
Convert Clock to use a C++ Class

ahcorde:
This PR builds on top of this other PR https://github.com/ros2/rclpy/pull/745

This converts the Clock functions to a C++ class. 

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/rclpy/rclpy/clock.py
+++ b/rclpy/rclpy/clock.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from rclpy.handle import Handle
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
 from .duration import Duration
@@ -84,15 +83,15 @@ class JumpHandle:
         if threshold.min_backward is not None:
             min_backward = threshold.min_backward.nanoseconds
 
-        with self._clock.handle as clock_capsule:
-            _rclpy.rclpy_add_clock_callback(
-                clock_capsule, self, threshold.on_clock_change, min_forward, min_backward)
+        with self._clock.handle:
+            self._clock.handle.add_clock_callback(
+                 self, threshold.on_clock_change, min_forward, min_backward)
 
     def unregister(self):
         """Remove a jump callback from the clock."""
         if self._clock is not None:
-            with self._clock.handle as clock_capsule:
-                _rclpy.rclpy_remove_clock_callback(clock_capsule, self)
+            with self._clock.handle:
+                self._clock.handle.remove_clock_callback(self)
             self._clock = None
 
 
@@ -105,7 +104,7 @@ class Clock:
             self = super().__new__(ROSClock)
         else:
             self = super().__new__(cls)
-        self.__handle = Handle(_rclpy.rclpy_create_clock(clock_type))
+        self.__clock = _rclpy.Clock(clock_type)
         self._clock_type = clock_type
         return self
 
@@ -115,15 +114,15 @@ class Clock:
 
     @property
     def handle(self):
-        return self.__handle
+        return self.__clock
 
     def __repr__(self):
         return 'Clock(clock_type={0})'.format(self.clock_type.name)
 
     def now(self):
         from rclpy.time import Time
-        with self.handle as clock_capsule:
-            rcl_time = _rclpy.rclpy_clock_get_now(clock_capsule)
+        with self.handle:
+            rcl_time = self.__clock.get_now()
         return Time(nanoseconds=rcl_time.nanoseconds, clock_type=self.clock_type)
 
     def create_jump_callback(self, threshold, *, pre_callback=None, post_callback=None):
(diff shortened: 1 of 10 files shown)


rclpy/CMakeLists.txt
```diff
@@ -185,6 +185,9 @@ clean_windows_flags(_rclpy_pybind11)
 # Extension with code for Action servers and clients
 pybind11_add_module(_rclpy_action SHARED
```
sloretz, line 186: Is there an issue with a merge commit? The only two modules should be `_rclpy_pybind11` and `_rclpy_signal_handler` after #743.

rclpy/rclpy/clock.py
```diff
@@ -84,15 +83,13 @@ def __init__(self, *, clock, threshold, pre_callback, post_callback):
         if threshold.min_backward is not None:
             min_backward = threshold.min_backward.nanoseconds
 
-        with self._clock.handle as clock_capsule:
-            _rclpy.rclpy_add_clock_callback(
-                clock_capsule, self, threshold.on_clock_change, min_forward, min_backward)
+        self._clock.handle.add_clock_callback(
```
sloretz, line 86: Please keep the `with` statement so that the clock can't be destroyed while this is called.

```python3
with self._clock.handle:
    self._clock.handle.add_clock_callback
```

rclpy/rclpy/clock.py
```diff
@@ -84,15 +83,13 @@ def __init__(self, *, clock, threshold, pre_callback, post_callback):
...
+            self, threshold.on_clock_change, min_forward, min_backward)
 
     def unregister(self):
         """Remove a jump callback from the clock."""
         if self._clock is not None:
-            with self._clock.handle as clock_capsule:
-                _rclpy.rclpy_remove_clock_callback(clock_capsule, self)
+            self._clock.handle.remove_clock_callback(self)
```
sloretz, line 92: same comment about keeping the `with`

rclpy/src/rclpy/_rclpy_action.cpp
```diff
@@ -436,10 +438,11 @@ rclpy_action_create_client(
  * \return Capsule named 'rcl_action_server_t', or
  * \return NULL on failure.
  */
+ #include "clock.hpp"
 py::capsule
 rclpy_action_create_server(
   py::capsule pynode,
-  py::capsule pyclock,
+  py::object pyclock,
```
sloretz, line 445: It would be a little simpler to change `py::object pyclock` to `const rclpy::Clock & clock`, then instead of `auto clock = static_cast<rcl_clock_t *>(...)` the code can use the rcl pointer with `clock.rcl_ptr()`.

rclpy/src/rclpy/clock.cpp
```diff
@@ -34,108 +34,68 @@ using pybind11::literals::operator""_a;
...
-  auto clock = std::unique_ptr<rcl_clock_t, decltype(deleter)>(
-    static_cast<rcl_clock_t *>(PyMem_Malloc(sizeof(rcl_clock_t))),
-    deleter);
-  if (!clock) {
-    throw std::bad_alloc();
-  }
+  // Create a client
+  rcl_clock_ = std::shared_ptr<rcl_clock_t>(
+    new rcl_clock_t,
+    [this](rcl_clock_t * clock)
```
sloretz, line 42: No need to capture `this` in the lambda because it only uses the passed `clock` pointer

rclpy/src/rclpy/clock.hpp
```diff
@@ -19,97 +19,115 @@
...
+   * Raises TypeError if argument of invalid type
+   *
+   * \param[in] clock_type enum of type ClockType
+   * This constructor creates a Clock object
+   */
+  explicit Clock(rcl_clock_type_t clock_type);
+
+  /// Returns the current value of the clock
+  /**
+   * Raises ValueError if pyclock is not a clock capsule
```
sloretz, line 49: No need for this line since `pyclock` is no longer an argument

sloretz: same for a few functions below

rclpy/src/rclpy/timer.cpp
```diff
@@ -41,10 +41,12 @@ Timer::destroy()
 }
 
 Timer::Timer(
-  py::capsule pyclock, py::capsule pycontext, int64_t period_nsec)
-: clock_handle_(std::make_shared<Handle>(pyclock))
+  py::object pyclock, py::capsule pycontext, int64_t period_nsec)
```
sloretz, line 44: `py::object pyclock` -> `const Clock & clock`

rclpy/src/rclpy/timer.cpp
```diff
@@ -41,10 +41,12 @@ Timer::destroy()
...
 {
-  auto clock = clock_handle_->cast<rcl_clock_t *>("rcl_clock_t");
+  clock_handle_ = std::make_shared<rcl_clock_t>(*pyclock.cast<Clock>().rcl_ptr());
```
sloretz, line 46: I looks like this is creating a copy of the clock, but I don't think that's desired. If the original copy of the clock goes out of scope, then the clock will be `fini()`'d, and the jump callback pointers on the `rcl_clock_t` itself will point to free'd memory.

Instead, I think `Clock` should have a method that returns an `std::shared_ptr<rcl_clock_t>`, maybe

```c++
std::shared_ptr<rcl_clock_t>
rcl_shared_ptr()
{
    return rcl_clock_
}
```

and then `Timer` can store that with `: clock_handle_(clock.rcl_shared_ptr())`

rclpy/src/rclpy/timer.cpp
```diff
@@ -170,7 +172,7 @@ void
 define_timer(py::object module)
 {
   py::class_<Timer, Destroyable>(module, "Timer")
-  .def(py::init<py::capsule, py::capsule, int64_t>())
+  .def(py::init<py::object, py::capsule, int64_t>())
```
sloretz, line 175: same comment about `const Clock &`

sloretz (changes requested):
Looks like there's a case for `Destroyable` tro

rclpy/src/rclpy/timer.cpp
sloretz: @ahcorde mind addressing this comment?

ahcorde:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14211/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9011/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11904/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14314/)

ahcorde:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14213/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9013/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11906/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14316/)

sloretz (approved):

sloretz:
CI LGTM, both windows CMake warnings are in cyclonedds. Merging :tada:

PR #754, opened 2021-04-05
Check if shutdown callback weak method is valid before calling it

ivanpauno:
Fixes https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_repeated/1555/testReport/junit/(root)/projectroot/test_client/.

The issue might also be related to some of the pybind11 migration changes, but checking if the weak method is valid before calling it seems reasonable to do.


ahcorde (approved):

fujitatomoya (approved):
LGTM

PR #756, opened 2021-04-06
Convert Publisher and Subscription to use C++ Classes

ahcorde:
Part of #665.

This converts the Publisher and Subscription functions to a C++ classes. 

I have some issue with the QoSEvent destructor in some of the tests:

```bash
Thread 29 "python3" received signal SIGSEGV, Segmentation fault.
[Cambiando a Thread 0x7fffa3fff700 (LWP 338069)]
0x00000000006152cf in _is_legal_capsule (invalid_capsule=0x6d7160 "PyCapsule_GetPointer called with invalid PyCapsule object", capsule=0x3) at ../Objects/capsule.c:84
84	../Objects/capsule.c: No existe el archivo o el directorio.
(gdb) bt
#0  0x00000000006152cf in _is_legal_capsule (invalid_capsule=0x6d7160 "PyCapsule_GetPointer called with invalid PyCapsule object", capsule=0x3) at ../Objects/capsule.c:84
#1  PyCapsule_GetPointer (o=<unknown at remote 0x3>, name=0x7ffff5985881 "rcl_node_t") at ../Objects/capsule.c:84
#2  0x00007ffff6c182fb in rclpy_handle_get_pointer_from_capsule () from /home/ahcorde/ros2_galactic/install/lib/librclpy_common.so
#3  0x00007ffff5932510 in rclpy::Handle::rcl_ptr(char const*) () from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
#4  0x00007ffff5918bc3 in rcl_node_t* rclpy::Handle::cast_or_warn<rcl_node_t*>(char const*) () from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
#5  0x00007ffff593f2c4 in rclpy::Publisher::Publisher(pybind11::capsule, pybind11::object, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, pybind11::object)::{lambda(rcl_publisher_t*)#1}::operator()(rcl_publisher_t*) const [clone .lto_priv.0] () from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
[...]

PR #757, opened 2021-04-06
[pub/sub 756 addition] use std::shared_ptr holder type and shared_from_this

sloretz:
Opening a PR for comment on top of #756. I think this fixes https://github.com/ros2/rclpy/pull/756#pullrequestreview-629124458 and https://github.com/ros2/rclpy/pull/756#issuecomment-814246782

This changes the [holder type of `Destroyable` subclasses use std::shared_ptr](https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html#std-shared-ptr) and makes the subclasses of `Destroyable` use `enable_shared_from_this`. This allows classes like `QoSEvent` to keep a shared_ptr to the publisher or subscription wrapper type without creating a copy of those instances.

Opening as draft because I think I should make a bug fix PR applying these changes to Timer/Clock.

### DIFF
--- a/rclpy/src/rclpy/client.cpp
+++ b/rclpy/src/rclpy/client.cpp
@@ -155,7 +155,7 @@ Client::take_response(py::object pyresponse_type)
 void
 define_client(py::object module)
 {
-  py::class_<Client, Destroyable>(module, "Client")
+  py::class_<Client, Destroyable, std::shared_ptr<Client>>(module, "Client")
   .def(py::init<py::capsule, py::object, const char *, py::object>())
   .def_property_readonly(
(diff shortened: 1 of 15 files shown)


ahcorde:
thank you @sloretz This PR fixes the issues with the destructors ( at least locally ). Happy to merge it on my PR

sloretz:
> Happy to merge it on my PR

@ahcorde awesome :) I pushed another commit to update the Timer use the clock wrapper too in 8a75498. I'd be happy to close #758 and do that fix in #756 instead.

ahcorde (approved):

PR #758, opened 2021-04-06
Use `std::shared_ptr` holder type for Destroyable subsclasses to keep grandparent wrapper instances alive

sloretz:
Part of #665 - alternative to #757

This makes `Destroyable` subclasses use the `std::shared_ptr` holder type and use `shared_from this`. This allows classes to keep the parent wrapper instance alive so that grandparent's are kept alive too.

Since `clock` has no grandparent types, there isn't actually an issue in `master` - it's not until #756 adds Publisher and Subscription that it becomes an issue.. Maybe it's fine to leave this as is, and include this fix via #757 to #756?

### DIFF
--- a/rclpy/src/rclpy/client.cpp
+++ b/rclpy/src/rclpy/client.cpp
@@ -155,7 +155,7 @@ Client::take_response(py::object pyresponse_type)
 void
 define_client(py::object module)
 {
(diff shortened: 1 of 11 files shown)


sloretz:
Closing in favor of #757 which will hopefully be merged or included into #756

PR #728, opened 2021-03-20
Implement signal handling in Python

sloretz:
Part of #665 because it removes the signal handler CPython extension

This implements signal handling in Python using the `signal` library rather than doing it in C. I think the benefits outweigh the drawbacks.

Benefits:
* Less C code to maintain
* Don't have to port the atomics usage in the sigint module to C++
* One less compilation unit depending on `rclpy_handle_...`
* Don't need to worry about signal safety in Python signal handler

Drawbacks
* The signal handler isn't run *exactly* when the signal happens - The real signal handler is setting a flag that tells CPython to call the Python signal handlers ASAP
* If pure C code takes a really really long time, the Python signal handler won't get called until the interpreter gets to run again
  * But C code taking a long time without releasing the GIL would already be a problem on its own
* It adds an extra frame to the stack-trace

With signal handling in C

```console
$ ros2 run examples_rclpy_minimal_publisher publisher_member_function
[INFO] [1616203329.647568651] [minimal_publisher]: Publishing: "Hello World: 0"
[INFO] [1616203330.133528959] [minimal_publisher]: Publishing: "Hello World: 1"
[INFO] [1616203330.633883396] [minimal_publisher]: Publishing: "Hello World: 2"
^CTraceback (most recent call last):
  File "/home/osrf/ws/ros2/install/examples_rclpy_minimal_publisher/lib/examples_rclpy_minimal_publisher/publisher_member_function", li
ne 11, in <module>
    load_entry_point('examples-rclpy-minimal-publisher', 'console_scripts', 'publisher_member_function')()
  File "/home/osrf/ws/ros2/build/examples_rclpy_minimal_publisher/examples_rclpy_minimal_publisher/publisher_member_function.py", line
43, in main
    rclpy.spin(minimal_publisher)
  File "/home/osrf/ws/ros2/install/rclpy/lib/python3.8/site-packages/rclpy/__init__.py", line 196, in spin
    executor.spin_once()
  File "/home/osrf/ws/ros2/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 693, in spin_once
    handler, entity, node = self.wait_for_ready_callbacks(timeout_sec=timeout_sec)
  File "/home/osrf/ws/ros2/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 679, in wait_for_ready_callbacks
    return next(self._cb_iter)
[...]

PR #771, opened 2021-04-15
Convert Node and Context to use C++ Classes

ahcorde:

This converts the Node and Context functions to a C++ class. 

This PR depends on this other PR https://github.com/ros2/rclpy/pull/728 and include the changes of this branch about [GuardCondition to use C++ classes](https://github.com/ros2/rclpy/tree/ahcorde/pybind11-guardcondition).

There is one test which is failing:

 - test_declare_qos_parameters_with_unhappy_callback

I will create this as a draft, we need to decide if we are going to backport the signal PR or find a workaround.


ahcorde:
The GuardCondition PR is in, I open this PR for review.

The test `test_declare_qos_parameters_with_unhappy_callback` is still failing:, do you mind to have a look @hidmic ?

rclpy/src/rclpy/node.hpp
```diff
@@ -17,157 +17,161 @@
...
+    bool enable_rosout);
+
+  /// Get the fully qualified name of the node.
+  /**
+   * Raises RCLError if name is not set
+   *
+   * \return String containing the fully qualified name of the node
+   */
+  const char *
+  get_node_fully_qualified_name();
```
hidmic, line 63: @ahcorde nit: do we need that `node_` token? Isn't it implied when doing `node.get_fully_qualified_name()`? Same for other APIs below.

rclpy/src/rclpy/node.hpp
```diff
@@ -17,157 +17,161 @@
...
+   * Raises RCLError if the names are unavailable.
+   *
+   * \param[in] get_enclaves specifies if the output includes the enclaves names
+   *            or not
+   * \return Python list of tuples, containing:
+   *  node name, node namespace, and
+   *  enclave if `get_enclaves` is true.
+   */
+  py::list
+  get_node_names_impl(bool get_enclaves);
```
hidmic, line 124: @ahcorde hmm, should this be private?

rclpy/test/test_destruction.py
```diff
@@ -38,8 +38,8 @@ def test_destroy_node_twice():
     try:
         node = rclpy.create_node('test_node2', context=context)
         node.destroy_node()
-        with pytest.raises(InvalidHandle):
-            node.destroy_node()
+        # with pytest.raises(InvalidHandle):
```
hidmic, line 41: @ahcorde hmm, doesn't it throw anymore? Can we destroy a node twice?

rclpy/test/test_subscription.py
```diff
@@ -48,7 +49,13 @@ def make_mock_subscription(namespace, topic_name, cli_args=None):
     ('/example/topic', 'ns', '/example/topic'),
 ])
 def test_get_subscription_topic_name(topic_name, namespace, expected):
-    sub = make_mock_subscription(namespace, topic_name)
+    node = Node('node_name', namespace=namespace, cli_args=None)
```
hidmic, line 52: @ahcorde why copy `make_mock_subscription` implementation here? Same below.

hidmic (commented):
First pass !

rclpy/test/test_subscription.py
ahcorde: the node only exists in the mock function, then the object will be destroy becuase there is no reference to the node. I can return the node too to keep the object alive.

rclpy/src/rclpy/node.cpp
```diff
@@ -179,15 +179,15 @@ Node::get_node_names_impl(bool get_enclaves)
 }
 
 py::list
-Node::get_node_names_and_namespaces()
+Node::get_names_and_namespaces()
```
hidmic, line 182: @ahcorde I'd suggest we keep `node` in the name here. It's not a reference to the object itself but to remote instances. Same below.

rclpy/src/rclpy/node.hpp
```diff
@@ -60,7 +60,7 @@ class Node : public Destroyable, public std::enable_shared_from_this<Node>
    * \return String containing the fully qualified name of the node
    */
   const char *
-  get_node_fully_qualified_name();
+  fully_qualified_name();
```
hidmic, line 63: @ahcorde nit: why `fully_qualified_name()` and `get_namespace()`? I mean, why not always putting `get_`? Or never?  Consistency would be best.

rclpy/test/test_destruction.py
```diff
@@ -38,7 +38,6 @@ def test_destroy_node_twice():
     try:
         node = rclpy.create_node('test_node2', context=context)
         node.destroy_node()
-        # with pytest.raises(InvalidHandle):
```
hidmic: @ahcorde https://github.com/ros2/rclpy/pull/771#discussion_r615032651 still holds.

rclpy/src/rclpy/node.hpp
```diff
@@ -17,157 +17,161 @@
...
+   * Raises UnknownROSArgsError if there are unknown CLI arguments given
+   * Raises MemoryError if memory could not be allocated for the node
+   *
+   * \param[in] node_name name of the node to be created
+   * \param[in] namespace namespace for the node
+   * \param[in] context Context
+   * \param[in] pycli_args a sequence of command line arguments for just this node, or None
+   * \param[in] use_global_arguments if true then the node will also use cli arguments on context
+   * \param[in] enable rosout if true then enable rosout logging
+   * \return Capsule of the pointer to the created rcl_node_t * structure
```
hidmic, line 46: @ahcorde hmm, leftover?
```suggestion
```

rclpy/test/test_destruction.py
```diff
@@ -162,11 +161,8 @@ def test_destroy_node_asap():
 
     try:
         node = rclpy.create_node('test_destroy_subscription_asap', context=context)
-        with node.handle:
-            node.destroy_node()
-            # handle valid because it's still being used
-            with node.handle:
-                pass
```
hidmic, line 169: @ahcorde why the change?

rclpy/test/test_destruction.py
```diff
@@ -38,8 +38,7 @@ def test_destroy_node_twice():
     try:
         node = rclpy.create_node('test_node2', context=context)
         node.destroy_node()
-        with pytest.raises(InvalidHandle):
-            node.destroy_node()
+        node.destroy_node()
```
hidmic, line 41: @ahcorde https://github.com/ros2/rclpy/pull/771#discussion_r615032651 still applies.

rclpy/test/test_handle.py
```diff
@@ -38,9 +38,7 @@ def test_handle_destroyed_when_not_used():
     try:
         node = rclpy.create_node('test_handle_destroyed_when_not_used', context=context)
         with node.handle:
-            node.handle.destroy()
-            with node.handle:
-                pass
```
hidmic, line 43: @ahcorde why the change?

rclpy/src/rclpy/node.cpp
```diff
@@ -534,22 +468,48 @@ create_node(
...
+    "Get the namespace of a node.")
+  .def(
+    "get_count_publishers", &Node::get_count_publishers,
+    "Returns the count of all the publishers known for that topic in the entire ROS graph.")
+  .def(
+    "get_count_subscribers", &Node::get_count_subscribers,
+    "Returns the count of all the subscribers known for that topic in the entire ROS graph.")
+  .def(
+    "get_names_impl", &Node::get_names_impl,
+    "Get the list of nodes discovered by the provided node")
```
hidmic, line 503: @ahcorde this is now a private method.

rclpy/src/rclpy/context.cpp
```diff
@@ -30,93 +30,93 @@
...
+      }
+      delete context;
+    });
+  *rcl_context_ = rcl_get_zero_initialized_context();
+}
 
+void
+Context::destroy()
+{
+  rcl_context_.reset();
```
hidmic, line 72: @ahcorde I see the test you mention `SIGABRT`'ing. I think it's because we're not ensuring destruction order. I see `Context` dependent objects keeping shared ownership of that `Context`, but that won't prevent `rcl_context_t` destruction if something calls `Context::destroy()` on it.

rclpy/test/test_destruction.py
sloretz: > `# handle valid because it's still being used`

This was an intentional change in #739 with `Destroyable`. Once someone asks to destroy a `Destroyable`, it will block new uses of it. The `Handle` class allowed new uses as long as someone else was already using it, but that could lead to two threads blocking destruction forever as long as their uses overlapped.

rclpy/src/rclpy/context.cpp
```diff
@@ -30,93 +30,93 @@
...
+        rcl_ret_t ret;
+        if (rcl_context_is_valid(context)) {
+          // shutdown first, if still valid
+          ret = rcl_shutdown(context);
+          if (RCL_RET_OK != ret) {
+            RCUTILS_SAFE_FWRITE_TO_STDERR(
+              "[rclpy|" RCUTILS_STRINGIFY(__FILE__) ":" RCUTILS_STRINGIFY(__LINE__) "]: "
+              "failed to shutdown rcl_context_t");
+            RCUTILS_SAFE_FWRITE_TO_STDERR(rcl_get_error_string().str);
+            RCUTILS_SAFE_FWRITE_TO_STDERR("\n");
```
sloretz, line 50: I recommend using Python warnings instead of writing to stderr

https://github.com/ros2/rclpy/blob/2bb0a8065d437668b145a71c165193365b116312/rclpy/src/rclpy/client.cpp#L68-L74

rclpy/src/rclpy/context.cpp
sloretz: This seems like a design problem.

`rcl_context_t` needs to outlive `rcl_node_t`
`rcl_node_t` needs to outlive `rcl_publisher_t`.

`rclpy::Publisher` keeps `rcl_publisher_t` alive, until destructed or `destroy()`
`rclpy::Publisher` keeps `rclpy::Node` alive, until destructed or `destroy()`

`rclpy::Node` keeps `rcl_node_t` alive, until destructed or `destroy()`
`rclpy::Node` keeps `rclpy::Context` alive, until destructed or `destroy()`

`rclpy::Context` keeps `rcl_context_t` alive, until destructed or `destroy()`


`rclpy::Publisher` holding `std::shared_ptr<rclpy::Node>` is assuming keeping `rclpy::Node` alive will keep `rcl_node_t` and `rcl_context_t` alive, but that isn't true.
`rclpy::Node` holding `std::shared_ptr<rclpy::Context>` is assuming keeping `rclpy::Context` alive will keep `rcl_context_t` alive, but that isn't true.

How do we fix this?

rclpy/src/rclpy/_rclpy_action.cpp
```diff
@@ -75,18 +76,12 @@ rclpy_action_get_rmw_qos_profile(const char * rmw_profile)
 
 py::object
 rclpy_action_get_client_names_and_types_by_node(
```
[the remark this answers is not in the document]
ahcorde: @sloretz, what about the other functions in `rclpy` that also take a node as the first argument? for example in `names.hpp` or `graph.cpp`. Should I move them to the Node class?

rclpy/rclpy/node.py
```diff
@@ -305,21 +304,27 @@ def handle(self):
...
         :raises: AttributeError if modified after creation.
         """
-        return self.__handle
+        return self.__node
 
     @handle.setter
     def handle(self, value):
         raise AttributeError('handle cannot be modified after node creation')
 
+    def __del__(self):
```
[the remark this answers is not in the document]
ahcorde: But we should call here `destroy_node()` otherwise the destruction fails

rclpy/test/test_destruction.py
hidmic: That makes total sense. I still think we should test that usage attempts after destruction requests fail.

rclpy/src/rclpy/context.cpp
hidmic: Hmm, to replicate the destruction semantics we have today, having each wrapper object hold shared ownership of its parents' underlying C structs seems like the only path forward. It's not ideal, and somewhat defeats the purpose of early destruction as long as you have some dependent object in the garbage collector hands, but that's what `rclpy_handle_t` did.

Alternatively, we could have each wrapper object hold weak ownership of its children, so as to propagate an early destruction request down the dependency tree. Thoughts?

rclpy/rclpy/node.py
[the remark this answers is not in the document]
sloretz: > But we should call here `destroy_node()` otherwise the destruction fails

What does that mean? What does the failure look like?

rclpy/src/rclpy/context.cpp
sloretz: > to replicate the destruction semantics we have today, having each wrapper object hold shared ownership of its parents' underlying C structs seems like a path forward.

This seems like a reasonable path forward. Children need some way to keep the `rcl` part of their parent's alive. One way is to put those `rcl` pointers into their own structs.

<details><summary>Psuedocode of holding parent's C structs</summary>

```C++
class rclpy::Node
{
  struct RclPtrs
  {
    std::shared_ptr<rclpy::Context::RclPtrs> context_ptrs;
    std::unique_ptr<rcl_node_t> rcl_node;
  };

  /// Return RCL pointers so another class can keep the rcl part alive
  std::shared_ptr<rclpy::Node::RclPtrs>
  get_rcl_ptrs() const
  {
    return rcl_ptrs_;
  }

  void
  destroy()
  {
    // Drop references here, and rely on Python `class Node` to destroy children so all shared references to struct are dropped
    rcl_ptrs_.reset();
  }

  std::shared_ptr<rclpy::Node::RclPtrs> rcl_ptrs_;
}


class rclpy::Publisher
{
  struct RclPtrs
  {
    // Store node_ptrs in here so rcl_event_t can keep node ptrs alive too
    std::shared_ptr<rclpy::Node::RclPtrs> node_ptrs;
    std::unique_ptr<rcl_publisher_t> rcl_publisher;
  };

  Publisher(const rclpy::Node & node)
  : rcl_ptrs_(new rclpy::Publisher::RclPtrs)
  {
    rcl_ptrs_->node_ptrs = node->get_rcl_ptrs();
    /// ...
  }

  void
  destroy()
  {
    rcl_ptrs_.reset();
  }

  std::shared_ptr<rclpy::Publisher::RclPtrs> rcl_ptrs_;
}
```
</details>

The downside of this way is the publisher instance no longer has access to the Node's API. Maybe the publisher can have both if it intentionally makes a copy of  `rclpy::Node`?

<details><summary>Psuedocode of intentional node copy</summary>

```c++
class rclpy::Node
{
  // Copy constructor
  Node(const Node & other)
  : rcl_context_(other.rcl_context_), rcl_node_(other.rcl_node_)
  {
    // Gives copy shared ownership of the rcl pointers
  }

  void
  destroy()
  {
    // Drop references if we can, and rely on Python `class Node` to destroy children so all copies are destroyed too
    rcl_ptrs_.reset();
  }

  // Shared pointers so ownership can be shared with copies
  std::shared_ptr<rcl_context_t> rcl_context_;
  std::shared_ptr<rcl_node_t> rcl_node_;
}


class rclpy::Publisher
{
  Publisher(const rclpy::Node & node)
  : node_(node)  // Make a copy
  {
  }

  void
  destroy()
  {
    rcl_publisher_.reset();
    node_.destroy();
  }

  // Publisher has both access to the `rclpy::Node` API, and ownership of the rcl pointers
  rclpy::Node node_;
  std::shared_ptr<rcl_publisher_t> rcl_publisher_;
}
```
</details>

> It's not ideal, and somewhat defeats the purpose of early destruction as long as you have some dependent object in the garbage collector hands, but that's what rclpy_handle_t did.

I think the expectation with `rclpy_handle_t` was the Python object, `class Node`, took care of that in `destroy_node()` by calling `destroy()` on all of it's children first. I think that expectation can stay the same.

> Alternatively, we could have each wrapper object hold weak ownership of its children, so as to propagate an early destruction request down the dependency tree. Thoughts?

Say `Node` has a weak pointer to `Publisher`, and a user calls `Node::destroy()`. `Node` doesn't call `Publisher::destroy()`, because what if someone is still using `Publisher`? If `Node` calls `Publisher::destroy_when_not_in_use()` and `Publisher` is busy, then what? Does the publisher use its reference to the `Node` so that it can call `Node::destroy()` again when the `Publisher` is destroyed?

Node` having a weak reference to `Publisher` means `Node` must be told when the publisher depends on it. Maybe that would be `Publisher` calling an API on `Node` to say it was constructed?

I wish `rcl` handled destruction order - maybe by blocking `rcl_node_t` from being fini'd as long as another `rcl_*_t` was using it.

ahcorde: I implemented the first solution that @sloretz mentioned in this commit https://github.com/ros2/rclpy/pull/771/commits/7b0c4dcbb4a6d8f0904087004a92a99221c3754e

But I still have some problems, in this case with `test_publisher`, this is the backtrace:

```
Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
0x00007fffce1b1b2f in std::enable_if<std::__and_<std::__not_<std::__is_tuple_like<rcl_publisher_t*> >, std::is_move_constructible<rcl_publisher_t*>, std::is_move_assignable<rcl_publisher_t*> >::value, void>::type std::swap<rcl_publisher_t*>(rcl_publisher_t*&, rcl_publisher_t*&) () from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
(gdb) bt
#0  0x00007fffce1b1b2f in std::enable_if<std::__and_<std::__not_<std::__is_tuple_like<rcl_publisher_t*> >, std::is_move_constructible<rcl_publisher_t*>, std::is_move_assignable<rcl_publisher_t*> >::value, void>::type std::swap<rcl_publisher_t*>(rcl_publisher_t*&, rcl_publisher_t*&) () from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
#1  0x00007fffce1b09e3 in std::unique_ptr<rcl_publisher_t, std::function<void (rcl_publisher_t*)> >::reset(rcl_publisher_t*) ()
   from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
#2  0x00007fffce1afd21 in rclpy::Publisher::destroy() () from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
#3  0x00007fffce18a5fc in rclpy::Destroyable::destroy_when_not_in_use() () from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
#4  0x00007fffce18b3e3 in pybind11::cpp_function::cpp_function<void, rclpy::Destroyable, , pybind11::name, pybind11::is_method, pybind11::sibling, char [74]>(void (rclpy::Destroyable::*)(), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&, char const (&) [74])::{lambda(rclpy::Destroyable*)#1}::operator()(rclpy::Destroyable*) const ()
   from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
#5  0x00007fffce18d105 in void pybind11::detail::argument_loader<rclpy::Destroyable*>::call_impl<void, pybind11::cpp_function::cpp_function<void, rclpy::Destroyable, , pybind11::name, pybind11::is_method, pybind11::sibling, char [74]>(void (rclpy::Destroyable::*)(), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&, char const (&) [74])::{lambda(rclpy::Destroyable*)#1}&, 0ul, pybind11::detail::void_type>(pybind11::cpp_function::cpp_function<void, rclpy::Destroyable, , pybind11::name, pybind11::is_method, pybind11::sibling, char [74]>(void (rclpy::Destroyable::*)(), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&, char const (&) [74])::{lambda(rclpy::Destroyable*)#1}&, std::integer_sequence<unsigned long, 0ul>, pybind11::detail::void_type&&) && ()
   from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
#6  0x00007fffce18c2fe in std::enable_if<std::is_void<void>::value, pybind11::detail::void_type>::type pybind11::detail::argument_loader<rclpy::Destroyable*>::call<void, pybind11::detail::void_type, pybind11::cpp_function::cpp_function<void, rclpy::Destroyable, , pybind11::name, pybind11::is_method, pybind11::sibling, char [74]>(void (rclpy::Destroyable::*)(), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&, char const (&) [74])::{lambda(rclpy::Destroyable*)#1}&>(pybind11::cpp_function::cpp_function<void, rclpy::Destroyable, , pybind11::name, pybind11::is_method, pybind11::sibling, char [74]>(void (rclpy::Destroyable::*)(), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&, char const (&) [74])::{lambda(rclpy::Destroyable*)#1}&) && ()
   from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
#7  0x00007fffce18bd79 in pybind11::cpp_function::initialize<pybind11::cpp_function::initialize<void, rclpy::Destroyable, , pybind11::name, pybind11::is_method, pybind11::sibling, char [74]>(void (rclpy::Destroyable::*)(), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&, char const (&) [74])::{lambda(rclpy::Destroyable*)#1}, void, rclpy::Destroyable*, pybind11::name, pybind11::is_method, pybind11::sibling, char [74]>(pybind11::cpp_function::initialize<void, rclpy::Destroyable, , pybind11::name, pybind11::is_method, pybind11::sibling, char [74]>(void (rclpy::Destroyable::*)(), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&, char const (&) [74])::{lambda(rclpy::Destroyable*)#1}&&, void (*)(rclpy::Destroyable*), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&, char const (&) [74])::{lambda(pybind11::detail::function_call&)#3}::operator()(pybind11::detail::function_call) const ()
   from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
#8  0x00007fffce18bde1 in pybind11::cpp_function::initialize<pybind11::cpp_function::initialize<void, rclpy::Destroyable, , pybind11::name, pybind11::is_method, pybind11::sibling, char [74]>(void (rclpy::Destroyable::*)(), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&, char const (&) [74])::{lambda(rclpy::Destroyable*)#1}, void, rclpy::Destroyable*, pybind11::name, pybind11::is_method, pybind11::sibling, char [74]>(pybind11::cpp_function::initialize<void, rclpy::Destroyable, , pybind11::name, pybind11::is_method, pybind11::sibling, char [74]>(void (rclpy::Destroyable::*)(), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&, char const (&) [74])::{lambda(rclpy::Destroyable*)#1}&&, void (*)(rclpy::Destroyable*), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&, char const (&) [74])::{lambda(pybind11::detail::function_call&)#3}::_FUN(pybind11::detail::function_call) ()
   from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
#9  0x00007fffce0f4095 in pybind11::cpp_function::dispatcher(_object*, _object*, _object*) () from /home/ahcorde/ros2_galactic/build/rclpy/test_rclpy/_rclpy_pybind11.cpython-38-x86_64-linux-gnu.so
#10 0x00000000005f2fb9 in cfunction_call_varargs (kwargs=<optimized out>, args=<optimized out>, func=<built-in method destroy_when_not_in_use of PyCapsule object at remote 0x7fffce2c4b40>)
    at ../Objects/call.c:772
#11 PyCFunction_Call (func=<built-in method destroy_when_not_in_use of PyCapsule object at remote 0x7fffce2c4b40>, args=<optimized out>, kwargs=<optimized out>) at ../Objects/call.c:772
#12 0x00000000005f3446 in _PyObject_MakeTpCall (callable=<built-in method destroy_when_not_in_use of PyCapsule object at remote 0x7fffce2c4b40>, args=<optimized out>, nargs=<optimized out>, 
    keywords=<optimized out>) at ../Include/internal/pycore_pyerrors.h:13
#13 0x000000000050aa8b in _PyObject_Vectorcall (kwnames=0x0, nargsf=<optimized out>, args=0x7fffcc06e6d0, callable=<built-in method destroy_when_not_in_use of PyCapsule object at remote 0x7fffce2c4b40>)
    at ../Include/cpython/abstract.h:125
#14 _PyObject_Vectorcall (kwnames=0x0, nargsf=<optimized out>, args=0x7fffcc06e6d0, callable=<built-in method destroy_when_not_in_use of PyCapsule object at remote 0x7fffce2c4b40>)
    at ../Include/cpython/abstract.h:115
#15 method_vectorcall (method=<optimized out>, args=0x7fffcc06e6d8, nargsf=<optimized out>, kwnames=0x0) at ../Objects/classobject.c:60
#16 0x000000000056ef97 in _PyObject_Vectorcall (kwnames=0x0, nargsf=<optimized out>, args=0x7fffcc06e6d8, callable=<method at remote 0x7ffff5b75180>) at ../Include/cpython/abstract.h:127
#17 call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>, tstate=0x962f70) at ../Python/ceval.c:4963
#18 _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at ../Python/ceval.c:3469
#19 0x00000000005f5e56 in PyEval_EvalFrameEx (throwflag=0, 
    f=Frame 0x7fffcc06e550, for file /home/ahcorde/ros2_galactic/src/ros2/rclpy/rclpy/rclpy/publisher.py, line 92, in destroy (self=<Publisher(_Publisher__publisher=<test_rclpy._rclpy_pybind11.Publisher at remote 0x7fffce01dfb0>, msg_type=<Metaclass_BasicTypes(__module__='test_msgs.msg._basic_types', __doc__="Message class 'BasicTypes'.", __slots__=['_bool_value', '_byte_value', '_char_value', '_float32_value', '_float64_value', '_int8_value', '_uint8_value', '_int16_value', '_uint16_value', '_int32_value', '_uint32_value', '_int64_value', '_uint64_value'], _fields_and_field_types={'bool_value': 'boolean', 'byte_value': 'octet', 'char_value': 'uint8', 'float32_value': 'float', 'float64_value': 'double', 'int8_value': 'int8', 'uint8_value': 'uint8', 'int16_value': 'int16', 'uint16_value': 'uint16', 'int32_value': 'int32', 'uint32_value': 'uint32', 'int64_value': 'int64', 'uint64_value': 'uint64'}, SLOT_TYPES=(<BasicType at remote 0x7fffce344790>, <BasicType at remote 0x7fffce3447c0>, <BasicType at remote 0x7fffce344...(truncated))
    at ../Python/ceval.c:741
#20 function_code_fastcall (globals=<optimized out>, nargs=<optimized out>, args=<optimized out>, co=<optimized out>) at ../Objects/call.c:283
#21 _PyFunction_Vectorcall (func=<optimized out>, stack=<optimized out>, nargsf=<optimized out>, kwnames=<optimized out>) at ../Objects/call.c:410
#22 0x000000000056a136 in _PyObject_Vectorcall (kwnames=0x0, nargsf=<optimized out>, args=0x7fffcc069b28, callable=<function at remote 0x7fffcc2edb80>) at ../Include/cpython/abstract.h:127
```

hidmic: >  If `Node` calls `Publisher::destroy_when_not_in_use()` and `Publisher` is busy, then what?

Then neither gets destroyed, but both are "scheduled" for destruction, blocking further use cases. Then, if you have dependent objects increasing/decreasing their dependencies' use count in topological order when theirs is (i.e. on `__enter__` and on `__exit__`), then they'll get destroyed once all ongoing tasks are complete.

> `Node` having a weak reference to `Publisher` means `Node` must be told when the `Publisher` depends on it. 

Indeed. But that's something we can do internally, when the `Publisher` is constructed through a non-binded API. We can also use friendship if we want to hide that API in C++.

sloretz: @ahcorde I pushed a few commits to this branch that get all tests except cppcheck to pass, but it's not quite complete. The ActionClient, Client, Timer, etc classes need to use the RclPtrs structs of their parents too or they won't keep the `rcl` part of them alive.

895ee4b prevents a Destroyable from being destroyed twice. This was happening because of the publisher being explicitly destroyed in `test_publisher`, and `Node.__del__` was trying to destroy the publisher again. I removed the `Node.__del__` in 21e80db. It shouldn't be needed if everything else is working correctly. The remaining bugs were fixed by reordering the pointers in the `RclPtrs` structs in ad53180. C++ destroys structs in reverse order, so pointers to parents need to be placed above children created from them.

After looking at it I think I like the intentional copy approach more. I'll try prototyping that and open a PR to this branch if it works.

ahcorde: @sloretz I prototyped the second option in this branch https://github.com/ros2/rclpy/pull/new/ahcorde/pybind11-node_option2, 

I cherry-picked two of your improvements but I'm having some issues with qos_event tests

sloretz: @ahcorde I opened https://github.com/ros2/rclpy/pull/773 which has a few differences, and all tests pass. Mind taking a look?

The qos_event tests are written in a way that makes them flaky when we do destruction ordering correctly, but I included a fix on my branch.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14459/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9229/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12130/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14561/)

ahcorde:
> @sloretz, what about the other functions in `rclpy` that also take a node as the first argument? for example in `names.hpp` or `graph.cpp`. Should I move them to the Node class?

This comment is still pending, what do you think @sloretz ?

sloretz:
>  @sloretz, what about the other functions in rclpy that also take a node as the first argument? for example in names.hpp or graph.cpp. Should I move them to the Node class?

I think those are fine actually. I would like to get rid of the `_rcl_action.cpp` file, but I think I was wrong to say everything should be a method on the Node. Maybe some of the action functions could be moved to `names.hpp` or `graph.hpp` themselves?

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14479/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9243/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12146/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14579/)

rclpy/rclpy/context.py
```diff
@@ -35,16 +35,21 @@ class Context:
...
         self._callbacks_lock = threading.Lock()
         self._logging_initialized = False
 
     @property
     def handle(self):
-        return self._handle
+        return self.__context
+
+    def __del__(self):
+        self.destroy()
```
sloretz, line 49: I'd recommend removing the `__del__` method. It shouldn't be necessary.

rclpy/src/rclpy/init.cpp
```diff
@@ -62,14 +62,8 @@ struct InitOptions
 };
 
 void
-init(py::list pyargs, py::capsule pycontext, size_t domain_id)
+init(py::list pyargs, Context & context, size_t domain_id)
```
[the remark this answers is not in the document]
sloretz: Is the idea to move the logic in `init(...)` into a constructor of `Context`? That sounds fine to me. I don't think the zero initialized context is used anywhere else.

rclpy/src/rclpy/publisher.cpp
```diff
@@ -19,9 +19,9 @@
 
 #include <memory>
 #include <string>
+#include <iostream>
```
sloretz, line 22: Unnecessary include?

rclpy/src/rclpy/subscription.hpp
```diff
@@ -22,9 +22,10 @@
 
 #include <memory>
 #include <string>
+#include <iostream>
```
sloretz, line 25: Unnecessary include?

rclpy/src/rclpy/timer.cpp
```diff
@@ -37,20 +37,14 @@ void
 Timer::destroy()
 {
   rcl_timer_.reset();
-  clock_handle_.reset();
+  clock_.destroy();
+  context_.destroy_when_not_in_use();
```
sloretz, line 41: I think `context_.destroy()` is safe to use here since the `Timer` owns a copy, but it probably doesn't change anything.

sloretz (commented):
Just some nitpicks, otherwise this PR looks pretty good!

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14488/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9248/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12154/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14588/)

rclpy/rclpy/node.py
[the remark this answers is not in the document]
sloretz: @ahcorde mind removing this method?

rclpy/rclpy/node.py
```diff
@@ -305,21 +304,27 @@ def handle(self):
...
+        self.destroy()
+
+    def destroy(self):
```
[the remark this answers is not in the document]
sloretz: @ahcorde mind removing this method?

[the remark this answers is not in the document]
ahcorde: @sloretz it's already removed https://github.com/ros2/rclpy/blob/ahcorde/pybind11-node/rclpy/rclpy/node.py#L310-L313

rclpy/rclpy/node.py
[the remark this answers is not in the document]
ahcorde: @sloretz it's already removed https://github.com/ros2/rclpy/blob/ahcorde/pybind11-node/rclpy/rclpy/node.py#L310-L313

rclpy/rclpy/node.py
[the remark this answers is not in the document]
sloretz: Ah weird, the github UI is still showing it for me. I'll look at the new commits locally :-/

rclpy/rclpy/context.py
```diff
@@ -60,28 +60,29 @@ def init(self,
...
 
         global g_logging_ref_count
-        with self._handle as capsule, self._lock:
+        with self._lock:
             if domain_id is not None and domain_id < 0:
                 raise RuntimeError(
                     'Domain id ({}) should not be lower than zero.'
                     .format(domain_id))
-            rclpy_implementation.rclpy_init(
+            self.__context = rclpy_implementation.Context(
```
sloretz, line 68: I think this should check if `self.__context` already exists, and error if that's the case. If we don't do that then it's possible to initialize two contexts such that the first context lives until it's garbage collected.

rclpy/test/test_init_shutdown.py
```diff
@@ -61,8 +61,7 @@ def test_double_init():
     context = rclpy.context.Context()
     rclpy.init(context=context)
     try:
-        with pytest.raises(RuntimeError):
-            rclpy.init(context=context)
+        rclpy.init(context=context)
```
sloretz, line 64: See above comment. I think the second call to init should throw.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14496/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9256/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12164/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14597/)

sloretz (approved):
LGTM! Thanks for iterating! :tada:

Issue #775, opened 2021-04-28
Bad performance of python logging function with severity below configured


ksuszka:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - offical foxy docker image
- Version or commit hash:
  - latest foxy branch, but it seems the same for master
- DDS implementation:
  - CycloneDDS but doesn't really matter
- Client library (if applicable):
  - rclcpy

#### Steps to reproduce issue

Behaviour observed when profiling slow performance of ros2 carla_ros_bridge package. After fixing some issues with data transfer between python/C (same issue as described in https://github.com/ros2/rclpy/issues/763), I've observed that log function takes a lot of CPU time even with log level set to some high value.

Top lines from profiler output:
```
     1000    0.005    0.000   44.910    0.045 /work/ws/install/carla_ros_bridge/lib/python3.8/site-packages/carla_ros_bridge/bridge.py:351(_update)
     1000    0.087    0.000   44.904    0.045 /work/ws/install/carla_ros_bridge/lib/python3.8/site-packages/carla_ros_bridge/actor_factory.py:133(update_actor_states)
    12000    0.062    0.000   41.855    0.003 /work/ws/install/carla_ros_bridge/lib/python3.8/site-packages/carla_ros_bridge/sensor.py:244(update)
     9000    0.161    0.000   35.317    0.004 /work/ws/install/carla_ros_bridge/lib/python3.8/site-packages/carla_ros_bridge/sensor.py:217(_update_synchronous_sensor)
     4000   15.996    0.004   20.793    0.005 /work/ws/install/carla_ros_bridge/lib/python3.8/site-packages/carla_ros_bridge/sensor.py:275(create_cloud)
     1000    0.068    0.000   11.404    0.011 /work/ws/install/carla_ros_bridge/lib/python3.8/site-packages/carla_ros_bridge/lidar.py:63(sensor_data_updated)
     1000    0.699    0.001   10.636    0.011 /work/ws/install/carla_ros_bridge/lib/python3.8/site-packages/carla_ros_bridge/lidar.py:132(sensor_data_updated)
    12000    0.026    0.000   10.333    0.001 /work/ws/install/ros_compatibility/lib/python3.8/site-packages/ros_compatibility/ros_compatible_node.py:281(logdebug)
    12000    0.077    0.000   10.304    0.001 /opt/ros/foxy/lib/python3.8/site-packages/rclpy/impl/rcutils_logger.py:326(debug)
[...]

PR #776, opened 2021-04-28
Break log function execution ASAP if configured severity is too high

ksuszka:
An attempt to solve https://github.com/ros2/rclpy/issues/775.

### DIFF
--- a/rclpy/rclpy/impl/rcutils_logger.py
+++ b/rclpy/rclpy/impl/rcutils_logger.py
@@ -274,12 +274,12 @@ class RcutilsLogger:
         :raises: ValueError on invalid parameters values.
         :rtype: bool
         """
-        from rclpy.logging import LoggingSeverity
-        severity = LoggingSeverity(severity)
-
         # Gather context info and check filters only if the severity is appropriate.
         if not self.is_enabled_for(severity):
             return False
+
+        from rclpy.logging import LoggingSeverity
+        severity = LoggingSeverity(severity)
 
         name = kwargs.pop('name', self.name)
 
--- a/rclpy/rclpy/impl/rcutils_logger.py
+++ b/rclpy/rclpy/impl/rcutils_logger.py
@@ -277,6 +277,10 @@ class RcutilsLogger:
         from rclpy.logging import LoggingSeverity
         severity = LoggingSeverity(severity)
 
+        # Gather context info and check filters only if the severity is appropriate.
+        if not self.is_enabled_for(severity):
+            return False
+
         name = kwargs.pop('name', self.name)
 
         # Infer the requested log filters from the keyword arguments
@@ -305,10 +309,6 @@ class RcutilsLogger:
                 if any(context[p] != kwargs.get(p, filter_params[p]) for p in filter_params):
                     raise ValueError(
                         'Logging filter parameters cannot be changed between calls.')
-
-        # Only check filters if the severity is appropriate.
-        if not self.is_enabled_for(severity):
-            return False
 
         # Check if any filter determines the message shouldn't be processed.
         # Note(dhood): even if a message doesn't get logged, a filter might still update its state


rclpy/rclpy/impl/rcutils_logger.py
```diff
@@ -277,6 +277,10 @@ def log(self, message, severity, **kwargs):
         from rclpy.logging import LoggingSeverity
         severity = LoggingSeverity(severity)
 
+        # Gather context info and check filters only if the severity is appropriate.
+        if not self.is_enabled_for(severity):
+            return False
+
```
clalancette, line 283: This is an obvious improvement, but I wonder if we can go a little further.  In particular, it looks like `is_enabled_for` already wraps the incoming `severity` in a LoggingSeverity object, so if we put this block of code as the very first thing in the method, we will save a tiny bit more amount of time.

clalancette (changes requested):
One minor change, but this seems reasonable to me.  @sloretz I'd appreciate thoughts from you as well.

rclpy/rclpy/impl/rcutils_logger.py
sloretz: If this PR is optimized enough for the contributor's use case, and we were fine with the performance before, then I would suggest merging as is.

ksuszka: I did some simple performance tests. Bear in mind, I'm not a python programmer, so there is a high chance I committed some blasphemy ;).

I wrote very simple test script:
```
import rclpy
import timeit
rclpy.init()

print(timeit.timeit('z.is_enabled_for(rclpy.logging.LoggingSeverity.FATAL)',setup='import rclpy\nz=rclpy.create_node("a").get_logger()',number=10000))
print(timeit.timeit('z.debug("ss")',setup='import rclpy\nz=rclpy.create_node("a").get_logger()',number=10000))
print(timeit.timeit('z.fatal("ss")',setup='import rclpy\nz=rclpy.create_node("a").get_logger()',number=10000))

```

For orignal code I get:
```
0.013552845921367407
9.574878340121359
9.609605725854635
```

For current modification I get:
```
0.014010991901159286
0.031755123753100634
9.542557789944112
```

If I move check above LoggingSeverity import (so import and conversion are done once if severity is not high enough) I get:
```
0.013952112291008234
0.0214669331908226
9.54146963218227
```
So there is some additional performance boost.

However, after that, I've checked what is the cost of doing this import every single time - I modified code, so LoggingSeverity class is imported in `__init__` method and stored in `self.LoggingSeverity` variable, and then it is used in is_enabled_for method. And the result is:
```
0.009306364692747593
0.017703140154480934
9.61737445788458
```

So there is some additional boost.

I'm not sure why we cannot import it on global level (I've tried but there is a message about rclpy.logging not fully initialized and possible circural dependency - this is probably the reason it is done in that way in current version), but my proposition is to:
- move severity check above import line in the log method,
- import LoggingSeverity only once in `__init__` method and store it in `self.LoggingSeverity` field, and use it in all other places in `RcutilsLogger` class

What do you think?

ksuszka: I've moved the check to the begging of the function.

As for referencing LoggingSeverity, after looking into logging.py I think more elegent solution would be to break this circural dependency, either by injecting LoggingSeverity into RcutilsLogger from logging.py, or even better, by elevating LoggingSeverity enum to rcutils_logger.py or to separate module as a dependency for both logging.py and rcutils_logger.py

clalancette: Yeah, I agree that it seems much nicer to break the circular dependency and just import the module once at the top.  But I defer to @sloretz 's judgement here.

clalancette: (also, thanks for spending the time to do the benchmarks, it really helps in informing what we should do here)

sloretz (approved):
LGTM with green CI. Thanks for the benchmarking!

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14503/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9260/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12168/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14602/)

clalancette (approved):

sloretz:
Thanks for the PR!

sloretz:
@Mergifyio backport galactic foxy

PR #798, opened 2021-05-18
Allow declaring uninitialized parameters

jacobperron:
Parameters can now be declared without a default value and without an override.
Attempting to access a statically typed parameter that does not have value will raise an exception.
Getting dynamically typed parameters will return an unset parameter value.

This change is equivalent to that made in rclcpp: https://github.com/ros2/rclcpp/pull/1673

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -612,11 +612,8 @@ class Node:
         if not self.has_parameter(name):
             return alternative_value
 
-        # Return alternative for uninitialized static parameters
-        if (
-            not self._descriptors[name].dynamic_typing and
-            self._parameters[name].type_ == Parameter.Type.NOT_SET
-        ):
+        # Return alternative for uninitialized parameters
+        if (self._parameters[name].type_ == Parameter.Type.NOT_SET):
             return alternative_value
 
         return self._parameters[name]
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -603,12 +603,23 @@ class Node:
 
         :param name: Fully-qualified name of the parameter, including its namespace.
         :param alternative_value: Alternative parameter to get if it had not been declared before.
-        :return: Requested parameter, or alternative value if it hadn't been declared before.
+        :return: Requested parameter, or alternative value if it hadn't been declared before or is
+          an uninitialized statically typed parameter.
         """
         if alternative_value is None:
             alternative_value = Parameter(name, Parameter.Type.NOT_SET)
 
-        return self._parameters.get(name, alternative_value)
+        if not self.has_parameter(name):
+            return alternative_value
+
+        # Return alternative for uninitialized static parameters
+        if (
+            not self._descriptors[name].dynamic_typing and
+            self._parameters[name].type_ == Parameter.Type.NOT_SET
+        ):
+            return alternative_value
+
+        return self._parameters[name]
 
     def get_parameters_by_prefix(self, prefix: str) -> Dict[str, Optional[Union[
         bool, int, float, str, bytes,
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -48,11 +48,11 @@ from rclpy.context import Context
 from rclpy.exceptions import InvalidParameterTypeException
 from rclpy.exceptions import InvalidParameterValueException
 from rclpy.exceptions import InvalidTopicNameException
-from rclpy.exceptions import NoParameterOverrideProvidedException
 from rclpy.exceptions import NotInitializedException
 from rclpy.exceptions import ParameterAlreadyDeclaredException
 from rclpy.exceptions import ParameterImmutableException
 from rclpy.exceptions import ParameterNotDeclaredException
+from rclpy.exceptions import ParameterUninitializedException
 from rclpy.executors import Executor
 from rclpy.expand_topic_name import expand_topic_name
 from rclpy.guard_condition import GuardCondition
@@ -364,7 +364,6 @@ class Node:
         parameters: List[Union[
             Tuple[str],
             Tuple[str, Parameter.Type],
-            Tuple[str, Any],
             Tuple[str, Any, ParameterDescriptor],
         ]],
         ignore_override: bool = False
@@ -467,9 +466,6 @@ class Node:
             if not ignore_override and name in self._parameter_overrides:
                 value = self._parameter_overrides[name].value
 
-            if value is None and not descriptor.dynamic_typing:
-                raise NoParameterOverrideProvidedException(name)
-
             if namespace:
                 name = f'{namespace}.{name}'
 
@@ -493,7 +489,9 @@ class Node:
             raise_on_failure=True,
             allow_undeclared_parameters=True
         )
-        return self.get_parameters([parameter.name for parameter in parameter_list])
+        # Don't call get_parameters() to bypass check for NOT_SET parameters
+        parameter_names = [parameter.name for parameter in parameter_list]
+        return [self._parameters[name] for name in parameter_names]
 
     def undeclare_parameter(self, name: str):
         """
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/node.py
```diff
@@ -493,7 +489,9 @@ def declare_parameters(
             raise_on_failure=True,
             allow_undeclared_parameters=True
         )
-        return self.get_parameters([parameter.name for parameter in parameter_list])
+        # Don't call get_parameters() to bypass check for NOT_SET parameters
+        parameter_names = [parameter.name for parameter in parameter_list]
+        return [self._parameters[name] for name in parameter_names]
```
jacobperron, line 494: I'm 95% sure this is okay (ie. we aren't losing anything from not calling `get_parameters`), but it deserves extra attention :sweat_smile:

rclpy/rclpy/node.py
```diff
@@ -364,7 +364,6 @@ def declare_parameters(
         parameters: List[Union[
             Tuple[str],
             Tuple[str, Parameter.Type],
-            Tuple[str, Any],
```
ivanpauno, line 367: wasn't this for passing a value, e.g.: `declare_parameters('', [('my_param', 5)])`?

rclpy/rclpy/node.py
ivanpauno: nit:

```suggestion
        return [self._parameters[parameter.name] for parameter in parameter_list]
```

rclpy/rclpy/node.py
```diff
@@ -562,6 +560,8 @@ def get_parameters(self, names: List[str]) -> List[Parameter]:
             undeclared parameters are allowed.
         :raises: ParameterNotDeclaredException if undeclared parameters are not allowed,
             and at least one parameter hadn't been declared beforehand.
+        :raises: ParameterUninitializedException if at least one parameter is statically typed and
```
ivanpauno, line 563: should that exception be raised?

I don't remember what rclcpp is exactly throwing, undeclaring a parameter was invalid when they are statically typed.
The exception should be the same no matter if it's uninitialzed or not, and shouldn't be `ParameterUninitializedException`.

It would be great to test this in both `rclcpp` and here

rclpy/rclpy/node.py
```diff
@@ -577,9 +577,18 @@ def get_parameter(self, name: str) -> Parameter:
             undeclared parameters are allowed.
```
ivanpauno, line 577: `get_parameter_or()` should return the `alternative` value if the parameter is uninitialized, it doesn't seem to be doing that.

rclpy/rclpy/node.py
```diff
@@ -941,7 +950,10 @@ def _apply_descriptor(
 
         if descriptor.dynamic_typing:
             descriptor.type = parameter.type_.value
-        elif descriptor.type != parameter.type_.value:
+        elif (
+            parameter.type_ != Parameter.Type.NOT_SET and
```
ivanpauno, line 954: IIUC, this is allowing to set an statically typed parameter to PARAMETER_NOT_SET after it was declared.
That shouldn't be valid, a test case is also a good idea.

rclpy/rclpy/node.py
wjwwood: This looks ok to me, but for sure others also look at it.

I think the apprehension comes from duplicating the logic slightly. Maybe a different way to do it would be to have a function that is called here for each parameter, which is also called from `self.get_parameter()`, just so they go through the same path.

My nascent concern is that we change `get_parameter` in the future, but not this code in some breaking way, because the future author didn't realize it could be accessed from more than one place.

wjwwood (approved):
lgtm, with green CI and some manual testing to check for sanity.

rclpy/rclpy/node.py
jacobperron: I think the annotation was incorrect before. IIUC, we only expect a tuple up to length `3`. The last value of the tuple can have type `ParameterDescriptor` or `Any` (for the case of passing a value).

rclpy/rclpy/node.py
jacobperron: I'm not sure I follow. I think this mimics the behavior of rclcpp. If a user calls `get_parameters(['foo', 'bar'])` and `bar` is declared, but not initialized, then `ParameterUninitializedException` is raised. Would you expect something else?

rclpy/rclpy/node.py
jacobperron: Should it? The docs state it will return the alternative value if the parameter is not *declared*. We could update the docs though if we want to change the behavior.

I hadn't considered this case for `rclcpp` either. I'll go back and check if we need to do anything differently.

rclpy/rclpy/node.py
jacobperron: You're right. I'm finding it tricky to handle this case with the current way the code is structured.

rclpy/rclpy/node.py
jacobperron: I've applied Ivan's suggestion (5aeceba).

I'm not sure we can reasonably add a common function, since `get_parameters` calls `get_parameter` for each name and does some validation (checks if declared and initialized). I think trying use a common function would make the implementation too convoluted .

rclpy/rclpy/node.py
jacobperron: This seems to do the trick: 5e8266f

Though, it seems a bit hacky to me to check if the parameter exists inside the "apply_descriptor" method.

rclpy/rclpy/node.py
wjwwood: > I'm not sure we can reasonably add a common function, since get_parameters calls get_parameter for each name and does some validation (checks if declared and initialized). I think trying use a common function would make the implementation too convoluted .

Ok, that's fine with me. To be clear, I was thinking a common function between `get_parameter` and this method, I was thinking `get_parameters` would still call `get_parameter`.

As I said, it was just an idea to address what I thought you were concerned about as well (we're essentially duplicating the accessor code without the validations). But I don't think it's that big of a deal to leave it as-is.

rclpy/rclpy/node.py
jacobperron: After playing around with some examples, I agree that it we should return the alternative value for uninitialized parameters. I've handled this case and added a test in 38f85c3

jacobperron: FWIW, I've in rclcpp if a static or dynamic parameter is NOT_SET then the `get_parameters_or` always returns the alternative value. I think that's fair, so I've updated this PR to match that behavior: 763ef63

No changes are required in rclcpp, but I'll open up a follow-up PR adding tests for completeness.

jacobperron:
Thanks for the reviews. I think I've addressed everything. PTAL.

rclpy/rclpy/node.py
ivanpauno: https://github.com/ros2/rclpy/blob/1b335d37e52f401d207fcfe6ec59fce2290ee524/rclpy/test/test_node.py#L630

We actually can be more specific than using `Any`, I think `Union[int, float, str, bytes, None]` would be fine

rclpy/rclpy/node.py
ivanpauno: lol, I thought this was the documentation of `undeclare_parameter()` sorry

rclpy/rclpy/node.py
ivanpauno: Sounds good to me

rclpy/rclpy/node.py
ivanpauno: yeah, the logic is a bit cumbersome, but I think it's doing the correct thing :+1:

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14660/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9413/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12324/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14804/)

jacobperron:
Thanks for triggering CI!
I thought the PR job would catch that flake8 error :man_shrugging:

PR #800, opened 2021-05-28
Call Context._logging_fini() in Context.try_shutdown()

ivanpauno:
Fixes https://github.com/ros2/rclpy/pull/518#discussion_r636538358

### DIFF
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -120,11 +120,12 @@ class Context:
         """Add a callback to be called on shutdown."""
         if not callable(callback):
             raise TypeError('callback should be a callable, got {}', type(callback))
-        if not self.ok():
-            callback()
-        else:
-            with self._callbacks_lock:
-                self._callbacks.append(weakref.WeakMethod(callback, self._remove_callback))
+        with self.__context, self._lock:
+            if not self.__context.ok():
+                callback()
+            else:
+                with self._callbacks_lock:
+                    self._callbacks.append(weakref.WeakMethod(callback, self._remove_callback))
 
     def _logging_fini(self):
         # This function must be called with self._lock held.
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -120,10 +120,10 @@ class Context:
         """Add a callback to be called on shutdown."""
         if not callable(callback):
             raise TypeError('callback should be a callable, got {}', type(callback))
-        with self._callbacks_lock:
-            if not self.ok():
-                callback()
-            else:
+        if not self.ok():
+            callback()
+        else:
+            with self._callbacks_lock:
                 self._callbacks.append(weakref.WeakMethod(callback, self._remove_callback))
 
     def _logging_fini(self):
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -101,8 +101,8 @@ class Context:
         # imported locally to avoid loading extensions on module import
         with self.__context, self._lock:
             self.__context.shutdown()
-        self._call_on_shutdown_callbacks()
-        self._logging_fini()
+            self._call_on_shutdown_callbacks()
+            self._logging_fini()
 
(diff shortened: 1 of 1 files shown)


rclpy/rclpy/context.py
```diff
@@ -101,8 +101,8 @@ def shutdown(self):
         # imported locally to avoid loading extensions on module import
         with self.__context, self._lock:
             self.__context.shutdown()
-        self._call_on_shutdown_callbacks()
-        self._logging_fini()
+            self._call_on_shutdown_callbacks()
```
wjwwood, line 104: Is this ok to do with the lock acquired?

I was thinking maybe leave this outside the lock, leave the lock usage in fini logging but make it an rlock?

rclpy/rclpy/context.py
```diff
@@ -111,6 +111,7 @@ def try_shutdown(self):
             if self.__context.ok():
                 self.__context.shutdown()
                 self._call_on_shutdown_callbacks()
```
fujitatomoya, line 113: same here, is this also okay to issue user callback with `_lock`? i think any system lock should not be acquired during user callback.

ivanpauno: this was already the case

rclpy/rclpy/context.py
ivanpauno: Calling `_call_on_shutdown_callbacks()` was supposed to be ok, as `try_shutdown()` was already doing that.
But we were taking the locks in the opposite order in another place, so that doesn't look fine.
https://github.com/ros2/rclpy/pull/800/commits/c9cba6a8a1e1c2d3098b2c4ef124582ca30fd8f9, https://github.com/ros2/rclpy/pull/800/commits/c2f71b575c43c2834cf7523fe5f38063c91958d0 should fix it.

> I was thinking maybe leave this outside the lock, leave the lock usage in fini logging but make it an rlock?

I prefer only using an `rlock` when it's actually needed and not because of an implementation detail.
`g_logging_configure_lock` is always locked after `self._lock`, so there's no deadlock.
Doing all the shutdown process atomically keeping `self._lock` held sounds like a good idea IMO.

sloretz: It looks like before this PR `on_shutdown` and `shutdown()` did not call callbacks with `self._lock` acquired, while `try_shutdown()` did.

Calling callbacks with `self._lock` seems fine to me because the context is already shutdown when it's called. There doesn't seem to be a useful method on a shutdown context that a callback might want to call. However, it looks like that makes `self._callbacks_lock` redundant. Anywhere `self._callbacks_lock` is used `self._lock` is already held.

I would recommend either making `try_shutdown()` not call callbacks with `self._lock`, or removing `self._callbacks_lock`. I personally like the second option because it means one less lock to keep track of when reviewing PRs to `Context`.

ivanpauno: >  However, it looks like that makes self._callbacks_lock redundant. Anywhere self._callbacks_lock is used self._lock is already held.

That's true

> I would recommend either making try_shutdown() not call callbacks with self._lock, or removing self._callbacks_lock. I personally like the second option because it means one less lock to keep track of when reviewing PRs to Context.

Yes, that makes sense to me.
There's not much benefit of having two locks here.

ivanpauno: Done in https://github.com/ros2/rclpy/pull/800/commits/d909a71e46d9c83deac6c2faff0e268a37c3bcd8

sloretz (approved):

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15067/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9799/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12752/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15234/)

PR #777, opened 2021-04-28
Merge Post galactic freeze into Rolling

sloretz:
Part of #665 - This includes all the pybind11 changes that we want to merge in Galactic patch 1. Before that, we've got to update rolling. All code here has already has been reviewed and approved.

I think rebase and merge would be good as that preserves the history of a bunch of different squash/merged PRs.

PR #803, opened 2021-06-23
[Galactic] Backport pybind11 changes

sloretz:
This backports the Pybind11 changes that were intended for Galactic but couldn't be merged due to the Galactic freeze. #777


sloretz:
Full CI (repos URL: `https://raw.githubusercontent.com/ros2/ros2/galactic/ros2.repos`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14803/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9552/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12469/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14956/)

sloretz:
CI LGTM! `ros2param.ros2param.test.test_verb_list.test_verb_list` and `ros2param.ros2param.test.test_verb_load.test_verb_load ` are known flaky tests - Here's a Galactic nigthly where they failed https://build.ros2.org/view/Gci/job/Gci__nightly-release_ubuntu_focal_amd64/8/

sloretz:
@ivanpauno May I ask you to double check this PR before I merge it?

rclpy/test/test_destruction.py
```diff
@@ -144,11 +143,7 @@ def test_destroy_subscription_asap():
             with sub.handle:
                 pass
 
-            with sub.handle:
-                node.destroy_subscription(sub)
-                # handle valid because it's still being used
-                with sub.handle:
-                    pass
+            node.destroy_subscription(sub)
```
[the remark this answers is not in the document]
ivanpauno: same question applies to the similar changes in this file and others

rclpy/test/test_subscription.py
```diff
@@ -48,7 +49,13 @@ def make_mock_subscription(namespace, topic_name, cli_args=None):
     ('/example/topic', 'ns', '/example/topic'),
 ])
 def test_get_subscription_topic_name(topic_name, namespace, expected):
-    sub = make_mock_subscription(namespace, topic_name)
```
ivanpauno, line 51: make_mock_subscription() definition can be deleted from the file

ivanpauno (commented):
I have only reviewed modifications in .py files, including tests, to detect possible API breaking changes.
I don't have time to look at all the changes in the cpp files, it would be good if someone that originally worked in these changes can take another look to this PR.

I'm okay with the backport if we're not breaking API and CI is green.

rclpy/test/test_destruction.py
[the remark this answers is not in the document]
sloretz: It was changed to prevent a case where destruction could be blocked forever. 
There's a longer description in the body of ros2/rclpy#739. In short, the subscription is still valid to existing users, but no new users are allowed. This prevents new users extending an object's lifetime indefinitely.

rclpy/test/test_subscription.py
sloretz: Opened https://github.com/ros2/rclpy/pull/809 deleting it from this file.

rclpy/test/test_destruction.py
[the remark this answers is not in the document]
ivanpauno: mmm, I don't understand exactly, I will have to look at the code a bit more.

What does happen with the old code snippet now?
Will it throw an InvalidHandle exception? Will it block?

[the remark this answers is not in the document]
sloretz: I would expect the inner `with sub.handle` to raise `InvalidHandle`.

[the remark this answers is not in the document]
ivanpauno: Okay, I understand the change now.
Do we still the `Handle` class? It seems to have been replaced with shared ptrs and Destroyable (I'm not sure if completely though).

There's a minimal chance this breaks pre-existing code, but I don't think anybody was relying on this behavior.

ivanpauno (approved):
LGTM with green full CI.

I would also prefer if someone involved in the original work can left an approval.

rclpy/rclpy/context.py
```diff
@@ -57,31 +57,36 @@ def init(self,
...
-            if initialize_logging and not self._logging_initialized:
-                with g_logging_configure_lock:
-                    g_logging_ref_count += 1
-                    if g_logging_ref_count == 1:
-                        rclpy_implementation.rclpy_logging_configure(capsule)
-                self._logging_initialized = True
+            try:
+                if self.__context is not None:
+                    raise RuntimeError
+            except AttributeError:
```
hidmic, line 71: @sloretz nit: I know this is in `master`, but it's still unusual. Why not set `self.__context` to `None` in `Context.__init__()`?

rclpy/src/rclpy/action_client.cpp
```diff
@@ -0,0 +1,321 @@
...
+      rcl_ret_t ret = rcl_action_client_fini(action_client, node.rcl_ptr());
+      if (RCL_RET_OK != ret) {
+        // Warning should use line number of the current stack frame
+        int stack_level = 1;
+        PyErr_WarnFormat(
+          PyExc_RuntimeWarning, stack_level, "Failed to fini publisher: %s",
+          rcl_get_error_string().str);
+        rcl_reset_error();
+      }
+      delete action_client;
```
hidmic, line 80: @sloretz meta: in retrospect, I wonder if we should've used a Python allocator-backed `std::allocator` for this.

rclpy/src/rclpy/context.cpp
```diff
@@ -20,103 +20,137 @@
...
-        stderr,
-        "[rclpy|" RCUTILS_STRINGIFY(__FILE__) ":" RCUTILS_STRINGIFY(__LINE__) "]: "
-        "failed to fini rcl_context_t (%d) during PyCapsule destructor: %s\n",
-        ret,
-        rcl_get_error_string().str);
-      rcl_reset_error();
-    }
-  }
-  PyMem_FREE(context);
+  rcl_context_.reset();
```
hidmic, line 106: @sloretz meta: one thing that's probably not great is that, if it weren't for the consistent `with obj:` where obj is a `Destroyable`, you could easily crash the process by calling APIs after an explicit `destroy()` -- even in a single threaded setting where there's no risk for concurrent destruction making that `with` block somewhat unnecessary.

hidmic (approved):
I skimmed through the PR. It looks good. I left a few comments, nothing to address here.

I do have to ask though: are we absolutely sure we want to backport over ~3000 lines of code into an stable distribution?

sloretz:
> I do have to ask though: are we absolutely sure we want to backport over ~3000 lines of code into an stable distribution?

That's a great question to ask. I think it's important in this case since it enables more backports, and makes the destruction handling of entities consistent.

rclpy/test/test_destruction.py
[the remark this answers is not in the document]
sloretz: On `master` the `Handle` class is gone, but I'm not sure which PR to backport. The only use I see of it on this branch is in `test_handle.py`.

rclpy/rclpy/context.py
sloretz: Hmm this is strange. I opened https://github.com/ros2/rclpy/pull/812 to try setting it to `None` in `__init__()`

rclpy/src/rclpy/action_client.cpp
sloretz: Oh yeah! You mean like the allocator `rclpy::Client` is using? Looks lt's the only one using it.

https://github.com/ros2/rclpy/blob/fbe9ad12b806345a3b3c0578af7e21d1d6bc7bea/rclpy/src/rclpy/client.cpp#L55-L70

rclpy/src/rclpy/context.cpp
sloretz: Hmm yeah, maybe that `with obj:` logic could be moved into the APIs themselves.  Maybe a `DestroyableScopeGuard` or something with https://pybind11.readthedocs.io/en/stable/advanced/functions.html#call-guard that did the equivalent of `__enter__` and `__exit__` could replace that, and get rid of one of the jumps between Python and C++ at the same time.

sloretz:
CI (galactic repos file + this branch, build: `--packages-above-and-dependencies rclpy` test: `--packages-above rclpy`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15028/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9765/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12718/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15191/)

sloretz:
CI LGTM, merging :tada:

rclpy/src/rclpy/action_client.cpp
hidmic: Indeed.

rclpy/src/rclpy/context.cpp
hidmic: An implicit call guard would do, yeah.

PR #816, opened 2021-08-17
Only add one done callback to a future

sloretz:
This seems to fix an issue in #605 

The issue is reproducible with the snippet below. It is using a `MultiThreadedExecutor` and sending sequential client requests to a server on a different node in the same executor. By the 3rd or 4th request, the observed behavior is the server stops responding. I observed with print statements in `MultiThreadedExecutor.spin_once_until_future_complete()` that the number of calls to it seems to grow very rapidly with each new client request.

I suspect that the previous future's done callbacks are being executed while the next future waits. The first future's done callbacks wake the executor, and that causes `MultiThreadedExecutor.spin_once_until_future_complete()` to be called again on the next future, which adds another done callback to the next future. That compounds and causes the next future to add even more done callbacks to the one after that. I'm not sure I understand the problem completely, but this change seems to resolve the issue.

This PR fixes the issue by only adding a single done callback in `spin_until_future_complete()`. 

<details><summary>Reproducing code</summary>

```python3
import rclpy

from example_interfaces.srv import AddTwoInts


rclpy.init()

def server_callback(cb_req, cb_rsp):
    print(f'Server callback request {cb_req.a} + {cb_req.b}')
    cb_rsp.sum = cb_req.a + cb_req.b
    return cb_rsp


# setup
server_node = rclpy.create_node("test_server")
server = server_node.create_service(AddTwoInts, "/test_server", server_callback)
client_node = rclpy.create_node(f"test_client")
client = client_node.create_client(AddTwoInts, "/test_server")

executor = rclpy.executors.MultiThreadedExecutor()
executor.add_node(server_node)
executor.add_node(client_node)

for ii in range(10):
    print(f'Begin iteration {ii}')

    # Send a request
    req = AddTwoInts.Request()
    req.a = ii
    req.b = 0
    call_future = client.call_async(req)

    # Wait for a response
    executor.spin_until_future_complete(call_future, timeout_sec=10.0)
    print(f'Call future done? {call_future.done()}')
    if call_future.done():
        print(f'Response {call_future.result()}')

    print(f'End iteration {ii}')


# Cleanup
client_node.destroy_client(client)

server_node.destroy_service(server)
server_node.destroy_node()
client_node.destroy_node()

rclpy.shutdown()
```
</details>

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -288,6 +288,9 @@ class Executor:
             end = start + timeout_sec
             timeout_left = timeout_sec
 
+            # Make sure the future wakes this executor when it is done
+            future.add_done_callback(lambda x: self.wake())
+
             while self._context.ok() and not future.done() and not self._is_shutdown:
                 self.spin_once_until_future_complete(future, timeout_left)
                 now = time.monotonic()
@@ -754,5 +757,4 @@ class MultiThreadedExecutor(Executor):
         self._spin_once_impl(timeout_sec)
 
     def spin_once_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
-        future.add_done_callback(lambda x: self.wake())
         self._spin_once_impl(timeout_sec, future.done)
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -280,6 +280,9 @@ class Executor:
 
     def spin_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
         """Execute callbacks until a given future is done or a timeout occurs."""
+        # Make sure the future wakes this executor when it is done
+        future.add_done_callback(lambda x: self.wake())
+
         if timeout_sec is None or timeout_sec < 0:
             while self._context.ok() and not future.done() and not self._is_shutdown:
                 self.spin_once_until_future_complete(future, timeout_sec)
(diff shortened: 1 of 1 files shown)


rclpy/rclpy/executors.py
```diff
@@ -288,6 +288,9 @@ def spin_until_future_complete(self, future: Future, timeout_sec: float = None)
             end = start + timeout_sec
             timeout_left = timeout_sec
 
+            # Make sure the future wakes this executor when it is done
+            future.add_done_callback(lambda x: self.wake())
```
ivanpauno, line 292: isn't this also needed also in the if case?
Maybe, it can be added above line L283

ivanpauno:
> The first future's done callbacks wake the executor, and that causes MultiThreadedExecutor.spin_once_until_future_complete() to be called again on the next future, which adds another done callback to the next future. That compounds and causes the next future to add even more done callbacks to the one after that

that's a good one :man_facepalming:

fujitatomoya (commented):
problem description and fix looks reasonable to me. good to go with @ivanpauno 's comment.

rclpy/rclpy/executors.py
sloretz: oops, moved it above line 283 in 
62c2ae4

ivanpauno (approved):
LGTM!

sloretz:
CI (build: `--packages-up-to rclpy` test: ` --packages-select rclpy`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15142/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9871/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12830/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15316/)

ivanpauno:
@mergifyio backport foxy galactic

PR #818, opened 2021-08-26
👩‍🌾 Check that header is not none before accessing it in test

chapulina:
`test_service_timestamps` has started failing lately with:

```
>                   self.assertNotEqual(0, header.source_timestamp)
E                   AttributeError: 'NoneType' object has no attribute 'source_timestamp'
```

See the test history here:

https://build.ros2.org/job/Rci__nightly-connext_ubuntu_focal_amd64/429/testReport/rclpy.rclpy.test.test_client/TestClient/test_service_timestamps/history/

I'm not sure what's the underlying issue that causes the header to be none, but catching that before trying to use the variable should allow the test to fail gracefully instead of throwing an error.

---

https://github.com/osrf/buildfarmer/issues/224

### DIFF
--- a/rclpy/test/test_client.py
+++ b/rclpy/test/test_client.py
@@ -108,6 +108,7 @@ class TestClient(unittest.TestCase):
                     result = srv.handle.service_take_request(srv.srv_type.Request)
                 if result is not None:
                     request, header = result
+                    self.assertTrue(header is not None)
                     self.assertNotEqual(0, header.source_timestamp)
                     return
                 else:


clalancette:
> I'm not sure what's the underlying issue that causes the header to be none

Looking at https://github.com/ros2/rclpy/blob/ad226136bd1675f886dcbc618044714c2edc8871/rclpy/src/rclpy/service.cpp#L104 , it looks like that can only happen if the underlying call to `rcl_take_request_with_info` returned `RCL_RET_SERVICE_TAKE_FAILED`.  Whether that gets returned is up to the individual RMW, which is `rmw_connextdds` in this case.  Still not sure why that is happening; @asorbini could you look into it?

> but catching that before trying to use the variable should allow the test to fail gracefully instead of throwing an error.

But yeah, I agree with this.

clalancette (approved):

j-rivero:
Problem appeared again in today's CI https://build.ros2.org/job/Rci__nightly-fastrtps-dynamic_ubuntu_focal_amd64/432/testReport/rclpy.rclpy.test.test_client/TestClient/test_service_timestamps/

chapulina:
> Problem appeared again in today's CI

Yay! The assertion works! :upside_down_face:

PR #819, opened 2021-08-30
Add convert function from ParameterValue to Python builtin

jacobperron:
This function is particular useful when processing a GetParameters service request.
Tools like ros2param implement this function locally, but it would be nice to have a common implementation to avoid duplication.

This function was adapted from the implementation in ros2param: https://github.com/ros2/ros2cli/blob/1f4a3aaf7b930830b3fa04773715bf93ea65693e/ros2param/ros2param/api/__init__.py#L31

### DIFF
--- a/rclpy/rclpy/parameter.py
+++ b/rclpy/rclpy/parameter.py
@@ -172,3 +172,39 @@ class Parameter:
 
     def to_parameter_msg(self):
         return ParameterMsg(name=self.name, value=self.get_parameter_value())
+
+
+def parameter_value_to_python(parameter_value: ParameterValue):
+    """
+    Get the Python value from a rcl_interfaces/msg/ParameterValue object.
+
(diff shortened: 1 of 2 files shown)


clalancette (approved):
I could definitely use this in some code I'm working on.

I think we should also probably update ros2cli to use this function while we are at it; otherwise we are likely to forget.

But this code as-is looks good to me with green CI.

jacobperron:
Testing `rclpy`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15136/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9865/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12824/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15308/)

Issue #822, opened 2021-09-14
Memory leak in client when using python wrapper to send request to service


crazyhank:
I found this problem in latest Galactic release, it is simple to reproduce, write a simple service(C++) and a client (Python), memory leak will definitly happen. It will not happen when using C++ in client.

1. Data struct:
```
int32 seq
uint64 time
byte[1048576] input_tensor
---
int32 seq
uint64 time
byte[1048576] output_tensor

```

2. Client side code:
```
import sys

from my_struct.srv import InferService
import rclpy
from rclpy.node import Node

import time


class MinimalClientAsync(Node):

    def __init__(self):
        super().__init__('minimal_client_async')
        self.cli = self.create_client(InferService, 'infer')
        while not self.cli.wait_for_service(timeout_sec=1.0):
            self.get_logger().info('service not available, waiting again...')
        self.req = InferService.Request()

    def send_request(self):
        self.req.seq = 2000
        self.req.time = 200
        self.future = self.cli.call_async(self.req)


def main(args=None):
    rclpy.init(args=args)

    minimal_client = MinimalClientAsync()
    for _ in range(100000):
        minimal_client.send_request()

        while rclpy.ok():
            rclpy.spin_once(minimal_client)
            if minimal_client.future.done():
                try:
                    response = minimal_client.future.result()
                except Exception as e:
                    minimal_client.get_logger().info(
                        'Service call failed %r' % (e,))
                else:
                    minimal_client.get_logger().info(
                        'Result of inference: resp seq %d' % (response.seq))
                break
            #time.sleep(0.5)


    minimal_client.destroy_node()
    rclpy.shutdown()


if __name__ == '__main__':
    main()
```

You can write a simple service code, just receive the request and do nothing and send a response to client.

aprotyas:
Do you mind explicating why you think there will be a memory leak?

crazyhank:
This problem is found in our project, I need time to seperate test code to reproduce it. As you see, there is a 1MB data member in response data struct, each time when I receive a response from servcer, I see a about 1MB memory used for client node, so a memory usage is increased as test continues.
I seems that reponse buffer is not released for python wrapper, no problem is found when I switch to C++ code.

aprotyas:
Thanks for the explanation! It would be great if you could provide a self-contained example that could be used for stress-testing.

crazyhank:
Hi, I list the key information, I think it enough and easy to reproduce the problem.

1、Request/Response data struct:
```
[...]

PR #828, opened 2021-09-28
Fix memory leak.

llapx:
Signed-off-by: Lei Liu <Lei.Liu.AP@sony.com>

### DIFF
--- a/rclpy/src/rclpy/client.cpp
+++ b/rclpy/src/rclpy/client.cpp
@@ -137,7 +137,6 @@ Client::take_response(py::object pyresponse_type)
   result_tuple[0] = header;
 
   result_tuple[1] = convert_to_py(taken_response.get(), pyresponse_type);
-  // result_tuple now owns the message
 
   return result_tuple;
 }
--- a/rclpy/src/rclpy/client.cpp
+++ b/rclpy/src/rclpy/client.cpp
@@ -138,7 +138,6 @@ Client::take_response(py::object pyresponse_type)
 
   result_tuple[1] = convert_to_py(taken_response.get(), pyresponse_type);
   // result_tuple now owns the message
-  taken_response.release();
 
   return result_tuple;
 }
--- a/rclpy/src/rclpy/service.cpp
+++ b/rclpy/src/rclpy/service.cpp
@@ -119,7 +119,6 @@ Service::service_take_request(py::object pyrequest_type)
 
   result_tuple[1] = header;
   result_tuple[0] = convert_to_py(taken_request.get(), pyrequest_type);
-  taken_request.release();
 
   return result_tuple;
 }


aprotyas (approved):
LGTM under #822's context.

Although, I wonder if there was a case to capture this concern through a unit test as well? I'm not sure, but this seems like a case that should be tested for (i.e. is the client/server correctly "taking" the response/request).

llapx:
In my option, it's a bug which may affect many test cases.

aprotyas:
For reviewers, just reproducing the memory usage charts from https://github.com/ros2/rclpy/issues/822#issuecomment-923655611
test_client_before
test_client_after

The same observations should hold for services using `rclpy`.
As such, I think this closes #822.

rclpy/src/rclpy/client.cpp
```diff
@@ -138,7 +138,6 @@ Client::take_response(py::object pyresponse_type)
 
   result_tuple[1] = convert_to_py(taken_response.get(), pyresponse_type);
   // result_tuple now owns the message
```
iuhilnehc-ynos, line 140: There is no reason to keep this comment, I suggest deleting it.

iuhilnehc-ynos (commented):
nitpick.

fujitatomoya (approved):
lgtm with @iuhilnehc-ynos 's comment.

fujitatomoya:
@aprotyas it seems that memory usage doe not change on the graph chart between before and after? it still increases the memory usage constantly on client side? just checking.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15324/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10035/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13004/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15517/)

aprotyas:
> @aprotyas it seems that memory usage doe not change on the graph chart between before and after? it still increases the memory usage constantly on client side? just checking.

No no, the first image is from before, and the second image is from after this PR.
In the second image, you see that the memory usage does not continuously increase (stays consistent around 60 MB). So I believe the PR fixes the memory leak.

fujitatomoya:
@aprotyas Ah, i see it. thanks!

iuhilnehc-ynos (approved):
LGTM.

Re-CI (error happened on Window because of rti package installation, rebuild with @fujitatomoya setting on ci_launcher):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15333/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10044/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13013/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15527/)

PR #814, opened 2021-08-11
Use pybind11 for signal handling, and delete now unused rclpy_common, pycapsule, and handle code

sloretz:
This uses pybind11 for the signal handler APIs. It replaces `rcutils` atomics with with `std::atomic` because C and C++ atomics headers can't be mixed ([until C++23](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2741.htm)?). Finally this deletes the now unused code for handle, pycapsule, and rclpy_common.

Replaces #728
Closes #665 (finally :tada:)

PR #830, opened 2021-09-30
Handle sigterm

ivanpauno:
Depends on https://github.com/ros2/rclpy/pull/814.
Equivalent to https://github.com/ros2/rclcpp/pull/1771.


ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15434/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10137/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13120/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15653/)

emersonknapp:
As you hit on in the conversation above, I have one concern with the behavior of this PR, and it's that you can't handle SIGINT and SIGTERM the same within your Python application. If I make a test as follows:

```
import rclpy
from demo_nodes_py.topics.talker import Talker

def main():
    rclpy.init()
    node = Talker()
    try:
        rclpy.spin(node)
    except KeyboardInterrupt:
        print('Got SIGINT')
    except rclpy.executors.ExternalShutdownException:
        print('Got SIGTERM')
    print('Destroying')
    node.destroy_node()
    rclpy.shutdown()


if __name__ == '__main__':
    main()
```

If I run this program, and send a SIGINT to the process, then the following happens:

```
[INFO] [1635816254.444592338] [talker]: Publishing: "Hello World: 0"
Got SIGINT
Destroying
```

However if you send a SIGTERM, then the following unclean shutdown happens:

```
[INFO] [1635815908.502049088] [talker]: Publishing: "Hello World: 0"
Got SIGTERM
Destroying
  File "./src/emersonknapp/ektest/src/ekpy", line 42, in <module>
    main()
  File "./src/emersonknapp/ektest/src/ekpy", line 38, in main
    rclpy.shutdown()
  File "/ws/install/gcc/rclpy/lib/python3.8/site-packages/rclpy/__init__.py", line 123, in shutdown
    _shutdown(context=context)
  File "/ws/install/gcc/rclpy/lib/python3.8/site-packages/rclpy/utilities.py", line 58, in shutdown
    return context.shutdown()
  File "/ws/install/gcc/rclpy/lib/python3.8/site-packages/rclpy/context.py", line 101, in shutdown
    self.__context.shutdown()
rclpy._rclpy_pybind11.RCLError: failed to shutdown: rcl_shutdown already called on the given context, at /ws/src/ros2/rcl/rcl/src/rcl/init.c:241
```

I don't think it makes sense to require different shutdown logic between these signals - did you come to a resolution for which one of the behaviors it should be?

ivanpauno:
> I don't think it makes sense to require different shutdown logic between these signals - did you come to a resolution for which one of the behaviors it should be?

I'm honestly not sure.

Python raises a `KeyboardInterrupt` from the sigint handler, the signal handler added by ros2 tried to respect that logic.
That seems reasonable, as it's familiar to python users.

We could also raise a `KeyboardInterrupt` from sigterm. But IMO that would confuse python users, who expect `KeyboardInterrupt` to only be raised when sigint happens.

Maybe we could raise a different exception when sigterm happens, which would be pretty similar to the sigint case (e.g. `rclpy.executors.ExternalShutdownException`, without actually shutting down contexts).
It's a bit problematic though, as for that we need the `signal` module, and `signal.signal()` can only be called from the main thread, which adds a limitation to `rclpy.init()` that currently doesn't exist.

---

The following code should work cleanly in both cases:

```python3
import rclpy
from demo_nodes_py.topics.talker import Talker

def main():
    rclpy.init()
    node = Talker()
    try:
        rclpy.spin(node)
    except KeyboardInterrupt:
        print('Got SIGINT')
        rclpy.shutdown()
    except rclpy.executors.ExternalShutdownException:
        print('Got SIGTERM')
    print('Destroying')
    node.destroy_node()


if __name__ == '__main__':
    main()
```

ivanpauno:
I think that the other alternative is to also shutdown asynchronously when sigint happens, so you don't really need to call rclpy.shutdown() in either case.
That would be a breaking change, though I think it's probably a good idea.

IIRC @hidmic preferred this alternative.

hidmic:
I think it makes sense for the signals that were configured as a shutdown event upon `rclpy` initialization to effectively shutdown. That's not mutually exclusive (or shouldn't be IMO) with the default signal handler behavior e.g. raise `KeyboardInterrupt` on SIGINT.

emersonknapp:
> I think that the other alternative is to also shutdown asynchronously when sigint happens, so you don't really need to call rclpy.shutdown() in either case.
> I think it makes sense for the signals that were configured as a shutdown event upon rclpy initialization to effectively shutdown.

Yeah, this also makes sense to me. Then, in practice would we say a ROS 2 Python `__main__` should look like:

```
try:
    rclpy.init()
    my_program()
except (KeyboardInterrupt, rclpy.executors.ExternalShutdownException):
    print('Got clean shutdown signal exception.')
else:
    rclpy.shutdown()
finally:
    my_specific_cleanup()
```

ivanpauno:
> Yeah, this also makes sense to me. Then, in practice would we say a ROS 2 Python __main__ should look like:

Exactly.
I will open a PR trying that, I guess I will have to open some PRs in other repos fixing regressions.

Issue #834, opened 2021-10-12
Timer created during async service handler never fires


jtbandes:
## Bug report

**Required Info:**

- Operating System: Ubuntu 20.04
- Installation type: binaries
- Version or commit hash: tested in Galactic and Rolling
- DDS implementation: default
- Client library (if applicable): rclpy

#### Steps to reproduce issue
Using an `async` function as a service handler [seems to be supported](https://github.com/ros2/rclpy/blob/38a408867719cb76c4b4f0b1caaf9cc191336970/rclpy/rclpy/executors.py#L369), but if a timer is added during the service handler (e.g. to respond to the service after an artificial delay), the timer never fires. I also tried using a `MultiThreadedExecutor(2)` with the same results.

```py
from example_interfaces.srv import AddTwoInts
import rclpy
from rclpy.node import Node


class MinimalService(Node):
    def __init__(self):
        super().__init__('minimal_service')
        self.srv = self.create_service(AddTwoInts, 'add_two_ints', self.add_two_ints_callback)

    async def add_two_ints_callback(self, request, response):
        self.get_logger().info(f'Executor: {str(self.executor)} ', )
        self.get_logger().info('Incoming request... a: %d b: %d' % (request.a, request.b))
        response.sum = request.a + request.b

        future = rclpy.task.Future()
        def handler():
            self.get_logger().info('Timer fired')
            future.set_result(None)
        timer = self.create_timer(2, handler)
        await future
        timer.destroy()
        self.get_logger().info('Finished request!  a: %d b: %d' % (request.a, request.b))
        return response


def main(args=None):
    rclpy.init(args=args)
    minimal_service = MinimalService()
    rclpy.spin(minimal_service)
    rclpy.shutdown()


if __name__ == '__main__':
    main()
```

```
ros2 service call /add_two_ints example_interfaces/srv/AddTwoInts "{a: 1, b: 2}"
```

#### Expected behavior

The service call should receive a response after a delay. While the executor is waiting for the timer, it should be possible for more services to be handled.

#### Actual behavior

The timer never fires and new incoming service requests are not logged.

Barry-Xu-2018:
> IMO, await is not “blocking” per se; it’s more like yielding control back to the executor.

Yes.   
My understanding is

https://github.com/ros2/rclpy/blob/6f4e42cb4b409cbd79273dd9dd12cc35198ff138/rclpy/rclpy/executors.py#L704-L710

[...]

PR #847, opened 2021-11-05
Change default callback group to a ReentrantCallbackGroup

jtbandes:
Per discussion in https://github.com/ros2/rclpy/issues/834, making the default callback group re-entrant makes adopting coroutines easier.

cc @sloretz @ivanpauno

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -38,7 +38,6 @@ from rcl_interfaces.msg import ParameterValue
 from rcl_interfaces.msg import SetParametersResult
 
 from rclpy.callback_groups import CallbackGroup
-from rclpy.callback_groups import MutuallyExclusiveCallbackGroup
 from rclpy.callback_groups import ReentrantCallbackGroup
 from rclpy.client import Client
 from rclpy.clock import Clock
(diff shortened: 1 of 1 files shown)


jtbandes:
Any interest in reviewing or accepting this change? cc @sloretz

fujitatomoya (commented):
I think using `MutuallyExclusiveCallbackGroup` in default is intention to avoid possible confliction in user application as default behavior. This is aligned with `rclcpp`.
besides, user application can change it into `MutuallyExclusiveCallbackGroup`.
I am not really inclined to take this fix to change the default behavior. any thoughts?

jtbandes:
Specific comment from prior discussion for reference: https://github.com/ros2/rclpy/issues/834#issuecomment-961331870

wjwwood:
I'd like to vote for not taking this change. I do not think that we should diverge from `rclcpp` on this.

Perhaps we could have some special logic that allows coroutines to be handled differently, but this change would affect non-coroutine versions as well.

wjwwood:
I talked with @sloretz about this and I think we could consider removing callback groups or removing typed callback groups, but I still wouldn't change the default here right now, for the reason I stated before about existing non-coroutine users being affected.

PR #647, opened 2020-12-11
added wake executor to Node.create_subscription to fix issue 628

ssumoo:
As @ivanpauno suggested, this PR fixes https://github.com/ros2/rclpy/issues/628

PR #855, opened 2021-12-01
Adding _wake_executor to create_subscription on foxy

lowellausen-karelics:
In short, same as PR #647 but on foxy branch.

I was facing a problem where, when creating a subscriber (`create_subscription`) for a node that is already spinning the callback would never run (when it should run). It would only run after calling another random node method, for example creating a timer.

After searching we found out that the problem is that `create_subscription` wasn't calling `_wake_executor` so the node wasn't really aware of the new callback, until some other method called `_wake_executor`.

This problem has been fixed already but only in the master branch  (on PR #647) , so I thought it could be important to bring this fix to the foxy branch as well.

Signed-off-by: Leonardo Oliveira Wellausen <leonardo.wellausen@karelics.fi>

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1220,6 +1220,7 @@ class Node:
             raise
         self.__subscriptions.append(subscription)
         callback_group.add_entity(subscription)
+        self._wake_executor()
 
         for event_handler in subscription.event_handlers:
             self.add_waitable(event_handler)


sloretz (approved):
LGTM with green CI!

sloretz:
CI
* Repos file: https://gist.githubusercontent.com/sloretz/167aff2ec60675548b1983ae3c04f1be/raw/049cb86ae399291271b308de326cac064a63c83d/ros2.repos
* build: `--packages-up-to rclpy`
* test: `--packages-select rclpy`

Jobs
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15791/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10469/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13453/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16048/)

sloretz:
CMake warnings appear unrelated. Thanks for the backport PR!

PR #858, opened 2021-12-09
Add Clock.sleep_until

### LINKED ISSUE #617: Add sleep function that respects ROS time
#### Feature description
This is a request for a function that sleeps the current thread while respecting ROS time.

There's some discussion about where a method like this should live in ros2/rclcpp#1248.
Historically this method has lived as [`rospy.sleep()` and `rospy.Rate.sleep()`](http://wiki.ros.org/rospy/Overview/Time#Sleeping_and_Rates)

Sleeping should respect the clock being used. I think this could be implemented by [waiting on a `threading.Event`](https://docs.python.org/3.8/library/threading.html#threading.Event.wait) that's woken by a time jump callback. Steady clock and system clocks don't have time jump callbacks, so the event would expire on the timeout being reached.

I'm assuming this would be a `sleep_for(duration)` rather than a `sleep_until(time)`. The latter is tricky when using system time because right now we don't have time jump callbacks for when system time changes.

Sleeping on ROS clocks would only be as accurate as the frequency of published time messages on the `/clock` topic. More frequent publishing has a drawback that it will wake the time jump callback more frequently before the sleep period has elapsed.

Related to ros2/rclcpp#1730

sloretz:
Fixes #617 

This works, and adds a method `Clock.sleep_until()` to sleep until a particular time on a clock is reached, respecting ROS time.

I chose to implement the time jump callbacks in Python because it's easier and reuses code, but it does mean the internal API `Event::wait_until_ros()` requires the caller to setup the required time jump callback or it will wait forever. Since this is an internal API, I think it's fine to document it and leave it as is. I couldn't think of a cleaner solution and user's won't be affected one way or the other.

Related to https://github.com/ros2/rclcpp/issues/1730


sloretz:
PR job requires https://github.com/ros2/rcl/pull/955 released to pass

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15767/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10448/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13433/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16023/)

sloretz:
CI (Re-run)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15770/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10451/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13435/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16025/)

sloretz:
CI Re-run
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15773/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10452/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13436/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16026/)

clalancette:
> PR job requires [ros2/rcl#955](https://github.com/ros2/rcl/pull/955) released to pass

I've merged https://github.com/ros/rosdistro/pull/31477, which should allow the Rpr job to pass (once the downstream packages are done compiling. which will take a while).

rclpy/src/rclpy/event.cpp
```diff
@@ -0,0 +1,120 @@
...
+// limitations under the License.
+
+#include <pybind11/pybind11.h>
+
+#include <rcl/error_handling.h>
+#include <rcl/rcl.h>
+#include <rcl/time.h>
+#include <rcl/types.h>
+
+#include <condition_variable>
```
clalancette, line 22: Nit: `#include <chrono>` for `std::chrono`.

rclpy/src/rclpy/event.cpp
```diff
@@ -0,0 +1,120 @@
...
+  const typename ClockType::time_point chrono_entry = ClockType::now();
+
+  rcl_duration_t delta_t;
+  rcl_ret_t ret = rcl_difference_times(&rcl_entry, &until, &delta_t);
+
+  if (RCL_RET_OK != ret) {
+    throw RCLError("failed to subtract times");
+  }
+
+  // Cast because system clock resolution is be too big for nanoseconds on Windows & OSX
```
clalancette, line 47:
```suggestion
  // Cast because system clock resolution is too big for nanoseconds on Windows & OSX
```

rclpy/test/test_clock.py
```diff
@@ -211,6 +216,137 @@ def test_triggered_clock_change_callbacks(self):
...
+
+def test_sleep_until_mismatched_clock_type(default_context):
+    clock = Clock(clock_type=ClockType.SYSTEM_TIME)
+    with pytest.raises(ValueError, match='.*clock type does not match.*'):
+        clock.sleep_until(Time(clock_type=ClockType.STEADY_TIME))
+
+
+def test_sleep_until_non_default_context(non_default_context):
+    clock = Clock()
+    assert clock.sleep_until(clock.now() + Duration(seconds=0.1), context=non_default_context)
```
clalancette, line 242: Is there a chance that we'll have spurious wakeups in a 100 millisecond wait (especially on Windows and macOS)?  That is, is this a potentially flaky test?

(the same question goes for all of the places where we actually sleep below)

rclpy/test/test_clock.py
```diff
@@ -211,6 +216,137 @@ def test_triggered_clock_change_callbacks(self):
...
+@pytest.mark.parametrize('ros_time_enabled', (True, False))
+def test_sleep_until_ros_time_toggled(default_context, ros_time_enabled):
+    clock = ROSClock()
+    clock._set_ros_time_is_active(not ros_time_enabled)
+
+    retval = None
+
+    def run():
+        nonlocal retval
+        retval = clock.sleep_until(clock.now() + Duration(seconds=999999))
```
clalancette, line 283: I'm a bit worried about this very long timeout.  In the case that a bug is introduced, this will wait for a *long* time.  Maybe we should just reduce this to 10 seconds or something?

rclpy/test/test_clock.py
```diff
@@ -211,6 +216,137 @@ def test_triggered_clock_change_callbacks(self):
...
+
+    t = threading.Thread(target=run, daemon=True)
+    t.start()
+
+    # wait for thread to get inside sleep_until call
+    time.sleep(0.2)
```
clalancette, line 289: Hm, I don't love sleeping an arbitrary amount here for the thread to startup.  It means that the test is slower (though 200 milliseconds isn't too bad), but it also means this is a potentially flaky test.  I unfortunately don't have a great solution, since we really do need to wait until we are in `sleep_until`, but I'm wondering if there is some other clever solution here.

clalancette (commented):
> I chose to implement the time jump callbacks in Python because it's easier and reuses code

Just as a question: what is the alternative?  To implement it on the C++ side?

Besides that, I've left some thoughts and comments inline.  Nothing major, really, other than the possible problems of flaky tests.

rclpy/src/rclpy/event.cpp
```diff
@@ -0,0 +1,120 @@
...
+  // Check if ROS time is enabled in C++ to avoid TOCTTOU with TimeSource by holding GIL
+  if (clock->get_ros_time_override_is_enabled()) {
+    // Could be a long wait, release the gil
+    py::gil_scoped_release release;
+    std::unique_lock<std::mutex> lock(mutex_);
+    // Caller must have setup a time jump callback to wake this event
+    cv_.wait(lock, [this]() {return state_;});
+  } else {
+    // ROS time not enabled is system time
+    wait_until<std::chrono::system_clock>(clock, until);
```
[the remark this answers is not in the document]
sloretz: ROS time disabled is always system time

https://github.com/ros2/rcl/blob/e2cf9fdc2fe462136bf01dfc676c6054dc111d4e/rcl/src/rcl/time.c#L68-L69

rclpy/test/test_clock.py
sloretz: I can shorten it. Right now if a bug is introduced, the max it will wait until ctest kills it is 120 seconds

https://github.com/ros2/rclpy/blob/3af017844a214a838a27891710cc9bee52331842/rclpy/CMakeLists.txt#L229

sloretz:
> >   I chose to implement the time jump callbacks in Python because it's easier and reuses code
>
> Just as a question: what is the alternative? To implement it on the C++ side?

Yeah, the alternative would be to implement it in C++ by calling [`rcl_clock_add_jump_callback()` directly](https://github.com/ros2/rcl/blob/e2cf9fdc2fe462136bf01dfc676c6054dc111d4e/rcl/include/rcl/time.h#L668). The advantage is the internal `rclpy::Event::wait_until_ros()` API would encapsulate all the ROS time stuff; the disadvantage is Python is easier to write. Since it's an internal API I went with the easier option.

rclpy/test/test_clock.py
sloretz: Shortened to 10 seconds in 0530656493a2de30baace5f2cde20e73527092bf

rclpy/src/rclpy/event.cpp
sloretz: Added missing includes in a07942dfbe2cbbda73e9407b7dfa866b65365568

rclpy/src/rclpy/event.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include <rcl/error_handling.h>
+#include <rcl/rcl.h>
+#include <rcl/time.h>
+#include <rcl/types.h>
+
+#include <condition_variable>
+#include <cstring>
+#include <memory>
+#include <stdexcept>
```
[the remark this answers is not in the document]
sloretz: Added missing includes in a07942d

rclpy/src/rclpy/event.cpp
sloretz: Removed 'be' in 9c93f3433ba76241c9c3b1932789d0adbd47e6a7

rclpy/test/test_clock.py
clalancette: > Is the question: will the underlying std::condition_variable::wait spuriously wake up and cause sleep_until() to return False?

Yeah, that is basically the question.

> The rclpy::Event::wait_...() are using the predicate version of std::condition_variable::wait() and std::condition_variable::wait_until() to prevent spurious wake ups.

Ah right.  OK, in that case this should be fine.  Thanks for the explanation.

rclpy/test/test_clock.py
clalancette: Oh right, I forgot that `rclpy` uses `ctest` and not `pytest`.  Still, a shorter timeout in the case of a bug is a good thing, I think.  The timeout of 10 seconds seems reasonable to me, thanks.

rclpy/test/test_clock.py
clalancette: > A really clever solution might be to inspect the thread to see what line of code it's executing. I don't think I can implement that robustly in a reasonable amount of time.

Oh, very interesting.  That is really clever.  Probably overkill, though.  I guess I'm OK with the current solution, though if we find it is flaky we'll have to spend some additional time on it.

rclpy/test/test_clock.py
```diff
@@ -211,6 +216,137 @@ def test_triggered_clock_change_callbacks(self):
...
+
+    clock._set_ros_time_is_active(ros_time_enabled)
+
+    # wait for thread to exit
+    time.sleep(0.2)
```
clalancette, line 294: If we *don't* daemonize the thread, then we could do a `join` on the thread here instead, which at least gets rid of one sleep in the tests.  What do you think?

clalancette:
> Yeah, the alternative would be to implement it in C++ by calling rcl_clock_add_jump_callback() directly. The advantage is the internal rclpy::Event::wait_until_ros() API would encapsulate all the ROS time stuff; the disadvantage is Python is easier to write. Since it's an internal API I went with the easier option.

Sounds reasonable to me.  I think the solution you've chosen is fine.

clalancette:

rclpy/src/rclpy/event.hpp
```diff
@@ -0,0 +1,73 @@
...
+namespace py = pybind11;
+
+namespace rclpy
+{
+class Event
+{
+public:
+  /// Wait until a time specified by a system or steady clock.
+  /// \param clock the clock to use for time synchronization with until
+  template<typename ClockType>
```
j-rivero, line 43: Nit: Do we need to document the `until` parameter here and in the next method?

j-rivero (commented):
I can not find any relevant issue or discrepancy in the code. Just a Nit about doxygen. I re-run CI trying to trigger the possible race condition in tests.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15780/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10458/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13442/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16034/)

rclpy/test/test_clock.py
sloretz: `join()`ing the threads to avoid the second sleep in a8ddedf

rclpy/rclpy/clock.py
```diff
@@ -172,6 +174,68 @@ def callback_shim(jump_dict):
...
+
+            # ROS time being activated or deactivated changes the epoch, so sleep
+            # time loses its meaning
+            time_source_changed = (
+                time_source_changed or
+                ClockChange.ROS_TIME_ACTIVATED == time_jump.clock_change or
+                ClockChange.ROS_TIME_DEACTIVATED == time_jump.clock_change)
+
+            if time_source_changed or self.now() >= until:
+                event.set()
```
ivanpauno, line 217: why is the event triggered when the time source changes?
shouldn't still be checking it `now >= until` with the new time source?

rclpy/src/rclpy/event.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include <rcl/rcl.h>
+#include <rcl/time.h>
+#include <rcl/types.h>
+
+#include <condition_variable>
+#include <cstring>
+#include <memory>
+#include <stdexcept>
+
+#include "event.hpp"
```
ivanpauno, line 27: there's already a bunch of different things called event here and there.

Examples:

- https://github.com/ros2/rcl/blob/master/rcl/include/rcl/event.h
- https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/event.hpp

Could we use a more specific name?
ClockEvent?

rclpy/test/test_clock.py
```diff
@@ -12,18 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import threading
```
ivanpauno, line 15: Some of my comments in the tests cases of https://github.com/ros2/rclpy/pull/864 also apply here

ivanpauno (commented):
I have some minor comments, otherwise lgtm!

rclpy/rclpy/clock.py
sloretz: I don't think it should because `until` is specified with the old time source. It's possible for ROS time to be enabled and then `/clock` to publish really small numbers that would cause sleep to not wake for years. This might happen when playing an old bag file, or connecting to a simulator where time may be close to zero.

rclpy/test/test_clock.py
sloretz: The 99999 timeout was also indirectly testing that the methods returned sooner than the duration of time they were asked to sleep. In 8615b7b I added explicit checks that they return in a small amount of time.

clalancette (commented):
I agree with @ivanpauno's comment about renaming `Event` -> `ClockEvent`.  I also have expressed concern about the possibility of flaky tests here, but I actually don't see a reasonable way to fix it given what this is trying to test.  So once the rename is done, I'm happy to approve.

rclpy/src/rclpy/event.cpp
sloretz: `Event` -> `ClockEvent` in cd90a7765306da3021c41629b21b640d74e2436f and dbf83fe13005b8cdb1e3dd83ecacb6706501b7e2

fujitatomoya (approved):
lgtm

rclpy/rclpy/clock.py
ivanpauno: Sounds good.
Is this consistent with rclpcpp behavior?

clalancette (approved):

rclpy/rclpy/clock.py
sloretz: >  Is this consistent with rclpcpp behavior?

Yup, rclcpp also returns `false` when the time source changes https://github.com/ros2/rclcpp/blob/b2b676d3172ada509e58fa552a676a446809d83c/rclcpp/test/rclcpp/test_time.cpp#L571-L585

rclpy/test/test_clock.py
ivanpauno: Of my comments in the other PR, I think that the only one relevant for the moment is using `thread.join()` instead of a sleep() call to wait the thread to exit.
The other comments can be ignored.

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15783/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10461/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13445/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16039/)

PR #864, opened 2021-12-13
Add Clock.sleep_for using sleep_until

### LINKED ISSUE #617: Add sleep function that respects ROS time

sloretz:
Resolves #617 
Requires #858

### DIFF
--- a/rclpy/rclpy/clock.py
+++ b/rclpy/rclpy/clock.py
@@ -236,6 +236,30 @@ class Clock:
 
         return self.now() >= until
 
+    def sleep_for(self, rel_time: Duration, context=None) -> bool:
+        """
+        Sleep for a specified duration.
+
+        Equivalent to:
+
+        .. code-block:: python
+
+            clock.sleep_until(clock.now() + rel_time, context)
+
+
+        When using a ROSClock, this may sleep forever if the TimeSource is misconfigured and the
+        context is never shut down.
+        ROS time being activated or deactivated causes this function to cease sleeping and return
+        False.
+
+        :param rel_time: Duration of time to sleep for.
+        :param context: Context which when shut down will cause this sleep to wake early.
+            If context is None, then the default context is used.
+        :return: True if the full duration was slept, or False if it woke for another reason.
+        :raises NotInitializedException: context has not been initialized or is shutdown.
+        """
+        return self.sleep_until(self.now() + rel_time, context)
+
 
 class ROSClock(Clock):
 
--- a/rclpy/test/test_clock.py
+++ b/rclpy/test/test_clock.py
@@ -245,10 +245,21 @@ def test_sleep_until_non_default_context(non_default_context):
     assert clock.sleep_until(clock.now() + Duration(seconds=0.1), context=non_default_context)
 
 
+def test_sleep_for_non_default_context(non_default_context):
+    clock = Clock()
+    assert clock.sleep_for(Duration(seconds=0.1), context=non_default_context)
+
+
 def test_sleep_until_invalid_context():
     clock = Clock()
     with pytest.raises(NotInitializedException):
         clock.sleep_until(clock.now() + Duration(seconds=0.1), context=Context())
+
+
+def test_sleep_for_invalid_context():
+    clock = Clock()
+    with pytest.raises(NotInitializedException):
+        clock.sleep_for(Duration(seconds=0.1), context=Context())
 
 
 @pytest.mark.parametrize(
@@ -264,6 +275,17 @@ def test_sleep_until_basic(default_context, clock_type):
 
 @pytest.mark.parametrize(
     'clock_type', (ClockType.SYSTEM_TIME, ClockType.STEADY_TIME, ClockType.ROS_TIME))
+def test_sleep_for_basic(default_context, clock_type):
+    clock = Clock(clock_type=clock_type)
+    sleep_duration = Duration(seconds=0.1)
+    start = clock.now()
+    assert clock.sleep_for(sleep_duration)
+    stop = clock.now()
+    assert stop - start >= sleep_duration
+
+
+@pytest.mark.parametrize(
+    'clock_type', (ClockType.SYSTEM_TIME, ClockType.STEADY_TIME, ClockType.ROS_TIME))
 def test_sleep_until_time_in_past(default_context, clock_type):
     clock = Clock(clock_type=clock_type)
     sleep_duration = Duration(seconds=-1)
@@ -271,6 +293,18 @@ def test_sleep_until_time_in_past(default_context, clock_type):
     assert clock.sleep_until(clock.now() + sleep_duration)
     stop = clock.now()
     assert stop - start < A_SMALL_AMOUNT_OF_TIME
+
+
+@pytest.mark.parametrize(
+    'clock_type', (ClockType.SYSTEM_TIME, ClockType.STEADY_TIME, ClockType.ROS_TIME))
+def test_sleep_for_negative_duration(default_context, clock_type):
+    clock = Clock(clock_type=clock_type)
+    sleep_duration = Duration(seconds=-1)
+    start = clock.now()
+    assert clock.sleep_for(sleep_duration)
+    stop = clock.now()
+    a_small_amount_of_time = Duration(seconds=0.01)
+    assert stop - start < a_small_amount_of_time
 
 
 @pytest.mark.parametrize('ros_time_enabled', (True, False))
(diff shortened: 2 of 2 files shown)


rclpy/test/test_clock.py
```diff
@@ -271,6 +293,18 @@ def test_sleep_until_time_in_past(default_context, clock_type):
...
+@pytest.mark.parametrize(
+    'clock_type', (ClockType.SYSTEM_TIME, ClockType.STEADY_TIME, ClockType.ROS_TIME))
+def test_sleep_for_negative_duration(default_context, clock_type):
+    clock = Clock(clock_type=clock_type)
+    sleep_duration = Duration(seconds=-1)
+    start = clock.now()
+    assert clock.sleep_for(sleep_duration)
+    stop = clock.now()
+    a_small_amount_of_time = Duration(seconds=0.01)
+    assert stop - start < a_small_amount_of_time
```
ivanpauno, line 305: This could cause some flakyness issues in CI.
I wouldn't expect much problem if we use a bigger tolerance, e.g. 0.5.

rclpy/test/test_clock.py
```diff
@@ -296,6 +330,31 @@ def run():
...
+
+    def run():
+        nonlocal retval
+        retval = clock.sleep_for(Duration(seconds=999999))
+
+    t = threading.Thread(target=run, daemon=True)
+    t.start()
+
+    # wait for thread to get inside sleep_for call
+    time.sleep(0.2)
```
ivanpauno, line 348: I don't like much a test depending on a race condition.
I would rather not have a test if this is the only way.

By the way, I don't get what's the intention of the test.

rclpy/test/test_clock.py
```diff
@@ -319,6 +378,28 @@ def run():
...
+    retval = None
+
+    def run():
+        nonlocal retval
+        retval = clock.sleep_for(Duration(seconds=999999), context=non_default_context)
+
+    t = threading.Thread(target=run, daemon=True)
+    t.start()
+
+    # wait for thread to get inside sleep_for call
```
ivanpauno, line 392: same

rclpy/test/test_clock.py
```diff
@@ -347,6 +428,35 @@ def run():
...
+    retval = None
+
+    def run():
+        nonlocal retval
+        retval = clock.sleep_for(sleep_duration)
+
+    t = threading.Thread(target=run, daemon=True)
+    t.start()
+
+    # wait for thread to get inside sleep_for call
```
ivanpauno, line 449: same here as well

rclpy/test/test_clock.py
ivanpauno: does is matter if `shutdown` was called before or after the thread got inside `sleep_for`?
It seems that it doesn't, in that case I would remove the sleep

rclpy/test/test_clock.py
```diff
@@ -319,6 +378,28 @@ def run():
...
+    time.sleep(0.2)
+
+    non_default_context.shutdown()
+
+    # wait for thread to exit
+    time.sleep(0.2)
```
ivanpauno, line 398: can we join the thread instead?

sloretz: Joining the thread in a496961

rclpy/test/test_clock.py
sloretz: Using 0.5s in 3dd61e7

rclpy/test/test_clock.py
sloretz: Yeah, it's not my favorite either. Another way may be to inspect the current frame of the other thread to see what code it's exiting, but I don't think I can implement that robustly in a reasonable time.

https://github.com/ros2/rclpy/pull/858#discussion_r768967247

---

The `ros_time_toggled` tests enable or disable ROS time while the method is sleeping. Since that switches from system time to `/clock` time, it's not clear what amount of time should be slept when that happens. This test makes sure the method wakes up and returns False.

rclpy/test/test_clock.py
sloretz: It does. This test checks that the sleep method will sleep on an initially valid context and then wake if it was shutdown. If `shutdown` was called before the thread got inside sleep_for then the method would raise `NotInitializedException` instead.

rclpy/test/test_clock.py
ivanpauno: A reasonable alternative is to toggle the clock type in a loop, wait until `retval is not None` and then assert that `retval is False`.
The loop could have a certain maximum duration, and we could make it "less busy" be sleeping for a bit between each time we toggle the clock type.


e.g.:


```python3
    retval = None

    def run():
        nonlocal retval
        retval = clock.sleep_for(Duration(seconds=999999))

    t = threading.Thread(target=run)
    t.start()

    start = time.now()
    # wait for thread to get inside sleep_for call
    while retval is None and time.now() < start + 0.5:
            clock._set_ros_time_is_active(ros_time_enabled)
            ros_time_enabled = not ros_time_enabled

    assert retval is False
    t.join()
```

Not really amazing, and it doens't make much sense to have the test parameterized with `ros_time_enabled` (as we really don't now which of the toggles made the clock stop waiting), but probably a bit more robust.

ivanpauno: anyway, I'm okay with the current approach, but we should track CI the first few days to see if the tests are reliable.

rclpy/test/test_clock.py
ivanpauno: I see, I don't have a suggestion of how to improve this one.
I'm not sure if the inconsistency makes sense, as while trying to shutdown asynchronously it would trigger different behaviors depending if the wait_until()/wait_for() called already happened or not.

I think this approach is acceptable for the moment.

ivanpauno (approved):

sloretz:
Rebased and incorporated feedback in tests from #858 that also apply here

sloretz:
CI (build: ` --packages-up-to rclpy` test: `--packages-select rclpy`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15784/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10462/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13446/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16041/)

rclpy/rclpy/clock.py
```diff
@@ -236,6 +236,30 @@ def on_time_jump(time_jump: TimeJump):
...
+        When using a ROSClock, this may sleep forever if the TimeSource is misconfigured and the
+        context is never shut down.
+        ROS time being activated or deactivated causes this function to cease sleeping and return
+        False.
+
+        :param rel_time: Duration of time to sleep for.
+        :param context: Context which when shut down will cause this sleep to wake early.
+            If context is None, then the default context is used.
+        :return: True if the full duration was slept, or False if it woke for another reason.
+        :raises NotInitializedException: context has not been initialized or is shutdown.
```
j-rivero, line 259: Looking into the `sleep_until` documentation, I think that we also want to add the `ValueError` exception here.
```suggestion
        :raises ValueError: until is specified for a different type of clock than this one.
        :raises NotInitializedException: context has not been initialized or is shutdown.
```

j-rivero (approved):
A minor Nit on the documentation. Looks good to me.

There is value in the comments about how to make the test more robust and I would like not to lose them after merging. Not sure if adding the URLs as comments in the code or opening an issue would help in the future to reach these comments if the test goes flaky at some point. I leave up to you @sloretz.

Thanks for the work!.

rclpy/rclpy/clock.py
sloretz: I don't think `sleep_for` can raise `ValueError` because the time `until` is coming from `self.now()` in `Clock.sleep_for()`. If the clock type doesn't match then there is a bug in either `Clock.now()` or in the `Time` class.

sloretz:
> There is value in the comments about how to make the test more robust and I would like not to lose them after merging. Not sure if adding the URLs as comments in the code or opening an issue would help in the future to reach these comments if the test goes flaky at some point. I leave up to you @sloretz.

They're valid comments about the test being potentially flaky. I think the right thing to do is to watch the repeated jobs for failures, especially Windows where IIUC the OS scheduler waits a while between context switches. I looked through the last 3 repeated jobs and I don't see any `sleep_until` failures, so it's looking good that these won't be flaky either. I linked the very latest if you'd like to click back through them too.

* Linux x86: https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/2505/#showFailuresLink 
* Linux ARM: https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_repeated/1811/#showFailuresLink
* OSX: https://ci.ros2.org/view/nightly/job/nightly_osx_repeated/2574/#showFailuresLink
* Windows: https://ci.ros2.org/view/nightly/job/nightly_win_rep/2464/#showFailuresLink

If they become flaky in the future, links to the discussion are always available by doing `git blame`, looking at the PR number in the title of the commit (github adds it during `Squash and merge`), and then opening it. An issue doesn't seem to be necessary because the tests aren't showing flaky behavior on CI. I think this is fine as is.

sloretz:
Resolved PR comments with Jose out-of-band

PR #868, opened 2021-12-22
Make rclpy.try_shutdown() behavior to follow rclpy.shutdown() more closely

ivanpauno:
Needed by https://github.com/ros2/ros2cli/pull/683.

I'm not sure if this is the ideal approach, maybe init()/shutdown() are doing too much (?).
I'm open to alternatives.

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -54,7 +54,7 @@ from rclpy.utilities import get_default_context
 from rclpy.utilities import get_rmw_implementation_identifier  # noqa: F401
 from rclpy.utilities import ok  # noqa: F401 forwarding to this module
 from rclpy.utilities import shutdown as _shutdown
-from rclpy.utilities import try_shutdown as _try_shutdown
+from rclpy.utilities import try_shutdown  # noqa: F401
 
 # Avoid loading extensions on module import
 if TYPE_CHECKING:
@@ -100,6 +100,13 @@ def get_global_executor() -> 'Executor':
         # imported locally to avoid loading extensions on module import
         from rclpy.executors import SingleThreadedExecutor
         __executor = SingleThreadedExecutor()
+        context = get_default_context()
+
+        def reset_executor():
+            global __executor
+            __executor.shutdown()
+            __executor = None
+        context.on_shutdown(reset_executor)
     return __executor
 
 
@@ -116,10 +123,6 @@ def shutdown(*, context: Context = None, uninstall_handlers: Optional[bool] = No
         If `True`, signal handlers will be uninstalled.
         If not, signal handlers won't be uninstalled.
     """
-    global __executor
-    if __executor is not None:
-        __executor.shutdown()
-        __executor = None
     _shutdown(context=context)
     if (
         uninstall_handlers or (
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/utilities.py
```diff
@@ -60,9 +60,19 @@ def shutdown(*, context=None):
...
-        context = get_default_context()
-    return context.try_shutdown()
+        # Replace the default context with a new one if shutdown was successful
+        # or if the context was already shutdown.
+        with g_context_lock:
+            if g_default_context is None:
+                g_default_context = Context()
+            g_default_context.try_shutdown()
+            if not g_default_context.ok():
+                g_default_context = None
```
sloretz, line 73: I wish the default context registered an on_shutdown callback to manage the global variable too.

rclpy/rclpy/__init__.py
```diff
@@ -129,6 +129,23 @@ def shutdown(*, context: Context = None, uninstall_handlers: Optional[bool] = No
...
+
+    :param context: The context to invalidate. If ``None``, then the default context is used
+        (see :func:`.get_default_context`).
+    """
+    global __executor
+    _try_shutdown(context=context)
+    if not ok():
+        if __executor is not None:
+            __executor.shutdown()
+            __executor = None
```
sloretz, line 146: I wish the default executor registered an `on_shutdown()` callback with the default Context and handled itself that way. Would you be willing to try that? If not, I'd approve this change as it does seem to make the two shutdown functions more alike.

rclpy/rclpy/utilities.py
ivanpauno: Currently all the work is done with the default context lock taken.
If we register an on_shutdown() callback I don't think that would work.

Maybe it's possible to do that with some extra trick.

rclpy/rclpy/__init__.py
ivanpauno: That seems to be working correctly, see https://github.com/ros2/rclpy/pull/868/commits/cb23bd7827a6e69c7348021d53dc1877bd9fc705

ivanpauno:
I had one extra issue related to https://github.com/ros2/rclpy/pull/868#discussion_r774045161 after doing some further testing.
The `reset_executor()` closure I was passing to `on_shutdown()` doesn't have any owner.
As `Context` only retains weak ownership of the callbacks being passed it was sometimes not being called.

https://github.com/ros2/rclpy/pull/868/commits/d65b4871a3122a927703c8cc94f1ccc2dae628a8 fixes that.
After that change, we retaing week ownership of methods and shared ownership of free functions or closures.
I think that's intuitive.
I could also see the case of always retaining shared ownership and the user explictily passing a weak object when they need, but that would be an unexpected change (particularly for methods).

rclpy/rclpy/context.py
```diff
@@ -127,7 +127,7 @@ def on_shutdown(self, callback: Callable[[], None]):
                 if ismethod(callback):
                     self._callbacks.append(weakref.WeakMethod(callback, self._remove_callback))
                 else:
-                    self._callbacks.append(weakref.ref(callback, self._remove_callback))
+                    self._callbacks.append(callback)
```
sloretz, line 130: > I could also see the case of always retaining shared ownership and the user explictily passing a weak object when they need that, but that would be an unexpected change (particularly for methods).

Oh hmm. This is interesting. I don't see a better way, and it does seem like the behavior would be intuitive.

sloretz (approved):

ivanpauno:
CI for this PR alone (I will rerun PR for https://github.com/ros2/ros2cli/pull/683 after merging this one):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15830/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10506/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13494/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16114/)

Issue #870, opened 2022-01-04
MultiThreadedExecutor race condition in client.call_async


augustelalande:
This is related to #564, but the race condition goes deeper than what was stated there.

Essentially, this line:
https://github.com/ros2/rclpy/blob/379ce0a2196dd1418db4a2a98c1a9b792d99d5c9/rclpy/rclpy/executors.py#L354
can execute before this line:
https://github.com/ros2/rclpy/blob/379ce0a2196dd1418db4a2a98c1a9b792d99d5c9/rclpy/rclpy/client.py#L135

The executor then ignores the request as 'cancelled'
https://github.com/ros2/rclpy/blob/379ce0a2196dd1418db4a2a98c1a9b792d99d5c9/rclpy/rclpy/executors.py#L355-L357
and the client is stuck waiting for the future to finish.

augustelalande:
Potentially, the solution would be to execute call_async with a lock and then require the same lock when fetching the pending_request.

jacobperron:
Resolved in #871 

I've also created a ticket for a similar race in the action client: https://github.com/ros2/rclpy/issues/878

PR #871, opened 2022-01-04
Fix multi-threaded race condition in client.call_async

augustelalande:
Proposal to fix multi-threaded race condition in client.call_async as described in #870

### DIFF
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -64,7 +64,7 @@ class Client:
         # True when the callback is ready to fire but has not been "taken" by an executor
         self._executor_event = False
 
-        self.call_lock = threading.Lock()
+        self._lock = threading.Lock()
 
     def call(self, request: SrvTypeRequest) -> SrvTypeResponse:
         """
@@ -128,7 +128,7 @@ class Client:
         if not isinstance(request, self.srv_type.Request):
             raise TypeError()
 
-        with self.call_lock:
+        with self._lock:
             with self.handle:
                 sequence_number = self.__client.send_request(request)
             if sequence_number in self._pending_requests:
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -132,7 +132,7 @@ class Client:
             with self.handle:
                 sequence_number = self.__client.send_request(request)
             if sequence_number in self._pending_requests:
-                raise RuntimeError('Sequence (%r) conflicts with pending request' % sequence_number)
+                raise RuntimeError(f'Sequence ({sequence_number}) conflicts with pending request')
 
             future = Future()
             self._pending_requests[sequence_number] = future
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -99,22 +99,6 @@ class Client:
             raise future.exception()
         return future.result()
 
-    def remove_pending_request(self, future: Future) -> None:
-        """
-        Remove a future from the list of pending requests.
-
-        This prevents a future from receiving a response and executing its done callbacks.
-
-        :param future: A future returned from :meth:`call_async`
-        """
-        for seq, req_future in self._pending_requests.items():
-            if future == req_future:
-                try:
-                    del self._pending_requests[seq]
-                except KeyError:
-                    pass
-                break
-
     def call_async(self, request: SrvTypeRequest) -> Future:
         """
         Make a service request and asyncronously get the result.
@@ -140,6 +124,33 @@ class Client:
             future.add_done_callback(self.remove_pending_request)
 
         return future
+
+    def get_pending_request(self, sequence_number: int) -> Future:
+        """
+        Get a future from the list of pending requests.
+
+        :param sequence_number: Number identifying the pending request.
+        :return: The future corresponding to the sequence_number.
+        :raises: KeyError if the sequence_number is not in the pending requests.
+        """
+        with self._lock:
+            return self._pending_requests[sequence_number]
+
+    def remove_pending_request(self, future: Future) -> None:
+        """
+        Remove a future from the list of pending requests.
+
+        This prevents a future from receiving a response and executing its done callbacks.
+
+        :param future: A future returned from :meth:`call_async`
+        """
+        for seq, req_future in self._pending_requests.items():
+            if future == req_future:
+                try:
+                    del self._pending_requests[seq]
+                except KeyError:
+                    pass
+                break
 
     def service_is_ready(self) -> bool:
         """
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -64,6 +64,8 @@ class Client:
         # True when the callback is ready to fire but has not been "taken" by an executor
         self._executor_event = False
 
+        self.call_lock = threading.Lock()
+
     def call(self, request: SrvTypeRequest) -> SrvTypeResponse:
         """
         Make a service request and wait for the result.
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/client.py
```diff
@@ -64,6 +64,8 @@ def __init__(
         # True when the callback is ready to fire but has not been "taken" by an executor
         self._executor_event = False
 
+        self.call_lock = threading.Lock()
```
fujitatomoya, line 67: simplify the lock name for class weak internal used just like others?

```suggestion

        self._lock = threading.Lock()
```

fujitatomoya (commented):
i think this makes sense to avoid the possible racy condition in theory.

rclpy/rclpy/client.py
augustelalande: Done

fujitatomoya (approved):

fujitatomoya:
@ivanpauno @jacobperron could you take a look at this when you have time?

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15920/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10596/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16245/)

jacobperron:
Tricky..

I suppose there are similar threading issues with the action client, wherever we are making service requests, for instance

https://github.com/ros2/rclpy/blob/289d7d32117c7ba9d7f2fec58f53f3ee2ed9cf2a/rclpy/rclpy/action/client.py#L435

has a race with

https://github.com/ros2/rclpy/blob/7a906afb0fe3b54e4e4958e4d2fee94e8823162e/rclpy/rclpy/executors.py#L383

It's possible the executor tries to process a request before the future is available in the list.

It's too bad we have to introduce a lock at the scope of sending the request, but I can't think of another way. Hopefully it won't have a significant impact to performance.

@sloretz Thoughts?

rclpy/rclpy/executors.py
```diff
@@ -351,7 +351,8 @@ async def _execute_client(self, client, seq_and_response):
         if header is not None:
             try:
                 sequence = header.request_id.sequence_number
-                future = client._pending_requests[sequence]
+                with client._lock:
+                    future = client._pending_requests[sequence]
```
jacobperron, line 355: IMO, it might be nice to move some of this logic inside `Client`. For example, the business about the lock and payload (header/response). It moves the Client more towards conforming to the Waitable interface:

```python
async def _execute_client(self, client, seq_and_response):
    header, response = seq_and_response
    if header is None:
        return
    future = client.get_request_future(header)
    future._set_executor(self)
    await client.execute(header, response)
```

Then we can use the lock inside `client.get_request_future()` and `client.execute()`.

Alternatively, we could let the execute method do everything:

```python
async def _execute_client(self, client, seq_and_response):
    await client.execute(self, seq_and_response)
```

This kind of refactor might be left to a follow-up PR as well.

Thoughts?

augustelalande: What should the function do if the sequence_number is not in the pending requests? Probably shouldn't let the KeyError exception fall through?

ivanpauno: > What should the function do if the sequence_number is not in the pending requests? Probably shouldn't let the KeyError exception fall through?

I think that letting `KeyError` fall through is fine.
It should be documented in the docstring.

Executor implementation can continue catching the exception in the same way it's currently being done.

augustelalande: Ok, I added the get_pending_request method. (I also took the liberty of moving around the remove_pending_request method so that they would be grouped together nicely).

augustelalande:
> Tricky..
> 
> I suppose there are similar threading issues with the action client, wherever we are making service requests, for instance
> 
> https://github.com/ros2/rclpy/blob/289d7d32117c7ba9d7f2fec58f53f3ee2ed9cf2a/rclpy/rclpy/action/client.py#L435
> 
> has a race with
> 
> https://github.com/ros2/rclpy/blob/7a906afb0fe3b54e4e4958e4d2fee94e8823162e/rclpy/rclpy/executors.py#L383
> 
> It's possible the executor tries to process a request before the future is available in the list.
> 
> It's too bad we have to introduce a lock at the scope of sending the request, but I can't think of another way. Hopefully it won't have a significant impact to performance.
> 
> @sloretz Thoughts?

Yes this race condition definitely exists in the ActionClient. Actually, even #564 still exists there. The question is why is the Client functionality being reimplemented in ActionClient? Why not reuse the existing Client functionality by creating sub-clients to handle goal requests, cancel requests etc.?

I guess this should be a new issue.

ivanpauno (approved):

ivanpauno:
> Yes this race condition definitely exists in the ActionClient. Actually, even #564 still exists there. The question is why is the Client functionality being reimplemented in ActionClient? Why not reuse the existing Client functionality by creating sub-clients to handle goal requests, cancel requests etc.?

:+1:, I'm not sure how easy is to do that but I think it would be good to reuse the existing client code.

> I guess this should be a new issue.

Yes, I don't think we need to address that here.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15937/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10610/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16259/)

jacobperron:
I think the Action Client doesn't share the same logic as the Client because it is using the `rcl_action` C-API under the hood. Perhaps it possible to refactor the code in rclpy such that they share the same high-level logic; I don't recall if this was considered when actions were implemented.

rclpy/rclpy/client.py
```diff
@@ -126,18 +112,46 @@ def call_async(self, request: SrvTypeRequest) -> Future:
...
+
+    def remove_pending_request(self, future: Future) -> None:
+        """
+        Remove a future from the list of pending requests.
+
+        This prevents a future from receiving a response and executing its done callbacks.
+
+        :param future: A future returned from :meth:`call_async`
+        """
+        for seq, req_future in self._pending_requests.items():
```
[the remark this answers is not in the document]
augustelalande: If we lock the whole thing I think we can drop the `try ... except KeyError` from the method?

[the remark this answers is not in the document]
jacobperron: Yeah I think so :+1: . I suppose the try-except is there because of the race.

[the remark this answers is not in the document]
augustelalande: Done.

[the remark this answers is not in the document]
ivanpauno: > If we lock the whole thing I think we can drop the try ... except KeyError from the method?
> Yeah I think so +1 . I suppose the try-except is there because of the race.

I don't think that was releated to the race.
The remove_pending_requests() method is used to avoid leaking when the server never answers and you're not interested in the response anymore (e.g. after waiting with a timeout).
The request might have been removed from the `_pending_requests` list before the user called the method, e.g. because a response just arrived between the timeout expired and the method call.

The try...except is needed, even with the lock.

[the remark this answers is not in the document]
ivanpauno: ah yes, I misread the code, thanks!

ivanpauno (approved):

jacobperron (approved):

jacobperron:
I've rebased on `master` to recent changes related to pybind11, hopefully this resolves CI failures:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15949/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10622/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16271/)

PR #874, opened 2022-01-06
Update to pybind11 2.7.1

clalancette:
We can remove most of the hacks we were carrying around, but we also need to add in one more overload of the Python executable so pybind11 properly detects things on Windows Debug.

Depends on https://github.com/ros2/pybind11_vendor/pull/10 (CI is over in that PR)

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -35,19 +35,14 @@ find_package(Python3 REQUIRED Interpreter Development)
 set(_PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE}")
 if(WIN32 AND CMAKE_BUILD_TYPE STREQUAL "Debug")
   # Set the python debug interpreter.
-  # pybind11 will setup the build for debug now.
+  # We need to set *both* PYTHON_EXECUTABLE (for use in this CMake file),
+  # as well as Python3_EXECUTABLE (which is what pybind11 uses).
   set(PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE_DEBUG}")
+  set(Python3_EXECUTABLE "${PYTHON_EXECUTABLE_DEBUG}")
 endif()
 
 find_package(pybind11_vendor REQUIRED)
 find_package(pybind11 REQUIRED)
-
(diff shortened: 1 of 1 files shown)


rclpy/CMakeLists.txt
```diff
@@ -30,49 +30,20 @@ find_package(rmw_implementation_cmake REQUIRED)
...
+find_package(Python3 REQUIRED Interpreter Development)
 
 set(_PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE}")
 if(WIN32 AND CMAKE_BUILD_TYPE STREQUAL "Debug")
   # Set the python debug interpreter.
-  # pybind11 will setup the build for debug now.
+  # We need to set *both* PYTHON_EXECUTABLE (for use in this CMake file),
+  # as well as Python3_EXECUTABLE (which is what pybind11 uses).
   set(PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE_DEBUG}")
+  set(Python3_EXECUTABLE "${PYTHON_EXECUTABLE_DEBUG}")
```
sloretz, line 41: I was just looking at something similar at tf2_py. Since `ament_cmake` uses `FindPython3`, If `Python3_EXECUTABLE` is set before `find_package(Python3)`, then all the uses of `PYTHON_EXECUTABLE` can be removed: https://github.com/ros2/geometry2/pull/494

I would recommend doing that here too by putting this code before any other `find_package()` call, then deleting `PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE}"` and  and `set(PYTHON_EXECUTABLE "${_PYTHON_EXECUTABLE}")`

```cmake
# Figure out Python3 debug/release before anything else can find_package it
if(WIN32 AND CMAKE_BUILD_TYPE STREQUAL "Debug")
  find_package(python_cmake_module REQUIRED)
  find_package(PythonExtra REQUIRED)
  # Force FindPython3 to use the debug interpretter where ROS 2 expects it
  set(Python3_EXECUTABLE "${PYTHON_EXECUTABLE_DEBUG}")
endif()
find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
```

sloretz (approved):

PR #876, opened 2022-01-07
Bugfix/duration to msg precision

### LINKED ISSUE #875: rclpy.duration.Duration.to_msg() loses precision
## Bug report

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binaries
- Version or commit hash:
  - ros-foxy-desktop 0.9.2-1focal.20211119.123452
  - rclpy

#### Steps to reproduce issue
```python
from rclpy.duration import Duration
d = Duration(nanoseconds=1234567890987654321)
m = d.to_msg()
print((d.nanoseconds % 10**9, m.nanosec))
```

#### Expected behavior
Prints `(987654321, 987654321)`.

#### Actual behavior
Prints `(987654321, 987654400)`.

suurjaak:
Fixes #875.

### DIFF
--- a/rclpy/docs/source/api.rst
+++ b/rclpy/docs/source/api.rst
@@ -7,6 +7,7 @@ API
 
    api/init_shutdown
    api/node
+   api/duration
    api/topics
    api/services
    api/actions
--- a/rclpy/docs/source/api.rst
+++ b/rclpy/docs/source/api.rst
@@ -7,7 +7,6 @@ API
 
    api/init_shutdown
    api/node
-   api/duration
    api/topics
    api/services
    api/actions
--- a/rclpy/rclpy/duration.py
+++ b/rclpy/rclpy/duration.py
@@ -23,7 +23,7 @@ class Duration:
         """
         Create an instance of :class:`Duration`, combined from given seconds and nanoseconds.
 
-        :param seconds: Time span seconds, if any, fractional part will be discarded.
+        :param seconds: Time span seconds, if any, fractional part will be included.
         :param nanoseconds: Time span nanoseconds, if any, fractional part will be discarded.
         """
         total_nanoseconds = int(seconds * 10**9)
--- a/rclpy/rclpy/duration.py
+++ b/rclpy/rclpy/duration.py
@@ -20,8 +20,15 @@ CONVERSION_CONSTANT = 10 ** 9
 
 
 class Duration:
+    """A period between two time points, with nanosecond precision."""
 
     def __init__(self, *, seconds=0, nanoseconds=0):
+        """
+        Create an instance of :class:`Duration`, combined from given seconds and nanoseconds.
+
+        :param seconds: Time span seconds, if any, fractional part will be discarded.
+        :param nanoseconds: Time span nanoseconds, if any, fractional part will be discarded.
+        """
         total_nanoseconds = int(seconds) * CONVERSION_CONSTANT
         total_nanoseconds += int(nanoseconds)
         if total_nanoseconds >= 2**63 or total_nanoseconds < -2**63:
@@ -75,11 +82,22 @@ class Duration:
         return NotImplemented
 
     def to_msg(self):
+        """
+        Get duration as :class:`builtin_interfaces.msg.Duration`.
+
+        :returns: duration as message
+        :rtype: builtin_interfaces.msg.Duration
+        """
         seconds, nanoseconds = divmod(self.nanoseconds, CONVERSION_CONSTANT)
         return builtin_interfaces.msg.Duration(sec=seconds, nanosec=nanoseconds)
 
     @classmethod
     def from_msg(cls, msg):
+        """
+        Create an instance of :class:`Duration` from a duration message.
+
+        :param msg: An instance of :class:`builtin_interfaces.msg.Duration`.
+        """
         if not isinstance(msg, builtin_interfaces.msg.Duration):
             raise TypeError('Must pass a builtin_interfaces.msg.Duration object')
         return cls(seconds=msg.sec, nanoseconds=msg.nanosec)
--- a/rclpy/rclpy/duration.py
+++ b/rclpy/rclpy/duration.py
@@ -16,9 +16,6 @@ import builtin_interfaces.msg
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
 
-CONVERSION_CONSTANT = 10 ** 9
-
-
 class Duration:
     """A period between two time points, with nanosecond precision."""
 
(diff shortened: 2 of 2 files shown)


rclpy/rclpy/duration.py
```diff
@@ -16,10 +16,20 @@
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
 
+CONVERSION_CONSTANT = 10 ** 9
```
DLu, line 19: Perhaps change this to `NANOSECONDS_PER_SECOND` and add a comment noting that the type is `int` and not `float` like `1e9`

suurjaak: Can do, of course, but I introduced this constant on the basis of https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/time.py#L20. Maybe better to keep these two sibling modules more alike?

aprotyas: > Maybe better to keep these two sibling modules more alike?

This is generally a good idea, but the constant name here is not very descriptive - why propagate bad style? Having said that:

```suggestion
NANOSECONDS_PER_SECOND = 10 ** 9
```

rclpy/rclpy/duration.py
```diff
@@ -16,10 +16,20 @@
...
 
     def __init__(self, *, seconds=0, nanoseconds=0):
-        total_nanoseconds = int(seconds * 1e9)
+        """
+        Create an instance of :class:`Duration`, combined from given seconds and nanoseconds.
+
+        :param seconds: Time span seconds, if any, fractional part will be discarded.
+        :param nanoseconds: Time span nanoseconds, if any, fractional part will be discarded.
+        """
+        total_nanoseconds = int(seconds) * CONVERSION_CONSTANT
```
aprotyas, line 32: This causes test failures like https://build.ros2.org/job/Rpr__rclpy__ubuntu_focal_amd64/674/testReport/junit/rclpy.rclpy.test.test_time/TestTime/test_duration_construction/

How about:
```suggestion
        total_nanoseconds = int(seconds * CONVERSION_CONSTANT)
```

This discards the fractional part after converting to nanoseconds, rather than before. I'm not sure if this is the intended semantics though.

rclpy/rclpy/duration.py
suurjaak: On second thought, dropped the constant altogether as the module does not need to mirror `rclpy.Time` that thoroughly.

rclpy/rclpy/duration.py
suurjaak: Made the logic mirror that in `rclpy.Time`. 

The semantics of it are a bit questionable, though. If `rclcpp` can be taken as the reference implementation, then there the [corresponding constructor](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/duration.cpp#L34) accepts only integer values. And in general it's a bad idea to use floats for sub-second precision. 

But that would go beyond this bugfix. 

Some tests, however, still fail. Looks like parts of the stack might already be coded to expect this specific floating-point behavior..

aprotyas: Yeah, I don't think the change I suggested is the way to go, but also as you mentioned other parts of the stack expect specific floating-point behavior.

suurjaak: The problem with the failing test is that it's poorly written, very much on the edge of failing due to sub-millisecond timing differences between different test runs. Some test runs like https://build.ros2.org/job/Rpr__rclpy__ubuntu_focal_amd64/677/ succeeded despite no functional differences in code. And looking at builds for other pull requests, this test fails frequently, for no reasons connected with changes made. 

I made another change - discovered and made use of https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/constants.py. No functional difference, but now all tests succeed. 

Do you think I should rewrite the test to be less sensitive to random timing differences? I would assert the length of callbacks in https://github.com/ros2/rclpy/blob/master/rclpy/test/test_timer.py#L104 to be in `(2, 3)`.

rclpy/docs/source/api.rst
```diff
@@ -7,6 +7,7 @@ API
 
    api/init_shutdown
    api/node
+   api/duration
```
iuhilnehc-ynos, line 10: It seems that you forgot to add the file `rclpy/docs/source/api/duration.rst`?

iuhilnehc-ynos (commented):
LGTM with the `rclpy.constants.S_TO_NS`

rclpy/docs/source/api.rst
suurjaak: Oops, that was an accidental commit. I was experimenting with building the docs to see what the generated documentation looks like. But since the result was rather useless, and I'm not familiar enough with Sphinx configuration to fix that, I did not plan to include it. 

Reverted.

rclpy/rclpy/duration.py
aprotyas: > The problem with the failing test is that it's poorly written, very much on the edge of failing due to sub-millisecond timing differences between different test runs

I agree that the nature of the test failures does not always reflect proposed changes.


> Do you think I should rewrite the test to be less sensitive to random timing differences? I would assert the length of callbacks in https://github.com/ros2/rclpy/blob/master/rclpy/test/test_timer.py#L104 to be in `(2, 3)`.

That's not a bad idea, but I reckon you should file a separate issue about the tests (if for nothing else but visibility) and then make PRs to close said issue. That'll also give a chance for maintainers to share their opinion about the timing tests.

aprotyas (approved):

rclpy/rclpy/duration.py
clalancette: >  > Do you think I should rewrite the test to be less sensitive to random timing differences? I would assert the length of callbacks in https://github.com/ros2/rclpy/blob/master/rclpy/test/test_timer.py#L104 to be in (2, 3).

> That's not a bad idea, but I reckon you should file a separate issue about the tests (if for nothing else but visibility) and then make PRs to close said issue. That'll also give a chance for maintainers to share their opinion about the timing tests.

Agreed.  We would very much appreciate a new issue and a new PR to make the timing tests more reliable.  Thanks!

clalancette (approved):
Looks good to me with green CI.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15914/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10589/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16238/)

clalancette:
FYI, I rebased and force-pushed this onto the latest so that I can get CI to run.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15914/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10590/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16239/)

fujitatomoya (approved):

fujitatomoya:
@Mergifyio backport galactic

fujitatomoya:
@Mergifyio backport foxy

PR #883, opened 2022-01-18
Address deadlock issue in client.call

augustelalande:
Proposal to address the deadlock issue when calling `client.call` from within another callback.

I realise people may have different points of view on whether `client.call` should exist, and whether this proposal makes sense. But based on https://github.com/ros2/rclpy/pull/873#issuecomment-1008082009 ideas are welcome, so I thought I would at least propose it.

### DIFF
--- a/rclpy/rclpy/parameter_service.py
+++ b/rclpy/rclpy/parameter_service.py
@@ -14,11 +14,12 @@
 
 import weakref
 
+from rcl_interfaces.msg import ListParametersResult
 from rcl_interfaces.msg import SetParametersResult
 from rcl_interfaces.srv import DescribeParameters, GetParameters, GetParameterTypes
 from rcl_interfaces.srv import ListParameters, SetParameters, SetParametersAtomically
 from rclpy.exceptions import ParameterNotDeclaredException, ParameterUninitializedException
-from rclpy.parameter import Parameter, PARAMETER_SEPARATOR_STRING
+from rclpy.parameter import Parameter
 from rclpy.qos import qos_profile_parameters
 from rclpy.validate_topic_name import TOPIC_SEPARATOR_STRING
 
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/client.py
```diff
@@ -81,20 +81,21 @@ def call(self, request: SrvTypeRequest) -> SrvTypeResponse:
...
-        future.add_done_callback(unblock)
 
-        # Check future.done() before waiting on the event.
-        # The callback might have been added after the future is completed,
-        # resulting in the event never being set.
-        if not future.done():
-            event.wait()
+        while not future.done():
+            with self.handle:
+                header, response = self.__client.take_response(self.srv_type.Response)
```
fujitatomoya, line 88: just comments, i did not run this code for test.

- if response is not ready for the request, this `while loop` will consume cpu resource until the response is received?
- if another thread requests with the same `Client` before, it might take out the unrelated response and return?

i may be missing something though.

augustelalande: > if response is not ready for the request, this `while loop` will consume cpu resource until the response is received?

Yes. This part I could use some help on. How does executor avoid polling clients which do not have responses ready?

> if another thread requests with the same `Client` before, it might take out the unrelated response and return?

I don't think this should be a problem. The `Client` processes responses the same way that the executor does so even if it processes an unrelated response it should be fine.

sloretz: It looks like this tries to get the client response while the executor is blocked. It will still deadlock if the service server is in the same executor, or if the service servier relies on information from something in the same executor before responding.

sloretz (changes requested):
This is in a way spinning a small part of the executor (just the client) within an executor. That's an interesting idea for some limited cases, but the cost of it being a busy loop seems too high.

PR #885, opened 2022-01-19
Update documentation version number

augustelalande:
The docs version number has not been updated since it was created.

Maybe it would be a good idea to read the version directly from package.xml?


clalancette (approved):
Looks good, I'll go ahead and merge this just based on the Rpr job.  Thanks.

> Maybe it would be a good idea to read the version directly from package.xml?

Possibly, yeah.  Though I'm not entirely sure how this is going to interact with rosdoc2 (which we are achingly close to landing Python support for).  So I would hold off on that for now.

PR #887, opened 2022-01-24
Check in declare_parameters if deduced type of given value fits to op…

### LINKED ISSUE #879: declare_parameters allows conflicting type definition

NikolasE:
…tionally declared type in ParameterDescriptor

Signed-off-by: Nikolas Engelhard <nikolas.engelhard@motius.de>

PR for master from https://github.com/ros2/rclpy/pull/880

Closes #879

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -460,7 +460,16 @@ class Node:
                 if not descriptor.dynamic_typing and value is not None:
                     # infer type from default value
                     if not isinstance(value, ParameterValue):
-                        descriptor.type = Parameter.Type.from_parameter_value(value).value
+                        deduced_type = Parameter.Type.from_parameter_value(value).value
+                        if descriptor.type != ParameterType.PARAMETER_NOT_SET:
+                            if deduced_type != descriptor.type:
+                                reason = f"Passed value is of type {deduced_type}, \
+                                            but ParameterDescriptor declared type {descriptor.type}"
+                                raise InvalidParameterValueException(
+                                            parameter=Parameter(name, value=value),
+                                            value=value,
+                                            reason=reason)
+                        descriptor.type = deduced_type
                     else:
                         if value.type == ParameterType.PARAMETER_NOT_SET:
                             raise ValueError(
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -55,6 +55,8 @@ from rclpy.exceptions import ParameterAlreadyDeclaredException
 from rclpy.exceptions import ParameterImmutableException
 from rclpy.exceptions import ParameterNotDeclaredException
 from rclpy.exceptions import ParameterUninitializedException
+from rclpy.exceptions import ParameterException
+
 from rclpy.executors import Executor
 from rclpy.expand_topic_name import expand_topic_name
 from rclpy.guard_condition import GuardCondition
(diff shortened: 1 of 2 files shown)


fujitatomoya (changes requested):
overall looks good to me. could you address failure https://build.ros2.org/job/Rpr__rclpy__ubuntu_focal_amd64/693/testReport/?

fujitatomoya:
@ivanpauno i think this makes sense to me, could you take a look to share your thought?

rclpy/rclpy/node.py
```diff
@@ -460,7 +460,16 @@ def declare_parameters(
                 if not descriptor.dynamic_typing and value is not None:
                     # infer type from default value
                     if not isinstance(value, ParameterValue):
-                        descriptor.type = Parameter.Type.from_parameter_value(value).value
+                        deduced_type = Parameter.Type.from_parameter_value(value).value
+                        if descriptor.type != ParameterType.PARAMETER_NOT_SET:
```
ivanpauno, line 464: `descriptor.type` and `descriptor.name` are ignored by all `declare_parameter()` methods.
The stored descriptor is updated with the actual name and type.

I don't think that approach makes much sense, but that's how it has always worked.

If a parameter of an specified type needs to be declared without a default value, you can use:

```python3
node.declare_parameter('name', Parameter.Type.MY_PARAM_TYPE)
```

ivanpauno (changes requested):

rclpy/rclpy/node.py
NikolasE: So do you propose to also add a check for the name or not checking for inconsistencies at all? 
I think the test case I fixed shows nicely that we should use all information given by the user to prevent surprises.

fujitatomoya: IMO, either of them looks good to me.

- throw exception since user input is inconsistent. (this PR suggestion)
- actual name and type will be honored against descriptor. (current behavior, now i am inclined to this one, see https://github.com/ros2/rclpy/issues/912#issuecomment-1069681520) but i would suggest that it would be better to add doc for this behavior to avoid the surprise.

i will keep this thread open to hear from others.

> The stored descriptor is updated with the actual name and type.

this behavior is same with `rclcpp`. (so if we change this, `rclcpp` should be changed accordingly.)

rclpy/test/test_node.py
```diff
@@ -800,6 +800,13 @@ def test_use_sim_time(self):
         self.assertTrue(temp_node.get_parameter(USE_SIM_TIME_NAME).value)
         temp_node.destroy_node()
 
+    def test_node_declare_parameters_value_and_type_mismatch(self):
```
fujitatomoya, line 803: either way we take, i think it would be worth to keep this test scenario to make sure implementation.

rclpy/rclpy/node.py
NikolasE: I'll extend the PR so that we also check for a name-mismatch and throw an exception.

ivanpauno (changes requested):

rclpy/rclpy/node.py
ivanpauno: >  how about below?

That sounds fine to me

PR #853, opened 2021-11-24
Fix automatically declared parameters descriptor type

ivanpauno:
Fixes https://github.com/ros2/ros2/issues/1173.
Fixes https://github.com/ros2/rclpy/issues/829.

PR #896, opened 2022-02-16
Avoid exception in Node constructor when use override for 'use_sim_time'

shumov-ag:
Allow to create a Node with 'use_sim_time' parameter in parameter overrides.
Avoid exception `
rclpy.exceptions.ParameterAlreadyDeclaredException: ('Parameter(s) already declared', ['use_sim_time'])
`

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -207,11 +207,7 @@ class Node:
             self._parameter_overrides.update({p.name: p for p in parameter_overrides})
 
         # Clock that has support for ROS time.
-        # Note: parameter overrides and parameter event publisher need to be ready at this point
-        # to be able to declare 'use_sim_time' if it was not declared yet.
         self._clock = ROSClock()
-        self._time_source = TimeSource(node=self)
-        self._time_source.attach_clock(self._clock)
 
         if automatically_declare_parameters_from_overrides:
             self.declare_parameters(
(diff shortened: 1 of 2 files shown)


clalancette (commented):
Hm, this particular thing was discussed and dismissed when https://github.com/ros2/rclpy/pull/853 was put in.  @ivanpauno can you take a look and give your thoughts?

rclpy/rclpy/node.py
```diff
@@ -207,11 +207,7 @@ def __init__(
...
 
         # Clock that has support for ROS time.
-        # Note: parameter overrides and parameter event publisher need to be ready at this point
-        # to be able to declare 'use_sim_time' if it was not declared yet.
         self._clock = ROSClock()
-        self._time_source = TimeSource(node=self)
-        self._time_source.attach_clock(self._clock)
 
         if automatically_declare_parameters_from_overrides:
             self.declare_parameters(
```
fujitatomoya, line 213: I believe that `ROSClock()` is required to prior to `declare_parameters` for sure.

`ParameterEvetns` from this call will be based on system timestamp, even if it overrides `use_sim_time` parameter into true, because TimeSource is not attached yet.

fujitatomoya (commented):
this looks good to me, @ivanpauno could you follow up just in case i miss something.

ivanpauno (approved):

ivanpauno:
@shumov-ag could you share a way to reproduce the current issue?
Could you also add an unit test that fails before this change and starts passing after?

Thanks!

shumov-ag:
Yes, @ivanpauno. I had already added the test "test_create_node_with_parameter_overrides"

ivanpauno (approved):

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16274/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10895/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16640/)

ivanpauno:
@Mergifyio backport galactic

PR #901, opened 2022-02-22
Bugfix to Node.destroy_rate() result

### LINKED ISSUE #900: rclpy.node.Node.destroy_rate() does not return bool
## Bug report

- Version or commit hash:
  - a5236920be273cdf91d8104485a3657b7b08ac55
  - rclpy

#### Steps to reproduce issue

```python
import rclpy

rclpy.init()
node = rclpy.create_node("node")
rate = node.create_rate(1)
print(node.destroy_rate(rate))
```

#### Expected behavior

Prints `True`.

#### Actual behavior

```python
>>> print(node.destroy_rate(rate))
None
```

suurjaak:
Fixes #900.


clalancette (commented):
So the main reason that calls like `destroy_guard_condition` return a boolean is because they are used to figure out whether we should continue processing a list during destroy_node: https://github.com/ros2/rclpy/blob/8c4892b3ccbbaec23f88cc4c8d10e5fc1f3ade91/rclpy/rclpy/node.py#L1655-L1688 .  Because we don't do that with `destroy_rate`, I don't really think it needs to return a boolean.  In that case, I would just change the documentation to have no `return` value (that is, just remove line 1650).

But pinging @sloretz for thoughts.

suurjaak:
The same rationale would then apply to `destroy_timer()` as well, because `destroy_rate()` depends on it. 

However, `destroy_timer()` quite explicitly handles failure and returns `False`.

suurjaak:
Also, these are not just internal methods used by `destroy_node()`. These are public methods that framework users invoke. If destroying a timer can fail, so can destroying a rate, and it would be wrong to swallow the failure.

clalancette:
> Also, these are not just internal methods used by `destroy_node()`. These are public methods that framework users invoke. If destroying a timer can fail, so can destroying a rate, and it would be wrong to swallow the failure.

I guess I generally don't like return values on cleanup operations.  If it fails, it is not at all clear what the user can or should do about it.  I would actually be in favor of not returning errors for any of these `destroy_` operations, except for the obvious fact that they are used by `destroy_node`.

I guess from that perspective, it might be better to be consistent and return a boolean from `destroy_rate`, even though I don't like the final API.  But I guess I'll wait to hear from Shane to get his perspective.

rclpy/test/test_rate.py
```diff
@@ -101,7 +101,8 @@ def test_rate_invalid_period(self):
 
     def test_destroy(self):
         rate = self.node.create_rate(FREQ)
-        self.node.destroy_rate(rate)
+        success = self.node.destroy_rate(rate)
```
fujitatomoya, line 104: what about line 112? we should probably check on that as well?

fujitatomoya (changes requested):
> If it fails, it is not at all clear what the user can or should do about it. 

agree.

> it might be better to be consistent and return a boolean from destroy_rate

i do think so too, this is more consistent with current implementation.

after all, i am okay to go with this fix.

rclpy/test/test_rate.py
suurjaak: Indeed. Made it so.

fujitatomoya (approved):

rclpy/rclpy/node.py
```diff
@@ -1649,8 +1649,10 @@ def destroy_rate(self, rate: Rate):
 
         :return: ``True`` if successful, ``False`` otherwise.
         """
-        self.destroy_timer(rate._timer)
-        rate.destroy()
+        success = self.destroy_timer(rate._timer)
+        if success:
+            rate.destroy()
+        return success
```
clalancette, line 1655: If we are going to do this, we also need to change the type annotation on line 1646 to have `-> bool`.

rclpy/rclpy/node.py
```diff
@@ -1649,8 +1649,10 @@ def destroy_rate(self, rate: Rate):
 
         :return: ``True`` if successful, ``False`` otherwise.
         """
-        self.destroy_timer(rate._timer)
-        rate.destroy()
+        success = self.destroy_timer(rate._timer)
+        if success:
```
sloretz, line 1653: I think `rate.destroy()` still needs to be called regardless of the result of destroying the timer. That will wake any threads sleeping on the `Rate`. Timers are removed from the node's `self._timers` list even when destroying them fails, so there isn't a second opportunity to destroy the rate instance and wake those threads.

rclpy/test/test_rate.py
```diff
@@ -101,14 +101,16 @@ def test_rate_invalid_period(self):
...
+        assert success is True, "destroy_rate() returned %r instead of True" % success
```
sloretz, line 105: These asserts are a bit too verbose in my opinion. `assert self.node.destroy_rate(rate)` is fine, especially with how verbose the output is from pytest when a test assertion fails.

sloretz (changes requested):
I'm on board with adding a `bool` indicating if destruction was successful. I don't know what a user would do with that info, but maybe it's helpful to know.

rclpy/rclpy/node.py
suurjaak: Fixed.

rclpy/rclpy/node.py
suurjaak: Fixed.

rclpy/test/test_rate.py
suurjaak: Made it so.

sloretz (approved):
LGTM with green CI

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16229/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10854/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16584/)

clalancette (approved):

PR #904, opened 2022-02-24
Make sure to take out contexts on Action{Client,Server}.

clalancette:
This ensures that the underlying object won't be freed when
e.g. a WaitSet is still using it.

This should solve https://github.com/ros2/geometry2/issues/503 .  Still in draft because @sloretz and I are still working out some of the details.

### DIFF
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -351,6 +351,19 @@ class ActionClient(Waitable):
     def add_to_wait_set(self, wait_set):
         """Add entities to wait set."""
         self._client_handle.add_to_waitset(wait_set)
+
+    def __enter__(self):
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16242/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10865/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16596/)

jacobperron (approved):
LGTM pending CI

clalancette:
CI was green (except for aarch64, which is a known problem with Eigen).  Going ahead and merging this one, we can follow up with a better solution later on if need be.

PR #905, opened 2022-02-26
Properly implement action server/client handle cleanup.

clalancette:
In particular, it should never really be the case that the
underlying handle that we are using is None.  That only seemed
to be happening because we were double destroying; once in
the explicit call to destroy(), and once in __del__.  But
we don't actually need __del__; during garbage collection, we'll
drop the reference to the handle and then the underlying object
will get freed anyway.  Just remove all of that extraneous
infrastructure here.

Besides being a lot cleaner, this is also likely responsible for warnings during tests on the buildfarm like:

```
04:24:41.497 144: [test_action_client-3] Exception ignored in: <function ActionClient.__del__ at 0x00000245450633A0>
04:24:41.497 144: [test_action_client-3] Traceback (most recent call last):
04:24:41.497 144: [test_action_client-3]   File "C:\ci\ws\install\Lib\site-packages\rclpy\action\client.py", line 588, in __del__
04:24:41.498 144: [test_action_client-3]   File "C:\ci\ws\install\Lib\site-packages\rclpy\action\client.py", line 581, in destroy
04:24:41.498 144: [test_action_client-3] rclpy._rclpy_pybind11.InvalidHandle: cannot use Destroyable because destruction was requested
```

(we'll carefully check the results of CI to make sure that this is gone with this fix)

### DIFF
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -583,6 +583,5 @@ class ActionClient(Waitable):
 
     def destroy(self):
         """Destroy the underlying action client handle."""
-        with self._node.handle:
-            self._client_handle.destroy_when_not_in_use()
+        self._client_handle.destroy_when_not_in_use()
         self._node.remove_waitable(self)
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -353,15 +353,9 @@ class ActionClient(Waitable):
         self._client_handle.add_to_waitset(wait_set)
 
     def __enter__(self):
-        if self._client_handle is None:
-            return None
-
         return self._client_handle.__enter__()
 
     def __exit__(self, t, v, tb):
-        if self._client_handle is None:
-            return
-
         self._client_handle.__exit__(t, v, tb)
 
     # End Waitable API
@@ -589,13 +583,6 @@ class ActionClient(Waitable):
 
     def destroy(self):
         """Destroy the underlying action client handle."""
-        if self._client_handle is None:
-            return
         with self._node.handle:
             self._client_handle.destroy_when_not_in_use()
-            self._node.remove_waitable(self)
-        self._client_handle = None
-
-    def __del__(self):
-        """Destroy the underlying action client handle."""
-        self.destroy()
+        self._node.remove_waitable(self)
--- a/rclpy/rclpy/action/server.py
+++ b/rclpy/rclpy/action/server.py
@@ -495,15 +495,9 @@ class ActionServer(Waitable):
             self._handle.add_to_waitset(wait_set)
 
     def __enter__(self):
-        if self._handle is None:
-            return None
-
         return self._handle.__enter__()
 
     def __exit__(self, t, v, tb):
-        if self._handle is None:
-            return
-
         self._handle.__exit__(t, v, tb)
 
     # End Waitable API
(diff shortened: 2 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16250/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10871/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16608/)

rclpy/rclpy/action/server.py
```diff
@@ -602,16 +596,8 @@ def register_execute_callback(self, execute_callback):
 
     def destroy(self):
         """Destroy the underlying action server handle."""
-        if self._handle is None:
-            return
-
         for goal_handle in self._goal_handles.values():
             goal_handle.destroy()
 
         self._handle.destroy_when_not_in_use()
```
fujitatomoya, line 602:
```suggestion
        with self._node.handle:
            self._handle.destroy_when_not_in_use()
```

is needed to call the following?

https://github.com/ros2/rclpy/blob/53771be14430b360a7efb62bdd64085b25bb8707/rclpy/src/rclpy/action_server.cpp#L32

clalancette: Hm, good question.  When I was discussing this with @sloretz , he suggested that it wasn't necessary here, but I'm honestly not sure.  @sloretz can you comment?

sloretz: The action server calling `node_.destroy()` is surprising to me. I'll have a closer look.

sloretz: There's no need to call `with self._node.handle` here, even if ``destroy_when_not_in_use()` ends up calling `ActionServer.destroy()` right away.

It's a little weird because in rclpy there's a desire both to have the option to destroy rcl objects ASAP and the requirement to gracefully garbage collect them when no one explicitly destroys them. What's happening is `ActionServer` has a copy of a `Node` object and a shared pointer to the `rcl_action_server_t` type. All the classes have this pattern: Have a copy of the type they need kept alive (`ActionServer` needs `Node` to be alive for it to work), and a shared pointer to the rcl type. The shared pointer has a destructor that free's the `rcl` type.

Destroying objects right away is done by releasing shared pointers to the rcl types from the bottom up: `ActionServer` releases the reference to the `rcl_action_server_t` type before asking `Node` to release the reference to it's type. Garbage collection works by keeping those shared pointers alive from the bottom up (`ActionServer` has a `Node` so that `ActionServer` will be garbage collected before `Node`).  The context manager for each says "I'm using the rcl functions of this thing right now, so don't destroy it please". `node_.destroy()` isn't using the `rcl_node_t` instance, it's getting rid of the reference to it so it never uses it again. That's why `with self._node.handle:` doesn't need to be called.

rclpy/rclpy/action/client.py
```diff
@@ -589,13 +583,6 @@ def wait_for_server(self, timeout_sec=None):
 
     def destroy(self):
         """Destroy the underlying action client handle."""
-        if self._client_handle is None:
-            return
         with self._node.handle:
```
sloretz, line 586: No need for `with self._node.handle` here either

clalancette: Fixed in 7882366

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16263/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10884/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16623/)

sloretz (approved):
LGTM with green CI

rclpy/rclpy/action/server.py
fujitatomoya: @sloretz thanks for the note!

fujitatomoya (approved):

clalancette:
Warnings on aarch64 are known, so going ahead and merging this one.

PR #906, opened 2022-02-28
Make rclpy dependencies explicit

clalancette:
While working on some other things in rclpy, I noticed that rclpy didn't properly declare all of its dependencies.  In particular, it depends on `rosidl_runtime_c` and `rmw`, but neither of these were explicitly called out in the package.xml or CMakeLists.txt.

To fix this, there are 3 patches in this series:

1. Does a pass for include-what-you-use everywhere in the rclpy code.
2. Adds the new dependencies.
3. Fixes up documentation problems noticed while doing the above.

All 3 patches are completely independent, and could all be taken separately.  But given that I noticed them all together, I figured I would open one big PR for it.  Let me know if you'd like me to split it up.

### DIFF
--- a/rclpy/src/rclpy/action_client.hpp
+++ b/rclpy/src/rclpy/action_client.hpp
@@ -17,7 +17,8 @@
 
 #include <pybind11/pybind11.h>
 
-#include <rcl_action/rcl_action.h>
+#include <rcl_action/action_client.h>
+#include <rmw/types.h>
 
 #include <memory>
 
--- a/rclpy/src/rclpy/action_client.hpp
+++ b/rclpy/src/rclpy/action_client.hpp
@@ -45,12 +45,12 @@ public:
    *
    * \param[in] node Node to add the action client to.
    * \param[in] pyaction_type Action module associated with the action client.
-   * \param[in] pyaction_name Python object containing the action name.
+   * \param[in] action_name The action name.
    * \param[in] goal_service_qos rmw_qos_profile_t object for the goal service.
    * \param[in] result_service_qos rmw_qos_profile_t object for the result service.
    * \param[in] cancel_service_qos rmw_qos_profile_t object for the cancel service.
-   * \param[in] feedback_qos rmw_qos_profile_t object for the feedback subscriber.
-   * \param[in] status_qos rmw_qos_profile_t object for the status subscriber.
+   * \param[in] feedback_topic_qos rmw_qos_profile_t object for the feedback subscriber.
+   * \param[in] status_topic_qos rmw_qos_profile_t object for the status subscriber.
    */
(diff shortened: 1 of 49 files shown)


rclpy/src/rclpy/action_server.cpp
```diff
@@ -11,16 +11,27 @@
...
+
 #include <memory>
+#include <stdexcept>
 #include <string>
 
-#include "pybind11/pybind11.h"
-
-#include "rcl/error_handling.h"
-#include "rcpputils/scope_exit.hpp"
+#include <rcpputils/scope_exit.hpp>
```
adityapande-1995, line 29: What is the general preference / policy for include tags with respect to quotes marks and angle brackets ?

rclpy/src/rclpy/action_client.hpp
```diff
@@ -205,15 +205,15 @@ class ActionClient : public Destroyable, public std::enable_shared_from_this<Act
    *        is_result_response_ready)
    */
   py::tuple
-  is_ready(WaitSet & wait_set);
+  is_ready(rclpy::WaitSet & wait_set);
```
adityapande-1995, line 208: The function implementations in ``action_client.cpp`` still have the signatures without ``rclpy::`` namespace. Maybe we can change those to match the ones in headers ?

adityapande-1995 (commented):
Can we split these into say 2 PRs ? Might make it easier to review individually.

rclpy/src/rclpy/action_client.hpp
adityapande-1995: https://github.com/ros2/rclpy/blob/b9ed8c96e002c1cf8da839cffdb0b63b6a28b375/rclpy/src/rclpy/action_client.cpp#L244

rclpy/src/rclpy/action_server.cpp
sloretz: I'm not sure we have a style rule for it, but I think a good rule is if it's in a different package use `<>`, if it's in the same package use `""`.

sloretz:
> Changes to fully-qualified object names within the pybind11 code, just to make it clear where things are coming from.

It looks like these changes are inside code in a `namespace rclpy {...}` block. I feel like that's enough context to have a pretty good intuition for where they come from.

clalancette:
> It looks like these changes are inside code in a `namespace rclpy {...}` block. I feel like that's enough context to have a pretty good intuition for where they come from.

That's fair.  I waffled on this anyway.  I'll remove all of that change from this PR (which should simplify it quite a bit as well).

clalancette:
CI is all green, and approved, so I'm going to go ahead and merge this in as a fix.

Issue #850, opened 2021-11-16
Add callback groups to executor


jacobperron:
#### Feature description

Support adding specific callback groups to an executor, instead of entire node objects.
This enables finer control over what is added to an executor so we can have multiple executors operating on the same node. 

Equivalent of https://github.com/ros2/rclcpp/pull/1218, but in Python.

#### Implementation considerations

I'm presuming this is a feature that we'd like to have, so contradict me otherwise. 

Judging by the size of the rclcpp PR, I don't expect this to be an easy enhancement to make, but figured I'd open this ticket to start a discussion.

PR #909, opened 2022-03-04
Move /clock callback processing to seperate thread.

### LINKED ISSUE #792: Insulate /clock callback on specific thread
#### Feature description

rclpy follow-up of https://github.com/ros2/rclcpp/issues/1542.

the current implementation of simulation time which listen to the /clock topic, this behavior breaks the symmetry between the usage of real time (system time internally accessed without the need of subscribing to a topic, and hence always updated) and simulation time (time received on the /clock topic and hence frozen in a callback).
the symmetry should be maintain in order to not draw false conclusions about the behavior of a code developed in simulation VS real time, and thus independently from good/bad code practice.

N.A

Flova:
This PR moves the spinning of the Executor for the /clock subscriber to a seperate thread. (#792)
As described in the issue this needs support for spinning a callback group in its own Executor. Support for this exists in rclcpp and was sugested for rclpy in #850. The current implementation is a bit hacky imo., so suggestions regarding a different way of implementing this are welcome.

Fixes #792 and partially #850

### DIFF
--- a/rclpy/rclpy/callback_groups.py
+++ b/rclpy/rclpy/callback_groups.py
@@ -39,6 +39,14 @@ class CallbackGroup:
         """
         self.entities.add(weakref.ref(entity))
 
+    def get_entites(self):
+        """
+        Get all entities from the callback group.
+
+        :return: A set containing weakrefs to the entities.
+        """
+        return self.entities
+
     def has_entity(self, entity) -> bool:
         """
         Determine if an entity has been added to this group.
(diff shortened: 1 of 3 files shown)


Flova:
Any suggestions how we might want to go forward with this issues? @clalancette 

It is currently quite annoying that tf2 as well as simulation time are broken and lead to freezes/timeouts. One fix is the usage of the multithreaded executor, which was quite slow in our testing. It also only partially addresses the problem as e.g. a 4 threaded executor could still block if it processes 4 normal messages at once and no time or tf message is able to get trough.

PR #910, opened 2022-03-08
time_until_next_call returns max if timer is canceled.

fujitatomoya:
aligned with https://github.com/ros2/rcl/pull/963 and https://github.com/ros2/rclcpp/pull/1893

### DIFF
--- a/rclpy/src/rclpy/timer.cpp
+++ b/rclpy/src/rclpy/timer.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
 
 #include <rcl/context.h>
 #include <rcl/error_handling.h>
@@ -105,11 +106,13 @@ void Timer::change_timer_period(int64_t period_nsec)
   }
 }
 
-int64_t Timer::time_until_next_call()
+std::optional<int64_t> Timer::time_until_next_call()
 {
   int64_t remaining_time;
   rcl_ret_t ret = rcl_timer_get_time_until_next_call(rcl_timer_.get(), &remaining_time);
-  if (ret != RCL_RET_OK) {
+  if (ret == RCL_RET_TIMER_CANCELED) {
+    return std::nullopt;
+  } else if (ret != RCL_RET_OK) {
     throw RCLError("failed to get time until next timer call");
   }
 
--- a/rclpy/src/rclpy/timer.cpp
+++ b/rclpy/src/rclpy/timer.cpp
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 #include <pybind11/pybind11.h>
-#include <pybind11/stl.h>
 
 #include <rcl/context.h>
 #include <rcl/error_handling.h>
--- a/rclpy/test/test_timer.py
+++ b/rclpy/test/test_timer.py
@@ -18,6 +18,7 @@ import time
 
 import pytest
 import rclpy
+from rclpy.constants import S_TO_NS
 from rclpy.executors import SingleThreadedExecutor
 
 
@@ -49,7 +50,7 @@ def test_zero_callback(period):
             executor = SingleThreadedExecutor(context=context)
             try:
                 executor.add_node(node)
-                # The first spin_once() takes  long enough for 1ms timer tests to fail
+                # The first spin_once() takes long enough for 1ms timer tests to fail
                 executor.spin_once(timeout_sec=0)
 
                 callbacks = []
@@ -78,7 +79,7 @@ def test_number_callbacks(period):
             executor = SingleThreadedExecutor(context=context)
             try:
                 executor.add_node(node)
-                # The first spin_once() takes  long enough for 1ms timer tests to fail
+                # The first spin_once() takes long enough for 1ms timer tests to fail
                 executor.spin_once(timeout_sec=0)
 
                 callbacks = []
@@ -110,7 +111,7 @@ def test_cancel_reset(period):
             executor = SingleThreadedExecutor(context=context)
             try:
                 executor.add_node(node)
-                # The first spin_once() takes  long enough for 1ms timer tests to fail
+                # The first spin_once() takes long enough for 1ms timer tests to fail
                 executor.spin_once(timeout_sec=0)
 
                 callbacks = []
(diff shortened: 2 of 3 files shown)


fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16304/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10924/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16671/)

rclpy/test/test_timer.py
```diff
@@ -148,3 +149,35 @@ def test_cancel_reset(period):
...
+                try:
+                    assert not timer.is_canceled()
+                    executor.spin_once(0.1)
+                    assert timer.time_until_next_call() <= (1 * S_TO_NS)
+                    timer.reset()
+                    assert not timer.is_canceled()
+                    assert timer.time_until_next_call() <= (1 * S_TO_NS)
+                    timer.cancel()
+                    assert timer.is_canceled()
+                    assert timer.time_until_next_call() >= (1 * S_TO_NS)
```
sloretz, line 175: The implementation seems ok, I'm just not sure about returning max time here. It seems not so easy in python to know if `time_until_next_call()` is returning a valid but really long time or if the timer was canceled I would recommend either raising if the timer is canceled, or returning `None` such that this test becomes 

```python3
assert timer.time_until_next_call() is None
```

fujitatomoya: @sloretz that totally makes sense, thanks 👍

fujitatomoya: Just checking that you are suggesting like following with changing API?

```
bool time_until_next_call(int64_t nsec);
```

so that `time_until_next_call` python can return None if false?

sloretz: I was thinking 

```c++
std::optional<int64_t> time_until_next_call();
```

Which I think [pybind11 supports](https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#c-17-library-containers) and we can use since [Rolling and Humble target c++ 17](https://www.ros.org/reps/rep-2000.html#rolling-ridley-june-2020-ongoing)

ivanpauno: I think it would be nice to have similar behaviors both in rclcpp and rclpy.
The problem is that returning `std::optional<int64_t>` in that case will be a breaking change.
Though this method is probably not used a lot, so maybe breaking it is not a hugh problem.

sloretz: > I think it would be nice to have similar behaviors both in rclcpp and rclpy.

I think it's fine for the client libraries to differ to take advantage of language features. In C++ the caller has access to `std::chrono::nanoseconds::max()`, so the return value is not too difficult to compare. In rclpy the caller is using Python and doesn't easily have access to `std::numeric_limits<...>::max()`, but Python's `None` fits pretty well.

>  Though this method is probably not used a lot, so maybe breaking it is not a hugh problem.

Breaking changes are fine in ROS Rolling

ivanpauno: > I think it's fine for the client libraries to differ to take advantage of language features. In C++ the caller has access to std::chrono::nanoseconds::max(), so the return value is not too difficult to compare. In rclpy the caller is using Python and doesn't easily have access to std::numeric_limits<...>::max(), but Python's None fits pretty well.

The rationale sounds good to me

rclpy/src/rclpy/timer.cpp
```diff
@@ -13,13 +13,15 @@
...
 #include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
 
 #include <rcl/context.h>
 #include <rcl/error_handling.h>
 #include <rcl/rcl.h>
 #include <rcl/timer.h>
 #include <rcl/types.h>
 
+#include <limits>
```
fujitatomoya, line 24: should be deleted, cz it does not need it.

rclpy/test/test_timer.py
fujitatomoya: thanks for the informative comments. this has been addressed with `std::optional<int64_t>` in C++, and python returns None if the timer is canceled.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16309/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10929/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16679/)

fujitatomoya:
@sloretz @ivanpauno friendly ping.

rclpy/test/test_timer.py
```diff
@@ -148,3 +149,35 @@ def test_cancel_reset(period):
...
+                    assert timer.is_canceled()
+                    assert timer.time_until_next_call() is None
+                finally:
+                    node.destroy_timer(timer)
+            finally:
+                executor.shutdown()
+        finally:
+            node.destroy_node()
+    finally:
+        rclpy.shutdown(context=context)
```
ivanpauno, line 183: Maybe you can avoid some of the nesting (?).
There might be a typo in my suggestion.

```suggestion
    node = None
    executor = None
    timer = None
    try:
        node = rclpy.create_node('test_time_until_next_call', context=context)
        executor = SingleThreadedExecutor(context=context)
        executor.add_node(node)
        executor.spin_once(timeout_sec=0)
        timer = node.create_timer(1, lambda: None)
        assert not timer.is_canceled()
        executor.spin_once(0.1)
        assert timer.time_until_next_call() <= (1 * S_TO_NS)
        timer.reset()
        assert not timer.is_canceled()
        assert timer.time_until_next_call() <= (1 * S_TO_NS)
        timer.cancel()
        assert timer.is_canceled()
        assert timer.time_until_next_call() is None
    finally:
        if timer is not None:
            node.destroy_timer(timer)
        if executor is not None:
            executor.shutdown()
        if node is not None:
                node.destroy_node()
        rclpy.shutdown(context=context)
```

ivanpauno (approved):
LGTM!

rclpy/test/test_timer.py
fujitatomoya: yeah, i am go ahead to borrow you code.

fujitatomoya:
i do not see failure https://ci.ros2.org/job/ci_linux/16309/console on my local environment, but i think header is missing in hpp file.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16310/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10930/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16680/)

fujitatomoya:
windows failure is unrelated to this fix.

PR #919, opened 2022-03-30
Update spin_until_complete

hliberacki:
Due to changes in rclcpp replace spin_until_future_complete with spin_until_complete, add spin.

* Deprecate spin_until_future_complete
* Add spin_until_complete
* Add spin_for method
* Udpdate unit tests

Signed-off-by: Hubert Liberacki <hliberacki@gmail.com>


hliberacki:
Due to changes in `rclcpp` I've pushed a PR updating pythong side as well. 
PR in rclcpp - https://github.com/ros2/rclcpp/pull/1874

rclpy/rclpy/executors.py
```diff
@@ -361,7 +401,8 @@ async def _execute_client(self, client, seq_and_response):
 
     def _take_service(self, srv):
         with srv.handle:
-            request_and_header = srv.handle.service_take_request(srv.srv_type.Request)
+            request_and_header = srv.handle.service_take_request(
+                srv.srv_type.Request)
```
fujitatomoya, line 405: is this required to change? (so are some other places in this file.)

rclpy/test/test_executor.py
```diff
@@ -29,7 +29,8 @@ class TestExecutor(unittest.TestCase):
     def setUp(self):
         self.context = rclpy.context.Context()
         rclpy.init(context=self.context)
-        self.node = rclpy.create_node('TestExecutor', namespace='/rclpy', context=self.context)
+        self.node = rclpy.create_node(
+            'TestExecutor', namespace='/rclpy', context=self.context)
```
fujitatomoya, line 33: we would want to avoid unnecessary change if it is not required to pass the tests?

rclpy/test/test_executor.py
```diff
@@ -348,7 +350,23 @@ def test_executor_add_node(self):
...
+        executor.add_node(self.node)
+
+        def timer_callback():
+            pass
+        timer = self.node.create_timer(0.003, timer_callback)
+
+        start = time.monotonic()
+        executor.spin_for(duration_sec=0.1)
+        end = time.monotonic()
+        self.assertGreaterEqual(end - start, 0.1)
```
fujitatomoya, line 365: i think this could be too small duration for some platforms might not work as we expect. how about having the following?

```suggestion
        timer = self.node.create_timer(0.1, timer_callback)

        start = time.monotonic()
        executor.spin_for(duration_sec=1.0)
        end = time.monotonic()
        self.assertGreaterEqual(end - start, 1.0)
```

rclpy/test/test_executor.py
```diff
@@ -361,16 +379,68 @@ def timer_callback():
...
+        timer.cancel()
+
+    def test_executor_spin_until_complete_condition_done(self):
+        self.assertIsNotNone(self.node.handle)
+        executor = SingleThreadedExecutor(context=self.context)
+        executor.add_node(self.node)
+
+        def timer_callback():
+            pass
+        timer = self.node.create_timer(0.003, timer_callback)
```
fujitatomoya, line 406: probably same here, how about having timer with `0.1`?

hliberacki: Done - plus I've updated other places where a similar pattern of short intervals was used.

hliberacki:
https://github.com/ros2/rclcpp/pull/1874#issuecomment-1086906375 Passing CI with all related PRs linked and build together.

rclpy/rclpy/executors.py
```diff
@@ -716,9 +755,14 @@ def spin_once(self, timeout_sec: float = None) -> None:
...
                 raise handler.exception()
 
-    def spin_once_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
+    def spin_once_until_complete(self, condition, timeout_sec: float = None) -> None:
         self.spin_once(timeout_sec)
 
+    """Deprecated in favor of spin_once_until_complete"""
+
+    def spin_once_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
+        self.spin_once_until_complete(timeout_sec)
```
wjwwood, line 764: This docstring looks misplaced, does it not need to be inside the method?

rclpy/rclpy/executors.py
```diff
@@ -761,5 +805,11 @@ def _spin_once_impl(
...
 
+    def spin_once_until_complete(self, condition, timeout_sec: float = None) -> None:
+        self._spin_once_impl(timeout_sec, condition if callable(
+            condition) else condition.done)
+
+    """Deprecated in favor of spin_once_until_complete"""
+
     def spin_once_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
-        self._spin_once_impl(timeout_sec, future.done)
+        self.spin_once_until_complete(timeout_sec, future.done)
```
wjwwood, line 815: Same here, the docstring is in a weird place.

wjwwood (changes requested):
This has some necessary changes. I think it can be decoupled from the https://github.com/ros2/rclcpp/pull/1874 pr.

rclpy/rclpy/__init__.py
```diff
@@ -241,10 +287,7 @@ def spin_until_future_complete(
     :param executor: The executor to use, or the global executor if ``None``.
     :param timeout_sec: Seconds to wait. Block until the future is complete
         if ``None`` or negative. Don't wait if 0.
+
+    Deprecated in favor of spin_until_complete
```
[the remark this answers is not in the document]
hliberacki: @wjwwood  Sorry for the late reply, I wasn't available last week. 

From what I can see there is no Deprecated decorator defined for functions/methods in `rclpy`. Shall I add this decorator to the scope of this commit? If so, could it be added inside of [rclpy/rclpy/utilities.py](https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/utilities.py)? 

Another option which I see would be to simply use `warn` as it's done in [rclpy/rclpy/node.py:446](https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/node.py#L446) - yet I am not sure whether it's correct in this case. 

While creating the initial commit I was trying to find a correct pattern, but all I could find was -> [rclpy/rclpy/impl/rcutils_logger.py:339](https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/impl/rcutils_logger.py#L339), and I've followed this approach back then.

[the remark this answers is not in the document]
hliberacki: @wjwwood I've simply used warnings - let me know whether it's ok - if not I'll change it.

ihasdapie:
Now that https://github.com/ros2/rclpy/pull/959 is merged this PR should update `rclpy/test/test_parameter_client.py` to use the new `spin_until_complete` methods instead of `spin_until_future_complete`

hliberacki:
> Now that #959 is merged this PR should update `rclpy/test/test_parameter_client.py` to use the new `spin_until_complete` methods instead of `spin_until_future_complete`

I was out of town, I will fix it shortly.

@wjwwood did you have a chance to take a look into new changes addressing your comments?

rclpy/rclpy/__init__.py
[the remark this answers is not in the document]
fujitatomoya: I believe adding warning should be good enough

christophebedard:
I've opened #1268 to replace this PR.

PR #922, opened 2022-04-08
Allow to create a subscription with a callback that also receives the message info

ivanpauno:
The cpp `Subscription.take_message()` was also updated to include `reception_sequence_number`/`publication_sequence_number` in the message info.

This should be merged AFTER the API freeze ends.

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -335,12 +335,16 @@ class Executor:
         with sub.handle:
             msg_info = sub.handle.take_message(sub.msg_type, sub.raw)
             if msg_info is not None:
-                return msg_info[0]
+                if sub.callback_type is Subscription.CallbackType.MessageOnly:
+                    return (msg_info[0], )
+                else:
+                    print(msg_info)
+                    return msg_info
         return None
 
-    async def _execute_subscription(self, sub, msg):
-        if msg:
-            await await_or_execute(sub.callback, msg)
+    async def _execute_subscription(self, sub, *args):
+        if args:
+            await await_or_execute(sub.callback, *args)
 
     def _take_client(self, client):
         with client.handle:
@@ -415,7 +419,10 @@ class Executor:
                 gc.trigger()
 
                 try:
-                    await call_coroutine(entity, arg)
+                    if isinstance(arg, tuple):
+                        await call_coroutine(entity, *arg)
+                    else:
+                        await call_coroutine(entity, arg)
                 finally:
                     entity.callback_group.ending_execution(entity)
                     # Signal that work has been done so the next callback in a mutually exclusive
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -421,10 +421,6 @@ class Executor:
 
                 try:
                     await call_coroutine(entity, *arg)
-                    # if isinstance(arg, tuple):
-                    #     await call_coroutine(entity, *arg)
-                    # else:
-                    #     await call_coroutine(entity, arg)
                 finally:
                     entity.callback_group.ending_execution(entity)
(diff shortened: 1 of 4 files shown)


rclpy/rclpy/executors.py
```diff
@@ -415,7 +419,10 @@ async def handler(entity, gc, is_shutdown, work_tracker):
                 gc.trigger()
 
                 try:
-                    await call_coroutine(entity, arg)
+                    if isinstance(arg, tuple):
+                        await call_coroutine(entity, *arg)
+                    else:
+                        await call_coroutine(entity, arg)
```
ivanpauno, line 425: This is sort of a hack, maybe I should modify all `take_*` messages to always return a tuple of arguments, and here we could always use `await call_coroutine(entity, *arg)`

ivanpauno: Actually, this change broke services, actions and waitables.
I'll take a look on Monday

ivanpauno: Fixed in https://github.com/ros2/rclpy/pull/922/commits/73f9ace78a3f23d1fb5eaba50504d669772be126

rclpy/rclpy/executors.py
```diff
@@ -415,7 +420,11 @@ async def handler(entity, gc, is_shutdown, work_tracker):
                 gc.trigger()
 
                 try:
-                    await call_coroutine(entity, arg)
+                    await call_coroutine(entity, *arg)
+                    # if isinstance(arg, tuple):
+                    #     await call_coroutine(entity, *arg)
+                    # else:
+                    #     await call_coroutine(entity, arg)
```
jacobperron, line 427: Remove this vestigial comment

rclpy/rclpy/subscription.py
```diff
@@ -69,6 +75,21 @@ def __init__(
...
+            return
+        except TypeError:
+            pass
+        try:
+            inspect.signature(self.callback).bind(object(), object())
+            self.callback_type = Subscription.CallbackType.WithMessageInfo
+            return
+        except TypeError:
+            pass
+        raise RuntimeError(
```
jacobperron, line 90: I think it would be good to add a positive and negative test for this new logic.

ivanpauno: See https://github.com/ros2/rclpy/pull/922/commits/74afcb34a36aedae36720a569194f6ad0b515745

jacobperron (approved):
LGTM

rclpy/rclpy/subscription.py
```diff
@@ -69,6 +75,21 @@ def __init__(
 
         self.event_handlers: QoSEventHandler = event_callbacks.create_event_handlers(
             callback_group, subscription_impl, topic)
+        self.callback_type = Subscription.CallbackType.MessageOnly
```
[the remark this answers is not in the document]
jacobperron: That sounds like something that would be good to handle; good catch.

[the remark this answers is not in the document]
ivanpauno: Fixed in https://github.com/ros2/rclpy/pull/922/commits/414539f0e5c2e593ed5441d1bf8d82514e30a5aa.
I didn't think that modifying `Subscription.callback` after creating the subscription was valid, but based on the code it seems it was possible.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16575/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11159/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16963/)

wjwwood (approved):

fujitatomoya (approved):

ivanpauno:
Humble has already be branched off, so merging!

Issue #921, opened 2022-04-06
ActionClient feedback


mart-karelics:
## Bug report

**Required Info:**

- Operating System:
  - 
- Installation type:
  - 
- Version or commit hash:
  - 
- DDS implementation:
  - 
- Client library (if applicable):
  - 
rclpy: 
#### Steps to reproduce issue

```

```

PR #927, opened 2022-04-15
remove feedback callback when the goal has been completed.

fujitatomoya:
address https://github.com/ros2/rclpy/issues/921

### DIFF
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -209,7 +209,10 @@ class ActionClient(Waitable):
     def _remove_pending_goal_request(self, future):
         seq = self._remove_pending_request(future, self._pending_goal_requests)
         if seq in self._sequence_number_to_goal_id:
+            goal_uuid = bytes(self._sequence_number_to_goal_id[seq].uuid)
             del self._sequence_number_to_goal_id[seq]
+            if goal_uuid in self._feedback_callbacks:
+                del self._feedback_callbacks[goal_uuid]
 
     def _remove_pending_cancel_request(self, future):
         self._remove_pending_request(future, self._pending_cancel_requests)
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -347,7 +347,6 @@ class ActionClient(Waitable):
                                 GoalStatus.STATUS_CANCELED == status or
                                 GoalStatus.STATUS_ABORTED == status):
                             del self._goal_handles[goal_uuid]
-                            #
                     else:
                         # Weak reference is None
                         del self._goal_handles[goal_uuid]
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -209,10 +209,7 @@ class ActionClient(Waitable):
     def _remove_pending_goal_request(self, future):
         seq = self._remove_pending_request(future, self._pending_goal_requests)
         if seq in self._sequence_number_to_goal_id:
-            goal_uuid = bytes(self._sequence_number_to_goal_id[seq].uuid)
             del self._sequence_number_to_goal_id[seq]
-            if goal_uuid in self._feedback_callbacks:
-                del self._feedback_callbacks[goal_uuid]
 
     def _remove_pending_cancel_request(self, future):
         self._remove_pending_request(future, self._pending_cancel_requests)
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -168,11 +168,13 @@ class ActionClient(Waitable):
         # key: goal request sequence_number, value: Future for goal response
         self._pending_goal_requests = {}
         # key: goal request sequence_number, value: UUID
-        self._sequence_number_to_goal_id = {}
+        self._goal_sequence_number_to_goal_id = {}
         # key: cancel request sequence number, value: Future for cancel response
         self._pending_cancel_requests = {}
         # key: result request sequence number, value: Future for result response
         self._pending_result_requests = {}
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/action/client.py
```diff
@@ -209,7 +209,10 @@ def _remove_pending_request(self, future, pending_requests):
     def _remove_pending_goal_request(self, future):
         seq = self._remove_pending_request(future, self._pending_goal_requests)
         if seq in self._sequence_number_to_goal_id:
+            goal_uuid = bytes(self._sequence_number_to_goal_id[seq].uuid)
             del self._sequence_number_to_goal_id[seq]
+            if goal_uuid in self._feedback_callbacks:
+                del self._feedback_callbacks[goal_uuid]
```
fujitatomoya, line 215: this changes behavior a bit. previously feedback callback still can be issued after corresponding goal has been completed. but with this fix, that cannot be done. I am not sure if this behavior is intention or not.

fujitatomoya:
@sloretz @ivanpauno any thoughts?

rclpy/rclpy/action/client.py
ivanpauno: what's the behavior in rclcpp?

fujitatomoya: thanks for asking, `rclcpp` could be reference.

if user is NOT aware of result (either result callback option is not provided or async_get_result is not called), feedback callback can be issued as long as user application keeps reference to `std::shared_future<typename GoalHandle::SharedPtr>`.

https://github.com/ros2/rclcpp/blob/c24e485084f421292734ebf6b63f1a105a2c7602/rclcpp_action/include/rclcpp_action/client.hpp#L421-L422

https://github.com/ros2/rclcpp/blob/c24e485084f421292734ebf6b63f1a105a2c7602/rclcpp_action/include/rclcpp_action/client.hpp#L651-L662

if user is aware of result in anyways, then once result is received from action server to client, `goal_handles_[goal_id]` is erased. in this case, feedback callback will not be called anymore.

https://github.com/ros2/rclcpp/blob/c24e485084f421292734ebf6b63f1a105a2c7602/rclcpp_action/include/rclcpp_action/client.hpp#L727

current code is totally wrong, it should be removed aligned with `_pending_result_requests`.

fujitatomoya: I will fix the code soon.

ivanpauno: > if user is NOT aware of result (either result callback option is not provided or async_get_result is not called), feedback callback can be issued as long as user application keeps reference to std::shared_future<typename GoalHandle::SharedPtr>.

Maybe, we could keep a weak reference to `ClientGoalHandle`  in https://github.com/ros2/rclpy/blob/f1f40d6cd42bcf1988b2c72c8d0ce03bc62ec530/rclpy/rclpy/action/client.py#L167
to get something equivalent (?).
I'm not sure.

Though removing the feedback callback when the result is get is also not a bad idea.

@jacobperron @sloretz @adityapande-1995 any opinions?

jacobperron: Removing the feedback callback after the goal is completed seems reasonable to me. I don't think there's anything in the design document or other documentation that indicates we should expect feedback after a goal is completed.

jacobperron: However, are we interpreting this future correctly? I thought that receiving the response for the goal request means that the goal was accepted, not necessarily completed. We don't know that the goal is completed unless we check the status or the result is received.

jacobperron:
Maybe we should prune any callbacks in this case:

https://github.com/ros2/rclpy/blob/f1f40d6cd42bcf1988b2c72c8d0ce03bc62ec530/rclpy/rclpy/action/client.py#L340-L344

rclpy/rclpy/action/client.py
```diff
@@ -339,6 +347,7 @@ async def execute(self, taken_data):
                                 GoalStatus.STATUS_CANCELED == status or
                                 GoalStatus.STATUS_ABORTED == status):
                             del self._goal_handles[goal_uuid]
+                            #
```
ivanpauno, line 350:
```suggestion
```

ivanpauno (approved):
LGTM!

jacobperron:
I think the proposed changes are okay, but I guess the `_feedback_callbacks` list can still grow indefinitely if the user doesn't request the result.

rclpy/rclpy/action/client.py
fujitatomoya: 😅

fujitatomoya:
> I think the proposed changes are okay, but I guess the _feedback_callbacks list can still grow indefinitely if the user doesn't request the result.

true, but i tried to align this with rclcpp behavior. https://github.com/ros2/rclpy/pull/927#issuecomment-1108801919 makes sense to me but this is a bit different from rclcpp behavior commented on https://github.com/ros2/rclpy/pull/927#discussion_r854624912

jacobperron (approved):
LGTM

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16604/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11188/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16994/)

fujitatomoya:
windows meet many errors which are unrelated to action.

CI: (retry)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16998/)

fujitatomoya:
@Mergifyio backport humble galactic foxy

Issue #823, opened 2021-09-14
Addition of a ``wait_for_node()`` function


adityapande-1995:
#### Feature description
It would be nice to have a ``wait_for_node(node_name, timeout)`` function which would wait for a particular node to show up, and block for a specified time. Further, we can have a set of ``wait_for_* `` functions which could wait for parameters, services, topics to show up. 

#### Implementation considerations
This feature was mentioned in https://github.com/ros2/launch_ros/pull/263. 
Related feature exists in ``rcl`` (https://github.com/ros2/rcl/pull/907) but it hasn't been ported to ``rclcpp`` (https://github.com/ros2/rclcpp/pull/1621) due to a design issue.

aprotyas:
+1 for a function like this, but if I understand correctly, #907 looks for publisher/subscriber counts on a named topic, right? Which is close enough in spirit, but I guess that isn't the API this issue would want to have wrapped in `rclpy`.

adityapande-1995:
Right. The implementation I have in https://github.com/ros2/launch_ros/pull/263 is pretty basic, and definitely not efficient. Not sure where to implement this feature though, maybe in ``rcl`` and then wrap it in ``rclpy`` and ``rclcpp``

aprotyas:
> maybe in `rcl` and then wrap it in `rclpy` and `rclcpp`

I think that's the way to go because I see utility for all of the `wait_for_*` methods in the client libraries.

adityapande-1995:
A basic implementation is added here : https://github.com/ros2/examples/blob/200a16c2290e09a26eaa91f201555887cbb3e379/launch_testing/launch_testing_examples/launch_testing_examples/check_multiple_nodes_launch_test.py#L84

clalancette:
Good call.  I'll go ahead and close this one out, thanks!

PR #930, opened 2022-04-27
Adds wait_for_node

gonzodepedro:
Related #823 

Signed-off-by: Gonzalo de Pedro <gonzalo@depedro.com.ar>

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1979,8 +1979,8 @@ class Node:
                          won't timeout.
         :return: True if the node was found, False if timeout.
         """
-        if not node_name.startswith('/'):
-            node_name = f'/{node_name}'
+        if not fully_qualified_node_name.startswith('/'):
+            fully_qualified_node_name = f'/{fully_qualified_node_name}'
 
         start = time.time()
         flag = False
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1954,17 +1954,20 @@ class Node:
     def wait_for_node(
         self,
         node_name: str,
-        timeout: float = 8.0
+        timeout: float
     ) -> bool:
         """
         Wait until node name is present in the system or timeout.
 
         :param node_name: name of the node to wait for.
-        :timeout: seconds to wait for the node to be present.
-        :return: true if the node was found, false if timeout.
+        :param timeout: seconds to wait for the node to be present. If negative, the function
+                         won't timeout.
+        :return: True if the node was found, False if timeout.
         """
         start = time.time()
         flag = False
+        # TODO refactor this implementation when we can react to guard condition events, or replace
+        # it entirely with an implementation in rcl. see https://github.com/ros2/rclpy/issues/929
         while time.time() - start < timeout and not flag:
             flag = node_name in self.get_node_names()
             time.sleep(0.1)
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1808,6 +1808,19 @@ class Node:
             names_ns = self.handle.get_node_names_and_namespaces()
         return [n[0] for n in names_ns]
 
+    def get_fully_qualified_node_names(self) -> List[str]:
+        """
+        Get a list of fully qualified names for discovered nodes.
+
+        Similar to ``get_node_names_namespaces()``, but concatenates the names and namespaces.
+        :return: List of fully qualified node names.
+        """
+        names_and_namespaces = self.get_node_names_and_namespaces()
+        return [
+            ns + ('' if ns.endswith('/') else '/') + name
+            for name, ns in names_and_namespaces
+        ]
+
     def get_node_names_and_namespaces(self) -> List[Tuple[str, str]]:
         """
         Get a list of names and namespaces for discovered nodes.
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/node.py
```diff
@@ -1949,3 +1950,22 @@ def get_subscriptions_info_by_topic(
             topic_name,
             no_mangle,
             _rclpy.rclpy_get_subscriptions_info_by_topic)
+
+    def wait_for_node(
+        self,
```
jacobperron, line 1955: Another approach we could take is to put the implementation into a free function, living in it's own module (e.g. `graph.py`), that takes a node as an argument. This makes it easier to do dependency injection on node-like objects (nice for testing). We can can still have node method too, that just calls the free function.

I think this is a very minor comment that can be ignored; it's easy enough to break out the implementation later without breaking API.

jacobperron (commented):
LGTM, I've left a bunch of minor comments below.

rclpy/rclpy/node.py
```diff
@@ -1949,3 +1950,22 @@ def get_subscriptions_info_by_topic(
...
+        node_name: str,
```
fujitatomoya, line 1956: how about adding namespace as well which could be empty if user application does not use.

fujitatomoya (approved):

rclpy/rclpy/node.py
```diff
@@ -1949,3 +1950,22 @@ def get_subscriptions_info_by_topic(
...
+        timeout: float = 8.0
+    ) -> bool:
+        """
+        Wait until node name is present in the system or timeout.
+
+        :param node_name: name of the node to wait for.
+        :timeout: seconds to wait for the node to be present.
```
jacobperron, line 1963: missing "param":
```suggestion
        :param timeout: seconds to wait for the node to be present.
```

rclpy/test/test_node.py
```diff
@@ -1896,6 +1896,18 @@ def test_static_dynamic_typing(self):
         self.assertEqual(self.node.get_parameter('int_param').value, 4)
         self.assertFalse(self.node.has_parameter('dynamic_param'))
 
+    def test_wait_for_node(self):
+        try:
+            node = rclpy.create_node(
+                TEST_NODE + '_create_node', namespace='/my_ns', context=self.context
+            )
+            self.assertTrue(self.node.wait_for_node('my_node'))
```
jacobperron, line 1904: We're missing the timeout argument to `wait_for_node` here.

jacobperron (commented):
LGTM after the test is fixed.

rclpy/rclpy/node.py
```diff
@@ -1949,3 +1950,25 @@ def get_subscriptions_info_by_topic(
...
+        :param timeout: seconds to wait for the node to be present. If negative, the function
+                         won't timeout.
+        :return: True if the node was found, False if timeout.
+        """
+        start = time.time()
+        flag = False
+        # TODO refactor this implementation when we can react to guard condition events, or replace
+        # it entirely with an implementation in rcl. see https://github.com/ros2/rclpy/issues/929
+        while time.time() - start < timeout and not flag:
+            flag = node_name in self.get_node_names()
```
ivanpauno, line 1972: The node name is not an unique identifier for a node, `get_node_names_and_namespaces()` should be used instead.
Having a `get_fully_qualified_node_names()` would be useful as well.

ivanpauno (changes requested):

ivanpauno:
@jacobperron are you planning to pick this up from Gonzo?
If not, I can address the remaining comments.

rclpy/rclpy/node.py
jacobperron: I've updated the documentation to clarify this should by a fully qualified name (ie. including the namespace).

rclpy/rclpy/node.py
jacobperron: Good catch. I've updated this PR to account for fully qualified node names.

rclpy/rclpy/node.py
```diff
@@ -1949,3 +1963,31 @@ def get_subscriptions_info_by_topic(
             topic_name,
             no_mangle,
             _rclpy.rclpy_get_subscriptions_info_by_topic)
+
+    def wait_for_node(
+        self,
+        node_name: str,
```
ivanpauno, line 1969: nit:

I would use:
```suggestion
        fq_node_name: str,
```
or `fully_qualified_node_name`, to make it more explicit.

rclpy/test/test_node.py
```diff
@@ -1896,6 +1896,18 @@ def test_static_dynamic_typing(self):
         self.assertEqual(self.node.get_parameter('int_param').value, 4)
         self.assertFalse(self.node.has_parameter('dynamic_param'))
 
+    def test_wait_for_node(self):
+        try:
+            node = rclpy.create_node(
+                'waiting_for_this_node', namespace='/my_ns', context=self.context
+            )
+            self.assertTrue(self.node.wait_for_node('/my_ns/waiting_for_this_node', 0.5))
```
ivanpauno, line 1904: I would use a larger timeout, so it works reliably in all platforms and with all rmw implementations:

```suggestion
            self.assertTrue(self.node.wait_for_node('/my_ns/waiting_for_this_node', 3))
```

ivanpauno (approved):
I have some minor suggestions.
LGTM!

ivanpauno:
@jacobperron friendly ping

rclpy/rclpy/node.py
jacobperron: Done: [90e8bf0](https://github.com/ros2/rclpy/pull/930/commits/90e8bf0f56903c40ee4a6741e85acbb5d5cedb0f)

rclpy/test/test_node.py
jacobperron: done [90e8bf0](https://github.com/ros2/rclpy/pull/930/commits/90e8bf0f56903c40ee4a6741e85acbb5d5cedb0f)

ivanpauno (approved):

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17110/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11658/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17527/)

Issue #937, opened 2022-05-04
[Humble]: Sometimes get a stack trace when shutting down nodes


clalancette:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04, arm64
- Installation type:
  - Binaries
- Version or commit hash:
  - Humble, rclpy package version 3.3.2-2
- DDS implementation:
  - Default, Fast-DDS
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

Follow the tutorial at https://docs.ros.org/en/humble/Tutorials/Tf2/Writing-A-Tf2-Static-Broadcaster-Py.html .  Run the example created:

```
ros2 run learning_tf2_py static_turtle_tf2_broadcaster mystaticturtle 0 0 1 0 0 0
```

When done, hit Ctrl-C

#### Expected behavior

Example exits without error.

#### Actual behavior

We get a Python stack trace in rclpy:

```
^CTraceback (most recent call last):
  File "/home/ubuntu/tf2_ws/install/learning_tf2_py/lib/learning_tf2_py/static_turtle_tf2_broadcaster", line 33, in <module>
    sys.exit(load_entry_point('learning-tf2-py==0.0.0', 'console_scripts', 'static_turtle_tf2_broadcaster')())
  File "/home/ubuntu/tf2_ws/install/learning_tf2_py/lib/python3.10/site-packages/learning_tf2_py/static_turtle_tf2_broadcaster.py", line 84, in main
    rclpy.shutdown()
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/__init__.py", line 126, in shutdown
    _shutdown(context=context)
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/utilities.py", line 58, in shutdown
    return context.shutdown()
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/context.py", line 102, in shutdown
    self.__context.shutdown()
rclpy._rclpy_pybind11.RCLError: failed to shutdown: rcl_shutdown already called on the given context, at ./src/rcl/init.c:241
[ros2run]: Process exited with failure 1
```

This seems to happen every time.

fujitatomoya:
@clalancette can you try the following patch to see if the problem still can be observed?

```diff
diff --git a/rclpy/src/rclpy/context.cpp b/rclpy/src/rclpy/context.cpp
index 07cfcc5..bb4aa2c 100644
[...]

PR #939, opened 2022-05-05
check if the context is already shutdown.

fujitatomoya:
address https://github.com/ros2/rclpy/issues/937

### DIFF
--- a/rclpy/src/rclpy/context.cpp
+++ b/rclpy/src/rclpy/context.cpp
@@ -34,6 +34,7 @@
 
 namespace
 {
+// g_contexts is a collection of valid contexts
 std::vector<rcl_context_t *> g_contexts;
 std::mutex g_contexts_mutex;
 }
@@ -48,6 +49,7 @@ void shutdown_contexts()
     rcl_ret_t ret = rcl_shutdown(c);
     (void)ret;
   }
+  g_contexts.clear();
 }
 
 Context::Context(py::list pyargs, size_t domain_id)
@@ -150,16 +152,20 @@ Context::shutdown()
 {
   {
     std::lock_guard guard{g_contexts_mutex};
-    g_contexts.erase(
-      std::remove(g_contexts.begin(), g_contexts.end(), rcl_context_.get()),
-      g_contexts.end());
+    if (std::find(g_contexts.begin(), g_contexts.end(), rcl_context_.get()) != g_contexts.end()) {
+      g_contexts.erase(
+        std::remove(g_contexts.begin(), g_contexts.end(), rcl_context_.get()),
+        g_contexts.end());
+    }
   }
-  // check if context is already shutdown by signal handler
+  // Just in case, check if context is valid
   if (rcl_context_is_valid(rcl_context_.get())) {
     rcl_ret_t ret = rcl_shutdown(rcl_context_.get());
     if (RCL_RET_OK != ret) {
       throw RCLError("failed to shutdown");
     }
+  } else {
+    throw RCLError("context is already shutdown");
   }
 }
 
--- a/rclpy/src/rclpy/context.cpp
+++ b/rclpy/src/rclpy/context.cpp
@@ -152,20 +152,15 @@ Context::shutdown()
 {
   {
     std::lock_guard guard{g_contexts_mutex};
-    if (std::find(g_contexts.begin(), g_contexts.end(), rcl_context_.get()) != g_contexts.end()) {
-      g_contexts.erase(
-        std::remove(g_contexts.begin(), g_contexts.end(), rcl_context_.get()),
-        g_contexts.end());
+    auto iter = std::find(g_contexts.begin(), g_contexts.end(), rcl_context_.get());
+    if (iter != g_contexts.end()) {
+      g_contexts.erase(iter);
     }
   }
-  // Just in case, check if context is valid
-  if (rcl_context_is_valid(rcl_context_.get())) {
-    rcl_ret_t ret = rcl_shutdown(rcl_context_.get());
-    if (RCL_RET_OK != ret) {
-      throw RCLError("failed to shutdown");
-    }
-  } else {
-    throw RCLError("context is already shutdown");
+
+  rcl_ret_t ret = rcl_shutdown(rcl_context_.get());
+  if (RCL_RET_OK != ret) {
+    throw RCLError("failed to shutdown");
   }
 }
 
--- a/rclpy/src/rclpy/context.cpp
+++ b/rclpy/src/rclpy/context.cpp
@@ -42,6 +42,7 @@ namespace rclpy
 {
 void shutdown_contexts()
 {
+  // graceful shutdown all contexts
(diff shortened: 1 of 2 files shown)


rclpy/src/rclpy/context.cpp
```diff
@@ -42,6 +42,7 @@ namespace rclpy
 {
 void shutdown_contexts()
 {
+  // graceful shutdown all contexts
   std::lock_guard guard{g_contexts_mutex};
   for (auto * c : g_contexts) {
     rcl_ret_t ret = rcl_shutdown(c);
```
fujitatomoya, line 48: another way we can take is, we can remove the context from `g_contexts` once `rcl_shutdown` issued on it. so that `Context::shutdown` can check if it needs to issue `rcl_shutdown` on itself.

clalancette (approved):
This seems to fix the bug for me, and seems like an improvement.  Thanks for looking into it!  I'll approve with green CI.

But I would be interested to get some feedback from @sloretz before we merge it.

fujitatomoya:
yeah, i am open to ideas, that would be appreciated! (btw, this fix is aligned with rclcpp, which checks the context is still valid via `is_valid` method right before `shutdown`.)

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16661/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11247/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17056/)

fujitatomoya:
@sloretz friendly ping.

rclpy/src/rclpy/context.cpp
sloretz: That seems like a reasonable solution too, and matches `Context::shutdown()` removing the context from `g_contexts`.

rclpy/src/rclpy/context.cpp
```diff
@@ -153,9 +154,12 @@ Context::shutdown()
       std::remove(g_contexts.begin(), g_contexts.end(), rcl_context_.get()),
       g_contexts.end());
```
sloretz, line 155: With this PR there's a difference between how `Context::shutdown()` and `rclpy::shutdown_contexts()` treat `g_contexts`. The first one treats the global variable as a collection of all valid contexts, while the second treats it as a collection of all created contexts. I'd recommend picking one or the other and documenting which in a comment on `g_contexts`.


If `g_contexts` is all contexts, I'd recommend moving this line to the destructor `Context::~Context()`. That has an advantage that there's help from the language to make sure the global variable doesn't get leaked, but means contexts may live in that list for longer.

If `g_contexts` is all valid contexts, this removal is fine but I'd recommend going with your other solution of clearing `g_contexts` in `rclpy::shutdown_contexts()`.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16695/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11278/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17092/)

fujitatomoya:
@sloretz requesting review again.

rclpy/src/rclpy/context.cpp
```diff
@@ -149,13 +152,20 @@ Context::shutdown()
...
-  if (RCL_RET_OK != ret) {
-    throw RCLError("failed to shutdown");
+  // Just in case, check if context is valid
+  if (rcl_context_is_valid(rcl_context_.get())) {
+    rcl_ret_t ret = rcl_shutdown(rcl_context_.get());
+    if (RCL_RET_OK != ret) {
+      throw RCLError("failed to shutdown");
+    }
+  } else {
+    throw RCLError("context is already shutdown");
```
sloretz, line 168: I would recommend against `RCLError` since `rcl_context_is_valid` won't set the error state, meaning the message will be something like error not set. See how `RCLError` reads the error message here: https://github.com/ros2/rclpy/blob/31147813b653b4820c0b7c16b12ee30f8ab29ccb/rclpy/src/rclpy/exceptions.cpp#L56

But also, see the `rcl_shutdown` comment above

rclpy/src/rclpy/context.cpp
```diff
@@ -149,13 +152,20 @@ Context::shutdown()
...
+      g_contexts.erase(
+        std::remove(g_contexts.begin(), g_contexts.end(), rcl_context_.get()),
+        g_contexts.end());
+    }
   }
-  rcl_ret_t ret = rcl_shutdown(rcl_context_.get());
-  if (RCL_RET_OK != ret) {
-    throw RCLError("failed to shutdown");
+  // Just in case, check if context is valid
+  if (rcl_context_is_valid(rcl_context_.get())) {
```
sloretz, line 162: `rcl_shutdown` already does this check, so if we're going to throw anyways then we don't need this check here.

https://github.com/ros2/rcl/blob/493ebf3a5d8fe2d861a93006a7d6908fd4e592d9/rcl/src/rcl/init.c#L241

rclpy/src/rclpy/context.cpp
```diff
@@ -149,13 +152,20 @@ Context::shutdown()
 {
   {
     std::lock_guard guard{g_contexts_mutex};
-    g_contexts.erase(
-      std::remove(g_contexts.begin(), g_contexts.end(), rcl_context_.get()),
-      g_contexts.end());
+    if (std::find(g_contexts.begin(), g_contexts.end(), rcl_context_.get()) != g_contexts.end()) {
+      g_contexts.erase(
+        std::remove(g_contexts.begin(), g_contexts.end(), rcl_context_.get()),
```
sloretz, line 157: I think this searches for the context twice, and will keep searching the list even though the context should only be in it once. I'd recommend this instead

```c++
auto iter = std::find(....)
if (iter != g_contexts.end()) {
    g_contexts.erase(iter);
}
```

sloretz (changes requested):

rclpy/src/rclpy/context.cpp
fujitatomoya: nice point 👍 thanks!

rclpy/src/rclpy/context.cpp
fujitatomoya: okay, i will remove this check.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16703/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11283/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17099/)

fujitatomoya:
@sloretz thanks for the review, i have addressed your comments.

sloretz (approved):
LGTM with green CI! Thanks for iterating

fujitatomoya:
@Mergifyio backport humble

PR #945, opened 2022-05-25
implement parameter_client

ihasdapie:
This PR ports the [`rclcpp` `parameter_client`](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/parameter_client.hpp) to `rclpy` and also moves some functionality out of `ros2param` into `rclpy`. Related changes in ros2param can be found in https://github.com/ros2/ros2cli/pull/716

I'd like some review on the proposed architecture & implementation before I clean things up and write tests etc.


ihasdapie:
Just a sanity check:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16751/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11334/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17151/)

rclpy/rclpy/parameter_client.py
```diff
@@ -0,0 +1,273 @@
...
+        :rtype:
+        """
+        request = SetParameters.Request()
+        request.parameters = [RclpyParameter(name=i).to_parameter_msg() for i in names]
+        future = self.set_parameter_client_.call_async(request)
+        if callback:
+            future.add_done_callback(callback)
+        return future
+
+    def load_parameter_file(self, parameter_file, use_wildcard, return_parameters=False):
```
jacobperron, line 229: Similar to the c++ implementation, I think it would be nice to separate the logic of parsing of the parameters file and loading a dict of parameters.

jacobperron (commented):
Looks reasonable to me! I have mostly minor comments

rclpy/rclpy/__init__.py
```diff
@@ -131,17 +132,16 @@ def shutdown(*, context: Context = None, uninstall_handlers: Optional[bool] = No
...
-
 def create_node(
     node_name: str,
     *,
-    context: Context = None,
-    cli_args: List[str] = None,
-    namespace: str = None,
+    context: Union[Context, None] = None,
+    cli_args: Union[List[str], None] = None,
+    namespace: Union[str, None] = None,
```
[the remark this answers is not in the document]
ihasdapie: Yup, I just realized this morning that `Optional` is a thing. I'll put these into a separate PR

rclpy/rclpy/parameter_client.py
```diff
@@ -0,0 +1,273 @@
...
+        """Wait for all parameter services to be available.
+
+        :param timeout_sec: Seconds to wait. If ``None``, then wait forever.
+        :type timeout_sec: Union[float, None]
+        :return:
+        :rtype:
+
+        :return: ``True`` if all services are available, False otherwise.
+        """
+        return all([
```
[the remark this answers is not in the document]
ihasdapie: I wanted to avoid the added complexity but I guess it's needed here

rclpy/rclpy/parameter_client.py
ihasdapie: Loading a dict of parameters is just calling `set_parameters` on the parsed dict's values so I don't think there is too much of a point

jacobperron: I think it might be more generally useful to have an API that reads an input parameters yaml and returns a dict. The implementation of such a function might be replaced by a call to an equivalent `rcl` function in the future.

Then it's true that this function's implementation becomes much simpler, but that's a good thing, IMO.

jacobperron: For example, this function becomes something like:

```py
def load_parameter_file(self, parameter_file, use_wildcard, return_parameters=False):
    params_dict = load_parameters_yaml_file(parameter_file, use_wildcard)
    parameters = parse_parameter_dict(namespace='', parameter_dict=param_dict)
    return self.set_parameters(parameters)
```

rclpy/rclpy/parameter_client.py
```diff
@@ -0,0 +1,273 @@
...
+                value = param_file[k]
+                if type(value) != dict or 'ros__parameters' not in value:
+                    raise RuntimeError('Invalid structure of parameter file for node {}'
+                                       'expected same format as provided by ros2 param dump'
+                                       .format(k))
+                param_dict.update(value['ros__parameters'])
+            parameters = parse_parameter_dict(namespace='', parameter_dict=param_dict)
+            future = self.set_parameters(parameters)
+            if return_parameters:
+                return (future, parameters)
```
jacobperron, line 252: I'm not sure I understand the use case of returning the future _and_ the parameters. Seems like we're trying to pack too much functionality into this function. IMO, it'd be better to have users call a separate free-function to get parameters from a file (e.g. `load_parameters_yaml_file`).

rclpy/rclpy/parameter_client.py
```diff
@@ -0,0 +1,273 @@
...
+            return future
+
+def parse_parameter_dict(namespace, parameter_dict):
```
jacobperron, line 255: I would mark this function as private to this module by prepending a double underscore

```suggestion
def __parse_parameter_dict(namespace, parameter_dict):
```

rclpy/rclpy/parameter_client.py
ihasdapie: That was to make something work nicer in ros2param but I realized that it was a little ugly. I'll have it pushed once I fix the linting errors

rclpy/rclpy/parameter_client.py
ihasdapie: I've since moved this to `parameter.py` since it makes more sense to provide the methods for loading in parameters from a dict there as opposed to in `parameter_client`.

ihasdapie: Regarding private: I've prepended a single underscore since 1) it's not a class method (I don't think non-class methods get name-mangled, do they?) and 2) though I'd like it to be private it is also fairly useful if you happen to have a dict of parameters from elsewhere.

rclpy/rclpy/parameter_client.py
ihasdapie: If done this way `params_dict` is largely useless since the parameters are still in string form at that point, `parse_parameter_dict` is what will turn them into a dict with ParameterMsg values. 

I agree that it would be nice to replace it with a `rcl` call. In the [rclcpp](https://github.com/ros2/rclcpp/blob/02802bcc385c3d6d814add93618f3388d44adec7/rclcpp/src/rclcpp/parameter_client.cpp#L295-L289) [implementaton](https://github.com/ros2/rclcpp/blob/02802bcc385c3d6d814add93618f3388d44adec7/rclcpp/src/rclcpp/parameter_map.cpp#L140-L130) it appears that the `rcl` call will directly return the parameter map which is what my [proposed approach](https://github.com/ros2/rclpy/blob/802adf1a36278fe3ce86a32e9c0a7dc47c786603/rclpy/rclpy/parameter.py#L265) does.

rclpy/test/test_parameter_client.py
```diff
@@ -0,0 +1,136 @@
...
+        result = future.result()
+        assert result is not None
+        assert not result.results[0].successful
+        assert result.results[0].reason == 'Static parameter cannot be undeclared'
+
+        future = self.client.delete_parameters(['delete_param_dynamic'])
+        self.executor.spin_until_future_complete(future)
+        result = future.result()
+        assert result is not None
+        assert result.results[0].successful
```
ihasdapie, line 121: We could just form the assertions with a direct comparsion to the expected result. Quick fix if we want it but I didn't go this route because it was unnecessairly verbose.

```python
        assert results == rcl_interfaces.srv.DescribeParameters.Response(
            descriptors=[rcl_interfaces.msg.ParameterDescriptor(name='int_arr_param',
                                                                type=7,
                                                                description='',
                                                                additional_constraints='',
                                                                read_only=False,
                                                                dynamic_typing=False,
                                                                floating_point_range=[],
                                                                integer_range=[])])
```

rclpy/test/test_parameter_client.py
```diff
@@ -0,0 +1,136 @@
...
+        assert results is not None
+        res = [i.successful for i in results.results]
+        assert all(res)
+
+    def test_get_parameter(self):
+        future = self.client.get_parameters(['int_arr_param', 'float/param//'])
+        self.executor.spin_until_future_complete(future)
+        results = future.result()
+        assert results is not None
+        assert list(results.values[0].integer_array_value) == [1, 2, 3]
```
ihasdapie, line 69: Not sure if there's an order guarantee? Would `float/param//` ever be the first element in the array?

ihasdapie:
ci jobs for this PR and related PR https://github.com/ros2/ros2cli/pull/716


* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16772/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11353/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17174/)

Note: build may fail due to unrelated linting nitpick in `rclpy`, fixed now as per https://github.com/ros2/rclpy/pull/945/commits/7894e96f4ad484b9cb026bc336897b520066a342

rclpy/test/test_parameter.py
```diff
@@ -213,6 +214,23 @@ def test_parameter_value_to_python(self):
...
+                param_str: "string"
+            """
+        expected = {
+            'param_1': Parameter('param_1', Parameter.Type.INTEGER, 1).to_parameter_msg(),
+            'param_str': Parameter('param_str', Parameter.Type.STRING, 'string').to_parameter_msg()
+        }
+
+        with NamedTemporaryFile() as f:
+            f.write(str.encode(yaml_string))
+            f.seek(0)
```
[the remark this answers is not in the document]
ihasdapie: Need to move the file ptr offset back to the beginning of the file or else with the way i'm doing it `f.read()` will only see the end of a file

rclpy/test/test_parameter.py
```diff
@@ -213,6 +214,23 @@ def test_parameter_value_to_python(self):
...
+            parameter_dict = parameter_dict_from_yaml_file(f.name)
```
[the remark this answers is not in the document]
ihasdapie: Won't work, this requires a filename and doesn't take a stream. This is what the `rclcpp` implementation does https://github.com/ros2/rclcpp/blob/790e529ba359b83100ed4f7ca64a773aaa608968/rclcpp/src/rclcpp/parameter_client.cpp#L290-L302

rclpy/test/test_parameter.py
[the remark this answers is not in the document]
jacobperron: But we're not reusing the file handle inside `parameter_dict_from_yaml_file`; it creates it's own file handle (presumably already starting from the beginning of the file). Or am I misunderstanding something about Python file handles?

rclpy/test/test_parameter.py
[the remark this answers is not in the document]
jacobperron: That's exactly my point: writing a test demonstrating how the function fails when we give it a bogus filename.

rclpy/test/test_parameter_client.py
```diff
@@ -0,0 +1,136 @@
...
+        assert results is not None
+        assert 'int_arr_param' in results.result.names
+        assert 'float/param//' in results.result.names
+
+    def test_describe_parameters(self):
+        future = self.client.describe_parameters(['int_arr_param'])
+        self.executor.spin_until_future_complete(future)
+        results = future.result()
+        assert results is not None
+        assert results.descriptors[0].type == 7
```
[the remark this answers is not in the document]
jacobperron: We're still comparing against a literal number. Can you update this to:

```suggestion
        assert results.descriptors[0].type == Parameter.Type.INTEGER_ARRAY
```

rclpy/test/test_parameter_client.py
```diff
@@ -0,0 +1,139 @@
...
+        assert results.descriptors[0].name == 'int_arr_param'
+
+    def test_get_paramter_types(self):
+        future = self.client.get_parameter_types(['int_arr_param'])
+        self.executor.spin_until_future_complete(future)
+        results = future.result()
+        assert results is not None
```
jacobperron, line 95: To avoid an invalid index error:
```suggestion
        assert results is not None
        assert len(results.types) == 1
```

rclpy/test/test_parameter_client.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def test_delete_parameters(self):
+        self.target_node.declare_parameter('delete_param', 10)
+        descriptor = rcl_interfaces.msg.ParameterDescriptor(dynamic_typing=True)
+        self.target_node.declare_parameter('delete_param_dynamic', 10, descriptor=descriptor)
+
+        future = self.client.delete_parameters(['delete_param'])
+        self.executor.spin_until_future_complete(future)
+        result = future.result()
+        assert result is not None
```
jacobperron, line 116:
```suggestion
        assert result is not None
        assert len(result.results) == 1
```

rclpy/test/test_parameter_client.py
```diff
@@ -0,0 +1,139 @@
...
+        assert not result.results[0].successful
+        assert result.results[0].reason == 'Static parameter cannot be undeclared'
+
+        future = self.client.delete_parameters(['delete_param_dynamic'])
+        self.executor.spin_until_future_complete(future)
+        result = future.result()
+        assert result is not None
```
jacobperron, line 123:
```suggestion
        assert result is not None
        assert len(result.results) == 1
```

rclpy/test/test_parameter.py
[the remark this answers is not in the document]
ihasdapie: Oh ok, I misinterpreted it was wanting to have `parameter_dict_from_yaml_file` take a stream instead

rclpy/test/test_parameter_client.py
```diff
@@ -0,0 +1,136 @@
...
+                param_1: 1
+                param_str: "string"
+            """
+        with NamedTemporaryFile() as f:
+            f.write(str.encode(yaml_string))
+            f.seek(0)
+            future = self.client.load_parameter_file(f.name)
+        self.executor.spin_until_future_complete(future)
+        result = future.result()
+        assert result is not None
```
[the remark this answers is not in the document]
jacobperron: This comment still stands.

rclpy/rclpy/parameter_client.py
```diff
@@ -0,0 +1,310 @@
...
+        self._describe_parameters_client = self.node.create_client(
+            DescribeParameters, f'{remote_node_name}/describe_parameters',
+            qos_profile=qos_profile, callback_group=callback_group
+        )
+        self._set_parameters_atomically_client = self.node.create_client(
+            SetParametersAtomically, f'{remote_node_name}/set_parameters_atomically',
+            qos_profile=qos_profile, callback_group=callback_group
+        )
+
+    def service_is_ready(self) -> bool:
```
jacobperron, line 100:
```suggestion
    def server_is_ready(self) -> bool:
```

reads better IMO, since there is more than one service.


Another option is

```suggestion
    def services_are_ready(self) -> bool:
```

rclpy/rclpy/parameter_client.py
```diff
@@ -0,0 +1,310 @@
...
+        return all([
+            self._list_parameter_client.service_is_ready(),
+            self._set_parameter_client.service_is_ready(),
+            self._get_parameter_client.service_is_ready(),
+            self._get_parameter_types_client.service_is_ready(),
+            self._describe_parameters_client.service_is_ready(),
+            self._set_parameters_atomically_client.service_is_ready(),
+        ])
+
+    def wait_for_service(self, timeout_sec: Optional[float] = None) -> bool:
```
jacobperron, line 115:
```suggestion
    def wait_for_server(self, timeout_sec: Optional[float] = None) -> bool:
```

or 

```suggestion
    def wait_for_services(self, timeout_sec: Optional[float] = None) -> bool:
```

As long as we're consistent with the `*_ready` method.

rclpy/rclpy/parameter_client.py
```diff
@@ -0,0 +1,310 @@
...
+        """
+        Wait for all parameter services to be available.
+
+        :param timeout_sec: Seconds to wait. If ``None``, then wait forever.
+        :type timeout_sec: Union[float, None]
```
jacobperron, line 120: This is already documented in the type annotation.
```suggestion
```

rclpy/rclpy/parameter_client.py
```diff
@@ -0,0 +1,310 @@
...
+        names: List[str],
+        callback: Optional[Callable] = None
+    ) -> Future:
+        """
+        Get parameter types given names.
+
+        The result after the returned future is complete
+        will be of type ``rcl_interfaces.srv.GetParameterTypes.Response``.
+
+        Parameter type definitions are given in rcl_interfaces.msg.ParameterType
```
jacobperron, line 236: These are [mirrored in rclpy](https://github.com/ros2/rclpy/blob/2c8ffe098eefebeec5a49e906e2506da4e0db434/rclpy/rclpy/parameter.py#L26)
```suggestion
        Parameter type definitions are given in Parameter.Type
```

rclpy/rclpy/parameter_client.py
```diff
@@ -0,0 +1,310 @@
...
+        if callback:
+            future.add_done_callback(callback)
+        return future
+
+    def delete_parameters(
+        self, names: List[str], callback: Optional[Callable] = None
+    ) -> Future:
+        """
+        Unset parameters with given names.
+
```
jacobperron, line 276: It might be worth mentioning here that only parameters declared as dynamically typed can be unset.

rclpy/rclpy/parameter_client.py
```diff
@@ -0,0 +1,310 @@
...
+    def load_parameter_file(
+        self,
+        parameter_file: str,
+        use_wildcard: bool = False,
+    ) -> Future:
+        """
+        Load parameters from a yaml file.
+
+        Wrapper around `parse_parameter_dict` and `load_dict`
```
jacobperron, line 299: I don't know if this line is still relevant (e.g. I don't know what `parser_parameter_dict` and `load_dict` are).

rclpy/test/test_parameter.py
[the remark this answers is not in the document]
ihasdapie: You're right, I was prototying the test elsewhere where I wasn't making a new file handle. Changed to flush now

rclpy/rclpy/parameter_client.py
```diff
@@ -0,0 +1,310 @@
...
+        request = GetParameters.Request()
+        request.names = names
+        future = self._get_parameter_client.call_async(request)
+        if callback:
+            future.add_done_callback(callback)
+        return future
+
+    def set_parameters(
+        self,
+        parameters: Sequence[Parameter],
```
jacobperron, line 183: I think this type annotation is wrong. Looks like we're expecting a message type (`rcl_interfaces.msg.Parameter`), and not `rclpy.parameter.Parameter`. Although, maybe it's more convenient to accept an rclpy `Parameter` instead of (or in addition to) the messsage type. For example, our call in the test could change like,

```diff
-        future = self.client.set_parameters([
-            Parameter('int_param', Parameter.Type.INTEGER, 88).to_parameter_msg(),
-            Parameter('string/param', Parameter.Type.STRING, 'hello world').to_parameter_msg(),
-        ])
+        future = self.client.set_parameters([
+            Parameter('int_param', Parameter.Type.INTEGER, 88),
+            Parameter('string/param', Parameter.Type.STRING, 'hello world'),
+        ])
```

And we could do the conversion `to_parameter_msg()` inside this function as-needed.

jacobperron (commented):
Another round of review!

thanks for iterating :)

rclpy/rclpy/parameter_client.py
ihasdapie: Not relevant anymore, i'll take it out.

rclpy/rclpy/parameter_client.py
ihasdapie: In this file `Parameter` refers to `rcl_interfaces.msg.Parameter`. But also makes sense to accept both and convert as-needed.

rclpy/rclpy/parameter_client.py
```diff
@@ -0,0 +1,315 @@
...
+        The result after the returned future is complete
+        will be of type ``rcl_interfaces.srv.SetParameters.Response``.
+
+        :param parameters: Sequence of parameters to set.
+        :param callback: Callback function to call when the request is complete.
+        :return: ``Future`` with the result of the request.
+        """
+        request = SetParameters.Request()
+        request.parameters = [
+            i.to_parameter_msg() if isinstance(i, Parameter) else i for i in parameters
```
jacobperron, line 197: tiny nitpick: renaming `i` to `param` is more meaningful.

jacobperron (approved):
LGTM

Did you have an example to add to the [demos](https://github.com/ros2/demos/tree/master/demo_nodes_py/demo_nodes_py) package? We could create a `parameters` directory there for `demo_nodes_py`.

rclpy/rclpy/parameter_client.py
ihasdapie: I guess so but it's already an ugly long list comprehension and that would push it _over_ 100 lines and I'd have to break it apart again. In any case it's fairly clear what `i` is

ihasdapie:
Yup I'll submit something to the demos package as well

rclpy/rclpy/parameter_client.py
jacobperron: Sure, I don't think this is any less readable:

```py
            param.to_parameter_msg()
            if isinstance(param, Parameter) else param
            for param in parameters
```

I think it would also be acceptable to wrap this in a function private to the module (since we use the same logic in a couple places). e.g. `__normalize_to_list_of_parameter_msgs`

I won't block the PR on this comment though.

ihasdapie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16814/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11392/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17219/)

Blast545:
:man_farmer: This PR introduced a test regression in the windows jobs of the buildfarm. 
See: [nightly_win_rel#2334](https://ci.ros2.org/view/nightly/job/nightly_win_rel/2334/testReport/junit/rclpy.rclpy.test.test_parameter/TestParameter/test_parameter_dict_from_yaml_file/) and [nightly_win_deb#2389](https://ci.ros2.org/view/nightly/job/nightly_win_deb/2389/testReport/junit/rclpy.rclpy.test.test_parameter/TestParameter/test_parameter_dict_from_yaml_file/)

I see that CI triggered for this PR already showed this error, why was it ignored?

Can you take a look @ihasdapie ?

ihasdapie:
Ah shoot, my bad. I'll push something to resolve it shortly

fujitatomoya:
@ihasdapie i think we need to revert this.

PR #946, opened 2022-05-26
Add py.typed to package

### LINKED ISSUE #936: Include py.typed in a package
#### Feature description

I use [mypy](https://github.com/python/mypy), one of the most famous static type checkers in Python, to check my Python code depending on rclpy.
However, mypy won't use the type annotation of `rclpy`, since mypy only uses the type annotation of packages that are marked as [PEP561-compatible](https://peps.python.org/pep-0561/) (i.e. Packages that have `py.typed` inside).
- https://mypy.readthedocs.io/en/stable/installed_packages.html

Therefore, I can't check the code like:
```sh
$ cat foo.py
import rclpy

reveal_type(rclpy)
reveal_type(rclpy.init)

rclpy.foo  # should be error
rclpy.init(foo="bar")  # should be error

$ mypy foo.py 
foo.py:1: error: Cannot find implementation or library stub for module named 'rclpy'
foo.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
foo.py:3: note: Revealed type is 'Any'
foo.py:4: note: Revealed type is 'Any'
Found 1 error in 1 file (checked 1 source file)
```

Would it be possible to include py.typed in `rclpy` package? (I am happy to work on this task!)

- The simplest solution is to create an empty file `py.typed` under `rclpy` directory
  - `rclpy` uses `ament_python_install_package` macro, and the macro just copies the directory tree into the install directory:
 https://github.com/ament/ament_cmake/blob/b84cf9e6f2a61d8f9fc5a90c02dc2b5cb63e7f76/ament_cmake_python/cmake/ament_python_install_package.cmake#L181-L186
  - Adding this empty `py.typed` file is common in Python (e.g. [numpy](https://github.com/numpy/numpy/blob/4e667cbeaf5b8b4112b6e8bf9140a3d186d490ea/numpy/py.typed))
- Another approach is to add the following line in `CMakeLists.txt`:
    ```cmake
    file(WRITE "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}/py.typed" "")
    ```
  - But I don't think it's a good idea since it's a bit harder to know whether the package is PEP561-compatible or not.

bonprosoft:
I added `py.typed` file to mark the package [PEP561-compatible](https://peps.python.org/pep-0561/).
This would be helpful for static type checkers.

Closes #936 

cc: @adityapande-1995

Signed-off-by: Yuki Igarashi <me@bonprosoft.com>


adityapande-1995 (commented):
Looks good ! Is there a way we could add a test for this using our CI ?

aprotyas (approved):

bonprosoft:
> Is there a way we could add a test for this using our CI?

I'm not sure.
Because `py.typed` is copied into the package dir by [`ament_cmake_python`](https://github.com/ros2/rclpy/blob/7ffe94865f38c67812799bf5751a8f0f3f918c25/rclpy/CMakeLists.txt#L49), in my opinion, it is not so reasonable to check if the file is located in the package directory by `pytest`, which is for testing python modules.

If such test is highly necessary, let me think about the approach :smiley:

aprotyas:
Yes, I agree that PEP561 compatibility shouldn't/can't really be tested like the rest of rclpy's test suite.

Regardless, running CI for this change.

Repos file: https://gist.github.com/aprotyas/af5130e8d82be0011957df611898ca7e
Build args: `--packages-above-and-dependencies rclpy`
Test args: `--packages-above rclpy`
Job: https://ci.ros2.org/job/ci_launcher/10341/

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16758/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11340/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17157/)

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/python-type-checking-in-ros2/28507/1

Issue #944, opened 2022-05-20
rclpy.shutdown() hangs when called from a callback


wjwwood:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - Source build
- Version or commit hash:
  - `3.3.1`
- DDS implementation:
  - `rmw_fastrtps_cpp`

#### Steps to reproduce issue

- Run this script:

```
import rclpy
from rclpy.node import Node
from std_msgs.msg import String

class MinimalSubscriber(Node):

    def __init__(self):
        super().__init__('minimal_subscriber')
        self.subscription = self.create_subscription(
            String,
            '/chatter',
            self.callback,
            10)

    def callback(self, msg):
        self.get_logger().info('I heard: "%s"' % msg.data)
        rclpy.shutdown()


if __name__ == '__main__':
    rclpy.init()
    minimal_subscriber = MinimalSubscriber()
    rclpy.spin(minimal_subscriber)
    rclpy.shutdown()

```
- Run this pub command: `ros2 topic pub /chatter std_msgs/msg/String '{data: asdf}'`

The script does not exit, it just hangs after printing the first message is received.

Output with ctrl-c after hanging:

```
$ RMW_IMPLEMENTATION=rmw_fastrtps_cpp python3 ./test.py 
[INFO] [1653073408.435204958] [minimal_subscriber]: I heard: "asdf"
^CTraceback (most recent call last):
  File "./test.py", line 23, in <module>
    rclpy.spin(minimal_subscriber)
  File "/path/to/ros2/source_build/lib/python3.8/site-packages/rclpy/__init__.py", line 222, in spin
    executor.spin_once()
  File "/path/to/ros2/source_build/lib/python3.8/site-packages/rclpy/executors.py", line 710, in spin_once
    handler()
  File "/path/to/ros2/source_build/lib/python3.8/site-packages/rclpy/task.py", line 239, in __call__
    self._handler.send(None)
  File "/path/to/ros2/source_build/lib/python3.8/site-packages/rclpy/executors.py", line 418, in handler
    await call_coroutine(entity, arg)
  File "/path/to/ros2/source_build/lib/python3.8/site-packages/rclpy/executors.py", line 343, in _execute_subscription
    await await_or_execute(sub.callback, msg)
  File "/path/to/ros2/source_build/lib/python3.8/site-packages/rclpy/executors.py", line 107, in await_or_execute
    return callback(*args)
  File "./test.py", line 17, in callback
[...]

PR #947, opened 2022-05-27
fix rclpy.shutdown() from hanging when triggered from callback

ihasdapie:
This PR closes https://github.com/ros2/rclpy/issues/944 by allowing for a `executor.shutdown` call to happen in a callback. 

A caveat of the proposed approach is that this does raise an `InvalidHandle` but it appears to be harmless and we just catch it. This is what is [currently used](https://github.com/ros2/rclpy/blob/7ffe94865f38c67812799bf5751a8f0f3f918c25/rclpy/rclpy/executors.py#L525-L564) in `rclpy`.  Alternatively we could track whether or not if a `GuardCondition` has been destroyed

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -79,20 +79,21 @@ class _WorkTracker:
             self._num_work_executing -= 1
             self._work_condition.notify_all()
 
-    def wait(self, timeout_sec=None):
+    def wait(self, timeout_sec: Optional[float] = None, is_shutdown: bool = False):
         """
         Wait until all work completes.
 
         :param timeout_sec: Seconds to wait. Block forever if None or negative. Don't wait if 0
         :type timeout_sec: float or None
+        :param: is_shutdown: if this is a shutdown call
         :rtype: bool True if all work completed
         """
-        if timeout_sec is not None and timeout_sec < 0:
+        if timeout_sec is not None and timeout_sec < 0.0:
             timeout_sec = None
         # Wait for all work to complete
         with self._work_condition:
             if not self._work_condition.wait_for(
-                    lambda: self._num_work_executing == 0, timeout_sec):
+                    lambda: self._num_work_executing == 0 or is_shutdown, timeout_sec):
                 return False
         return True
 
@@ -205,7 +206,7 @@ class Executor:
                 # Tell executor it's been shut down
                 self._guard.trigger()
 
-        if not self._work_tracker.wait(timeout_sec):
+        if not self._work_tracker.wait(timeout_sec, is_shutdown=self._is_shutdown):
             return False
 
         # Clean up stuff that won't be used anymore
@@ -425,7 +426,13 @@ class Executor:
                     entity.callback_group.ending_execution(entity)
                     # Signal that work has been done so the next callback in a mutually exclusive
                     # callback group can get executed
-                    gc.trigger()
+
+                    # Catch expected error where calling executor.shutdown()
+                    # from callback causes the GuardCondition to be destroyed 
+                    try:
+                        gc.trigger()
+                    except InvalidHandle:
+                        pass
         task = Task(
             handler, (entity, self._guard, self._is_shutdown, self._work_tracker),
             executor=self)
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -428,7 +428,7 @@ class Executor:
                     # callback group can get executed
 
                     # Catch expected error where calling executor.shutdown()
-                    # from callback causes the GuardCondition to be destroyed 
+                    # from callback causes the GuardCondition to be destroyed
                     try:
                         gc.trigger()
                     except InvalidHandle:
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -190,7 +190,7 @@ class Executor:
         # Task inherits from Future
         return task
 
-    def shutdown(self, timeout_sec: float = 0) -> bool:
+    def shutdown(self, timeout_sec: float = None) -> bool:
         """
         Stop executing callbacks and wait for their completion.
 
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -79,7 +79,7 @@ class _WorkTracker:
             self._num_work_executing -= 1
             self._work_condition.notify_all()
 
-    def wait(self, timeout_sec: Optional[float] = None, is_shutdown: bool = False):
+    def wait(self, timeout_sec: Optional[float] = None):
         """
         Wait until all work completes.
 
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/executors.py
```diff
@@ -79,20 +79,21 @@ def __exit__(self, t, v, tb):
...
         :rtype: bool True if all work completed
         """
-        if timeout_sec is not None and timeout_sec < 0:
+        if timeout_sec is not None and timeout_sec < 0.0:
             timeout_sec = None
         # Wait for all work to complete
         with self._work_condition:
             if not self._work_condition.wait_for(
-                    lambda: self._num_work_executing == 0, timeout_sec):
+                    lambda: self._num_work_executing == 0 or is_shutdown, timeout_sec):
```
wjwwood, line 96: So, if `is_shutdown` is true, why even wait in the first place? The predicate really only needs to test things that change. Taking a step back, why call this function at all if `is_shutdown` is true?

wjwwood (commented):
I'm not sure about the approach.

My (perhaps unpopular) opinion is that `rclpy.shutdown()` should never block (at least not on work in callbacks of executors), because that's how it works in rclcpp (it would be better to be consistent between client libraries), and the whole point is to _signal_ shutdown not wait for its completion (because waiting for shutdown's completion is complicated and leads to deadlocks often enough).

So in addition to making a "shutdown and wait for shutdown" kind of method work from within a callback (which this pr addresses for some, but perhaps not all, scenarios), I think the default behavior of `rclpy.shutdown()` should change to just signaling executors to stop, not actually waiting for them to stop. That definitely avoids deadlocks and is sufficient in most cases.

In fact, I'm not sure why `executor.shutdown()` and `rclpy.shutdown()` wait for work to finish in rclpy. I'm sure there's history there that I don't remember or never was part of. Maybe @sloretz has some insight, as I know he worked on the executors in rclpy quite a bit at some point.

ihasdapie:
I agree that it would be best to keep the implementation consistent between client libraries. The `rclpy` `Executor.shutdown` takes a `timeout_sec` which defaults to waiting forever. Perhaps we can change the default behavior to immediately shutting down instead? Or just deprecate the option to wait at all -- it'd be good to hear from @sloretz if there's any reason why that would be a bad idea.

wjwwood:
My initial thought was the same, keep the option to block, but change the default to non-blocking, but again I don't know the context for why it was made like that to begin with. I know there are some race-conditions with executor "cancelling" vs "shutdown" that came up, but I don't remember the details either.

rclpy/rclpy/executors.py
```diff
@@ -205,7 +206,7 @@ def shutdown(self, timeout_sec: float = None) -> bool:
                 # Tell executor it's been shut down
                 self._guard.trigger()
 
-        if not self._work_tracker.wait(timeout_sec):
+        if not self._work_tracker.wait(timeout_sec, is_shutdown=self._is_shutdown):
```
ihasdapie, line 209: Maybe don't call wait if we're shutting down

fujitatomoya: if we take this path, probably we do not need to call this method.

fujitatomoya (commented):
> My (perhaps unpopular) opinion is that rclpy.shutdown() should never block (at least not on work in callbacks of executors), because that's how it works in rclcpp (it would be better to be consistent between client libraries), and the whole point is to signal shutdown not wait for its completion (because waiting for shutdown's completion is complicated and leads to deadlocks often enough).

agree on this.

ihasdapie:
I've changed the default to _not_ wait and to not even call the hanging function in the first place.

rclpy/test/test_executor.py
```diff
@@ -456,6 +457,29 @@ def timer_callback():
...
+
+        tmr = self.node.create_timer(timer_period, timer_callback)
+        executor.add_node(self.node)
+
+        # Use process so we can terminate it if it hangs
+        p = mp.Process(target=executor.spin_once, daemon=True)
+        p.start()
+        time.sleep(1 + timer_period)
+        self.assertFalse(hung_on_shutdown)
+        p.terminate()
```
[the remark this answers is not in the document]
ihasdapie: Yup I've since changed it to just use a `Thread`

rclpy/rclpy/executors.py
```diff
@@ -79,15 +79,16 @@ def __exit__(self, t, v, tb):
...
             self._work_condition.notify_all()
 
-    def wait(self, timeout_sec=None):
+    def wait(self, timeout_sec: Optional[float] = None):
         """
         Wait until all work completes.
 
         :param timeout_sec: Seconds to wait. Block forever if None or negative. Don't wait if 0
         :type timeout_sec: float or None
+        :param: is_shutdown: if this is a shutdown call
```
jacobperron, line 88:
```suggestion
```

rclpy/test/test_executor.py
```diff
@@ -456,6 +456,25 @@ def timer_callback():
...
+        shutdown_event = threading.Event()
+
+        def timer_callback():
+            nonlocal shutdown_event, executor
+            executor.shutdown()
+            shutdown_event.set()
+
+        tmr = self.node.create_timer(timer_period, timer_callback)
+        executor.add_node(self.node)
+        t = threading.Thread(target=executor.spin_once, daemon=True)
```
jacobperron, line 473: To guarantee we get the timer callback, I would call `spin` (continuously process callbacks) instead of `spin_once` (process one callback).
```suggestion
        t = threading.Thread(target=executor.spin, daemon=True)
```

jacobperron (commented):
LGTM, a have a couple minor comments

ihasdapie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16786/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11367/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17195/)

jacobperron:
I'm not sure, but it doesn't look like Windows is happy with this change. Maybe there was a legitimate reason for waiting during shutdown afterall.

rclpy/rclpy/executors.py
```diff
@@ -190,7 +190,7 @@ def create_task(self, callback: Union[Callable, Coroutine], *args, **kwargs) ->
         # Task inherits from Future
         return task
 
-    def shutdown(self, timeout_sec: float = None) -> bool:
+    def shutdown(self, timeout_sec: float = 0) -> bool:
```
jacobperron, line 193: I would revert this change and see if that resolves the Windows CI failures.

ihasdapie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16805/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11384/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17210/)

jacobperron (approved):
LGTM

The windows warnings are unrelated (a known issue).

PR #950, opened 2022-06-04
Change sphinx theme to readthedocs

ihasdapie:
ROS2's sphinx-generated documentation is overwhelmingly in the Read the Docs theme. This PR moves `rclpy` from the current built-in alabaster theme to read the docs. RTD is also (at least in my experience) nicer to use.

### DIFF
--- a/rclpy/docs/source/conf.py
+++ b/rclpy/docs/source/conf.py
@@ -32,7 +32,7 @@
 
 
 # -- Project information -----------------------------------------------------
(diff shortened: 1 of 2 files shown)


clalancette (changes requested):

clalancette (approved):
Since we don't run this during CI, there should be no need to run CI here.  The Rpr job gives us enough confidence that you can just merge this.

PR #956, opened 2022-06-15
fix load_parameter_dict_from_yaml test failure on windows

ihasdapie:
Resolves windows test failure for `load_parameter_dict_from_yaml_file` caused by windows handling of tempfiles. (See: https://github.com/ros2/rclpy/pull/945#issuecomment-1156703563)

I've also bundled in a small fix in this PR to change the default depth of `list_parameters` to the full depth instead of 1 which to me is a saner default.

### DIFF
--- a/rclpy/rclpy/parameter_client.py
+++ b/rclpy/rclpy/parameter_client.py
@@ -22,7 +22,6 @@ from rcl_interfaces.srv import GetParameterTypes
 from rcl_interfaces.srv import ListParameters
 from rcl_interfaces.srv import SetParameters
 from rcl_interfaces.srv import SetParametersAtomically
-
 from rclpy.callback_groups import CallbackGroup
 from rclpy.node import Node
 from rclpy.parameter import Parameter as Parameter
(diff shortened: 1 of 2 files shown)


fujitatomoya:
Basically I believe that each PR should have complete fix w/o build break for maintenance. this could be problem when we do cherry-pick.

@clalancette @Blast545 what do you think?

ihasdapie:
I think that would be good too, especially since I forgot to hit squash merge on the original PR 😅. But reverting would involve rewriting history which wouldn't be good

clalancette:
> I think that would be good too, especially since I forgot to hit squash merge on the original PR sweat_smile. But reverting would involve rewriting history which wouldn't be good

Not quite.  We can do a `git revert`, which doesn't rewrite history but makes a new commit that is the opposite of the original commit.  Then we could do a new PR, including the original PR and this fix.  That maintains history, and also keeps what @fujitatomoya is asking for.

I don't feel strongly about it either way, but I would strongly prefer that by end of day we either a) get the revert in, or b) get this fix in (I notice that it is failing all CI).

ihasdapie:
@fujitatomoya I've created a revert PR at https://github.com/ros2/rclpy/pull/958 and I will submit a new PR with the new feature & all the fixes included

PR #957, opened 2022-06-15
name and type in descriptor(s) is ignored via declare_parameter(s).

fujitatomoya:
address https://github.com/ros2/rclpy/issues/879

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -349,6 +349,9 @@ class Node:
         This method, if successful, will result in any callback registered with
         :func:`add_on_set_parameters_callback` to be called.
 
+        The name and type in the given descriptor is ignored, and should be specified using
+        the name argument to this function and the default value's type instead.
+
         :param name: Fully-qualified name of the parameter, including its namespace.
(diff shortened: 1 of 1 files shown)


fujitatomoya:
@ivanpauno follow up for https://github.com/ros2/rclpy/issues/879, could you take a look?

rclpy/rclpy/node.py
```diff
@@ -349,6 +349,9 @@ def declare_parameter(
         This method, if successful, will result in any callback registered with
         :func:`add_on_set_parameters_callback` to be called.
 
+        The name and type in the given descriptor is ignored, and should be specified using
+        the name argument to this function and the default value's type instead.
```
ivanpauno, line 353: I think we need a tutorial about typed parameters without a default value and how to get things like an empty array of strings as a default value, which it's not well documented and it's a bit tricky.
But that's unrelated to this contribution.

ivanpauno (approved):
LGTM!

ivanpauno:
Rpr failures are unrelated and the changes are only in the docs, so merging!

PR #963, opened 2022-06-26
Avoid calling send() on awaiting native coroutine objects

### LINKED ISSUE #962: Runtime error
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - Binaries
- Version or commit hash:
  - galactic 1.0.8-1focal.20220209.144341
- DDS implementation:
  - Fast-RTPS
  - rclpy

#### Steps to reproduce issue

```python
import asyncio
import logging
import rclpy

logging.getLogger('asyncio').setLevel(logging.DEBUG)

async def cb(msg):
    print('Received a message:')
    await asyncio.sleep(0.1)
    print(msg.data)

async def main():
    rclpy.init()
    node = rclpy.create_node('my_node')
    node.create_subscription(String, '/test', cb, 10)
    while True:
        rclpy.spin_once(node, timeout_sec=0.001)
        await asyncio.sleep(0.001)

asyncio.run(main())
```

Run the subscriber and then run:
```
ros2 topic pub -1 /test std_msgs/msg/String "data: hello"
```

#### Expected behavior
When published to /test, the node should print `Received a message:`, wait for 0.1 seconds, and then print the message data.

From https://github.com/ros2/rclpy/pull/166#issue-285849213:

> Any callback can now be a coroutine function (async def). If the callback is a coroutine and it becomes blocked the executor will move on and resume executing it later. The details of this are handled in the class Task which is a class very similar to [asyncio.Task](https://docs.python.org/3/library/asyncio-task.html#task). asyncio can't be used directly because it is not thread safe.

#### Actual behavior
```
Executing <Task pending name='Task-1' coro=<main() running at my_listener.py:19> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f2ed81ddb80>()] created at /usr/lib/python3.8/asyncio/base_events.py:422> cb=[_run_until_complete_cb() at /usr/lib/python3.8/asyncio/base_events.py:184] created at /usr/lib/python3.8/asyncio/base_events.py:595> took 0.199 seconds
Received a message:
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/mostafa/.vscode/extensions/ms-python.python-2022.8.0/pythonFiles/lib/python/debugpy/__main__.py", line 45, in <module>
    cli.main()
  File "/home/mostafa/.vscode/extensions/ms-python.python-2022.8.0/pythonFiles/lib/python/debugpy/../debugpy/server/cli.py", line 444, in main
    run()
  File "/home/mostafa/.vscode/extensions/ms-python.python-2022.8.0/pythonFiles/lib/python/debugpy/../debugpy/server/cli.py", line 285, in run_file
    runpy.run_path(target_as_str, run_name=compat.force_str("__main__"))
  File "/usr/lib/python3.8/runpy.py", line 265, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/usr/lib/python3.8/runpy.py", line 97, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "my_listener.py", line 21, in <module>
    asyncio.run(main())
  File "/usr/lib/python3.8/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "my_listener.py", line 18, in main
    rclpy.spin_once(node, timeout_sec=0.001)
  File "/home/mostafa/dev/github/rclpy/issues/long-await/install/rclpy/lib/python3.8/site-packages/rclpy/__init__.py", line 176, in spin_once
    executor.spin_once(timeout_sec=timeout_sec)
  File "/home/mostafa/dev/github/rclpy/issues/long-await/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 712, in spin_once
    raise handler.exception()
  File "/home/mostafa/dev/github/rclpy/issues/long-await/install/rclpy/lib/python3.8/site-packages/rclpy/task.py", line 242, in __call__
    self._handler.send(None)
  File "/home/mostafa/dev/github/rclpy/issues/long-await/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 418, in handler
    await call_coroutine(entity, arg)
  File "/home/mostafa/dev/github/rclpy/issues/long-await/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 343, in _execute_subscription
    await await_or_execute(sub.callback, msg)
  File "/home/mostafa/dev/github/rclpy/issues/long-await/install/rclpy/lib/python3.8/site-packages/rclpy/executors.py", line 104, in await_or_execute
    return await callback(*args)
  File "my_listener.py", line 10, in cb
    await asyncio.sleep(0.1)
  File "/usr/lib/python3.8/asyncio/tasks.py", line 659, in sleep
    return await future
RuntimeError: await wasn't used with future
```

##### Observation

Reducing the timeout in the asyncio.sleep line makes the issue disappear. In the example on provided, on my system, a timeout of ~0.002 seconds or smaller works fine.

##### Guess

This piece might play a role:

https://github.com/ros2/rclpy/blob/c76a6a1d98869f37155798cdb4777d9e44965743/rclpy/rclpy/task.py#L234-L256

My theory is that `self._handler.send(None)` is non-blocking, causing the line `self._executing = False` to be executed before the async callback is finished (while `self.done` is still `False`). So this causes the function to be assumed as *paused* (kind of like a yielded generator), so the already-running coro object gets called again by this piece:

https://github.com/ros2/rclpy/blob/c76a6a1d98869f37155798cdb4777d9e44965743/rclpy/rclpy/executors.py#L470-L482

This causes some kind of clash, hence `RuntimeError: await wasn't used with future`.

So as a quick test, I commented out the `self._executing = False` line. This caused the second call to be avoided, (so no runtime error anymore), but the callback didn't finish (the message wasn't printed out).

m2-farzan:
Fixes #962

#### Rationale
If a native coroutine (i.e. `async def`) is to be called manually (i.e. using `u = f(); while True: u.send(None)`, instead of `asyncio.create_task`), like in the case of our `Task` class, then the `send` call should only be done when the coroutine is ready to execute code (i.e. when it has finished an `await` line). This can be examined by storing the future object returned by the `send` function.

#### Disclaimer
Please review carefully (especially if this is to be backported), as I'm not too confident about the side-effects this one might have. Thanks!

### DIFF
--- a/rclpy/rclpy/task.py
+++ b/rclpy/rclpy/task.py
@@ -212,6 +212,7 @@ class Task(Future):
             self._handler = handler(*args, **kwargs)
             self._args = None
             self._kwargs = None
+            self._future = None
         # True while the task is being executed
         self._executing = False
         # Lock acquired to prevent task from executing in parallel with itself
@@ -236,7 +237,10 @@ class Task(Future):
             if inspect.iscoroutine(self._handler):
                 # Execute a coroutine
                 try:
-                    self._handler.send(None)
+                    if self._future is None or self._future.done():
+                        if self._future and self._future.exception():
+                            raise self._future.exception()
+                        self._future = self._handler.send(None)
                 except StopIteration as e:
                     # The coroutine finished; store the result
                     self._handler.close()


rclpy/rclpy/task.py
```diff
@@ -236,7 +237,10 @@ def __call__(self):
             if inspect.iscoroutine(self._handler):
                 # Execute a coroutine
                 try:
-                    self._handler.send(None)
+                    if self._future is None or self._future.done():
+                        if self._future and self._future.exception():
+                            raise self._future.exception()
+                        self._future = self._handler.send(None)
```
sloretz, line 243: I don't think this can assume a coroutine will return a future. It seems like the behavior is specific to the async library being used (see https://github.com/ros2/rclpy/issues/962#issuecomment-1179457414). This change is interesting though because it seems like it would both satisfy an `rclpy.Future` by calling `send(None)` until it raised `StopIteration`, and would only call `asyncio.Future.__await__` once.

It still needs a way to wake the executor when an asyncio future completes, and given that `asyncio` isn't thread safe I bet there would be problems if a multithreaded executor was used.

sloretz:
I'll close this one in favor of #971 for now. It can be reopened if this ends up looking like a better way forward.

Issue #968, opened 2022-07-16
Race condition between cleanup in executor shutdown and _wait_for_ready_callbacks.


sloretz:
## Bug report

**Required Info:**

- Operating System:
  - Jammy
- Installation type:
  - source
- Version or commit hash:
  - Latest https://github.com/ros2/rclpy/commit/649bdec92e74eee6ebb45bd2815df9ea9ff48654
- DDS implementation:
  - n/a
- Client library (if applicable):
  - rclpy
 
#### Steps to reproduce issue
1. Add a print statement or some other slow operation between [these lines](https://github.com/ros2/rclpy/blob/649bdec92e74eee6ebb45bd2815df9ea9ff48654/rclpy/rclpy/executors.py#L475-L517)
2. `colcon test --event-handlers console_direct+ --packages-select rclpy --ctest-args -R test_executor`

#### Expected behavior

All tests would pass

#### Actual behavior

`test_executor_immediate_shutdown` fails.

(line numbers don't add up because of local changes, but I'm pretty sure this issue exists on `rolling`)

```
14: E               Traceback (most recent call last):
14: E                 File "/usr/lib/python3.10/threading.py", line 1009, in _bootstrap_inner
14: E                   self.run()
14: E                 File "/usr/lib/python3.10/threading.py", line 946, in run
14: E                   self._target(*self._args, **self._kwargs)
14: E                 File "/home/osrf/ws/ros2/src/ros2/rclpy/rclpy/rclpy/executors.py", line 303, in spin
14: E                   self.spin_once()
14: E                 File "/home/osrf/ws/ros2/src/ros2/rclpy/rclpy/rclpy/executors.py", line 743, in spin_once
14: E                   handler, entity, node = self.wait_for_ready_callbacks(timeout_sec=timeout_sec)
14: E                 File "/home/osrf/ws/ros2/src/ros2/rclpy/rclpy/rclpy/executors.py", line 729, in wait_for_ready_callbacks
14: E                   return next(self._cb_iter)
14: E                 File "/home/osrf/ws/ros2/src/ros2/rclpy/rclpy/rclpy/executors.py", line 587, in _wait_for_ready_callbacks
14: E                   context_stack.enter_context(gc.handle)
14: E               AttributeError: 'NoneType' object has no attribute 'handle'
```

#### Additional information

Shutdown waits for all work tracked by the work tracker to finish

[...]

PR #975, opened 2022-07-22
Fix race condition between cleanup in executor shutdown and _wait_for…

llapx:
(Edit by @fujitatomoya )

…_ready_callbacks.

Signed-off-by: Lei Liu <Lei.Liu.AP@sony.com>

address https://github.com/ros2/rclpy/issues/968

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -562,6 +562,11 @@ class Executor:
                         guard_handles.append(gc.handle)
                     except InvalidHandle:
                         entity_count.num_guard_conditions -= 1
+                    except AttributeError:
+                        if self._is_shutdown:
+                            raise ShutdownException()
+                        else:
+                            raise ExternalShutdownException()
 
                 for waitable in waitables:
                     try:


llapx:
The https://build.ros2.org/job/Rpr__rclpy__ubuntu_jammy_amd64/147/testReport/junit/rclpy.rclpy.test/test_timer/test_zero_callback_0_001_/ is not related to the fix, can anybody fix it ?

rclpy/rclpy/executors.py
```diff
@@ -562,6 +562,11 @@ def _wait_for_ready_callbacks(
                         guard_handles.append(gc.handle)
                     except InvalidHandle:
                         entity_count.num_guard_conditions -= 1
+                    except AttributeError:
```
Barry-Xu-2018, line 565: From my opinions, I don't think it should throw an exception.     
#968 is possible to occur. So code should be updated to handle this case to avoid exception during shutdown.

fujitatomoya: right, this does not address root cause which is racy condition for `guards`.

llapx: this PR is under discussing, see https://github.com/ros2/rclpy/issues/968#issuecomment-1196160697 for detail.

fujitatomoya: I actually think this patch makes sense now.

I was considering the following.

```patch
diff --git a/rclpy/rclpy/executors.py b/rclpy/rclpy/executors.py
index 3c5bea4..73b2ba5 100644
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -514,8 +514,13 @@ class Executor:
             if timeout_timer is not None:
                 timers.append(timeout_timer)

-            guards.append(self._guard)
-            guards.append(self._sigint_gc)
+            with self._shutdown_lock:
+                if self._guard and self._sigint_gc:
+                    # make sure to append valid GuardConditions
+                    guards.append(self._guard)
+                    guards.append(self._sigint_gc)
+                else:
+                    raise ShutdownException()

             entity_count = NumberOfEntities(
                 len(subscriptions), len(guards), len(timers), len(clients), len(services))
```

but this does not guarantee the valid GuradConditions here,

https://github.com/ros2/rclpy/blob/649bdec92e74eee6ebb45bd2815df9ea9ff48654/rclpy/rclpy/executors.py#L561-L562

rclpy/rclpy/executors.py
```diff
@@ -562,6 +562,11 @@ def _wait_for_ready_callbacks(
...
+                        if self._is_shutdown:
+                            raise ShutdownException()
+                        else:
+                            raise ExternalShutdownException()
```
fujitatomoya, line 569: I think `SingleThreadedExecutor::spin_once` needs to catch this exception.

rclpy/rclpy/executors.py
Barry-Xu-2018: We think current design is not good.  
Only catching the exception isn't a good solution.   

We want to  
- Executor.shutdown() only set _is_shutdown and notify executor by guard trigger.
- Real shutdown process is moved to Executor.wait_for_ready_callbacks().  

This thought need to be discussed.  A comment is added to https://github.com/ros2/rclpy/issues/968#issuecomment-1196160697

fujitatomoya:
@llapx why are we closing this PR?

llapx:
@fujitatomoya 

Oh, sorry, I have refork the `rclpy` for another work (the old one not contain with rolling branch).

Issue #969, opened 2022-07-18
Action server/client should not use node logger


tonynajjar:
#### Feature description

When using the DEBUG logging level for my Python nodes with action servers/clients, I get spammed with [debug messages](https://github.com/ros2/rclpy/blob/c3a679666cce40379dadd96f9b8b4c7c237e0b5f/rclpy/rclpy/action/server.py#L272) from the action server. That's because the action server/client uses the node's logger.

In rclcpp, [another logger is created](https://github.com/ros2/rclcpp/blob/b953bdddf8de213b4a051ecf2d668dad65ff9f89/rclcpp_action/src/server.cpp#L100).

Could we make use of another logger in rclpy similarly to rclcpp?

I could work on that myself. Thanks

tonynajjar:
See https://github.com/ros2/rclpy/compare/galactic...pixel-robotics:rclpy:galactic as POC

fujitatomoya:
@tonynajjar I think this makes sense, are you willing to create PR to address this?

tonynajjar:
Done https://github.com/ros2/rclpy/pull/982

PR #982, opened 2022-07-27
Create sublogger for action server and action client

tonynajjar:
Addresses https://github.com/ros2/rclpy/issues/969

### DIFF
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -22,6 +22,7 @@ from action_msgs.srv import CancelGoal
 
 from rclpy.executors import await_or_execute
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+from rclpy.logging import get_logger
 from rclpy.qos import qos_profile_action_status_default
 from rclpy.qos import qos_profile_services_default
 from rclpy.qos import QoSProfile
@@ -180,6 +181,7 @@ class ActionClient(Waitable):
 
         callback_group.add_entity(self)
         self._node.add_waitable(self)
+        self._logger = get_logger("rclpy_action")
 
     def _generate_random_uuid(self):
         return UUID(uuid=list(uuid.uuid4().bytes))
@@ -300,7 +302,7 @@ class ActionClient(Waitable):
 
                 self._pending_goal_requests[sequence_number].set_result(goal_handle)
             else:
-                self._node.get_logger().warning(
+                self._logger.warning(
                     'Ignoring unexpected goal response. There may be more than '
                     f"one action server for the action '{self._action_name}'"
                 )
@@ -310,7 +312,7 @@ class ActionClient(Waitable):
             if sequence_number in self._pending_cancel_requests:
                 self._pending_cancel_requests[sequence_number].set_result(cancel_response)
             else:
-                self._node.get_logger().warning(
+                self._logger.warning(
                     'Ignoring unexpected cancel response. There may be more than '
                     f"one action server for the action '{self._action_name}'"
                 )
@@ -320,7 +322,7 @@ class ActionClient(Waitable):
             if sequence_number in self._pending_result_requests:
                 self._pending_result_requests[sequence_number].set_result(result_response)
             else:
-                self._node.get_logger().warning(
+                self._logger.warning(
                     'Ignoring unexpected result response. There may be more than '
                     f"one action server for the action '{self._action_name}'"
                 )
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -181,7 +181,7 @@ class ActionClient(Waitable):
 
         callback_group.add_entity(self)
         self._node.add_waitable(self)
-        self._logger = get_logger("rclpy_action")
+        self._logger = self._node.get_logger().get_child('action_client')
 
     def _generate_random_uuid(self):
         return UUID(uuid=list(uuid.uuid4().bytes))
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -22,7 +22,6 @@ from action_msgs.srv import CancelGoal
 
 from rclpy.executors import await_or_execute
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
-from rclpy.logging import get_logger
 from rclpy.qos import qos_profile_action_status_default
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):
lgtm with DCO fix.

fujitatomoya:
@sloretz @ivanpauno either of you could do another review on this?

tonynajjar:
> @sloretz @ivanpauno either of you could do another review on this?

Please do, it was a quick fix for us, I did not test this extensively or gave it much more thought

rclpy/rclpy/action/client.py
```diff
@@ -180,6 +181,7 @@ def __init__(
 
         callback_group.add_entity(self)
         self._node.add_waitable(self)
+        self._logger = get_logger("rclpy_action")
```
ivanpauno, line 184:
```suggestion
        self._logger = self.node.get_logger().get_child('action_client')
```

rclpy/rclpy/action/server.py
```diff
@@ -262,14 +263,15 @@ def __init__(
 
         callback_group.add_entity(self)
         self._node.add_waitable(self)
+        self._logger = get_logger("rclpy_action")
```
ivanpauno, line 266: same

rclpy/rclpy/action/client.py
ivanpauno: I think it makes sense to create a logger in the same namespace than the node logger, but not an independent one

tonynajjar: I see, but that makes my use case more difficult because now each logger has a different name. I need them to have the same name to filter them out more easily.

I'm able to filter out unwanted DEBUG logs like this:

```
log_options = [
    "--ros-args",
    "--log-level",
   "debug",
    "--log-level",
    "rmw_cyclonedds_cpp:=info",
    "--log-level",
    "rcl_action:=info",
    "--log-level",
    "rclpy_action:=info",
],

Node(
    package="some_package",
    executable="some_executable",
    arguments=log_options
),
```

With the change you propose, I'd have do replace `rclpy_action` with `namespace.some_executable.action_client`. Then each Node has a different argument

tonynajjar: In case you are wondering why I'm setting all the loggers to debug and then selectively setting some to info, instead of:

```
Node(
    package="some_package",
    executable="some_executable",
    arguments=[
    "--ros-args",
    "--log-level",
   "namespace.some_executable:=debug",
    ]
),
```

That's because:

1- To avoid setting different argument for each node
2- The namespace is a variable LaunchConfiguration (I guess parsing it is still possible)
3- I don't necessarily know all the name of all the loggers in the node

tonynajjar: Lastly, your proposition diverges from the rclcpp implementation [where an independent logger is used](https://github.com/ros2/rclcpp/blob/b953bdddf8de213b4a051ecf2d668dad65ff9f89/rclcpp_action/src/server.cpp#L100)

sloretz: > I see, but that makes my use case more difficult because now each logger has a different name. I need them to have the same name to filter them out more easily.

What if I only want to view the logs of one action server?

Which log messages are you hoping to see in your use case?

All entities seem to use the node's logger at at the moment ([I was surprised by this one](https://github.com/ros2/rclpy/blob/7685ee21a21fcb682ec42785544e03d05f83e851/rclpy/src/rclpy/subscription.cpp#L152-L159)). I'd lean towards nesting them further so that they become individually fiterable.

sloretz: > Lastly, your proposition diverges from the rclcpp implementation [where an independent logger is used](https://github.com/ros2/rclcpp/blob/b953bdddf8de213b4a051ecf2d668dad65ff9f89/rclcpp_action/src/server.cpp#L100)

That's an error message in a unique pointer deleter that doesn't have access to the logger. [rclcpp uses the node's logger name plus `rclcpp_action`](https://github.com/ros2/rclcpp/blob/b953bdddf8de213b4a051ecf2d668dad65ff9f89/rclcpp_action/src/server.cpp#L91) for [most log messages](https://github.com/ros2/rclcpp/blob/b953bdddf8de213b4a051ecf2d668dad65ff9f89/rclcpp_action/src/server.cpp#L355).

tonynajjar: Okay I understand better now. Having it as a child logger makes generally more sense. I'll find another workaround for my use case (or write another issue to support glob for selecting loggers)

rclpy/rclpy/action/client.py
```diff
@@ -22,6 +22,7 @@
 
 from rclpy.executors import await_or_execute
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+from rclpy.logging import get_logger
```
ivanpauno, line 25: I don't think you need this:

```suggestion
```

rclpy/rclpy/action/server.py
```diff
@@ -21,6 +21,7 @@
 
 from rclpy.executors import await_or_execute
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+from rclpy.logging import get_logger
```
ivanpauno, line 24: same

ivanpauno (approved):
lgtm

rclpy/rclpy/action/client.py
tonynajjar: fixed thanks

ivanpauno (approved):
LGTM!

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17041/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11601/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17466/)

tonynajjar:
To get this in galactic and humble what do I need to do? Create other PRs targetting those branches?

hello-amal:
Could we get a Humble backport of this? 

Humble has a bug where in certain error cases, it tries to access `self._logger`, but `self._logger` doesn't exist (e.g., [here](https://github.com/ros2/rclpy/blob/091f7eb64af8a5e52db4c8fbc6818499ec563a0a/rclpy/rclpy/action/server.py#L314)). This PR would address that.

Issue #983, opened 2022-08-03
Exception never retrieved when using MultiThreadedExecutor and Timers


mr337:
## Bug report

**Required Info:**

- Operating System:
Ubuntu 20.04.4
- Installation type:
binaries
- Version or commit hash:
foxy
- DDS implementation:
default
- Client library (if applicable):
rclpy

#### Steps to reproduce issue
I am encountering an issue with exceptions being thrown in timer callbacks using `MultiThreadedExecutor` going silent.

The sample shows that when a an exception is generated in a timer callback using the `SingleThreadExecutor`, implicitly I believe, the exception rises if uncaught, as expected.

When using the `MultiThreadedExecutor` the exception gets lost and silently fails, sometimes causing head scratching.
```
#!/usr/bin/env python3
import rclpy
from rclpy.executors import MultiThreadedExecutor
from rclpy.node import Node


class TestNode(Node):
    def __init__(self):
        super().__init__('zf9p')
        self.tmr = self.create_timer(1, self.tmr_callback)

    def tmr_callback(self):
        1/0
        print('Timer callback fired')


def single_executor():
    test_node = TestNode()

    while rclpy.ok() is True:
        rclpy.spin(test_node)

    test_node.destroy_timer(test_node.tmr)


def multi_executor():
    test_node = TestNode()

    # Moved to multi threaded executor so none of Actions block publishing of msgs
    executor = MultiThreadedExecutor(num_threads=1)
    executor.add_node(test_node)

    while rclpy.ok() is True:
        executor.spin()

    test_node.destroy_timer(test_node.tmr)


if __name__ == '__main__':
    rclpy.init()

    # single_executor()  # un comment to verify proper operation 
    multi_executor()

    rclpy.shutdown()
```

#### Expected behavior
Unhandled exception to rise to the top
```
$ python3 timer_example
Traceback (most recent call last):
  ...
  File "/opt/ros/foxy/lib/python3.8/site-packages/rclpy/executors.py", line 343, in _execute_timer
    await await_or_execute(tmr.callback)
  File "/opt/ros/foxy/lib/python3.8/site-packages/rclpy/executors.py", line 118, in await_or_execute
    return callback(*args)
  File "timer_example", line 13, in tmr_callback
    1/0
ZeroDivisionError: division by zero
```

#### Actual behavior
Exception seems to silently fail not handled, but after 10 or 30 seconds the following message does show.
```
$ python3 timer_example
The following exception was never retrieved: division by zero
[...]

PR #984, opened 2022-08-04
Raise user handler exception in MultiThreadedExecutor.

fujitatomoya:
addresses https://github.com/ros2/rclpy/issues/983

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -33,6 +33,7 @@ from typing import TypeVar
 from typing import Union
 
 
+import rclpy
 from rclpy.client import Client
 from rclpy.clock import Clock
 from rclpy.clock import ClockType
@@ -40,6 +41,7 @@ from rclpy.context import Context
 from rclpy.exceptions import InvalidHandle
 from rclpy.guard_condition import GuardCondition
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
+from rclpy.impl.logging_severity import LoggingSeverity
 from rclpy.service import Service
 from rclpy.signals import SignalHandlerGuardCondition
 from rclpy.subscription import Subscription
@@ -720,7 +722,11 @@ class SingleThreadedExecutor(Executor):
         else:
             handler()
             if handler.exception() is not None:
-                raise handler.exception()
+                rclpy.logging._root_logger.log(
+                    'Exception: ' + str(handler.exception()),
+                    LoggingSeverity.WARN,
+                    name='SingleThreadedExecutor',
+                )
 
     def spin_once_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
         self.spin_once(timeout_sec)
@@ -765,7 +771,11 @@ class MultiThreadedExecutor(Executor):
             def handler_wrapper(handler):
                 handler()
                 if handler.exception() is not None:
-                    raise handler.exception()
+                    rclpy.logging._root_logger.log(
+                        'Exception: ' + str(handler.exception()),
+                        LoggingSeverity.WARN,
+                        name='MultiThreadedExecutor',
+                    )
             self._executor.submit(handler_wrapper(handler))
 
     def spin_once(self, timeout_sec: float = None) -> None:
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -33,7 +33,6 @@ from typing import TypeVar
 from typing import Union
 
 
-import rclpy
 from rclpy.client import Client
 from rclpy.clock import Clock
 from rclpy.clock import ClockType
@@ -41,7 +40,6 @@ from rclpy.context import Context
 from rclpy.exceptions import InvalidHandle
 from rclpy.guard_condition import GuardCondition
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
-from rclpy.impl.logging_severity import LoggingSeverity
 from rclpy.service import Service
 from rclpy.signals import SignalHandlerGuardCondition
 from rclpy.subscription import Subscription
@@ -722,11 +720,7 @@ class SingleThreadedExecutor(Executor):
         else:
             handler()
             if handler.exception() is not None:
-                rclpy.logging._root_logger.log(
-                    'Exception: ' + str(handler.exception()),
-                    LoggingSeverity.WARN,
-                    name='SingleThreadedExecutor',
-                )
+                raise handler.exception()
 
     def spin_once_until_future_complete(self, future: Future, timeout_sec: float = None) -> None:
         self.spin_once(timeout_sec)
@@ -771,11 +765,7 @@ class MultiThreadedExecutor(Executor):
             def handler_wrapper(handler):
                 handler()
                 if handler.exception() is not None:
-                    rclpy.logging._root_logger.log(
-                        'Exception: ' + str(handler.exception()),
-                        LoggingSeverity.WARN,
-                        name='MultiThreadedExecutor',
-                    )
+                    raise handler.exception()
             self._executor.submit(handler_wrapper(handler))
 
     def spin_once(self, timeout_sec: float = None) -> None:
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -308,6 +308,7 @@ class Executor:
 
         :param timeout_sec: Seconds to wait. Block forever if ``None`` or negative.
             Don't wait if 0.
+        :raise RuntimeError: Unexpected failure.
         """
         raise NotImplementedError()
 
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/executors.py
```diff
@@ -762,7 +762,11 @@ def _spin_once_impl(
         except ConditionReachedException:
             pass
         else:
-            self._executor.submit(handler)
+            def handler_wrapper(handler):
```
fujitatomoya, line 765: create object to raise the exception if there is, instead of managing the `future.result()` via `submit`.

fujitatomoya:
@sloretz @ivanpauno could you take a look?

rclpy/rclpy/executors.py
```diff
@@ -762,7 +762,11 @@ def _spin_once_impl(
...
+                handler()
+                if handler.exception() is not None:
+                    raise handler.exception()
```
ivanpauno, line 768: I'm not sure if it's a great idea to throw an exception from the executor when any of the callbacks fail.
It would make everything fail when only one callback failed.

Maybe something else like logging an error is a good idea, I'm not sure.

mr337: That does bring up a good point. The main issue I had in https://github.com/ros2/rclpy/issues/983 is silently failing, and the message that no retrieval of exception could be very delayed leading to head scratching.

If we decide to not to throw exception from the executor, would logging it out be better, assuming that would be immediate? As a preventative measure all the timer call backs in our code are wrapped in a `try except` block but was a odd experience figuring that out from a user perspective.

ivanpauno: > If we decide to not to throw exception from the executor, would logging it out be better, assuming that would be immediate? 

Yes, logging is better than nothing.
And if you log from here that should be immediate.

sloretz: +1 to logging. Logging and ignoring it has some precedent [in concurrent.futures](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.add_done_callback).

An enhancement might be to allow the user to set a custom exception handler, as [`asyncio` seems to allow people to do](https://docs.python.org/3/library/asyncio-eventloop.html#error-handling-api).

fujitatomoya: addressed in https://github.com/ros2/rclpy/pull/984/commits/cda6aeeccb78777101235fafcd4ca6a14acc70d9

rclpy/rclpy/executors.py
```diff
@@ -720,7 +722,11 @@ def spin_once(self, timeout_sec: float = None) -> None:
         else:
             handler()
             if handler.exception() is not None:
-                raise handler.exception()
+                rclpy.logging._root_logger.log(
+                    'Exception: ' + str(handler.exception()),
+                    LoggingSeverity.WARN,
+                    name='SingleThreadedExecutor',
+                )
```
fujitatomoya, line 729: I think we should provide the same experience for `SingleThreadedExecutor`.

fujitatomoya:
the followings are the output with https://github.com/ros2/rclpy/issues/983

- SingleThreadedExecutor

```bash
# ros2 run prover_rclpy rclpy_983
[WARN] [1660060486.243490700] [SingleThreadedExecutor]: Exception: division by zero
[WARN] [1660060487.218409974] [SingleThreadedExecutor]: Exception: division by zero
[WARN] [1660060488.218467063] [SingleThreadedExecutor]: Exception: division by zero
[WARN] [1660060489.218468133] [SingleThreadedExecutor]: Exception: division by zero
[WARN] [1660060490.218465565] [SingleThreadedExecutor]: Exception: division by zero
...<snip>
```

- MultiThreadedExecutor

```bash
# ros2 run prover_rclpy rclpy_983
[WARN] [1660060398.085604463] [MultiThreadedExecutor]: Exception: division by zero
[WARN] [1660060399.065598330] [MultiThreadedExecutor]: Exception: division by zero
[WARN] [1660060400.065630959] [MultiThreadedExecutor]: Exception: division by zero
[WARN] [1660060401.065543777] [MultiThreadedExecutor]: Exception: division by zero
[WARN] [1660060402.065565788] [MultiThreadedExecutor]: Exception: division by zero
<...snip>
```

rclpy/rclpy/executors.py
ivanpauno: makes sense to me

ivanpauno (approved):
LGTM!

Thanks for the contribution

iuhilnehc-ynos:
I don't like this fix that only uses a warning log. What if users want to exit if any callback got an exception?
It's totally the users' responsibility to catch the exceptions either in the callback or calling spin.

fujitatomoya:
@ivanpauno @sloretz 

i am just concerned that what if user application expects the exception from user callback? there would be the case that user application intentionally generates the exception and catch that in user code. with this fix, it cannot be supported.
yet, i think this PR works for some cases, to keep the executors running with warnings.

just want to hear opinion about that before we push this fix further.

CC: @clalancette @wjwwood

ivanpauno:
> i am just concerned that what if user application expects the exception from user callback? there would be the case that user application intentionally generates the exception and catch that in user code. with this fix, it cannot be supported.

I would say it's uncommon to throw an exception from a callback and catch it from `Executor.spin()`.
It's not a pattern that I have seen and I wouldn't encourage it, but it was possible for the `SingleThreadedExecutor`.

So it's true that in that case it can break pre-existing code.
If we do the opposite (throwing an exception by default), we could break the `MultiThreadedExecutor`, which by default did nothing.

I think consistency is important, so I would do the same in both.
I don't mind the default, but logging may be fine (what does rclcpp do?).
Having a customizable exception handler like @sloretz sounds like a good idea as well (which by default, would only log).

fujitatomoya:
> what does rclcpp do?

good question, it just raises exception from user callback. (both `SingleThreadedExecutor` and `MultiThreadedExecutor`)

> I would say it's uncommon to throw an exception from a callback and catch it from Executor.spin().

agree.

i think we need to do,
- run CI for this PR. (will do it right now.)
- apply the same patch for `rclcpp` to log the exception message once executor
- consider more enhancement like customizable exception handler

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17113/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11661/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17530/)

wjwwood:
In my opinion, the executors (applies to both c++ and Python), exceptions should not be caught by default. If there is an exception that the user did not catch themselves, it should go to the unhandled exception handler (which aborts the application). This matches what happens in plain python and c++. 

I think that assuming that "when an exception occurs in a callback, it is correct to catch it and continue execution" is not a good assumption and isn't safe either. 

Instead we should allow users to set their own executor level unhandled exception handler. This would be called when an exception is thrown in the callbacks called by the executor.

For situations like component mode containers we could install a custom unhandled exception handler that logs the exception and then unloads the node. This avoids taking down the whole application when an unhandled exception occurs in a single node. 

There can also be exceptions inside the executor that aren't part of the callback being called necessarily. But I think those cases can be pushed into two buckets: things that should be caught by wrapping the spin call, and providing callbacks in the entities when errors specific to them occur, e.g. a timeout occurs when waiting for a service response times out. 

So I wouldn't do this pr as is. I think both the multi-threaded and single-threaded executors should exit when an exception occurs in the user callback. 

I'm not sure if we should reroute the exception to the thread that called spin in the multi threaded case. But it should take the application down for sure. 

An unhandled exception in a thread is something that is different for python and c++. I need to research that more but I don't think it affects my previous statement.

wjwwood (changes requested):
See my previous comment.

fujitatomoya:
appreciate informative comments, folks! after all, i am inclined to have the default behavior to be `raise exception`.

- "when an exception occurs in a callback, it is correct to catch it and continue execution" is not really good assumption.
- user application still can manage the exception raised from executor. (means that , it can be more flexible in this way.)

(note) probably we can add doc description for each executor of `rclcpp` and `rclpy`.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17124/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11671/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17541/)

fujitatomoya:
@wjwwood can you take a look?

ivanpauno (approved):
LGTM!

ivanpauno:
> user application still can manage the exception raised from executor. (means that , it can be more flexible in this way.)

It would be nice to have an unhandled exception handler in `Executor`, as @sloretz and @wjwwood mentioned, which is not the same as that.

The part about raising an exception from the executor instead of logging and ignoring the exception seems like a safer default behavior, so LGTM!

ivanpauno:
@fujitatomoya please check the DCO before merging

fujitatomoya:
@ivanpauno thanks, i missed the DCO on revert commit, it's been addressed.

wjwwood (approved):
Lgtm, thanks for iterating

fujitatomoya:
@Mergifyio backport humble galactic foxy

PR #986, opened 2022-08-08
Use Python3FindABI, avoid using python_cmake_module

ivanpauno:
This will require updating rep2000 ...

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.16)
 
 project(rclpy)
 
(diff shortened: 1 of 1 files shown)


ivanpauno:
Debug:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17098/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11647/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17515/)

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17099/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11648/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17516/)

ivanpauno:
I have to check the docs better :man_facepalming:  ....

> Note This hint is useful only on POSIX systems. So, on Windows systems, when Python3_FIND_ABI is defined, Python distributions from [python.org](https://www.python.org/) will be found only if value for each flag is OFF or ANY.

It would still be nice to be able to replace `python_cmake_module` with Python3, which should be possible.

clalancette:
> It would still be nice to be able to replace `python_cmake_module` with Python3, which should be possible.

I agree that it should be possible, but I think we are going to have to use a different approach than in this PR.  Given that and the age of this, I'm going to close this one out for now.

PR #993, opened 2022-08-19
fix timers in task

### LINKED ISSUE #985: Timer not executing when created from within a task
## Bug report

When using `spin_until_future_complete` with a task that `await` on a future set in a timer, the future is never completed. Interestingly, this only happens when setting a timeout on `spin_until_future_complete`, if a timeout is not provided, or if `None` is provided, then the task completes successfully.

**Required Info:**

- Operating System:
  - ubuntu22.04
- Installation type:
  - Binary
- Version or commit hash:
  - 3.3.4-1jammy.20220620.181239
- DDS implementation:
  - default
  - rclpy

#### Steps to reproduce issue
```python
import rclpy
import rclpy.node
import rclpy.executors

rclpy.init()
node = rclpy.node.Node("test_node")
executor = rclpy.executors.SingleThreadedExecutor()
executor.add_node(node)

fut = rclpy.Future()


async def work():
    node.create_timer(0.1, lambda: fut.set_result(True))
    await fut


task = executor.create_task(work())
executor.spin_until_future_complete(task, 1)  # works when timeout is None
print(task.done())
rclpy.shutdown()
```

#### Expected behavior
Future is completed

#### Actual behavior
Future is not completed

#### Additional information
Looking at the code of `spin_until_future_complete`, it behaves differently depending on if a timeout is provided.
```python
        if timeout_sec is None or timeout_sec < 0:
            while self._context.ok() and not future.done() and not self._is_shutdown:
                self.spin_once_until_future_complete(future, timeout_sec)
        else:
            start = time.monotonic()
            end = start + timeout_sec
            timeout_left = timeout_sec

            while self._context.ok() and not future.done() and not self._is_shutdown:
                self.spin_once_until_future_complete(future, timeout_left)
                now = time.monotonic()

                if now >= end:
                    return

                timeout_left = end - now
```

This all looks good but furthur investigation of `wait_for_ready_callbacks` (which is eventually called later down the chain),
```python
        while True:
            if self._cb_iter is None or self._last_args != args or self._last_kwargs != kwargs:
                # Create a new generator
                self._last_args = args
                self._last_kwargs = kwargs
                self._cb_iter = self._wait_for_ready_callbacks(*args, **kwargs)

            try:
                return next(self._cb_iter)
            except StopIteration:
                # Generator ran out of work
                self._cb_iter = None
```

It creates a new generator whenever `_last_args` or `_last_kwargs` is different, the "timeout" path of `spin_until_future_complete` does exactly that, each `spin_once_until_future_complete` is passed a different timeout.

My guess is that somehow creating new generators every "tick" causes new "work" to be created, and executing a "work" causes a new generator to be created, resulting in a infinite list of pending "work", which causes the actual task to never be executed.

I also notice that when adding a `print` statement to `Future.__await__()`, the same future is being yielded endlessly very quickly.

```python
    def __await__(self):
        # Yield if the task is not finished
        while not self._done:
            print(id(self), "yielding")
            yield
        return self.result()
```

output
```
139875185274560 yielding
139875185274560 yielding
139875185274560 yielding
139875185274560 yielding
139875185274560 yielding
139875185274560 yielding
139875185274560 yielding
... thousands of lines
```

koonpeng:
Fixes #985.

I am not familiar with the executor logic in rclpy, so this is a hacky fix I did. Instead of passing a new timeout each tick, I added `self._spin_until_future_complete_timer` and `self._spin_until_future_complete_timeout` to wake the executor when the timeout is reached. Also added `test_timer_in_task.py` to test the new behavior.

---

Looking at the code of `spin_until_future_complete`, it behaves differently depending on if a timeout is provided.
```python
        if timeout_sec is None or timeout_sec < 0:
            while self._context.ok() and not future.done() and not self._is_shutdown:
                self.spin_once_until_future_complete(future, timeout_sec)
        else:
            start = time.monotonic()
            end = start + timeout_sec
            timeout_left = timeout_sec

            while self._context.ok() and not future.done() and not self._is_shutdown:
                self.spin_once_until_future_complete(future, timeout_left)
                now = time.monotonic()

                if now >= end:
                    return

                timeout_left = end - now
```

This all looks good but furthur investigation of `wait_for_ready_callbacks` (which is eventually called later down the chain),
```python
        while True:
            if self._cb_iter is None or self._last_args != args or self._last_kwargs != kwargs:
                # Create a new generator
                self._last_args = args
                self._last_kwargs = kwargs
                self._cb_iter = self._wait_for_ready_callbacks(*args, **kwargs)

            try:
                return next(self._cb_iter)
            except StopIteration:
                # Generator ran out of work
                self._cb_iter = None
```

It creates a new generator whenever `_last_args` or `_last_kwargs` is different, the "timeout" path of `spin_until_future_complete` does exactly that, each `spin_once_until_future_complete` is passed a different timeout.

My guess is that somehow creating new generators every "tick" causes new "work" to be created, and executing a "work" causes a new generator to be created, resulting in a infinite list of pending "work", which causes the actual task to never be executed.

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -186,6 +186,7 @@ if(BUILD_TESTING)
       test/test_time_source.py
       test/test_time.py
       test/test_timer.py
+      test/test_timer_in_task.py
       test/test_topic_or_service_is_hidden.py
       test/test_topic_endpoint_info.py
       test/test_type_support.py
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -19,7 +19,6 @@ import multiprocessing
 from threading import Condition
 from threading import Lock
 from threading import RLock
-import time
 from typing import Any
 from typing import Callable
 from typing import Coroutine
@@ -169,6 +168,8 @@ class Executor:
         self._clock = Clock(clock_type=ClockType.STEADY_TIME)
         self._sigint_gc = SignalHandlerGuardCondition(context)
         self._context.on_shutdown(self.wake)
+        self._spin_until_future_complete_timer: Optional[Timer] = None
+        self._spin_until_future_complete_timeout = False
 
     @property
     def context(self) -> Context:
@@ -287,18 +288,16 @@ class Executor:
             while self._context.ok() and not future.done() and not self._is_shutdown:
                 self.spin_once_until_future_complete(future, timeout_sec)
         else:
-            start = time.monotonic()
-            end = start + timeout_sec
-            timeout_left = timeout_sec
-
-            while self._context.ok() and not future.done() and not self._is_shutdown:
-                self.spin_once_until_future_complete(future, timeout_left)
-                now = time.monotonic()
-
-                if now >= end:
-                    return
-
-                timeout_left = end - now
+            if self._spin_until_future_complete_timer is not None:
+                # this should not happen
+                raise RuntimeError('Executor already spinning')
+            self._spin_until_future_complete_timer = Timer(
+                None, None, timeout_sec_to_nsec(timeout_sec),
+                self._clock, context=self._context)
+            self._spin_until_future_complete_timeout = False
+            while not future.done() and not self._spin_until_future_complete_timeout:
+                self.spin_once()
+            self._spin_until_future_complete_timeout = None
 
     def spin_once(self, timeout_sec: float = None) -> None:
         """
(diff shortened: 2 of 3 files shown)


fujitatomoya (commented):
adding timer based on timeout argument looks okay to avoid to keep creating generator, but i cannot really say what approach is reasonable.

@sloretz @ivanpauno could you share your thoughts?

ivanpauno (changes requested):
I think there's something weird somewhere else, this doesn't look like the right fix.

Based on your comments in #985, I think we should be fixing what `wait_for_ready_callbacks()`/`_wait_for_ready_callbacks()` are doing.
I haven't exactly figured out what changes are needed, I will try to look again soon.

Issue #1016, opened 2022-09-30
Synchronous service client call broken in timer callback


molasses11:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binaries
- Version or commit hash:
  - foxy (the 2022-09-28 release that was just pushed through the apt repos)
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclpy (of course)

#### Steps to reproduce issue

Use the minimal service server from the examples
``` python
from example_interfaces.srv import AddTwoInts

import rclpy
from rclpy.node import Node


class MinimalService(Node):

    def __init__(self):
        super().__init__('minimal_service')
        self.srv = self.create_service(AddTwoInts, 'add_two_ints', self.add_two_ints_callback)

    def add_two_ints_callback(self, request, response):
        response.sum = request.a + request.b
        self.get_logger().info('Incoming request\na: %d b: %d' % (request.a, request.b))

        return response


def main(args=None):
    rclpy.init(args=args)

    minimal_service = MinimalService()

    rclpy.spin(minimal_service)

    rclpy.shutdown()


if __name__ == '__main__':
    main()
```
and a slightly modified service client:
``` python
from example_interfaces.srv import AddTwoInts
import rclpy
from rclpy.node import Node

from rclpy.executors import MultiThreadedExecutor
from rclpy.callback_groups import MutuallyExclusiveCallbackGroup


class MinimalClient(Node):
    def __init__(self):
        super().__init__("minimal_client_async")
        self.cb = None
        self.timer_cb = MutuallyExclusiveCallbackGroup()
        self.cli = self.create_client(
            AddTwoInts, "add_two_ints", callback_group=self.cb
        )
        while not self.cli.wait_for_service(timeout_sec=1.0):
            self.get_logger().info("service not available, waiting again...")
        self.req = AddTwoInts.Request()

        # Create a timer for the main loop execution
        self.timer = self.create_timer(
            1.0, self.main_loop, callback_group=self.timer_cb
        )

    def send_request(self, a, b):
        self.get_logger().info("send_request: enter")
        self.req.a = a
        self.req.b = b

        # Only works once, then never executes the time again
        # self.future = self.cli.call_async(self.req)
        # rclpy.spin_until_future_complete(self, self.future)
        # self.get_logger().info("send_request: exit")
        # return self.future.result()

        # Used to work, but blocks now
        return self.cli.call(self.req)

[...]

PR #1017, opened 2022-10-03
Revert "Raise user handler exception in MultiThreadedExecutor. (#984)"

fujitatomoya:
This reverts commit 6c0847a70df2a359c28ffae7e49ece2e62c1794d.

Signed-off-by: Tomoya Fujita <Tomoya.Fujita@sony.com>

PR #1019, opened 2022-10-04
recursive synchronous handler blocks on ThreadPoolExecutor::submit().

fujitatomoya:
address https://github.com/ros2/rclpy/issues/1016

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -14,6 +14,7 @@
 
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import ExitStack
+from functools import partial
 import inspect
 import multiprocessing
 from threading import Condition
@@ -769,7 +770,7 @@ class MultiThreadedExecutor(Executor):
                 handler()
                 if handler.exception() is not None:
                     raise handler.exception()
-            self._executor.submit(handler_wrapper(handler))
+            self._executor.submit(partial(handler_wrapper, handler))
 
     def spin_once(self, timeout_sec: float = None) -> None:
         self._spin_once_impl(timeout_sec)


fujitatomoya:
This is draft PR to get feedback since this regenerates the problem https://github.com/ros2/rclpy/issues/983.

fujitatomoya:
@wjwwood @ivanpauno @sloretz any thoughts? asking for help...

wjwwood:
Is it possible to make a regression test for the issue in https://github.com/ros2/rclpy/issues/1016? That would be the best way to prove that fixes the problem.

fujitatomoya:
> Is it possible to make a regression test for the issue in https://github.com/ros2/rclpy/issues/1016?

yeah, i think so. Using `MultiThreadedExecutor` to call sync client request via timer callback will reproduce the issue.

But before that, i do not have good idea to address https://github.com/ros2/rclpy/issues/1016 and https://github.com/ros2/rclpy/issues/983 atm...

fujitatomoya:
CC: @Barry-Xu-2018 @iuhilnehc-ynos

ivanpauno (approved):
Makes sense!

ivanpauno:
> But before that, i do not have good idea to address https://github.com/ros2/rclpy/issues/1016 and https://github.com/ros2/rclpy/issues/983 atm

aren't both issues fixed after this PR?

fujitatomoya:
> aren't both issues fixed after this PR?

No, unfortunately. This PR regenerates https://github.com/ros2/rclpy/issues/983, that is why I keep this under draft.

rclpy/rclpy/executors.py
```diff
@@ -769,7 +770,7 @@ def handler_wrapper(handler):
                 handler()
                 if handler.exception() is not None:
                     raise handler.exception()
-            self._executor.submit(handler_wrapper(handler))
+            self._executor.submit(partial(handler_wrapper, handler))
```
fujitatomoya, line 773: This also works the same.

```suggestion
            self._executor.submit(handler_wrapper, handler)
```

ivanpauno: Well, I didn't notice it in my original https://github.com/ros2/rclpy/pull/984 review, but that PR is broken.
It's not submiting `handler` for execution in the thread pool.
It's execution the handler immediately and submitting `None` to the executor ...

The only way to detect exceptions is to store the future returned by the executor and check if they finished successfully or with an exception.
We cannot block until the future is completed, so in each iteration of `spin()` we could poll all the stored futures and check if one of them failed (removing the ones that completed).
When a failure is detecter, it's thrown.

---

I'm okay with reverting https://github.com/ros2/rclpy/pull/984 temporally, as the solution to the actual problem #984 tried to solve isn't that trivial.

ivanpauno: @wjwwood what do you think?

fujitatomoya: > I'm okay with reverting https://github.com/ros2/rclpy/pull/984 temporally, as the solution to the actual problem https://github.com/ros2/rclpy/pull/984 tried to solve isn't that trivial.

agree.

ivanpauno: BTW, thanks for the PR @fujitatomoya !
I didn't undestand what was the issue until this :sweat_smile:

fujitatomoya: Reopen revert commit https://github.com/ros2/rclpy/pull/1017

fujitatomoya:
I will close this draft, instead will do the following.

- https://github.com/ros2/rclpy/pull/1017 (revert https://github.com/ros2/rclpy/pull/984) to address https://github.com/ros2/rclpy/issues/1016
- Backport https://github.com/ros2/rclpy/pull/1017 to humble / galactic / foxy
- Reopen https://github.com/ros2/rclpy/issues/983, will address with new PR. (Suggested fix, see https://github.com/ros2/rclpy/pull/1019#discussion_r988178727)
- Give it some time window before backport humble / galactic / foxy at this time.

PR #1031, opened 2022-10-28
Set the default number of threads of the MultiThreadedExecutor to 2

fujitatomoya:
https://github.com/ros2/rclcpp/pull/2030 for `rclpy`

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -734,7 +734,8 @@ class MultiThreadedExecutor(Executor):
 
     :param num_threads: number of worker threads in the pool.
         If ``None``, the number of threads will be automatically set by querying the underlying OS
-        for the CPU affinity of the process space. If it is not set, defaults to 2.
+        for the CPU affinity of the process space.
+        If the OS doesn't provide this information, defaults to 2.
     :param context: The context associated with the executor.
     """
 
@@ -742,7 +743,7 @@ class MultiThreadedExecutor(Executor):
         super().__init__(context=context)
         if num_threads is None:
             # On Linux, it will try to use the number of CPU this process has access to.
-            # On other platforms, os.sched_getaffinity() doesn't exist so we use the number of CPU.
+            # Other platforms, os.sched_getaffinity() doesn't exist so we use the number of CPUs.
             if hasattr(os, 'sched_getaffinity'):
                 num_threads = len(os.sched_getaffinity(0))
             else:
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -32,6 +32,7 @@ from typing import TYPE_CHECKING
 from typing import TypeVar
 from typing import Union
 
+import warnings
 
 from rclpy.client import Client
 from rclpy.clock import Clock
@@ -732,8 +733,8 @@ class MultiThreadedExecutor(Executor):
     Runs callbacks in a pool of threads.
 
     :param num_threads: number of worker threads in the pool. If ``None``, the number of threads
-        will use :func:`multiprocessing.cpu_count`. If that's not implemented the number of threads
-        defaults to 1.
+        will use :func:`multiprocessing.cpu_count` (minimum of 2).
+        If that's not implemented, the number of threads defaults to 2.
     :param context: The context associated with the executor.
     """
 
@@ -741,9 +742,13 @@ class MultiThreadedExecutor(Executor):
         super().__init__(context=context)
         if num_threads is None:
             try:
-                num_threads = multiprocessing.cpu_count()
+                num_threads = max(multiprocessing.cpu_count(), 2)
             except NotImplementedError:
-                num_threads = 1
+                num_threads = 2
+        if num_threads == 1:
+            warnings.warn(
+                'MultiThreadedExecutor is used with a single thread.\n'
+                'Use the SingleThreadedExecutor instead.')
         self._executor = ThreadPoolExecutor(num_threads)
 
     def _spin_once_impl(
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -15,7 +15,7 @@
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import ExitStack
 import inspect
-import multiprocessing
+import os
 from threading import Condition
 from threading import Lock
 from threading import RLock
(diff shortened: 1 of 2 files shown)


fujitatomoya:
@clalancette can you take a look?

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17556/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12094/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18123/)

rclpy/rclpy/executors.py
```diff
@@ -732,18 +733,22 @@ class MultiThreadedExecutor(Executor):
...
 
     def __init__(self, num_threads: int = None, *, context: Context = None) -> None:
         super().__init__(context=context)
         if num_threads is None:
             try:
-                num_threads = multiprocessing.cpu_count()
+                num_threads = max(multiprocessing.cpu_count(), 2)
             except NotImplementedError:
-                num_threads = 1
+                num_threads = 2
```
clalancette, line 747: Looking at this closer, I think we should actually refactor this a bit.

First, on Linux it is more appropriate to use `len(os.get_schedaffinity(0))`, which will tell you the number of cores this process has access to, not just how many are in the machine (see https://docs.python.org/3/library/os.html#os.sched_getaffinity and https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count).  That said, `os.get_schedaffinity` doesn't work on either macOS or Windows, so we need another solution there.

But that solution doesn't need to import the multiprocessing library.  In particular, we can use `os.cpu_count()` (which is what `multiprocessing.cpu_count()` calls anyway.  The only difference between the two is that os.cpu_count() returns None if it can't work, while `multiprocessing.cpu_count()` throws an exception.

So with all of that said, and the change you want to make, I think we should make this:

```
if num_threads is None:
     # On Linux, os.sched_getaffinity() returns the number of cores this process has access to.
     # On all other platforms, os.sched_getaffinity() doesn't exist so we use the number of cores.
     if hasattr(os, 'sched_getaffinity'):
        num_threads = len(os.sched_getaffinity(0))
    else:
        num_threads = os.cpu_count()

    # The calls above may still return None if they aren't supported; in that case, just use a hard-coded 2.
    if num_threads is None:
        num_threads = 2
```

(and with that change, we can then remove the `import multiprocessing` at the top of the file)

rclpy/rclpy/executors.py
```diff
@@ -732,18 +733,22 @@ class MultiThreadedExecutor(Executor):
     Runs callbacks in a pool of threads.
 
     :param num_threads: number of worker threads in the pool. If ``None``, the number of threads
-        will use :func:`multiprocessing.cpu_count`. If that's not implemented the number of threads
-        defaults to 1.
+        will use :func:`multiprocessing.cpu_count` (minimum of 2).
+        If that's not implemented, the number of threads defaults to 2.
```
clalancette, line 737:
```suggestion
        will be automatically set by querying the underlying OS for the number of CPUs (with a minimum of 2).
```

clalancette (changes requested):

rclpy/rclpy/executors.py
fujitatomoya: Thanks for pointing that out, sounds reasonable to me.

If the CPU affinity is set, it should be honored here as well.
I am not sure `multiprocessing.cpu_count()` or `os.cpu_count()` collects offline (fallback) cpu counts, but using `len(os.sched_getaffinity(0))` at 1st place does make sense to me.

I will fix it according, i will borrow some of your code! thanks 👍

fujitatomoya: i think follow-up required for `rclcpp` as well.

fujitatomoya: and another question is should we honor the affinity set by user? it the user set the affinity to only one CPU, we should honor that setting, so minimum can be 1.

clalancette: > I am not sure `multiprocessing.cpu_count()` or `os.cpu_count()` collects offline (fallback) cpu counts, but using `len(os.sched_getaffinity(0))` at 1st place does make sense to me.

As far as I understand, `cpu_count()` ends up calling [sysconf(_SC_NPROCESSORS_ONLN)](https://github.com/python/cpython/blob/87b5fd9d9de11c68b9940d3a1dba902138129109/Modules/posixmodule.c#L13382), so it should only collect online CPUs.

> i think follow-up required for `rclcpp` as well.

That's true, it would be better to make this change for `rclcpp` as well.  It's going to be more work there, as we are going to have to invoke platform-specific logic to get it.  Probably we should implement a function down in `rcutils` to get this information, and then use that in rclcpp.

> and another question is should we honor the affinity set by user? it the user set the affinity to only one CPU, we should honor that setting, so minimum can be 1.

Yes, the new code should do that, at least on Linux.  That is, `os.sched_getaffinity` will return 1 if the user set the affinity to only one CPU, and then they will get the warning that this isn't a good idea.  But it should still run.  On Windows and macOS, I have no idea how to do this.

rclpy/rclpy/executors.py
```diff
@@ -731,19 +732,28 @@ class MultiThreadedExecutor(Executor):
     """
     Runs callbacks in a pool of threads.
 
-    :param num_threads: number of worker threads in the pool. If ``None``, the number of threads
-        will use :func:`multiprocessing.cpu_count`. If that's not implemented the number of threads
-        defaults to 1.
+    :param num_threads: number of worker threads in the pool.
+        If ``None``, the number of threads will be automatically set by querying the underlying OS
+        for the CPU affinity of the process space. If it is not set, defaults to 2.
```
clalancette, line 737: Sorry, I think we should make this a bit clearer:

```suggestion
        If ``None``, the number of threads will be automatically set by querying the underlying OS
        for the CPU affinity of the process space. If the OS doesn't provide this information, defaults to 2.
```

rclpy/rclpy/executors.py
```diff
@@ -731,19 +732,28 @@ class MultiThreadedExecutor(Executor):
...
 
     def __init__(self, num_threads: int = None, *, context: Context = None) -> None:
         super().__init__(context=context)
         if num_threads is None:
-            try:
-                num_threads = multiprocessing.cpu_count()
-            except NotImplementedError:
-                num_threads = 1
+            # On Linux, it will try to use the number of CPU this process has access to.
+            # On other platforms, os.sched_getaffinity() doesn't exist so we use the number of CPU.
```
clalancette, line 745:
```suggestion
            # On other platforms, os.sched_getaffinity() doesn't exist so we use the number of CPUs.
```

clalancette (changes requested):
Two small fixes for comments, then this will be good to go.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17558/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12096/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18125/)

clalancette (approved):

fujitatomoya:
@clalancette just fyi, https://github.com/ros2/rclcpp/issues/2037 as follow-up for `rclcpp`

Issue #1030, opened 2022-10-27
ParameterUninitializedException on accessing parameter that is not yet set crashes node


Achllle:
## Bug report

I'm not sure if this is a bug or a feature. When you try to access a parameter on a node from the CLI or from another node, the node hosting the parameter will crash rather than return an error message.

**Required Info:**

- Operating System:
  - ros:galactic and ros:rolling docker images
- Installation type:
  - docker pull ros:rolling
- DDS implementation:
  - defaults
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

1. follow instructions from official tutorial
2. modify the code to use an uninitialized parameter as follows:

```
import rclpy
from rclpy.node import Node
from rclpy.parameter import Parameter
from rcl_interfaces.msg import SetParametersResult


class MinimalParam(Node):
    def __init__(self):
        super().__init__('minimal_param_node')
        self.log = self.get_logger()
        self.declare_parameter('my_parameter', Parameter.Type.STRING)

        self.add_on_set_parameters_callback(self.callback)

    def callback(self, parameters):
        for p in parameters:
            if p.name == 'my_parameter':
                self.log.info(f"Got my_parameter: {p.value}")
        return SetParametersResult(successful=True)


def main():
    rclpy.init()
    node = MinimalParam()
    rclpy.spin(node)

if __name__ == '__main__':
    main()
```
3. build and run this node
4. get the parameter from a different terminal using the CLI: `ros2 param get /minimal_param_node my_parameter`

#### Expected behavior

* The set param call returns with `successful=False` and `reason="parameter 'my_parameter' not initialized yet"`
* The node doesn't crash

#### Actual behavior

* The set param call hangs
* The node crashes with log:

```
Traceback (most recent call last):   
  File "/root/ros2_ws/install/python_parameters/lib/python_parameters/minimal_param_node", line 33, in <module>
    sys.exit(load_entry_point('python-parameters', 'console_scripts', 'minimal_param_node')())
  File "/root/ros2_ws/build/python_parameters/python_parameters/python_parameters_node.py", line 25, in main 
    rclpy.spin(node)      
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/__init__.py", line 222, in spin                                                                                                                                 
    executor.spin_once()                            
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 712, in spin_once                                                                                                                           
    raise handler.exception()                                                                                   
[...]

PR #1032, opened 2022-10-30
Fix incorrect comparsion on whether parameter type is NOT_SET

Barry-Xu-2018:
According to current code, if the type of declared parameter is 'NOT_SET', it throws an exception.

https://github.com/ros2/rclpy/blob/3d8547ad80d9fa72335b9ee0c105a80fcaf275c1/rclpy/rclpy/node.py#L472-L475

Actually, If write code like below, it doesn't throw any exception.

```rclpy
self.declare_parameter('my_parameter2', Parameter.Type.NOT_SET)
```

It is because second_arg is an instance of Parameter.Type and it doesn't have `value`. So it is always not equal.

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -470,7 +470,7 @@ class Node:
                 descriptor.dynamic_typing = True
 
             if isinstance(second_arg, Parameter.Type):
-                if second_arg.value == Parameter.Type.NOT_SET:
+                if second_arg == Parameter.Type.NOT_SET:
                     raise ValueError(
                         f'Cannot declare parameter {{{name}}} as statically typed of type NOT_SET')
                 if descriptor.dynamic_typing is True:


Barry-Xu-2018:
@clalancette @fujitatomoya @iuhilnehc-ynos please review this PR

rclpy/rclpy/node.py
```diff
@@ -470,7 +470,7 @@ def declare_parameters(
                 descriptor.dynamic_typing = True
 
             if isinstance(second_arg, Parameter.Type):
-                if second_arg.value == Parameter.Type.NOT_SET:
+                if second_arg == Parameter.Type.NOT_SET:
```
iuhilnehc-ynos, line 473: Is it necessary to update https://github.com/ros2/rclpy/blob/594669deb91451a7569e4f859cbc5d20ce93c200/rclpy/rclpy/node.py#L480?

iuhilnehc-ynos: The type of `descriptor.type` is `'type': 'uint8',`, so getting the value from an enum is correct.

iuhilnehc-ynos (approved):

rclpy/rclpy/node.py
Barry-Xu-2018: yes

fujitatomoya (approved):
lgtm, test would be ideal.

fujitatomoya:
part of https://github.com/ros2/rclpy/issues/1030

fujitatomoya:
@Barry-Xu-2018 @iuhilnehc-ynos can we run CI for this?

iuhilnehc-ynos:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17572/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12112/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18139/)

fujitatomoya:
@ivanpauno can you take a look at this?

fujitatomoya:
I would like to have another maintainer review on this.

ivanpauno (approved):
LGTM!
Thanks for the fix and for improving code coverage

PR #1036, opened 2022-11-10
Document rclpy.utilities.remove_ros_args

sloretz:
I noticed this function was missing a docstring while telling someone about it.

### DIFF
--- a/rclpy/rclpy/utilities.py
+++ b/rclpy/rclpy/utilities.py
@@ -15,14 +15,9 @@
 import os
 import sys
 import threading
+from typing import List
 from typing import Optional
-
-if sys.version_info < (3, 9, 0):
-    from typing import List
-    from typing import Sequence
-else:
-    List = list
-    from collections.abc import Sequence
+from typing import Sequence
 
 import ament_index_python
 
--- a/rclpy/rclpy/utilities.py
+++ b/rclpy/rclpy/utilities.py
@@ -15,6 +15,14 @@
 import os
 import sys
 import threading
+from typing import Optional
+
+if sys.version_info < (3, 9, 0):
+    from typing import List
+    from typing import Sequence
+else:
+    List = list
+    from collections.abc import Sequence
 
 import ament_index_python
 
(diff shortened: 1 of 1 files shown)


clalancette (approved):
It looks reasonable to me, but I'll suggest running CI (including on RHEL) to make sure this works everywhere.

adityapande-1995:
CI : 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17626/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12174/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18201/)

rclpy/rclpy/utilities.py
```diff
@@ -15,6 +15,14 @@
...
 import sys
 import threading
+from typing import Optional
+
+if sys.version_info < (3, 9, 0):
+    from typing import List
+    from typing import Sequence
+else:
+    List = list
+    from collections.abc import Sequence
```
fujitatomoya, line 25: @sloretz can you enlighten me why this is added for the future reference?

fujitatomoya (approved):
lgtm for doc section.

rclpy/rclpy/utilities.py
sloretz: Ah, I saw `typing.List` and `typing.Sequence` were deprecated in Python 3.9, and assumed it would result in a deprecation warning.

Some local testing and a glance at the [cpython source code](https://github.com/python/cpython/blob/8fe1a88e6ac3df20ad6bb96dd7e25aa18f2f438a/Lib/typing.py#L2705) show's there's no deprecation warning, so I'll get rid of this if/else logic.

sloretz: removed in 0ad3a49

sloretz:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17627/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12175/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18202/)

rclpy/rclpy/utilities.py
fujitatomoya: @sloretz thanks for the information!

clalancette (approved):

PR #1040, opened 2022-11-15
Document rclpy.time.Time class

sloretz:
This adds documentation for the Time class. Here's how it looks.

image

### DIFF
--- a/rclpy/rclpy/time.py
+++ b/rclpy/rclpy/time.py
@@ -59,7 +59,7 @@ class Time:
         """
         Get time separated into seconds and nanoseconds components.
 
-        :returns: 2-tuple seconds and nanoseconds
+        :return: 2-tuple seconds and nanoseconds
         """
         nanoseconds = self.nanoseconds
         return (nanoseconds // CONVERSION_CONSTANT, nanoseconds % CONVERSION_CONSTANT)
--- a/rclpy/rclpy/time.py
+++ b/rclpy/rclpy/time.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Tuple
+
 import builtin_interfaces.msg
 from rclpy.duration import Duration
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/time.py
```diff
@@ -41,21 +51,22 @@ def __init__(
         self._time_handle = _rclpy.rcl_time_point_t(total_nanoseconds, clock_type)
 
     @property
-    def nanoseconds(self):
+    def nanoseconds(self) -> int:
+        """:return: the total number of nanoseconds since the clock's epoch."""
```
clalancette, line 55: I see that some places we use `:return:`, and others we use `:returns:`.  Assuming they both work, can we be consistent (at least within this file)?

rclpy/rclpy/time.py
```diff
@@ -41,21 +51,22 @@ def __init__(
...
         return self._time_handle.nanoseconds
 
-    def seconds_nanoseconds(self):
+    def seconds_nanoseconds(self) -> Tuple[int, int]:
         """
-        Get time as separate seconds and nanoseconds components.
+        Get time separated into seconds and nanoseconds components.
 
         :returns: 2-tuple seconds and nanoseconds
-        :rtype: tuple(int, int)
```
clalancette, line 52: I notice that here you've removed `rtype`, but in some other places you added it.  This is again a place I'd hope we can be consistent, either always having the `rtype`, or always omitting it.  That said, there may be Python standards here I don't know about, so take this comment with a grain of salt.

sloretz: I went with what looked best in the generated documentation. Type annotations worked fine here, but not for returning `builtin_interfaces.msg.Time` because sphinx shortens it to `Time`, so I added `:rtype:` there.

rclpy/rclpy/time.py
sloretz: Sure, though it might not matter. They [seem to be aliases](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists), and the generated documentation looks identical to me.

sloretz: b14ca2c

clalancette (approved):
Despite the fact that this is a documentation-only change, I'll still suggest running CI on it (particularly for Windows and RHEL, since those have our oldest Python versions).

fujitatomoya (approved):
i am good to go!

sloretz:
CI (build: `--packages-up-to rclpy` test: ` --packages-select rclpy`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17646/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12201/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18220/)

PR #1048, opened 2022-11-28
Add in a warning for a depth of 0 with KEEP_LAST.

clalancette:
It really doesn't make much sense to have a KeepLast depth of 0; no data could possibly be stored. Indeed, the underlying DDS implementations don't actually support this. It currently "works" because a KeepLast depth of 0 is assumed to be system default, so the RMW layer typically chooses "1" instead. But this isn't something we should be encouraging users to do, so add a warning.

This is the rclpy companion PR to https://github.com/ros2/rclcpp/pull/2048

### DIFF
--- a/rclpy/rclpy/qos.py
+++ b/rclpy/rclpy/qos.py
@@ -164,6 +164,12 @@ class QoSProfile:
     @depth.setter
     def depth(self, value):
         assert isinstance(value, int)
+
+        if self.history == QoSHistoryPolicy.KEEP_LAST and value == 0:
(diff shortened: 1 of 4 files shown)


adityapande-1995 (approved):
Looks good with green CI

clalancette:
CI for this is in https://github.com/ros2/rclcpp/pull/2048#issuecomment-1330693880

mjcarroll (approved):

clalancette:
CI for this was clean, so merging.

PR #1058, opened 2022-12-20
 Hook up the inconsistent topic event inside of rclpy

clalancette:
This PR is part of https://github.com/ros2/ros2/issues/1361 . In particular, what this does is take the inconsistent topic event as it comes from the rmw/rcl layer, and allows the user to register for a callback to find that out.

Along the way, some refactoring was in order.  In particular, rclpy assumed that all events were QoS events, which is not really the case. This PR does some renaming of things to make it clear that these events are not just for QoS, but can be for other things as well. With that done, the last patch in the series actually implements the inconsistent topic callback.

Part of https://github.com/ros2/ros2/issues/1361

Depends on https://github.com/ros2/rmw/pull/339
Depends on https://github.com/ros2/rcl/pull/1024


fujitatomoya (approved):
lgtm

PR #1062, opened 2023-01-02
Add default preset qos profile

ivanpauno:
Minor addition, will be used in https://github.com/ros2/ros2cli/pull/788.


ivanpauno:
Check CI in https://github.com/ros2/ros2cli/pull/788#issuecomment-1369167929

fujitatomoya (approved):
@ivanpauno this was actually just missing?

ivanpauno:
> @ivanpauno this was actually just missing?

I guess, I don't see a reason to not have this option.

PR #1070, opened 2023-01-15
Call destroy() before removing ServerGoalHandle

otamachan:
Signed-off-by: Tamaki Nishino <otamachan@gmail.com>

As reported by @longjie0723 here https://answers.ros.org/question/411394/rclpy-action-server-seems-slowing-down-and-consuming-memory-in-long-term/ ,Python `ActionServer` could keep `Future` objects while it's running, which causes memory consumption increase and the performance issue.
It seems that it's because `ServerGoalHandler` [registers `Future`](https://github.com/ros2/rclpy/blob/fb1058935f25f20a5ba9552994b9b63a3edd36f9/rclpy/rclpy/action/server.py#L74) when it is constructed and unregisters when `destroy()` is called, but `destroy()` is not called when the object is removed [here](https://github.com/ros2/rclpy/blob/fb1058935f25f20a5ba9552994b9b63a3edd36f9/rclpy/rclpy/action/server.py#L408) .

This PR calls `destroy()` before `ServerGoalHandler` is deleted.

I confirmed the [the original issue](https://answers.ros.org/question/411394/rclpy-action-server-seems-slowing-down-and-consuming-memory-in-long-term/) is resolved by this PR.

### DIFF
--- a/rclpy/rclpy/action/server.py
+++ b/rclpy/rclpy/action/server.py
@@ -405,6 +405,7 @@ class ActionServer(Waitable):
     async def _execute_expire_goals(self, expired_goals):
         for goal in expired_goals:
             goal_uuid = bytes(goal.goal_id.uuid)
+            self._goal_handles[goal_uuid].destroy()
             del self._goal_handles[goal_uuid]
 
     def _send_result_response(self, request_header, future):


rclpy/rclpy/action/server.py
```diff
@@ -405,6 +405,7 @@ async def _execute_get_result_request(self, request_header_and_message):
     async def _execute_expire_goals(self, expired_goals):
         for goal in expired_goals:
             goal_uuid = bytes(goal.goal_id.uuid)
+            self._goal_handles[goal_uuid].destroy()
```
fujitatomoya, line 408: I see this works and addresses memory leak.

how about adding dtor in `ServerGoalHandle` to call `destroy` method internally? so that once this object is deleted, it automatically issues dtor to destroy the `goal_handle`?

fujitatomoya (commented):
@otamachan thanks for the contribution.

rclpy/rclpy/action/server.py
fujitatomoya: CC: @sloretz

otamachan: Thanks for the comment. I thought so too at the first time, but I fount only a few classes have `__del__` method, so I assumed that the most of the classes in `rclpy` are expected to be called `destroy`. I'm ok to change to call `destroy` in `__del__` but I'd like to ask maintainers' opinion as well.

```bash
$ git grep __del__
rclpy/rclpy/executors.py:    def __del__(self):
rclpy/rclpy/signals.py:    def __del__(self):
rclpy/rclpy/task.py:    def __del__(self):
```

fujitatomoya: this is not hard requirement, so that we can do the refactor later, but lets wait and see other opinion.

CC: @sloretz @clalancette @ivanpauno

otamachan: I see. Thank you for your suggestion. I'll try that!

apockill:
I'm running into this memory leak in production- but the concern is less in the memory aspect but the fact that action call rate slows down rather dramatically after a few hundred calls (from 300 calls per second to 50 to 10).

I've put together a small reproducible example that shows that ActionServer._futures is indeed growing at a high rate, but unfortunately for some reason, `_execute_expire_goals` is never being called by the ActionServer, and this fix is therefor not working.

Could someone help me understand what triggers the calling of `_execute_expire_goals`, so I can better help debug this (and my) issue?  Thanks!

fujitatomoya:
@otamachan are you working on https://github.com/ros2/rclpy/pull/1070#discussion_r1080592437?

dcconner:
I am also interested in the status of this fix

apockill:
Hey folks! This issue seems to be pretty key to any action running in ROS- especially if you rely on ROS actions being relatively fast to call, and if you plan on running a robot for longer periods of time. I understand that https://github.com/ros2/rclpy/pull/1070#discussion_r1080592437 might not be the ideal fix, but wouldn't it be worth merging it in the interim until the proposed refactor can be complete?

fujitatomoya:
@apockill Just FYI, i created https://github.com/ros2/rclpy/pull/1113.

clalancette:
Closing this one in favor of #1113

Issue #1039, opened 2022-11-12
Callbacks are executed mutually exclusive, even tho ReentrantCallbackGroup and MultiThreadedExecutor are used.


Flova:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - latest binaries
- Version or commit hash:
  - 3.7.0-1jammy.20220914.144405
- DDS implementation:
  - Fast-RTPS as well as CycloneDDS
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

Launch the following node

```python
import time
import rclpy
from rclpy.executors import MultiThreadedExecutor
from rclpy.callback_groups import ReentrantCallbackGroup
from rclpy.node import Node
from std_msgs.msg import Empty


class TestNode(Node):

    def __init__(self) -> None:
        super().__init__('test_node')

        my_callback_group = ReentrantCallbackGroup()

        self.create_subscription(Empty, 'trigger1', self.callback1, 1, callback_group=my_callback_group)
        self.create_subscription(Empty, 'trigger2', self.callback2, 1, callback_group=my_callback_group)
        self.create_timer(1, self.callback3, callback_group=my_callback_group)

    def callback1(self, data) -> None:
        self.get_logger().warning("Start Trigger 1")
        time.sleep(10)
        self.get_logger().warning("End of Trigger 1")

    def callback2(self, data) -> None:
        self.get_logger().warning("Trigger 2")

    def callback3(self) -> None:
        self.get_logger().warning("Trigger Timer")


def main(args=None):
    rclpy.init(args=args)
    node = TestNode()
    ex = MultiThreadedExecutor()
    ex.add_node(node)
    ex.spin()
    node.destroy_node()
    rclpy.shutdown()

if __name__ == "__main__":
    main()
```

Manually publish the trigger messages. First, start the following command to keep publishing trigger messages.

```bash
ros2 topic pub /trigger2 std_msgs/Empty
```

After a few seconds run

``` bash
ros2 topic pub /trigger1 std_msgs/Empty -1
```

#### Expected behavior

The node should log both `Trigger 2` and `Trigger Timer` every second. Triggering the `trigger1` should not affect this. 
The log would look something like this:
```
Trigger 2
Trigger Timer
Trigger 2
Trigger Timer
Trigger 2
Trigger Timer
Trigger 2
Trigger Timer
Start Trigger 1
Trigger 2
Trigger Timer
...
Trigger 2
Trigger Timer
[...]

PR #1073, opened 2023-01-25
Fix #983 by saving future and checking for + raising any exceptions

### LINKED ISSUE #983: Exception never retrieved when using MultiThreadedExecutor and Timers

Achllle:
Fixes #983 while preserving functionality as reported in #1016 

Implementation follows @fujitatomoya 's [recommendation](https://github.com/ros2/rclpy/issues/983#issuecomment-1271782711). Rather than save the executor's returned future, I save the handler since the future doesn't seem to have access to the exception.

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -777,13 +777,10 @@ class MultiThreadedExecutor(Executor):
         else:
             self._executor.submit(handler)
             self.futures.append(handler)
-            # check for any exceptions
-            for future in self.futures:
+            for future in self.futures:  # check for any exceptions
                 if future.done():
-                    if future._exception:
-                        self.futures.remove(future)
-                        raise future._exception
                     self.futures.remove(future)
+                    future.result()
 
     def spin_once(self, timeout_sec: float = None) -> None:
         self._spin_once_impl(timeout_sec)
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -755,6 +755,7 @@ class MultiThreadedExecutor(Executor):
             warnings.warn(
                 'MultiThreadedExecutor is used with a single thread.\n'
                 'Use the SingleThreadedExecutor instead.')
+        self.futures = []
         self._executor = ThreadPoolExecutor(num_threads)
 
     def _spin_once_impl(
@@ -775,6 +776,14 @@ class MultiThreadedExecutor(Executor):
             pass
         else:
             self._executor.submit(handler)
+            self.futures.append(handler)
+            # check for any exceptions
+            for future in self.futures:
+                if future.done():
+                    if future._exception:
+                        self.futures.remove(future)
+                        raise future._exception
+                    self.futures.remove(future)
 
     def spin_once(self, timeout_sec: float = None) -> None:
         self._spin_once_impl(timeout_sec)
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -755,7 +755,7 @@ class MultiThreadedExecutor(Executor):
             warnings.warn(
                 'MultiThreadedExecutor is used with a single thread.\n'
                 'Use the SingleThreadedExecutor instead.')
-        self.futures = []
+        self._futures = []
         self._executor = ThreadPoolExecutor(num_threads)
 
     def _spin_once_impl(
@@ -776,10 +776,10 @@ class MultiThreadedExecutor(Executor):
             pass
         else:
             self._executor.submit(handler)
-            self.futures.append(handler)
-            for future in self.futures:  # check for any exceptions
+            self._futures.append(handler)
+            for future in self._futures:  # check for any exceptions
                 if future.done():
-                    self.futures.remove(future)
+                    self._futures.remove(future)
                     future.result()
 
     def spin_once(self, timeout_sec: float = None) -> None:


fujitatomoya:
@Achllle could you confirm this fixes https://github.com/fujitatomoya/ros2_test_prover/blob/master/prover_rclpy/src/rclpy_983.py ? as far as i see, it still does not catch the exception.

btw, regression test is already in upstream, https://github.com/ros2/rclpy/pull/1044. this can confirm that there is no regression such as https://github.com/ros2/rclpy/issues/1039.

Achllle:
Hmm, it seems like it's only a partial fix. 
Without this fix, the test case will 'hang' for ~10s, then burst out a stream of
```
The following exception was never retrieved: division by zero
The following exception was never retrieved: division by zero
...
```
and then repeat 10 seconds later, each time without any traceback.

With this fix, the script will fail immediately upon raising the `ZeroDivisionError` with traceback:
```
  File "/workspaces/dockerfiles/rclpy_src/src/rclpy_983.py", line 48, in <module>
    main()
  File "/workspaces/dockerfiles/rclpy_src/src/rclpy_983.py", line 42, in main
    multi_executor()
  File "/workspaces/dockerfiles/rclpy_src/src/rclpy_983.py", line 33, in multi_executor
    executor.spin()
  File "/workspaces/rolling/install/rclpy/local/lib/python3.10/dist-packages/rclpy/executors.py", line 280, in spin
    self.spin_once()
  File "/workspaces/rolling/install/rclpy/local/lib/python3.10/dist-packages/rclpy/executors.py", line 789, in spin_once
    self._spin_once_impl(timeout_sec)
  File "/workspaces/rolling/install/rclpy/local/lib/python3.10/dist-packages/rclpy/executors.py", line 785, in _spin_once_impl
    raise future._exception
  File "/workspaces/rolling/install/rclpy/local/lib/python3.10/dist-packages/rclpy/task.py", line 239, in __call__
    self._handler.send(None)
  File "/workspaces/rolling/install/rclpy/local/lib/python3.10/dist-packages/rclpy/executors.py", line 424, in handler
    await call_coroutine(entity, *arg)
  File "/workspaces/rolling/install/rclpy/local/lib/python3.10/dist-packages/rclpy/executors.py", line 334, in _execute_timer
    await await_or_execute(tmr.callback)
  File "/workspaces/rolling/install/rclpy/local/lib/python3.10/dist-packages/rclpy/executors.py", line 108, in await_or_execute
    return callback(*args)
  File "/workspaces/dockerfiles/rclpy_src/src/rclpy_983.py", line 12, in tmr_callback
    1/0
ZeroDivisionError: division by zero
The following exception was never retrieved: division by zero
```
but will still produce a `The following exception was never retrieved: division by zero`. I will dig in a bit more to see if I can rectify that, in the meantime any suggestions are welcome.

Achllle:
@fujitatomoya Can you take a look again? The latest commit should fix the `The following exception was never retrieved: division by zero` issue by calling `future.result()`. Tested on #983 test.

rclpy/rclpy/executors.py
```diff
@@ -755,6 +755,7 @@ def __init__(self, num_threads: int = None, *, context: Context = None) -> None:
             warnings.warn(
                 'MultiThreadedExecutor is used with a single thread.\n'
                 'Use the SingleThreadedExecutor instead.')
+        self.futures = []
```
fujitatomoya, line 758: this meant to be internal use only?

```suggestion
        self._futures = []
```

fujitatomoya (commented):
one minor comment needs to be addressed, overall lgtm.

btw, i confirmed that this can fix the issue to raise the exception right away.

fujitatomoya:
@wjwwood minor fix, this can address the issue https://github.com/ros2/rclpy/issues/983, can you take a look?

rclpy/rclpy/executors.py
```diff
@@ -775,6 +776,11 @@ def _spin_once_impl(
             pass
         else:
             self._executor.submit(handler)
+            self.futures.append(handler)
```
Achllle, line 779:
```suggestion
            self._futures.append(handler)
```

rclpy/rclpy/executors.py
```diff
@@ -775,6 +776,11 @@ def _spin_once_impl(
...
+            for future in self.futures:  # check for any exceptions
```
Achllle, line 780:
```suggestion
            for future in self._futures:  # check for any exceptions
```

rclpy/rclpy/executors.py
```diff
@@ -775,6 +776,11 @@ def _spin_once_impl(
...
+                if future.done():
+                    self.futures.remove(future)
```
Achllle, line 782:
```suggestion
                    self._futures.remove(future)
```

fujitatomoya (approved):
lgtm

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18046/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12575/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18691/)

tonynajjar:
I happened to stumble on the same bug, I tested the fix (on top of humble) and it worked for me. Looking forward for the merge and backport please :grin:

Achllle:
@fujitatomoya @wjwwood Anything I can do to move this along? Linux-aarch64 CI failure looks unrelated to me.

adityapande-1995:
@Achllle I'll trigger CI once more, and if that passes, I'll merge this.

clalancette:
For what it is worth, we shouldn't merge things in with failing CI.  In this particular case it didn't matter (CI was happy overnight), but it isn't our usual development workflow.

fujitatomoya:
@Mergifyio backport humble

fujitatomoya:
@Mergifyio backport foxy

jdlangs:
This works but it relies on another callback being available for execution before the exception is raised since `wait_for_ready_callbacks` can block indefinitely. If you have a node that just provides services, for example, you won't get the exception until someone makes another service call after the failing one.

Would it make sense to modify this to use a timeout and keep checking the futures on a regular basis?

froly:
@Achllle @fujitatomoya I believe there's a typo  in this patch
```
self._futures.append(handler)
```
should be
```
future = self._executor.submit(handler)
self._futures.append(future)
```

we want to store and check the `Future` returned by `executor.submit()` to make this work and propagate an exception

PR #1075, opened 2023-01-31
Use RCPPUTILS_SCOPE_EXIT to cleanup unparsed_indices_c.

clalancette:
The unique_ptr way to do this works, but clang static analysis complains that data could be leaked if the construction of the unique_ptr fails.  Since this code is really old, switch to using RCPPUTILS_SCOPE_EXIT, which is our more modern way to deal with this problem.


fujitatomoya (approved):

clalancette:
CI for this is in https://github.com/ros2/rcl/pull/1031#issuecomment-1412130409

PR #1079, opened 2023-02-06
Document that spin_once() should not be called from multiple threads

### LINKED ISSUE #1078: Document that `spin_once()` should not be called from multiple threads
I don't think it's obvious that `spin_once()` shouldn't be called from multiple threads. Let's improve the documentation.

Link to the current documentation:  https://docs.ros2.org/latest/api/rclpy/api/init_shutdown.html

answers.ros.org question that helped me figure this out:  https://answers.ros.org/question/356686/python-not-able-to-echo-ros2-topics-published-on-thread/

AndyZe:
Fixes #1078


fujitatomoya (commented):
I think this depends on arguments, if user application calls `rclpy::spin_once` with the same nodes and two different executor, this would be problem since we have a node in 2 wait sets. but we can create multiple threads to call `SingleThreadedExecutor::snip_once:` for each node, that is usually what user would do.

AndyZe:
Can you please suggest better wording for this PR?

`but we can create multiple threads to call SingleThreadedExecutor::snip_once: for each node, that is usually what user would do.`

This is what happened in my experience and it didn't work.

I'd actually suggest to deprecate `spin_once()` since it has these pitfalls that aren't well understood and recommend `spin()` only   ¯\_(ツ)_/¯

fujitatomoya:
To be honest, i am not sure how to address this as documentation. How about adding `node which is not already added to any other executors` in the description of `node` argument? that is the situation that we need to avoid here. there has been discussion related about this with `rclcpp`, https://github.com/ros2/rclcpp/issues/1953. probably node should be aware that it is already added to executor or not, or we can add the same node to different executor (wait sets).

> I'd actually suggest to deprecate spin_once() since it has these pitfalls that aren't well understood and recommend spin() only ¯_(ツ)_/¯

i believe we can have the same problem with `spin_once()`, that is not gonna be the solution.

> This is what happened in my experience and it didn't work.

i understand, for user experience, probably `please do not use those with multi-thread` would be reasonable for now. instead, we can recommend `MultiThreadedExecutor`. i will keep this open so that we can have more feedback.

thanks for iterating.

rclpy/rclpy/__init__.py
```diff
@@ -192,6 +192,8 @@ def spin_once(node: 'Node', *, executor: 'Executor' = None, timeout_sec: float =
     It is possible the work done is for a node other than the one provided if the global executor
     has a partially completed coroutine.
 
+    This method should not be called from multiple threads.
+
```
sloretz, line 196: How about

```
This method should not be called from multiple threads with the same node argument or executor argument.
```

It leaves out the details of why (avoiding calling `rcl_wait` on the same entity from multiple threads), but I think that's fine for such a high level method.

rclpy/rclpy/executors.py
```diff
@@ -307,6 +307,8 @@ def spin_once(self, timeout_sec: float = None) -> None:
 
         A custom executor should use :meth:`wait_for_ready_callbacks` to get work.
 
+        This method should not be called from multiple threads.
```
sloretz, line 310: I think this wording on `executor.spin_once` is fine.

rclpy/rclpy/__init__.py
AndyZe: Sure. If you make this a suggestion, I'd be happy to accept it. (I don't have a local copy of this very old PR any longer)

sloretz: I added it as a new commit!

sloretz (approved):

sloretz:
@mergifyio rebase

sloretz:
CI ([repos file](https://gist.githubusercontent.com/sloretz/02865267b5619b781be14c197ac7dae9/raw/08c80d9768339ac919aac6e8b12a903691f3a614/ros2.repos) build: `--packages-up-to rclpy` test: `--packages-select rclpy`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19657/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14164/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20380/)

adityapande-1995 (approved):

Issue #1095, opened 2023-03-24
Arm64 cannot link atomic


CursedRock17:
**Required Info:**

- Operating System:
   MacOS Ventura 13.2.1 arm64
- Installation type:
   Source
- Version or commit hash:
  rolling release branch
- DDS implementation:
   default
- Client library (if applicable):
   rclpy (CMakeLists.txt)

#### Steps to reproduce issue
This error occurs when starting the install process from source on MacOS, with the listed Operating system, probably anything after 10.14. The person who is affected by this issue should be installing with [this](https://docs.ros.org/en/rolling/Installation/Alternatives/macOS-Development-Setup.html) tutorial from the ROS 2 documentation and will most likely run into this issue.

#### Expected behavior
The rclpy package shouldn't fail during the colcon build stage of the installation.

#### Actual behavior
std::atomic fails to link

#### Additional information
I have fixed this with my forked repository, I just need to list the issue before throwing out a pull request. The pull request will cite that when using `APPLE` build systems, atomic will not be linked at that point in the process, but continues to be supported not causing any errors in the repository.

PR #1096, opened 2023-03-24
Solving Atomic undefined on OSX with clang

CursedRock17:
This is related to issue #1095 where building ROS 2 from source would fail at this package on OSX Ventura 13.2.1 with clang. I know MacOS is a tier 3 platform, but this pull request is very minuscule and doesn't rupture the build or its times significantly, just opening up other opportunities. Apple Clang cannot process atomic early in the build process so it just can't be included in the main `CMakeLists.txt` but it doesn't cause any issues later in this package so it is included elsewhere.

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -106,7 +106,7 @@ pybind11_add_module(_rclpy_pybind11 SHARED
 )
 
 if(CMAKE_C_COMPILER_ID MATCHES "Clang" AND NOT APPLE)
-  target_link_libraries(_rclpy_pybind11 PRIVATE atomic) 
+  target_link_libraries(_rclpy_pybind11 PRIVATE atomic)
 endif()
 
 target_include_directories(_rclpy_pybind11 PRIVATE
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
to be honest, I am not sure how this problem happens. if that works for OSX, that sounds reasonable and helpful for OSX user.

one thing, can you fix DCO error?

clalancette:
Yeah, I think this is generally OK, but it needs a `Signed-off-by` line.

mjcarroll:
> to be honest, I am not sure how this problem happens. 

For whatever reason, clang doesn't automatically link `atomic` on macOS.  This is something I have run into before.  There are several other github issues mentioning it, but I haven't found a conclusive reason.  The best argument I saw was that `libatomic` is a GCC library, so Clang doesn't include it by default on macOS.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18335/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12857/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19004/)

clalancette (approved):

Issue #1080, opened 2023-02-07
404 for API Docs Link on ROS Index 


ottojo:
The [ros index page for rclpy](https://index.ros.org/p/rclpy/github-ros2-rclpy/#humble) contains a "API Docs" link `http://docs.ros.org/en/humble/p/rclpy`, but the docs are not found at that URL.

Using google, i could only find docs for foxy (https://docs.ros2.org/foxy/api/rclpy/index.html) and galactic (https://docs.ros2.org/galactic/api/rclpy/index.html)

The CI job for generating (humble) docs seems to run, but docs are nowhere to be found: https://build.ros2.org/view/Hdoc/job/Hdoc__rclpy__ubuntu_jammy_amd64/

clalancette:
You did a good job tracking down how all of that works.  In short, you are correct that we aren't currently generating anything for rclpy properly.

In order to get there, we need to land https://github.com/ros-infrastructure/rosdoc2/pull/28 (which still needs some work), as well as do some custom configuration in `rclpy` itself.  I'll leave this open as the issue to track the addition of that configuration.

Dario-Mantegazza:
Hi,
it is probably too early to ask since the pull request cited above was just merged yesterday, but, when can we expect that the doc (for humble at least) will be published?
Thanks

clalancette:
> it is probably too early to ask since the pull request cited above was just merged yesterday, but, when can we expect that the doc (for humble at least) will be published?

While we did get the PR landed, there are still additional follow-up fixes that are needed to rosdoc2.  And we did not get the custom configuration in rclpy landed.  And then we need to backport it to humble.  We are slowly getting there, but it is still going to be some time to get it all working end-to-end.

ottojo:
Since the iron release is now out and the api-docs are built and correctly linked in the index, i'm happy with that resolution so far! I won't close this yet since backporting to humble is mentioned above, but if that should be moved to another issue, feel free to close this :+1:

PR #1103, opened 2023-04-02
Support documentation generation using rosdoc2

Yadunund:
Addresses https://github.com/ros2/rclpy/issues/1080
Requires https://github.com/ros-infrastructure/rosdoc2/pull/51

Here we add a custom rosdoc2.yaml file which forces rosdoc2 to generate documentation for this package as if it were an `ament_python` package. No underlay needs to be sourced.

rclpy_docs

### Cavats:
1. If I set `sphinx_sourcedir` to `docs/source` in the rosdoc2.yaml file, ie, if we provide the custom conf.py file there several hundred warnings reported by autodoc. I noticed that these warnings go away when using the default conf.py file that rosdoc2 generates and hence left out the custom sourcedir. Perhaps this can be done for future work.

2. I had to add a few `rclpy` submodules as `doc_depends` such that autodoc will mock these modules when trying to import them. Without this, there are several type errors thrown upon import as these submodules load constants from python C modules.
Open to any suggestions here!

Detailed errors below:

```
  File "home/yadunund/ws_rosdoc2/docs_build/rclpy/rclpy/duration.py", line 31, in __init__
    total_nanoseconds += int(nanoseconds)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'RMW_DURATION_INFINITE'

  File "/home/yadunund/ws_rolling/src/ros2/rclpy/rclpy/rclpy/qos.py", line 444, in <module>
    qos_profile_unknown = QoSProfile(**_rclpy.rmw_qos_profile_t.predefined(
  File "/usr/local/google/home/yadunund/ws_rolling/src/ros2/rclpy/rclpy/rclpy/qos.py", line 82, in __init__
    raise InvalidQoSProfileException('History and/or depth settings are required.')
rclpy.qos.InvalidQoSProfileException: ('Invalid QoSProfile', 'History and/or depth settings are required.')

  File "/home/yadunund/ws_rolling/src/ros2/rclpy/rclpy/rclpy/clock.py", line 126, in __new__
    if not isinstance(clock_type, ClockType):
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

### DIFF
--- a/rclpy/package.xml
+++ b/rclpy/package.xml
@@ -48,9 +48,11 @@
   <test_depend>test_msgs</test_depend>
 
   <doc_depend>python3-sphinx</doc_depend>
+  <doc_depend>python3-sphinx-autodoc-typehints</doc_depend>
   <doc_depend>python3-sphinx-rtd-theme</doc_depend>
 
   <export>
     <build_type>ament_cmake</build_type>
+    <rosdoc2>docs/rosdoc2.yaml</rosdoc2>
   </export>
 </package>
--- a/rclpy/package.xml
+++ b/rclpy/package.xml
@@ -23,6 +23,7 @@
   <build_depend>rcutils</build_depend>
   <build_depend>rmw_implementation_cmake</build_depend>
 
+  <depend>lifecycle_msgs</depend>
   <depend>rcl</depend>
   <depend>rcl_interfaces</depend>
   <depend>rcl_lifecycle</depend>
@@ -37,7 +38,6 @@
   <exec_depend>action_msgs</exec_depend>
   <exec_depend>ament_index_python</exec_depend>
   <exec_depend>builtin_interfaces</exec_depend>
-  <exec_depend>lifecycle_msgs</exec_depend>
   <exec_depend>rosgraph_msgs</exec_depend>
   <exec_depend>rpyutils</exec_depend>
 
--- a/rclpy/package.xml
+++ b/rclpy/package.xml
@@ -50,7 +50,6 @@
   <test_depend>test_msgs</test_depend>
 
   <doc_depend>python3-sphinx</doc_depend>
-  <doc_depend>python3-sphinx-autodoc-typehints</doc_depend>
   <doc_depend>python3-sphinx-rtd-theme</doc_depend>
   <!-- We add these modules as doc_depends in order to include them into the
   autodoc_mock_imports list. This prevents TypeErrors from being thrown
--- a/rclpy/package.xml
+++ b/rclpy/package.xml
@@ -33,8 +33,10 @@
   <depend>rosidl_runtime_c</depend>
   <depend>unique_identifier_msgs</depend>
 
+  <exec_depend>action_msgs</exec_depend>
   <exec_depend>ament_index_python</exec_depend>
   <exec_depend>builtin_interfaces</exec_depend>
+  <exec_depend>lifecycle_msgs</exec_depend>
   <exec_depend>rcl_interfaces</exec_depend>
   <exec_depend>rosgraph_msgs</exec_depend>
   <exec_depend>rpyutils</exec_depend>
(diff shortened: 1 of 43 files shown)


rclpy/package.xml
```diff
@@ -34,8 +34,10 @@
   <depend>rosidl_runtime_c</depend>
   <depend>unique_identifier_msgs</depend>
 
+  <exec_depend>action_msgs</exec_depend>
   <exec_depend>ament_index_python</exec_depend>
   <exec_depend>builtin_interfaces</exec_depend>
+  <exec_depend>lifecycle_msgs</exec_depend>
```
clalancette, line 40: Actually, `lifecycle_msgs` should be a `<depend>`; it is required for building the C++ side of things as well.

rclpy/rosdoc2.yaml
```diff
@@ -0,0 +1,79 @@
...
+    ## that were run, and may contain links to things like build farm jobs for
+    ## this package or links to other versions of this package.
+
+    ## If false, you can still include content that would have been in the index
+    ## into one of your '.rst' files from your Sphinx project, using the
+    ## '.. include::' directive in Sphinx.
+    ## For example, you could include it in a custom 'index.rst' so you can have
+    ## the standard information followed by custom content.
+
+    ## TODO(wjwwood): provide a concrete example of this (relative path?)
```
clalancette, line 20: I think we can remove this whole block; it isn't relevant for this package.

clalancette (changes requested):

rclpy/package.xml
Yadunund: Nice catch [37333a3](https://github.com/ros2/rclpy/pull/1103/commits/37333a34bc51482f07b907e51bfd06f2b6e74192)

rclpy/rosdoc2.yaml
Yadunund: Removed [37333a3](https://github.com/ros2/rclpy/pull/1103/commits/37333a34bc51482f07b907e51bfd06f2b6e74192)

rclpy/package.xml
```diff
@@ -48,9 +50,18 @@
   <test_depend>test_msgs</test_depend>
 
   <doc_depend>python3-sphinx</doc_depend>
+  <doc_depend>python3-sphinx-autodoc-typehints</doc_depend>
```
clalancette, line 53: After discussion, this doesn't seem to be necessary, so let's just remove it.

clalancette (changes requested):

rclpy/package.xml
Yadunund: Done [609605c](https://github.com/ros2/rclpy/pull/1103/commits/609605cf7340335beb0c256e1f3b77d9524d94cd)

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18419/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12965/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19102/)

rclpy/package.xml
```diff
@@ -49,8 +51,16 @@
 
   <doc_depend>python3-sphinx</doc_depend>
   <doc_depend>python3-sphinx-rtd-theme</doc_depend>
+  <!-- We add these modules as doc_depends in order to include them into the
+  autodoc_mock_imports list. This prevents TypeErrors from being thrown
+  when these modules import constants from rclpy C modules.
+  TODO: Remove these depends once a better solution is found.-->
+  <doc_depend>rclpy.duration</doc_depend>
+  <doc_depend>rclpy.qos</doc_depend>
+  <doc_depend>rclpy.clock</doc_depend>
```
wjwwood, line 60: Is there any discussion about this anywhere? I'd like to understand it better, because it seems... like an abuse of the dependency system.

Yadunund: Not apart from what was mentioned in the PR description. I agree we're taking advantage of the `doc_depend` tag but I couldn't find a better way to deal with the import errors. I'm all ears for a better solution!

PR #1114, opened 2023-04-14
Don't crash the action server if the client goes away.

clalancette:
While doing some testing locally, I had a broken action client that would send a goal, and then crash.  That meant it wasn't available to receive the goal response, nor the result response.  In this situation, my action server was throwing an exception and crashing.

But that doesn't seem right.  The action server should continue to run so other action clients can connect.  There are at least 2 different ways we could accomplish this:

1. Document that the user needs to catch these exceptions and deal with it themselves.
2. Handle these exceptions at the ActionServer layer.

There is some benefit to 1, in that the user has full control. However, the problem is that the exceptions we throw in rclpy aren't detailed enough for the user to disambiguate the errors. So instead, we go for option 2 here.

### DIFF
--- a/rclpy/rclpy/action/server.py
+++ b/rclpy/rclpy/action/server.py
@@ -306,7 +306,14 @@ class ActionServer(Waitable):
         response_msg = self._action_type.Impl.SendGoalService.Response()
         response_msg.accepted = accepted
         response_msg.stamp = goal_info.stamp
-        self._handle.send_goal_response(request_header, response_msg)
+
+        try:
+            # If the client goes away anytime before this, sending the goal response may fail.
+            # Catch the exception here and go on so we don't crash.
+            self._handle.send_goal_response(request_header, response_msg)
+        except RCLError:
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18594/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13134/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19331/)

mhidalgo-bdai:
@clalancette howdy! Can we backport this to Humble and Iron? Seems backwards compatible to me.

clalancette:
> @clalancette howdy! Can we backport this to Humble and Iron? Seems backwards compatible to me.

Yeah, I think we can backport this.

clalancette:
@Mergifyio backport iron humble

Issue #1106, opened 2023-04-09
Make rclpy.context.Context be a context manager


felixdivo:
#### Feature description
`rclpy.context.Context` would be useful to have as a [context manager](https://docs.python.org/3/reference/datamodel.html#context-managers). This would allow for:
```python
with Context() as context:
    executor = MultiThreadedExecutor(context=context)
    do_something_cool(executor)
```

Instead of
```python
context = Context()
try:
    context.init()
    executor = MultiThreadedExecutor(context=context)
    do_something_cool(executor)
finally:
    context.try_shutdown()
```
The first approach is shorter and (for some) conceptually easier. In both cases `context` can be referenced afterwards although it is invalid. However, in the first case it is much more clear that any sane use is within the `with context: ...` construct. And, one cannot forget to close it in the end. It is arguably more Pythonic.

#### Implementation considerations
One could use `shutdown` or `try_shutdown` at the end of the context block. The latter makes more sense, since it allows for more complex scenarios where the context is already in some of the cases.

clalancette:
This is a good idea, thanks.  We probably aren't going to work on this, but we would definitely review pull requests that implemented it.

felixdivo:
Okay nice. Is wrapping around `context.init()`/`context.try_shutdown()` even the right thing to do? Then maybe I can have a look. 😄

mjcarroll:
Yep, I think it maps the ROS concept cleanly into a more pythonic pattern.  Happy to take a look if you open a PR.

felixdivo:
There is also #1118, for `rclpy.executor.Executor`

clalancette:
We merged the PR implementing this a long time ago, so closing this out.

PR #1117, opened 2023-04-22
Turn Context into a ContextManager

felixdivo:
Implements  #1106

### DIFF
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -21,6 +21,7 @@ from typing import ContextManager
 from typing import Literal
 from typing import List
 from typing import Optional
+from typing import Type
 import weakref
 
 
@@ -161,8 +162,8 @@ class Context(ContextManager["Context"]):
 
     def __exit__(
         self,
-        exc_type: type[BaseException] | None,
-        exc_val: BaseException | None,
-        exc_tb: TracebackType | None,
-    ) -> Literal[False]:
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> None:
         self.try_shutdown()
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -34,6 +34,13 @@ class Context(ContextManager['Context']):
 
     Context objects should not be reused, and are finalized in their destructor.
     Wraps the `rcl_context_t` type.
+
+    :Example:
+        >>> from rclpy.context import Context
+        >>> with Context() as context:
+        >>>     context.ok()
+        True
+
     """
 
     def __init__(self):
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -18,7 +18,6 @@ import threading
 from types import TracebackType
 from typing import Callable
 from typing import ContextManager
-from typing import Literal
 from typing import List
 from typing import Optional
 from typing import Type
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -15,7 +15,10 @@
 from inspect import ismethod
 import sys
 import threading
+from types import TracebackType
 from typing import Callable
+from typing import ContextManager
+from typing import Literal
 from typing import List
 from typing import Optional
 import weakref
@@ -25,7 +28,7 @@ g_logging_configure_lock = threading.Lock()
 g_logging_ref_count = 0
 
 
-class Context:
+class Context(ContextManager["Context"]):
     """
     Encapsulates the lifecycle of init and shutdown.
 
@@ -149,3 +152,18 @@ class Context:
             raise RuntimeError('Context must be initialized before it can have a domain id')
         with self.__context, self._lock:
             return self.__context.get_domain_id()
+
+    def __enter__(self) -> "Context":
+        # We do not accept parameters here. If one wants to customize the init() call,
+        # they would have to call it manaully and not use the ContextManager convenience
+        self.init()
+        return self
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> Literal[False]:
+        self.try_shutdown()
+        return False  # Reraise possible exceptions
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -28,7 +28,7 @@ g_logging_configure_lock = threading.Lock()
 g_logging_ref_count = 0
 
 
-class Context(ContextManager["Context"]):
+class Context(ContextManager['Context']):
     """
     Encapsulates the lifecycle of init and shutdown.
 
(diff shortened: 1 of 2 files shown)


felixdivo:
Question: Wouldn't it be nice to have a similar thing for the global `init()`/`shutdown()` cycle of rclpy? It's easy to miss the `KeyboardInterrupt` handling or the `shutdown()` call for beginners, and for more experienced users it's onnying and distracting boilerplate.

Concretely, I'd propose something allowing to replace:

```python
import rclpy
from local_stuff import MyNodeClass

rclpy.init(args=args)

try:
    rclpy.spin(MyNodeClass())
except KeyboardInterrupt:
    pass
finally:
    rclpy.shutdown()
```

with 

```python
if __name__ == "__main__":
    with rclpy.console_runner() as executor:
        executor.spin(MyNodeClass())
```

Internally, we'd have `rclpy.console_runner()` call `rcly.get_global_executor()` and use the context in that. 

Maybe, `Executor` should also be a context manager?

rclpy/rclpy/context.py
```diff
@@ -149,3 +152,18 @@ def get_domain_id(self):
...
+        return self
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> Literal[False]:
+        self.try_shutdown()
+        return False  # Reraise possible exceptions
```
fujitatomoya, line 169: how about the following? (just a suggestion) could be more redundant?

```suggestion
        return exc_type is None  # Reraise possible exceptions
```

fujitatomoya (approved):
lgtm with comment.

rclpy/rclpy/context.py
felixdivo: Well, I think this could be explained clearer [in the Python docs](https://docs.python.org/3/library/stdtypes.html#contextmanager.__exit__). But I think always returning False is completely fine. If there was an exception, reraise it, and if not, the ouput does not matter. I like `False` because it is simple. I dpn't have strong opinions here.

felixdivo:
CC @clalancette

felixdivo:
Regarding my question above: I think that a `Executor` should be a context manager too. I'd add that to this PR. I also think that the other idea (`rclpy.console_runner()`) is worth a bit more discussion and should be deferred to an issue. If you agree that I should add the exector-context-manager and that it should be added to this PR, please give me a "👍" and I'll update this PR.

sloretz:
>  I think that a `Executor` should be a context manager too. I'd add that to this PR

I'd recommend making it a separate PR. I find it easier to review and merge multiple smaller PRs.

rclpy/rclpy/context.py
```diff
@@ -149,3 +152,18 @@ def get_domain_id(self):
             raise RuntimeError('Context must be initialized before it can have a domain id')
         with self.__context, self._lock:
             return self.__context.get_domain_id()
+
+    def __enter__(self) -> "Context":
+        # We do not accept parameters here. If one wants to customize the init() call,
+        # they would have to call it manaully and not use the ContextManager convenience
+        self.init()
```
sloretz, line 159: Ah, this makes me wish we had put the arguments to `init()` in `__init__` instead. Then users could have done:

```python3

with Context(initialize_logging=false) as ctx:
   ...
```

Not asking for you to fix this in this Pr, but I'm interested if you have any ideas.

rclpy/rclpy/context.py
```diff
@@ -149,3 +152,18 @@ def get_domain_id(self):
...
+        return self
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
```
sloretz, line 164: IIUC RHEL 9 and Windows still have Python < 3.10, so we'll need to use `Optional` here

* https://www.ros.org/reps/rep-2000.html#iron-irwini-may-2023-november-2024
* https://peps.python.org/pep-0604/

rclpy/rclpy/context.py
sloretz, line 169: Nit, IIUC no need for a return statement. Returning `False` looks to be the the same as the default return of `None`. https://docs.python.org/3/reference/datamodel.html#object.__exit__

sloretz (commented):
Just a couple nitpicks - otherwise I think the concept is great.

rclpy/rclpy/context.py
felixdivo: Whhop, missed that.

rclpy/rclpy/context.py
felixdivo: Yeah, and for me personally, the many `init()` calls are quite unexpected as a whole. As a Python user, I usually expect the `__init__()` to actually give me a fully initialized instance. I guess that's kind of like Resource acquisition is initialization (RAII) that I expect.

I guess you could gracefully transition to the `__init__()` by having the following interim solution for one release or something:
- Have `__init__()` take the arguments that `init()` takes, and if they are actually passed set things up right away. If they are not passed (which might need a trick to be detected), the user would need to call `init()`.
- The `init()` complains if the `__init__()` already set things up. If that was not the case, set things up here but show a deprecation warning pointing the user to move to `__init__()`.

Overall, I'm not sure whether it would even impact too many non-core ROS2 packages out there. I suppose there is often no need to handle `Context` object directly.

Maybe it makes sense to move this discussion into its own issue.

felixdivo:
The reviews have been addressed. If have no more to add.

felixdivo:
> Question: Wouldn't it be nice to have a similar thing for the global `init()`/`shutdown()` cycle of rclpy?

I opened the issue #1119 for discussing that.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18614/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13150/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19350/)

fujitatomoya:
@felixdivo can you address flake8 errors? see https://ci.ros2.org/job/ci_linux/18614/testReport/junit/(root)/projectroot/flake8/

felixdivo:
It's a bit weird to have `All checks have passed` in GitHub but there still are problems ^^

> @felixdivo can you address flake8 errors? 

Done. 🙂

felixdivo:
I added some docs. I'm unsure whether these doctests are actually run or how they are rendered in the docs.

clalancette:
> It's a bit weird to have `All checks have passed` in GitHub but there still are problems ^^

Yes, agreed.  The `Rpr` jobs don't install all of the same flake8 plugins as the https://ci.ros2.org jobs, so they can have different errors.  In the past, it wasn't possible to fix it due to some technical reasons, but those reasons have now gone away.  We just haven't had time to update the Rpr jobs to have the right plugins, but it is on our TODO list.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18639/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13169/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19369/)

sloretz (approved):

Issue #1097, opened 2023-03-27
Feature request: Get all declared parameters (and print them)


zkytony:
#### Feature description

I liked that in ROS1, when running roslaunch, it prints out all the parameters and their values. It is great for logging and debugging. I hope to do something like this more easily with ROS 2.

I wish there is a 'get_parameter_names' that returns all declared (and other important) parameters when I don't pass in anything. Or, when running 'ros2 launch', it prints out all parameters of each node being launched.

Currently (with ROS2 Humber), there is [get_parameters](https://docs.ros2.org/latest/api/rclpy/api/node.html?highlight=parameter#rclpy.node.Node.get_parameters) but it requires a list of names as input.

clalancette:
> I wish there is a 'get_parameter_names' that returns all declared (and other important) parameters when I don't pass in anything. Or, when running 'ros2 launch', it prints out all parameters of each node being launched.

The underlying functionality is all there; we have the call for `list_parameters` in rclcpp.  So this just needs an rclpy implementation.  If you are willing to take a shot, then you can look at https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp for some inspiration on how to implement it.

fujitatomoya:
@leeminju531 AFAIK, nobody is working on this. if you are willing to contribute, i am happy to do review 👍

leeminju531:
>  The underlying functionality is all there; we have the call for list_parameters in rclcpp. So this just needs an rclpy implementation. 

I believe that the functionality for listing all declared parameters is already present in rclpy. The implementation can be found in https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/parameter_service.py#L100#L140 .
From my understanding, this implementation looks for parameters based on a given prefix or returns all parameters if no prefix is provided. therefore, it may not be necessary to implement this feature.

If not, please let me know your thoughts on how to proceed.

fujitatomoya:
@leeminju531 

> I believe that the functionality for listing all declared parameters is already present in rclpy. The implementation can be found in https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/parameter_service.py#L100#L140 .

this is parameter service callback. so that it can prints all parameters via service.

i think the feature request here is to implement the method that belongs to `Node`? to print the all parameters that the node owns.

PR #1124, opened 2023-05-03
add list_parameters & test

leeminju531:
This addresses the issue (https://github.com/ros2/rclpy/issues/1097) by implementing the functionality for list_parameters that the node owns. 
This feature should help with logging and debugging by allowing users to easily retrieve all declared parameters and their values, similar to the functionality in ROS1.

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -38,6 +38,8 @@ from rcl_interfaces.msg import ParameterEvent
 from rcl_interfaces.msg import ParameterType
 from rcl_interfaces.msg import ParameterValue
 from rcl_interfaces.msg import SetParametersResult
+from rcl_interfaces.msg import ListParametersResult
+from rcl_interfaces.srv import ListParameters
 
 from rclpy.callback_groups import CallbackGroup
 from rclpy.callback_groups import MutuallyExclusiveCallbackGroup
@@ -953,6 +955,66 @@ class Node:
 
             # call post set parameter registered callbacks
             self._call_post_set_parameters_callback(parameter_list)
+
+        return result
+
+    def list_parameters(
+        self,
+        prefixes: List[str],
+        depth: int
+    ) -> ListParametersResult:
+        """
+        Get a list of parameter names and their prefixes.
+
+        :param prefixes: A list of prefixes to filter the parameter names.
+            Only the parameter names that start with any of the prefixes are returned.
+            If empty, all parameter names are returned.
+        :param depth: The depth of nested parameter names to include.
+        :return: The list of parameter names and their prefixes.
+        :raises: TypeError if the type of the passed argument is not expected types.
+        :raises: ValueError if depth is negative integer.
+        """
+        if not isinstance(prefixes, list):
+            raise TypeError('The prefixes argument must be a list')
+        if not all(isinstance(prefix, str) for prefix in prefixes):
+            raise TypeError('All prefixes must be instances of type str')
+        if not isinstance(depth, int):
+            raise TypeError('The depth must be instance of type int')
+        if depth < 0:
+            raise ValueError('The depth must be greater than or equal to zero')
+
+        result = ListParametersResult()
+
+        separator_less_than_depth: Callable[[str], bool] = \
+            lambda str: str.count(PARAMETER_SEPARATOR_STRING) < depth
+
+        recursive: bool = \
+            (len(prefixes) == 0) and (depth == ListParameters.Request.DEPTH_RECURSIVE)
+
+        for param_name in self._parameters.keys():
+            if not recursive:
+                get_all: bool = (len(prefixes) == 0) and (separator_less_than_depth(param_name))
+                if not get_all:
+                    prefix_matches = any(
+                        param_name == prefix or
+                        (
+                            param_name.startswith(prefix+PARAMETER_SEPARATOR_STRING) and
+                            (
+                                depth == ListParameters.Request.DEPTH_RECURSIVE or
+                                separator_less_than_depth(param_name[len(prefix)+1:])
+                            )
+                        )
+                        for prefix in prefixes
+                    )
+                    if not prefix_matches:
+                        continue
+
+            result.names.append(param_name)
+            last_separator = param_name.rfind(PARAMETER_SEPARATOR_STRING)
+            if last_separator != -1:
+                prefix = param_name[:last_separator]
+                if prefix not in result.prefixes:
+                    result.prefixes.append(prefix)
 
         return result
 
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -32,13 +32,13 @@ import weakref
 
 from rcl_interfaces.msg import FloatingPointRange
 from rcl_interfaces.msg import IntegerRange
+from rcl_interfaces.msg import ListParametersResult
 from rcl_interfaces.msg import Parameter as ParameterMsg
 from rcl_interfaces.msg import ParameterDescriptor
 from rcl_interfaces.msg import ParameterEvent
 from rcl_interfaces.msg import ParameterType
 from rcl_interfaces.msg import ParameterValue
 from rcl_interfaces.msg import SetParametersResult
-from rcl_interfaces.msg import ListParametersResult
 from rcl_interfaces.srv import ListParameters
 
 from rclpy.callback_groups import CallbackGroup
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -971,8 +971,8 @@ class Node:
             If empty, all parameter names are returned.
         :param depth: The depth of nested parameter names to include.
         :return: The list of parameter names and their prefixes.
-        :raises: TypeError if the type of the passed argument is not expected types.
-        :raises: ValueError if depth is negative integer.
+        :raises: TypeError if the type of any of the passed arguments is not an expected type.
+        :raises: ValueError if depth is a negative integer.
         """
         if not isinstance(prefixes, list):
             raise TypeError('The prefixes argument must be a list')
--- a/rclpy/test/test_node.py
+++ b/rclpy/test/test_node.py
@@ -28,6 +28,7 @@ from rcl_interfaces.msg import ParameterDescriptor
 from rcl_interfaces.msg import ParameterType
 from rcl_interfaces.msg import ParameterValue
 from rcl_interfaces.msg import SetParametersResult
+from rcl_interfaces.msg import ListParametersResult
 from rcl_interfaces.srv import GetParameters
 import rclpy
 from rclpy.clock import ClockType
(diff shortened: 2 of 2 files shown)


rclpy/test/test_node.py
```diff
@@ -991,6 +991,83 @@ def test_node_get_parameters_by_prefix(self):
...
+            ('foo_prefix.baz', 2.41),
+            ('bar_prefix.foo', 1),
+            ('bar_prefix.bar', 12.3),
+            ('bar_prefix.baz', 'world'),
+        ]
+        self.node.declare_parameters('', parameters)
+
+        parameter_names = self.node.list_parameters()
+        self.assertIsInstance(parameter_names, list)
+        self.assertEqual(len(parameter_names), 7)
```
fujitatomoya, line 1007:
```suggestion
        # Currently the only default parameter is 'use_sim_time', but that may change.
        # e.g) QoS override parameters
        self.assertEqual(len(parameter_names), len(parameters)+1)
```

rclpy/test/test_node.py
```diff
@@ -991,6 +991,83 @@ def test_node_get_parameters_by_prefix(self):
...
+                'bar_prefix.bar',
+                'bar_prefix.baz',
+                USE_SIM_TIME_NAME
+            ]
+        )
+
+        parameter_names = self.node.list_parameters(
+            prefixes=['foo_prefix', 'bar_prefix'])
+        self.assertIsInstance(parameter_names, list)
+        self.assertEqual(len(parameter_names), 6)
```
fujitatomoya, line 1024:
```suggestion
        self.assertEqual(len(parameter_names), len(parameters))
```

rclpy/test/test_node.py
```diff
@@ -991,6 +991,83 @@ def test_node_get_parameters_by_prefix(self):
         self.assertFalse(parameters)
         self.assertIsInstance(parameters, dict)
 
+    def test_node_list_parameters(self):
```
fujitatomoya, line 994: how about adding `self.assertRaises` test cases?

rclpy/test/test_node.py
```diff
@@ -991,6 +991,83 @@ def test_node_get_parameters_by_prefix(self):
...
+                'foo_prefix.bar',
+                'foo_prefix.baz',
+                'bar_prefix.foo',
+                'bar_prefix.bar',
+                'bar_prefix.baz'
+            ]
+        )
+
+        parameter_names = self.node.list_parameters(
+            prefixes=['foo_prefix'])
```
fujitatomoya, line 1038: how about adding the test case that prefix includes `.`? like `prefixes=['foo_prefix.']`

fujitatomoya (commented):
implementation looks good to me, probably we can add more test cases.

rclpy/rclpy/node.py
```diff
@@ -657,6 +657,61 @@ def get_parameters(self, names: List[str]) -> List[Parameter]:
...
+        DEPTH_RECURSIVE = 0
+
+        if prefixes is not None and not all(isinstance(prefix, str) for prefix in prefixes):
+            raise TypeError(
+                'All prefixes must be instances of type str or none')
+        if not isinstance(depth, int):
+            raise TypeError('The depth must be instances of type int')
+
+        if depth < 0:
+            depth = DEPTH_RECURSIVE
```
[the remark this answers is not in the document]
leeminju531: fixed.

rclpy/rclpy/node.py
```diff
@@ -657,6 +657,61 @@ def get_parameters(self, names: List[str]) -> List[Parameter]:
...
+        """
+        Get a list of parameter names.
+
+        :param prefixes: A list of prefixes to filter the parameter names.
+            Only the parameter names that start with any of the prefixes are returned.
+            If None, all parameter names are returned. Defaults to None.
+        :param depth: The depth of nested parameter names to include.
+            A value of 0 means that all levels are included. Defaults to 0.
+        :return: Fully-qualified names of the parameters, including their namespaces.
+        """
```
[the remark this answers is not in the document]
leeminju531: added.

rclpy/test/test_node.py
leeminju531: fixed.

rclpy/test/test_node.py
leeminju531: fixed

rclpy/test/test_node.py
leeminju531: added

rclpy/test/test_node.py
leeminju531: Thanks for the good test case. i think a prefix including `.` also be a valid option. right?
For this, i replaced the line
`if name.startswith(prefix + PARAMETER_SEPARATOR_STRING):` with 
`if parameter_name.startswith(prefix):`

I noticed that the current implementation in `parameter_service.py` still uses
the old line. I would be happy to update it if my proposed change is indeed correct.

fujitatomoya (approved):

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18692/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13218/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19422/)

Relativiteit:
@fujitatomoya Hey, I want to contribute for the first time, is there still a need for more testcases for this issue ?

rclpy/rclpy/node.py
```diff
@@ -657,6 +657,65 @@ def get_parameters(self, names: List[str]) -> List[Parameter]:
             raise TypeError('All names must be instances of type str')
         return [self.get_parameter(name) for name in names]
 
+    def list_parameters(self, prefixes: Optional[List[str]] = None, depth: int = 0) -> List[str]:
```
clalancette, line 660: I know we aren't totally consistent with this in `rclpy`, but I do think that we should in general aim to make these APIs work like the `rclcpp` ones.  In this case, I think that means that we should return a message of type `rcl_interfaces/msg/ListParametersResult`, rather than just a list of strings.

clalancette (changes requested):

fujitatomoya:
@leeminju531 sorry to be late to get back to this. https://github.com/ros2/rclpy/pull/1124#discussion_r1290195294 makes sense to me, could you address that? after that we can start the CI again.

leeminju531:
It appears to be a reasonable suggestion, and I'm fully prepared to make the necessary updates in the near future. 
Once these adjustments are in place, I will notify @clalancette  and @fujitatomoya to request a review.

leeminju531:
I have made revisions according to the suggestions. I kindly request @clalancette and @fujitatomoya to review the changes.

fujitatomoya (approved):

fujitatomoya:
@clalancette your comment is addressed, can you check again?

clalancette:
The more I look at this one, the more I'm curious why the implementation can't look like the [rclcpp one](https://github.com/ros2/rclcpp/blob/ba175922d3008704c67b10fe82a2f7c0cbee2be4/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L1032-L1087).  If it was more-or-less a Python port of that, it would be much easier to read (and reason about), since the logic would be the same.

@leeminju531 Can you maybe describe why the current implementation is like it is, rather than a direct port of the rclcpp code?

leeminju531:
I referred to [_list_parameters_callback](https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/parameter_service.py#L100#L140), where a similar process I intend to adopt has already been implemented ;however, i understood and agree your comments.

I've observed that the [service callbacks of ParameterService ](https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/parameter_service.py#L67#L166) mostly directly call the functions owned by the node, except for [_list_parameters_callback](https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/parameter_service.py#L100#L140), which is implemented directly. i  believe it should also referred to the function the node owns for consistency and maintainability.

I believe that the modification should be made in conjunction with your proposed comment.
@clalancette  What is your opinion on this?

clalancette:
> I've observed that the [service callbacks of ParameterService ](https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/parameter_service.py#L67#L166) mostly directly call the functions owned by the node, except for [_list_parameters_callback](https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/parameter_service.py#L100#L140), which is implemented directly. i believe it should also referred to the function the node owns for consistency and maintainability.
> 
> I believe that the modification should be made in conjunction with your proposed comment. @clalancette What is your opinion on this?

Yes, I think that all makes a lot of sense to me.  To recap:
1.  Update the implementation of `list_parameters` in node.py so it looks similar to the rclcpp version of the same.
2.  Update the implementation of `_list_parameters_callback` so that it calls the node.py implementation of `list_parameters`.

Please go ahead and do that, and I'm happy to review the result!

rclpy/rclpy/node.py
```diff
@@ -956,6 +958,66 @@ def _set_parameters_atomically_common(
...
+            if not recursive:
+                get_all: bool = (len(prefixes) == 0) and (separator_less_than_depth(param_name))
+                if not get_all:
+                    prefix_matches = any(
+                        param_name == prefix or
+                        (
+                            param_name.startswith(prefix+PARAMETER_SEPARATOR_STRING) and
+                            (
+                                depth == ListParameters.Request.DEPTH_RECURSIVE or
+                                separator_less_than_depth(param_name[len(prefix)+1:])
```
leeminju531, line 1004: The `separator_less_than_depth(param_name[len(prefix)+1:])` is the only different line compared to the [one in rclcpp](https://github.com/ros2/rclcpp/blob/ba175922d3008704c67b10fe82a2f7c0cbee2be4/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L1062-L1063).

I believe [the test case below](https://github.com/ros2/rclpy/pull/1124/files#diff-e2598f8cd62aef1be16dd72039bf6a457f42521b41f54dedd3831b2b1ee01425R1275-R1296) should be available since the depth is relative to the given prefixes. However, if this is present in `separator_less_than_depth(param_name[len(prefix):])`, it fails. (it passes if the depth is 2 or more)
If this understanding is correct, the one in rclcpp also should be fixed.

leeminju531:
Please review the changes at your convenience. @clalancette

rclpy/rclpy/node.py
clalancette: That is a really good call.  You are right that that test should work.

Do you mind submitting a pull request to `rclcpp` that fixes the bug and adds a test for it?

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19461/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13984/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20184/)

rclpy/rclpy/node.py
leeminju531: :ok_hand: Please check [this](https://github.com/ros2/rclcpp/pull/2300).

clalancette: Thank you!

Issue #1123, opened 2023-05-02
Race condition in  `ActionClient.*_goal_async` functions which result in a futures that never complete


apockill:
## Bug report

Hi all, thank you for maintaining this wonderful library! 

I've been encountering an intermittent bug that has left my head scratching. Unfortunately recently it started occurring more often, so I sat down and spent some time tracking down what was going on. Now, I've got a good idea of the problem- but I need some more expert eyes to decide on a good solution. I'll lay out all my findings and some perceived paths forward in this bug report.

**Required Info:**

- Operating System:
  - Ubuntu 22.04, within a docker image that is also running Ubuntu 22.04
- Installation type:
  - Binaries as installed from ros2 apt.
- Version or commit hash:
  - rclpy==3.3.7, python3.10
- DDS implementation:
  - CycloneDDS
- Client library (if applicable):
  - rclpy==3.3.7, python3.10

#### Steps to reproduce issue

There are multiple places where this race condition manifests itself. I will demonstrate how to reproduce it when calling `send_goal_async`. Take the `ActionClient.send_goal_async` and add a `sleep(#)`  directly below the line `sequence_number = self._client_handle.send_goal_request(request)`. For example:

```python3
    def send_goal_async(self, goal, feedback_callback=None, goal_uuid=None):
        """
        Send a goal and asynchronously get the result.

        The result of the returned Future is set to a ClientGoalHandle when receipt of the goal
        is acknowledged by an action server.

        :param goal: The goal request.
        :type goal: action_type.Goal
        :param feedback_callback: Callback function for feedback associated with the goal.
        :type feedback_callback: function
        :param goal_uuid: Universally unique identifier for the goal.
            If None, then a random UUID is generated.
        :type: unique_identifier_msgs.UUID
        :return: a Future instance to a goal handle that completes when the goal request
            has been accepted or rejected.
        :rtype: :class:`rclpy.task.Future` instance
        :raises: TypeError if the type of the passed goal isn't an instance of
          the Goal type of the provided action when the service was
          constructed.
        """

        if not isinstance(goal, self._action_type.Goal):
            raise TypeError()

        request = self._action_type.Impl.SendGoalService.Request()
        request.goal_id = (
            self._generate_random_uuid() if goal_uuid is None else goal_uuid
        )
        request.goal = goal
        sequence_number = self._client_handle.send_goal_request(request)
        sleep(5)   # <--- this line causes this race condition nearly 100% of the time

[...]

PR #1125, opened 2023-05-04
draft: avoid race condition in action client

aditya2592:
work in progress PR to address https://github.com/ros2/rclpy/issues/1123

### DIFF
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -30,6 +30,7 @@ from rclpy.type_support import check_for_type_support
 from rclpy.waitable import NumberOfEntities, Waitable
 
 from unique_identifier_msgs.msg import UUID
+WAIT_TIMEOUT = 15.0
 
 class ClientGoalHandle():
     """Goal handle for working with Action Clients."""
@@ -286,7 +287,7 @@ class ActionClient(Waitable):
         call any user-defined callbacks (e.g. feedback).
         """
         if 'goal' in taken_data:
-            if self._goal_event.wait(timeout=5.0):
+            if self._goal_event.wait(timeout=WAIT_TIMEOUT):
                 self._goal_event.clear()
                 sequence_number, goal_response = taken_data['goal']
                 if sequence_number in self._goal_sequence_number_to_goal_id:
@@ -314,7 +315,7 @@ class ActionClient(Waitable):
                 )
 
         if 'cancel' in taken_data:
-            if self._cancel_event.wait(timeout=5.0):
+            if self._cancel_event.wait(timeout=WAIT_TIMEOUT):
                 self._cancel_event.clear()
                 sequence_number, cancel_response = taken_data['cancel']
                 if sequence_number in self._pending_cancel_requests:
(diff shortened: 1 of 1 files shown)


clalancette:
Thank you for the contribution.

First, can you please retarget this to rolling?  We'll review it there first, then we can consider backporting it to `humble`.

Second, can you please rebase your commits and add in a `Signed-off-by` line?  The DCO bot has to pass for us to accept the commits.

Finally, can you please add some context to the initial description saying exactly what this is fixing?  That would help us review it.

apockill:
HI @clalancette - I'm just following up to say this PR is related to the bug described in https://github.com/ros2/rclpy/issues/1123, which I posted yesterday.

To summarize my bug report here: There are race conditions in the "async" functions on the ActionClient for sending goals, cancelling goals, and getting results. The core of the problem is that the underling `_client_handle` can sometimes finish its request before the `sequence_number` has been tracked in the `_goal_sequence_number_to_goal_id` (or any number of other dicts that are initialized for the purpose of tracking ongoing requests of different types.

The symptom ends up being that you have a request that never completes (blocking forever), and you see the following log in console:

```
Ignoring unexpected goal response. There may be more than one action server for the action '{ACTION_NAME}'
```

This PR seems to try out a solution where all of those dictionaries for tracking requests on the python side are protected by a lock. I think there might be multiple ways to tackle this, and this is one of them.

aditya2592:
Apologies for not linking the issue, I am currently still iterating on the fix and the PR is not ready yet. I have it directed to humble to test easily with my setup and planning to redirect to rolling if everything works and fix aligns with recommendation of the repo authors.

dcconner:
You might consider doing a squash merge of your development efforts, and then based your PR off of the single commit.

fujitatomoya:
> You might consider doing a squash merge of your development efforts, and then basing your PR off of the single commit.

i do not think this is needed since that is own development branch, and we can do so when we take the PR to mainline.

apockill:
Any thoughts on this proposal as a resolution? https://github.com/ros2/rclpy/issues/1123#issuecomment-1551836772 

If anyone whose familiar with the ActionClient has opinions on their preferred resolution method, I'm happy to make a PR based off of them! This race condition affects humble up to current.

rclpy/rclpy/action/client.py
```diff
@@ -30,7 +30,7 @@
 from rclpy.waitable import NumberOfEntities, Waitable
 
 from unique_identifier_msgs.msg import UUID
-
+WAIT_TIMEOUT = 15.0
```
fujitatomoya, line 33: never used?

rclpy/rclpy/action/client.py
```diff
@@ -177,6 +176,11 @@ def __init__(
         self._result_sequence_number_to_goal_id = {}
         # key: UUID in bytes, value: callback function
         self._feedback_callbacks = {}
+        self._data_lock = threading.Lock()
```
fujitatomoya, line 179: I would suggest we can introduce the threading lock against the specific member data object, instead of having giant lock in the `ActionClient`. that can be also more performative.

see https://github.com/ros2/rclcpp/blob/e2965831d51e9be03470cb07f8721012afcade9b/rclcpp_action/src/client.cpp#L112-L119

rclpy/rclpy/action/client.py
```diff
@@ -177,6 +176,11 @@ def __init__(
...
+        self._goal_event = threading.Event()
+        self._result_event = threading.Event()
+        self._cancel_event = threading.Event()
```
fujitatomoya, line 182: never used?

fujitatomoya (commented):
1st we need to target this to `rolling` as @clalancette requested.

fujitatomoya:
https://github.com/ros2/rclpy/pull/1308 replaces this PR.

PR #1129, opened 2023-05-21
Fix iteration over modified list

felixdivo:
See [the referenced StackOverflow article](https://stackoverflow.com/questions/1207406/how-to-remove-items-from-a-list-while-iterating) for why this change is necessary.

It is possible [to actually trigger this fault](https://github.com/felixdivo/ros2-easy-test/actions/runs/5035009110/jobs/9030212775#step:9:116).


rclpy/rclpy/executors.py
```diff
@@ -802,10 +802,13 @@ def _spin_once_impl(
         else:
             self._executor.submit(handler)
             self._futures.append(handler)
-            for future in self._futures:  # check for any exceptions
+            self._futures.append(handler)
```
clalancette, line 805: This looks like a typo; we shouldn't append the handler to tlist of `futures` twice, I don't think.

clalancette (changes requested):

rclpy/rclpy/executors.py
```diff
@@ -802,10 +802,13 @@ def _spin_once_impl(
...
+            # make a copy of the list that we iterate over while modifying it
+            # (https://stackoverflow.com/q/1207406/3753684)
+            for future in self._futures[:]:
                 if future.done():
                     self._futures.remove(future)
```
fujitatomoya, line 810: Just a comment to share. this will be slower, since it has to scan the list every iteration to remove.

fujitatomoya (commented):
@felixdivo thanks for the PR.

Do you happen to know how https://github.com/felixdivo/ros2-easy-test/actions/runs/5035009110/jobs/9030212775#step:9:116 can happen in theory? i could not figure this out how even this happens...

rclpy/rclpy/executors.py
felixdivo: Yes, you are right! Sorry. Also suspicious that the tests did not catch this ...

```suggestion
```

rclpy/rclpy/executors.py
felixdivo: Jup, true. That's why I also commented on it, to explain why you'd do something like this.

felixdivo:
> Do you happen to know how https://github.com/felixdivo/ros2-easy-test/actions/runs/5035009110/jobs/9030212775#step:9:116 can happen in theory? i could not figure this out how even this happens...

I guess in my case it could be a race condition, since I access the executor from different threads, where I think it makes sense that it can pccur. I accidentially found #1079, so maybe that is not supported anyways. It also does not happen determinisitcally. It just caused me to stare at these lines for a while and that's where I realized that there might be an issue.

rclpy/rclpy/executors.py
felixdivo: One could run some benchmarks to see if it affects performance. I don't know whether this is done regularly. 🤷 I don't plan on doing this.

felixdivo:
Side note: Why are these signoff things required (I did read [CONTRIBUTING.md](https://github.com/ros2/rclpy/blob/rolling/CONTRIBUTING.md))? I see basically no other projects use them and it makes contribting kind of a pain. E.g. we cannot add changes via comments and the defaults in most IDEs do not include this piece of text. Sure, no big deal but if I'm changing one line (e.g. some small typo, type annotation, ...) online and it has me clone the repo, undo commits, add the sign-offs, ... it makes contributing many times slower and often keeps me from fixing stuff because it is plainly annoying. Is this really legally required?

rclpy/rclpy/executors.py
fujitatomoya: that was just a comment, i do not think we need to benchmark test for this.

fujitatomoya (approved):

clalancette:
> Side note: Why are these signoff things required (I did read [CONTRIBUTING.md](https://github.com/ros2/rclpy/blob/rolling/CONTRIBUTING.md))?

In short, it is the lightest-weight thing we have to assure that contributors are allowed to contribute the code that they are providing.  The other option would be a CLA, which is far more cumbersome and invasive.

> E.g. we cannot add changes via comments and the defaults in most IDEs do not include this piece of text. 

For what it is worth, you can always type it in by hand when making the commit message, and that will make the bot happy.  Presumably you can also convince your IDE to append it to the commit message by default, particularly since it is a standard git commit flag (`-s`).

fujitatomoya:
@felixdivo if you are comfortable with https://github.com/ros2/rclpy/pull/1129#issuecomment-1559931633, could you address DCO error? and i will start the CI.

felixdivo:
@clalancette @fujitatomoya Thank you for explaining this. It's really a shame that we have to do this. I know you probably thought a lot about this, but a CLA is only signed once, which IMHO makes it easier. And regarding the legal consequences, it should not be more invasive.

Anyways, let's drop this discussion here since it has nothing to do with the changes in this PR. 😅

fujitatomoya (approved):

fujitatomoya:
@clalancette are you good to go?

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18810/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13343/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19553/)

PR #1132, opened 2023-06-07
Fix/param namespace association 894

### LINKED ISSUE #894: Cannot load parameters declared with namespace from yaml or command line
When using the following syntax to create parameters in a namespace, the parameters will not be loaded with values from either the command line or config yaml params file. Specifically the namespace= argument seems to be the issue.
```python
        self.declare_parameters(
            namespace="ns1",
            parameters=[("bar", "default_bar1")],
        )
```

If we instead use this syntax, simply adding the ns1 prefix to the parameter itself then both params file and command line works. According to documentation this should be equivalent to the first mode.
```python
        self.declare_parameters(
            namespace="",
            parameters=[("ns1.bar", "default_bar1")],
        )
```

**Required Info:**

- Operating System:
  - Ubuntu 20.04 on x86_64 and arm_64 (Rpi4)
- Installation type:
  - binaries from apt repo
- Version or commit hash:
  - latest galactic on amd64 (ros-galactic-rclpy/now 1.9.0-1focal.20211222.234056)
  - latest galactic on arm64 RPi4 (ros-galactic-rclpy/focal,now 1.9.0-1focal.20220203.003739)
- DDS implementation:
  - Cyclone
  - rclpy

#### Steps to reproduce issue
Retrieve the sample test from this github gist:
https://gist.github.com/guru-florida/37c9d89adb082a2329dec709e58a045d

#### Expected behavior
Both declaration modes should have equivalent behavior. Both modes do in fact result in the nodes _parameters
collection having the expected namespace prefix.
This example program should set params {foo, ns1.bar, ns2.bar} to {p1, p2, p3} respectively.

#### Actual behavior
Parameters using the namespace argument of declare_parameters() cannot be loaded from sysargs (using -p) or
from yaml config files.
This example program outputs params {foo, ns1.bar, ns2.bar} as {p1, default_bar1, p3} respectively.

#### Additional information
Inspecting at runtime, it seems that both modes add the parameter to the node._parameters dict with the proper prefix (so same behavior) but when the declare_parameters() eventually calls set_parameters() it is unable to find the ns1.bar parameter even though it is supplied. Furthermore, if you supply the parameter on the command line or in the params file without the namespace then it will find and use this parameter and place into the namespaced parameter. (I.e. In the example command line, change "-p ns1.bar:=p2" to "-p bar:=wrong" and it "ns1.bar" will be set to "wrong" on program output)

Also, there is a ROS Answers question alluding to the same issue:
https://answers.ros.org/question/395375/ros2-how-to-specify-parameter-with-namespace-in-python-launch-file/

I am working on diagnosing further but figured I had enough to trigger an issue.

devrite:
Fixes #894

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -453,13 +453,19 @@ class Node:
             # Get the values from the tuple, checking its types.
             # Use defaults if the tuple doesn't contain value and / or descriptor.
             name = parameter_tuple[0]
-            second_arg = parameter_tuple[1] if 1 < len(parameter_tuple) else None
-            descriptor = parameter_tuple[2] if 2 < len(parameter_tuple) else ParameterDescriptor()
-
             if not isinstance(name, str):
                 raise TypeError(
                         f'First element {name} at index {index} in parameters list '
                         'is not a str.')
+            if namespace:
+                name = f'{namespace}.{name}'
+
+            # Note(jubeira): declare_parameters verifies the name, but set_parameters doesn't.
+            validate_parameter_name(name)
+
+            second_arg = parameter_tuple[1] if 1 < len(parameter_tuple) else None
+            descriptor = parameter_tuple[2] if 2 < len(parameter_tuple) else ParameterDescriptor()
+
             if not isinstance(descriptor, ParameterDescriptor):
                 raise TypeError(
                     f'Third element {descriptor} at index {index} in parameters list '
@@ -502,12 +508,6 @@ class Node:
             # Get value from parameter overrides, of from tuple if it doesn't exist.
             if not ignore_override and name in self._parameter_overrides:
                 value = self._parameter_overrides[name].value
-
-            if namespace:
-                name = f'{namespace}.{name}'
-
-            # Note(jubeira): declare_parameters verifies the name, but set_parameters doesn't.
-            validate_parameter_name(name)
 
             parameter_list.append(Parameter(name, value=value))
             descriptors.update({name: descriptor})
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):
lgtm, @devrite thanks for the contribution.

@jacobperron can you do another review?

rclpy/rclpy/node.py
```diff
@@ -453,6 +453,11 @@ def declare_parameters(
             # Get the values from the tuple, checking its types.
             # Use defaults if the tuple doesn't contain value and / or descriptor.
             name = parameter_tuple[0]
+            if namespace:
+                name = f'{namespace}.{name}'
+
+            # Note(jubeira): declare_parameters verifies the name, but set_parameters doesn't.
+            validate_parameter_name(name)
```
clalancette, line 460: I think we need to move a bit more code around.  In particular, if `parameter_tuple[0]` was not passed as a `string`, then we could potentially have an issue here.  I'll suggest that we move the `if not isinstance(name, str)` check from lines 464 below to right after we assign `parameter_tuple[0]` to `name`.

clalancette (changes requested):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18953/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13485/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19706/)

clalancette:
@devrite Due to some unrelated changes, this needs to be rebased onto the `rolling` branch to pass CI.  Let me know once you've done that and I'll kick off another round.

devrite:
Cleared for launch!

PR #1138, opened 2023-07-02
Add in the ability to start timers paused.

### LINKED ISSUE #1137: add autostart timer option for the Timer class
#### Feature description

`rclpy` counterpart of https://github.com/ros2/rclcpp/pull/2005

see more details for https://github.com/ros2/rclcpp/issues/1980

`rclpy` implementation also depends on https://github.com/ros2/rcl/pull/1004, with adding `autostart` option flag to `Timer` class.

clalancette:
This was possible in the past by creating a timer
and then immediately canceling it.  But that could lead to a potential race where the timer fired once before it was canceled.  By allowing for the user
to create it paused, we avoid that race.

This is the rclpy version of https://github.com/ros2/rclcpp/pull/2005

@Voldivh @fujitatomoya FYI.  This fixes #1137

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1676,7 +1676,7 @@ class Node:
         callback: Callable,
         callback_group: Optional[CallbackGroup] = None,
         clock: Optional[Clock] = None,
-        autostart: Optional[bool] = True,
+        autostart: bool = True,
     ) -> Timer:
         """
         Create a new timer.
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1676,6 +1676,7 @@ class Node:
         callback: Callable,
         callback_group: Optional[CallbackGroup] = None,
         clock: Optional[Clock] = None,
+        autostart: Optional[bool] = True,
     ) -> Timer:
         """
         Create a new timer.
@@ -1694,7 +1695,9 @@ class Node:
             callback_group = self.default_callback_group
         if clock is None:
             clock = self._clock
-        timer = Timer(callback, callback_group, timer_period_nsec, clock, context=self.context)
+        timer = Timer(
+            callback, callback_group, timer_period_nsec, clock, context=self.context,
+            autostart=autostart)
 
         callback_group.add_entity(timer)
         self._timers.append(timer)
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1681,14 +1681,18 @@ class Node:
         """
         Create a new timer.
 
-        The timer will be started and every ``timer_period_sec`` number of seconds the provided
-        callback function will be called.
+        If autostart is ``True`` (the default), the timer will be started and every
+        ``timer_period_sec`` number of seconds the provided callback function will be called.
+        If autostart is ``False``, the timer will be created but not started; it can then be
+        started by calling ``reset()`` on the timer object.
 
-        :param timer_period_sec: The period (s) of the timer.
+        :param timer_period_sec: The period (in seconds) of the timer.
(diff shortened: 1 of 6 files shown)


rclpy/rclpy/node.py
```diff
@@ -1676,6 +1676,7 @@ def create_timer(
         callback: Callable,
         callback_group: Optional[CallbackGroup] = None,
         clock: Optional[Clock] = None,
+        autostart: Optional[bool] = True,
```
EsipovPA, line 1679: Maybe it is better just to do?
```suggestion
        autostart: bool = True,
```
As I get it from the description of the `Optional` [here](https://docs.python.org/3/library/typing.html#typing.Optional), it should represent a parameter with default `None`. I'd do it the same way in `timer.py`, in the `__init__` parameters section, [here](https://github.com/ros2/rclpy/pull/1138/files#diff-0860061fc25c955062f5fde6aeafebddccc61ebb3728746c781521106b0937e6R38).

rclpy/rclpy/timer.py
```diff
@@ -14,19 +14,34 @@
...
 class Timer:
 
-    def __init__(self, callback, callback_group, timer_period_ns, clock, *, context=None):
+    def __init__(
+        self,
+        callback: Callable,
+        callback_group: CallbackGroup,
+        timer_period_ns: int,
+        clock: Clock,
+        *,
```
EsipovPA, line 36: Should not it be `*args`? And should not it be positioned after all the other parameters? If so, you will not need to combine named and unnamed parameters [here](https://github.com/ros2/rclpy/pull/1138/files#diff-610fc172e919c8a045bd8f7aec9732a4830668c8bace69bc95132bd07aedbf31R1699)

rclpy/test/test_waitable.py
```diff
@@ -131,7 +131,7 @@ def __init__(self, node):
         period_nanoseconds = 10000
         with self._clock.handle, node.context.handle:
             self.timer = _rclpy.Timer(
-                self._clock.handle, node.context.handle, period_nanoseconds)
+                self._clock.handle, node.context.handle, period_nanoseconds, True)
```
EsipovPA, line 134: Whoul'd not this `True` value be added to the parameters list of this `__init__` method? Maybe with default `True` value?

rclpy/rclpy/node.py
clalancette: Yes, you are totally correct.  Fixed in 14192f30c7313543868b48b562367b32ff5f26c9

rclpy/rclpy/timer.py
clalancette: No, there should be no changes here.  We want this to be a kwargs, just like `context`.

rclpy/test/test_waitable.py
clalancette: I don't really see a need to.  This is for a specific test, and this test wants the timer to be automatically started, so there is no reason in my opinion to make it configurable.

fujitatomoya (approved):
lgtm 👍 

maybe we could add doc section that says timer with `autostart = false` can be started with calling `reset` method.

clalancette:
> maybe we could add doc section that says timer with `autostart = false` can be started with calling `reset` method.

Excellent point.  I've added documentation in fa71994.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19064/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13597/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19816/)

PR #1143, opened 2023-07-17
Fix spin_once_until_future_complete to quit when the future finishes.

### LINKED ISSUE #989: SingleThreadedExecutor::spin_once_until_future_complete continues to spin even if future is completed
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - source build
- Version or commit hash:
  - https://github.com/ros2/ros2/commit/5a35507c37f3b0449e31cb4073cf0352bfd52b0b
- DDS implementation:
  - Any (Fast-RTPS, RTI Connext)
  - rclpy

#### Expected behavior

`SingleThreadedExecutor::spin_once_until_future_complete` should return if the future is completed

#### Additional information

https://github.com/ros2/rclpy/blob/9a5d55ec513a096e63e2549daee6cade9126ba67/rclpy/rclpy/executors.py#L725-L726

clalancette:
This makes it match the function name much more closely. While we are in here, also fix a bug where the multi-threaded version would not quit immediately after the future completes. And also add tests for both of these situations.

This should fix #989

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -745,8 +745,6 @@ class SingleThreadedExecutor(Executor):
     def __init__(self, *, context: Optional[Context] = None) -> None:
         super().__init__(context=context)
 
-        self._futures = []
-
     def _spin_once_impl(
         self,
         timeout_sec: Optional[float] = None,
@@ -766,13 +764,7 @@ class SingleThreadedExecutor(Executor):
             if handler.exception() is not None:
                 raise handler.exception()
 
-            self._futures.append(handler)
-            # make a copy of the list that we iterate over while modifying it
-            # (https://stackoverflow.com/q/1207406/3753684)
-            for future in self._futures[:]:
-                if future.done():
-                    self._futures.remove(future)
-                    future.result()  # raise any exceptions
+            handler.result()  # raise any exceptions
 
     def spin_once(self, timeout_sec: Optional[float] = None) -> None:
         self._spin_once_impl(timeout_sec)
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -772,7 +772,7 @@ class SingleThreadedExecutor(Executor):
             for future in self._futures[:]:
                 if future.done():
                     self._futures.remove(future)
-                    future.result()  # re-raise any exceptions
+                    future.result()  # raise any exceptions
 
     def spin_once(self, timeout_sec: Optional[float] = None) -> None:
         self._spin_once_impl(timeout_sec)
@@ -844,7 +844,7 @@ class MultiThreadedExecutor(Executor):
             for future in self._futures[:]:
                 if future.done():
                     self._futures.remove(future)
-                    future.result()  # re-raise any exceptions
+                    future.result()  # raise any exceptions
 
     def spin_once(self, timeout_sec: Optional[float] = None) -> None:
         self._spin_once_impl(timeout_sec)
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -745,24 +745,45 @@ class SingleThreadedExecutor(Executor):
     def __init__(self, *, context: Optional[Context] = None) -> None:
         super().__init__(context=context)
 
-    def spin_once(self, timeout_sec: Optional[float] = None) -> None:
+        self._futures = []
+
+    def _spin_once_impl(
+        self,
+        timeout_sec: Optional[float] = None,
+        wait_condition: Callable[[], bool] = lambda: False
+    ) -> None:
         try:
-            handler, entity, node = self.wait_for_ready_callbacks(timeout_sec=timeout_sec)
+            handler, entity, node = self.wait_for_ready_callbacks(
+                timeout_sec, None, wait_condition)
         except ShutdownException:
             pass
         except TimeoutException:
+            pass
+        except ConditionReachedException:
             pass
         else:
             handler()
             if handler.exception() is not None:
                 raise handler.exception()
 
+            self._futures.append(handler)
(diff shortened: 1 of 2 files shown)


wjwwood (approved):
lgtm, I had one question, but I don't think it's necessarily blocking.

rclpy/rclpy/executors.py
```diff
@@ -745,24 +745,45 @@ class SingleThreadedExecutor(Executor):
...
         try:
-            handler, entity, node = self.wait_for_ready_callbacks(timeout_sec=timeout_sec)
+            handler, entity, node = self.wait_for_ready_callbacks(
+                timeout_sec, None, wait_condition)
         except ShutdownException:
             pass
         except TimeoutException:
             pass
+        except ConditionReachedException:
+            pass
```
fujitatomoya, line 763: just a comment to make sure if i understand correctly.

if the condition (future) is reached, leads to this exception to pass. this is the one of the case that `wait_condition` is already completed. (directly related to https://github.com/ros2/rclpy/issues/989)

rclpy/rclpy/executors.py
```diff
@@ -745,24 +745,45 @@ class SingleThreadedExecutor(Executor):
...
             if handler.exception() is not None:
                 raise handler.exception()
 
+            self._futures.append(handler)
+            # make a copy of the list that we iterate over while modifying it
+            # (https://stackoverflow.com/q/1207406/3753684)
+            for future in self._futures[:]:
+                if future.done():
+                    self._futures.remove(future)
+                    future.result()  # re-raise any exceptions
```
fujitatomoya, line 775: Do we need to re-raise the exception here again? the exception comes from user callback is already managed above for `SingleThreadedExecutor` case?

fujitatomoya (commented):
@clalancette thanks for the PR. I just have a question before lgtm.

rclpy/rclpy/executors.py
```diff
@@ -745,24 +745,45 @@ class SingleThreadedExecutor(Executor):
...
+
+    def spin_once(self, timeout_sec: Optional[float] = None) -> None:
+        self._spin_once_impl(timeout_sec)
```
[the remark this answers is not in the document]
clalancette: I'm not sure I totally understand the question, but let me explain what I understand happens to the futures in both `spin_once` and `spin_once_until_future_complete`.

When you are calling `spin_once`, then what happens is that our `wait_condition` for `_spin_once_impl` is the default `lambda: False`.  That means that that condition will never cause the work loop (in `_wait_for_ready_callbacks`) to return early, and so that function will always find one piece of work (assuming it doesn't time out or shutdown).  When it does find that piece of work, it executes it immediately in `_spin_once_impl`.  After that we add it to the `_futures` list to check on whether it is done.  We then *also* look at any other futures that were pending, and if those are done we execute the done handlers for those and remove them from the list.

When you are calling `spin_once_until_future_complete`, the logic is largely the same except that our `wait_condition` is the done callback on the future.  So if the future finishes before we find other work to do, that can cause `wait_for_ready_callbacks` to return early.  If that is the case, then it actually raises `ConditionReachedException`, because we have already called the done callback.  And thus it is never added to `_futures`.

Does that make sense to you?  And does that answer your question at all?

rclpy/rclpy/executors.py
clalancette: Yes, exactly.  From what I can tell looking at the code, the reason it is done this way is because we had to call the `wait_condition` to discover whether it was completed.  If it *was* completed, then presumably it changed some state already (like setting the future to done, etc), and so we don't want to make that call again.

rclpy/rclpy/executors.py
clalancette: I *think* we need this.  The reason is that the handler exception above is only for the current handler; if we had other futures in this list, then it may be possible that one of them is "done" because of an exception, and we need to notify the user about that.

methylDragon: I guess it's less so a re-raise as it is just raising the exception? (Since if the exception was raised in line 767, the handler wouldn't get added to the future at all?)

fujitatomoya: I may be mistaken, but do we need to have `SingleThreadedExecutor::_futures`? handler is executed on this thread in line 765 with checking the exception to re-raise it for user application every?

clalancette: > I guess it's less so a re-raise as it is just raising the exception? (Since if the exception was raised in line 767, the handler wouldn't get added to the futures list at all?)

Yeah, I think that is correct.  I copied that comment from the `MultiThreadedExecutor` implementation, but it is indeed just a "raise".  I'll update the comment to reflect that.

clalancette: > I may be mistaken, but do we need to have `SingleThreadedExecutor::_futures`? handler is executed on this thread in line 765 with checking the exception to re-raise it for user application every?

Actually, this is a good question.  In the `SingleThreadedExecutor` case, we are executing the handler directly (not submitting it to a ThreadPool to "eventually" be run).  So there is really no reason to have `self._futures` at all; we can just run the handler, then raise any exceptions immediately by just running `handler.result`.  I'll try that out.

clalancette: OK, yeah, that seems to work.  So that is now done in fd30442.

fujitatomoya (approved):
lgtm

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19262/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13805/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19993/)

PR #1153, opened 2023-08-23
Cleanup of test_node.py.

clalancette:
These tests weren't always properly cleaning up after themselves, and were sometimes improperly using the setUp{Class} fixtures.  They also didn't always clean up between tests, so sometimes the same node was reused over and over again.  Clean all that up here, which should make the tests much more robust.

### DIFF
--- a/rclpy/test/test_node.py
+++ b/rclpy/test/test_node.py
@@ -66,14 +66,18 @@ class TestNodeAllowUndeclaredParameters(unittest.TestCase):
     def setUpClass(cls):
         cls.context = rclpy.context.Context()
         rclpy.init(context=cls.context)
-        cls.node = rclpy.create_node(
-            TEST_NODE, namespace=TEST_NAMESPACE, context=cls.context,
-            allow_undeclared_parameters=True)
 
     @classmethod
     def tearDownClass(cls):
-        cls.node.destroy_node()
         rclpy.shutdown(context=cls.context)
+
+    def setUp(self):
+        self.node = rclpy.create_node(
+            TEST_NODE, namespace=TEST_NAMESPACE, context=self.context,
+            allow_undeclared_parameters=True)
+
+    def tearDown(self):
+        self.node.destroy_node()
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19382/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13907/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20102/)

rclpy/test/test_node.py
```diff
@@ -66,15 +66,19 @@ class TestNodeAllowUndeclaredParameters(unittest.TestCase):
...
         rclpy.shutdown(context=cls.context)
 
+    def setUp(self):
+        self.node = rclpy.create_node(
+            TEST_NODE, namespace=TEST_NAMESPACE, context=self.context,
+            allow_undeclared_parameters=True)
+
+    def tearDown(self):
+        self.node.destroy_node()
+
```
fujitatomoya, line 81: totally agree on this. especially adding test case into this class, probably this is likely that developers would expect.

rclpy/test/test_node.py
```diff
@@ -550,12 +508,66 @@ def test_node_has_parameter_services(self):
         )
 
 
+class TestExecutor(unittest.TestCase):
```
fujitatomoya, line 511: this could move to `test_executor.py`? but it looks like `test_executor.py` tests executor implementation, but here does interaction from node instance. so probably we can ignore this comment.

fujitatomoya (approved):
lgtm

rclpy/test/test_node.py
clalancette: Yeah, that one is a bit subjective.  I could see it going into either file.  For the sake of not making this PR any bigger, I'm going to leave this where it is; we can always move it later.

Thanks for the review!

Issue #954, opened 2022-06-14
Expose remapped service name for service clients


Bjoernolav:
#### Expose remapped service name for service clients
I'm writing a node connecting to two services. For the sake of generality, I'm using a name in the private namespace of the node, and intend to use remapping to connect to the actual service. Furthermore, I'm using `wait_for_service` to make sure that the service is available before continuing.

In the case where the service is unavailable, I would like to inform the user that the service was unavailable, and also including the remapped name in the log statement. It seems like it is not possible to get the remapped name for service clients yet?

Example:
```python
import rclpy
from rclpy.node import Node
from std_srvs.srv import Trigger

class TestNode(Node):
    def __init__(self, **kwargs) -> None:
        if 'node_name' not in kwargs:
            kwargs['node_name'] = 'test_node'
        super().__init__(**kwargs)
        self.client = self.create_client(Trigger, '~/service_name')
        if not self.client.wait_for_service(1.0):
            msg = f'No connection to service: {self.client.srv_name}'
            self.get_logger().fatal(msg)
            raise RuntimeError(msg)
        ....
```

Running this using `ros2 run pkg_name exec_name --ros_args -r '~/service_name:=/test'`, it seems like `self.client.srv_name` returns the original service name, and not the remapped one. 

`rclpy.publisher.Publisher` implements the [topic_name property](https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/publisher.py#L82) which seems to resolve the name. Is it possible to do the same for service clients?

fujitatomoya:
How about using this one?

https://github.com/ros2/rclpy/blob/5c5393d96cbcebea768916aa38dd12298695da94/rclpy/rclpy/node.py#L1228-L1240

Bjoernolav:
Oops, missed that one - works using that one yes, thanks!

But is there a reason why the `Client` implementation does not expose a `service_name` property in the same way as the `Publisher` class exposes a `topic_name` property? I guess it would be nice for them to expose similar APIs if possible.

amalnanavati:
Adding onto the comment about similar APIs, whereas `create_publisher` and `create_subscription` call `resolve_topic_name` by default, `create_client` and `create_service` do not call `resolve_service_name` by default. I'd say this is unexpected functionality. Imho, when someone calls the default functions to create topics/services from a node, I think it's reasonable to expect that both topic names and service names will be remapped.

fujitatomoya:
[...]

PR #1156, opened 2023-08-28
Support to get remapped service name

Barry-Xu-2018:
Address #954  

`srv_name` in service & client save real service name (remapped name).

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1622,7 +1622,7 @@ class Node:
 
         client = Client(
             self.context,
-            client_impl, srv_type, srv_name, qos_profile,
+            client_impl, srv_type, client_impl.service_name, qos_profile,
             callback_group)
         callback_group.add_entity(client)
         self._clients.append(client)
@@ -1667,7 +1667,7 @@ class Node:
 
         service = Service(
             service_impl,
-            srv_type, srv_name, callback, callback_group, qos_profile)
+            srv_type, service_impl.name, callback, callback_group, qos_profile)
         callback_group.add_entity(service)
         self._services.append(service)
         self._wake_executor()
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1622,7 +1622,7 @@ class Node:
 
         client = Client(
             self.context,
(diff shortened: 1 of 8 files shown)


Barry-Xu-2018:
`srv_name` isn't set as the property of service & client.   
So maybe it is better to make new property `service_name` for service & client class to save real service name. `srv_name` only save service name input by user.

rclpy/rclpy/node.py
```diff
@@ -1622,7 +1622,7 @@ def create_client(
 
         client = Client(
             self.context,
-            client_impl, srv_type, srv_name, qos_profile,
+            client_impl, srv_type, client_impl.service_name, qos_profile,
```
fujitatomoya, line 1625: this changes the current behavior but i think the intention is to store the actual name with implementation. so i think changing into `client_impl.service_name` is the correct to address the bug.

fujitatomoya (approved):
lgtm

fujitatomoya:
@iuhilnehc-ynos can i use another review? if you are good to go, please start the CI.

iuhilnehc-ynos:
> `srv_name` isn't set as the property of service & client. So maybe it is better to make new property `service_name` for service & client class to save real service name. `srv_name` only save service name input by user.

Yeah. This style is also what I expected. Please see
https://github.com/ros2/rclpy/blob/1843c2259260787d94975f658ed195a82f6eb90c/rclpy/rclpy/publisher.py#L81-L84

I prefer to use a new property named 'service_name' for the Service and Client.

BTW. I have no idea if users want to use the original `srv_name` passed into Service and Client even though there aren't relative names in the Publisher and Subscriber. Let's just keep it.

Barry-Xu-2018:
@iuhilnehc-ynos Thanks for your comments.   
I will update codes to use new property `service_name` for service & client class.

iuhilnehc-ynos (approved):

iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19419/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13947/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20140/)

Issue #878, opened 2022-01-14
Race condition in Action Client service calls


jacobperron:
Like https://github.com/ros2/rclpy/issues/870, there is a similar race with action clients.

See discussion here: https://github.com/ros2/rclpy/pull/871#issuecomment-1010322339

augustelalande:
This is one is tricky. Technically, all three functions `send_goal_async`, `_cancel_goal_async`, and `_get_result_async` can cause a race condition. To address this, we would need to acquire a lock from `self._client_handle.send_*_request` until `self.add_future(future)`.

Then we would need to acquire the same lock at `for future in waitable._futures`. This can be done by implementing a function `waitable.get_futures` which acquires the lock before returning the futures.

This is the first difficulty. Logically, this function should be implemented by the `Waitable` class which already implements the methods `add_future` and `remove_future`, and also initializes the `_futures` list. However, the lock is required by the subclass `ActionClient` which needs to acquire it for the aforementioned service calls. Two solutions:
1. `Waitable` creates the lock and acquires it for `get_futures` and `ActionClient` also acquires it for the service calls.
2. `ActionClient` implements `get_futures`.

This should avoid any race condition given the current implementation. However, this may not solve it in general. This solution would ensure that 
https://github.com/ros2/rclpy/blob/0bafc8ae187682ce8c62fe83da8be76badbf4abc/rclpy/rclpy/executors.py#L382
cannot run before all futures have been added to the _futures list. By extension, it also ensures that
https://github.com/ros2/rclpy/blob/0bafc8ae187682ce8c62fe83da8be76badbf4abc/rclpy/rclpy/executors.py#L384
cannot run before all futures have been added to the `_pending_*_requests` dictionary. However, if the `waitable.get_futures` method is for some reason not called before `waitable.execute` then it could lead to a race condition and ultimately a KeyError.

PR #1164, opened 2023-09-17
Address race conditions in action client

augustelalande:
This addresses the race conditions described in #878, by implementing essentially the same solution as #871.

Additionally, I also implement #838 which was implemented in the regular client but not the action client.

The one thing is does not address is that 
https://github.com/ros2/rclpy/blob/0a0239598a635f59d934851e2af3479fabc1c5da/rclpy/rclpy/executors.py#L453-L454
can potentially execute without the future having been added to the list, in which case the future's callbacks will execute immediately instead of being delegated to the task queue.


augustelalande:
@sloretz @adityapande-1995 any help?

PR #1178, opened 2023-10-02
Update executors.py - Check for _sigint_gc attribute in __del__ function

banzo:
Avoid the following error: AttributeError: 'SingleThreadedExecutor' object has no attribute '_sigint_gc'

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -239,7 +239,7 @@ class Executor(ContextManager['Executor']):
         return True
 
     def __del__(self):
-        if self._sigint_gc is not None:
+        if hasattr(self, '_sigint_gc') and self._sigint_gc is not None:
             self._sigint_gc.destroy()
 
     def add_node(self, node: 'Node') -> bool:


rclpy/rclpy/executors.py
```diff
@@ -239,7 +239,7 @@ def shutdown(self, timeout_sec: Optional[float] = None) -> bool:
         return True
 
     def __del__(self):
-        if self._sigint_gc is not None:
+        if hasattr(self, '_sigint_gc') and self._sigint_gc is not None:
```
clalancette, line 242: `_sigint_gc` is created in the `__init__` method, and is never unset on the class.  Can you explain how we could get into the situation where it is not an attribute on the class?

banzo: We are trying to generate unit tests using [Pynguin](https://github.com/se2p/pynguin/ , I am guessing it does some unorthodox use of rclpy. 

I can put together a Docker image to reproduce the issue minimally, if that can help.

clalancette: It seems like a bug in `Pynguin`, though.  In theory you could undefine *any* attribute on the class.  But I don't think we should add checks everywhere for that.

banzo: ok, thanks for the feedback, I'll see with the Pynguin guys then.

Issue #1179, opened 2023-10-03
wait_for_message Accumulating CPU Load After Repeated Calls


KKSTB:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - Binaries
- Version or commit hash:
  - Iron
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue
1. Setup and spin single threaded node executor in a new thread
2. Call wait_for_message in main thread or another thread:
```python3
for i in range(100):
    is_success, msg = rclpy.wait_for_message.wait_for_message(msg_type, rclpy_node, topic, time_to_wait=time_to_wait)
```

#### Expected behavior
CPU% stays the same after the for loop
#### Actual behavior
CPU% increased and does not drop
#### Additional information
`py-spy record -n -r 10 -p [PID]` indicates high CPU load at _take_subscription (rclpy/executors.py 338 https://github.com/ros2/rclpy/blob/e9bd428ed562019596e8d3964489bab90a85b203/rclpy/rclpy/executors.py#L338C25-L338C43).

KKSTB:
Fix is simple: just unsubscribe before return (tested). This should work assuming that there is no exception in between.

clalancette:
> Fix is simple: just unsubscribe before return (tested). This should work assuming that there is no exception in between.

Yes, we should definitely destroy the subscription.  I think we should put it into a `try..finally` block, so we make sure it is always destroyed.  Would you submit a pull request to do that?

fujitatomoya:
@KKSTB i was thinking probably we want to backport this to iron and humble, since this affects user platform. what do you think? besides, the issue was originally created with iron.

CC: @Yadunund @clalancette

KKSTB:
Ah you are right. Thank you for reminding me.

FelixLittleFei:
@KKSTB @fujitatomoya @clalancette 
Dear ROS2 developers, 
Could you please backport to ROS2 Humble for this wait_for_message function?
![Screenshot from 2023-12-20 15-49-31](https://github.com/ros2/rclpy/assets/89890055/8664ae15-276e-40dd-898d-59a00d94f876)

I noticed that it already exists in Rolling and Iron.
Thank you!

KKSTB:
Hi @FelixLittleFei 

Actually I tested that the wait_for_message function is also affected by this issue:
https://github.com/ros2/rclpy/issues/1206

[...]

PR #1180, opened 2023-10-04
 Fix to issue https://github.com/ros2/rclpy/issues/1179

KKSTB:
Fix to issue https://github.com/ros2/rclpy/issues/1179


rclpy/rclpy/wait_for_message.py
```diff
@@ -39,24 +41,27 @@ def wait_for_message(
...
-            if msg_info is not None:
-                return (True, msg_info[0])
+        if subs_ready:
+            if sub.handle.pointer in subs_ready:
+                msg_info = sub.handle.take_message(sub.msg_type, sub.raw)
+                if msg_info is not None:
+                    return True, msg_info[0]
 
-    return (False, None)
+        return False, None
```
clalancette, line 65: I think this should be moved after the `finally` block.

In practice, I don't think it will make much of a difference; if anything threw an exception, then the exception would propagate after the `finally` block is run.  But I think it is easier to tell that this function always returns *something* if this is outside of the `try...finally` block.

clalancette (changes requested):
I've left two small things to fix; once those are done, I'm happy with this.  Thank you for submitting the PR!

KKSTB:
Agree. This code (`return` at the end) looks nicer

clalancette (approved):
Thanks for iterating!  This looks good to me.

clalancette:
Oh, actually, before we can merge this, we need one more thing.  Can you please rebase this and add a `Signed-off-by` line?  You can do it with `git commit -s`.  Thanks.

fujitatomoya (approved):

KKSTB:
I think I gained some git experience in the process. Thanks

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19630/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14138/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20353/)

Issue #1181, opened 2023-10-10
Service client sometimes deadlock due to 'client will not receive response' of newly launched service server node


KKSTB:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - binaries
- Version or commit hash:
  - Iron
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp, rclpy, controller_manager

#### Steps to reproduce issue
1. Programmatically launch a launch python containing controller_manager using method https://github.com/ros2/launch/issues/724
2. Create controller_manager/load_controller client, wait_for_service() and call the service with a valid controller
3. destroy_client()

#### Expected behavior
Execute as normal
#### Actual behavior
Client stuck at client.call(), with the following printout:
```
my program print('robot/controller_manager/load_controller before wait_for_service()') --> already waiting for service
[INFO] [ros2_control_node-1]: process started with pid [25427]
[INFO] [robot_state_publisher-2]: process started with pid [25429]
[INFO] [spawner-3]: process started with pid [25431]
[robot_state_publisher-2] [INFO] [1696905559.365691573] [robot.robot_state_publisher]: got segment robot_base_ref
[robot_state_publisher-2] [INFO] [1696905559.365703944] [robot.robot_state_publisher]: got segment robot_link1
[robot_state_publisher-2] [INFO] [1696905559.365716391] [robot.robot_state_publisher]: got segment robot_link2
[robot_state_publisher-2] [INFO] [1696905559.365727908] [robot.robot_state_publisher]: got segment robot_link3
[robot_state_publisher-2] [INFO] [1696905559.365739118] [robot.robot_state_publisher]: got segment robot_link4
[robot_state_publisher-2] [INFO] [1696905559.365751414] [robot.robot_state_publisher]: got segment robot_link5
[robot_state_publisher-2] [INFO] [1696905559.365761952] [robot.robot_state_publisher]: got segment robot_link6
[robot_state_publisher-2] [INFO] [1696905559.365773261] [robot.robot_state_publisher]: got segment robot_link_base
[robot_state_publisher-2] [INFO] [1696905559.365784672] [robot.robot_state_publisher]: got segment robot_link_eef
[ros2_control_node-1] [WARN] [1696905559.372291758] [robot.controller_manager]: [Deprecated] Passing the robot description parameter directly to the control_manager node is deprecated. Use '~/robot_description' topic from 'robot_state_publisher' instead.
[...]

PR #1188, opened 2023-10-17
Add an optional timeout_sec input to Client.call() to fix issue https://github.com/ros2/rclpy/issues/1181

KKSTB:
Add an optional timeout_sec input so that the `call()` function can break away from the deadlock if RMW fails to response. The `call()` function returns future.result() which is `None`.

Another approach of adding timeout to the future returned by `call_async()` is much more difficult to implement, requiring changes from the `Future` class and upward. Also there is no timeout input for the rclcpp version of `async_send_request()`. So I would prefer not taking this approach instead.

### DIFF
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -69,7 +69,11 @@ class Client:
 
         self._lock = threading.Lock()
 
-    def call(self, request: SrvTypeRequest, timeout_sec: Optional[float] = None) -> SrvTypeResponse:
+    def call(
+        self,
+        request: SrvTypeRequest,
+        timeout_sec: Optional[float] = None
+    ) -> Optional[SrvTypeResponse]:
         """
         Make a service request and wait for the result.
 
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -98,6 +98,7 @@ class Client:
         # resulting in the event never being set.
         if not future.done():
             if not event.wait(timeout_sec):
+                # Timed out. remove_pending_request() to free resources
                 self.remove_pending_request(future)
         if future.exception() is not None:
             raise future.exception()
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -69,14 +69,14 @@ class Client:
 
         self._lock = threading.Lock()
 
-    def call(self, request: SrvTypeRequest) -> SrvTypeResponse:
+    def call(self, request: SrvTypeRequest, timeout_sec: Optional[float] = None) -> SrvTypeResponse:
         """
         Make a service request and wait for the result.
 
-        .. warning:: Do not call this method in a callback or a deadlock may occur.
+        .. warning:: Do not call this method in a callback, or a deadlock or timeout may occur.
 
         :param request: The service request.
-        :return: The service response.
+        :return: The service response, or None if timed out.
         :raises: TypeError if the type of the passed request isn't an instance
           of the Request type of the provided service when the client was
           constructed.
@@ -97,7 +97,8 @@ class Client:
         # The callback might have been added after the future is completed,
         # resulting in the event never being set.
         if not future.done():
-            event.wait()
+            if not event.wait(timeout_sec):
+                self.remove_pending_request(future)
         if future.exception() is not None:
             raise future.exception()
         return future.result()
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -76,6 +76,7 @@ class Client:
         .. warning:: Do not call this method in a callback, or a deadlock or timeout may occur.
 
         :param request: The service request.
+        :param timeout_sec: Seconds to wait. If ``None``, then wait forever.
         :return: The service response, or None if timed out.
         :raises: TypeError if the type of the passed request isn't an instance
           of the Request type of the provided service when the client was
--- a/rclpy/test/test_client.py
+++ b/rclpy/test/test_client.py
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 import platform
-import time
 import threading
+import time
 import traceback
 import unittest
 
--- a/rclpy/test/test_client.py
+++ b/rclpy/test/test_client.py
@@ -14,13 +14,14 @@
 
 import platform
 import time
-import unittest
 import threading
+import traceback
+import unittest
 
 from rcl_interfaces.srv import GetParameters
 import rclpy
-import rclpy.node
 import rclpy.executors
+import rclpy.node
 from rclpy.utilities import get_rmw_implementation_identifier
 from test_msgs.srv import Empty
 
@@ -59,7 +60,10 @@ class TestClient(unittest.TestCase):
             node.destroy_node()
 
     @staticmethod
-    def _spin_rclpy_node(rclpy_node: rclpy.node.Node, rclpy_executor: rclpy.executors.SingleThreadedExecutor) -> None:
+    def _spin_rclpy_node(
+        rclpy_node: rclpy.node.Node,
+        rclpy_executor: rclpy.executors.SingleThreadedExecutor
+    ) -> None:
         try:
             rclpy_executor.spin()
         except rclpy.executors.ExternalShutdownException:
@@ -209,28 +213,7 @@ class TestClient(unittest.TestCase):
             executor_thread = threading.Thread(
                 target=TestClient._spin_rclpy_node, args=(self.node, executor))
             executor_thread.start()
-            result = cli.call(GetParameters.Request(), 5)
-            self.assertTrue(result is not None)
-            executor.shutdown()
-            executor_thread.join()
-        finally:
-            self.node.destroy_client(cli)
-            self.node.destroy_service(srv)
-
-    def test_sync_call_slow(self):
-        def _service(request, response):
-            time.sleep(5)
-            return response
-        cli = self.node.create_client(GetParameters, 'get/parameters')
-        srv = self.node.create_service(GetParameters, 'get/parameters', _service)
-        try:
-            self.assertTrue(cli.wait_for_service(timeout_sec=20))
-            executor = rclpy.executors.SingleThreadedExecutor(context=self.context)
-            executor.add_node(self.node)
-            executor_thread = threading.Thread(
-                target=TestClient._spin_rclpy_node, args=(self.node, executor))
-            executor_thread.start()
-            result = cli.call(GetParameters.Request(), 10)
+            result = cli.call(GetParameters.Request(), 0.5)
             self.assertTrue(result is not None)
             executor.shutdown()
             executor_thread.join()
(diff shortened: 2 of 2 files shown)


fujitatomoya (commented):
IMO, we could deprecate those sync APIs like this as `rclpy`, because `rclpy` provides async APIs and user can manage from there, including these timeout behavior as they like.

On the other hand, I see this is one of the enhancements, at lease at this moment.

any thoughts?

KKSTB:
No idea too. I think it depends on how rclpy in ROS2 is positioned:
1. if rclpy is to provide the fundamentals only, then the sync APIs are too much to exist
2. if rclpy is to also provide implementations that others would normally do and would likely frequently do, then the sync APIs can provide a neat solution, and probably help avoid common problems that would probably encounter by having a proper implementation and with issues fixed

rclpy/rclpy/client.py
```diff
@@ -69,14 +69,19 @@ def __init__(
...
+    def call(
+        self,
+        request: SrvTypeRequest,
+        timeout_sec: Optional[float] = None
+    ) -> Optional[SrvTypeResponse]:
         """
         Make a service request and wait for the result.
 
-        .. warning:: Do not call this method in a callback or a deadlock may occur.
+        .. warning:: Do not call this method in a callback, or a deadlock or timeout may occur.
```
sloretz, line 80: nit, this warning is about the danger of blocking a thread in the executor. I don't think the timeout needs to be mentioned here.

sloretz (commented):
LGTM with one nit addressed

rclpy/rclpy/client.py
KKSTB: Trivial but its better be here just in case someone have this problem and don't understand why

fujitatomoya:
address https://github.com/ros2/rclpy/issues/1181

fujitatomoya:
@KKSTB can you add the test for timeout case to https://github.com/ros2/rclpy/blob/rolling/rclpy/test/test_client.py, then i will start the CI.

KKSTB:
This is a bit difficult now as I cannot build rclpy successfully due to recent change in rcl_node_type_description_service_init() https://github.com/ros2/rcl/commit/1b79535fa07fcc77c873adf5261308dac7a821ce

Have to wait for rolling update.

KKSTB:
@fujitatomoya thanks. Although it takes a long time to build from source.

I was originally using binary install of rolling, therefore failed to build.

rclpy/test/test_client.py
```diff
@@ -15,9 +15,11 @@
 import platform
 import time
 import unittest
+import threading
```
fujitatomoya, line 18: @KKSTB there are several flake8 errors, https://build.ros2.org/job/Rpr__rclpy__ubuntu_jammy_amd64/480/testReport/ could you address them?

rclpy/rclpy/client.py
fujitatomoya: @sloretz i think this is okay to add? before this timeout is added, it will block the executor thread, so eventually will be deadlock. but with timeout, i think this will be either timeout (if timeout is specified) or deadlock?

rclpy/test/test_client.py
```diff
@@ -183,6 +197,68 @@ def test_get_service_name_after_remapping(self):
...
+            executor_thread.start()
+            result = cli.call(GetParameters.Request(), 5)
+            self.assertTrue(result is not None)
+            executor.shutdown()
+            executor_thread.join()
+        finally:
+            self.node.destroy_client(cli)
+            self.node.destroy_service(srv)
+
+    def test_sync_call_slow(self):
```
fujitatomoya, line 220: Do we really need to have this test? this is technically the same test with previous one w/o `sleep(5)`? i am inclined to remove this to decrease the CI test period, if possible.

rclpy/test/test_client.py
```diff
@@ -183,6 +197,68 @@ def test_get_service_name_after_remapping(self):
...
+        cli = self.node.create_client(GetParameters, 'get/parameters')
+        srv = self.node.create_service(GetParameters, 'get/parameters', _service)
+        try:
+            self.assertTrue(cli.wait_for_service(timeout_sec=20))
+            executor = rclpy.executors.SingleThreadedExecutor(context=self.context)
+            executor.add_node(self.node)
+            executor_thread = threading.Thread(
+                target=TestClient._spin_rclpy_node, args=(self.node, executor))
+            executor_thread.start()
+            result = cli.call(GetParameters.Request(), 5)
```
fujitatomoya, line 254: `500msec` would be enough for timeout to reduce the test completion time?

```suggestion
            result = cli.call(GetParameters.Request(), 0.5)
```

rclpy/test/test_client.py
```diff
@@ -183,6 +197,68 @@ def test_get_service_name_after_remapping(self):
...
+            self.assertTrue(result is not None)
+            executor.shutdown()
+            executor_thread.join()
+        finally:
+            self.node.destroy_client(cli)
+            self.node.destroy_service(srv)
+
+    def test_sync_call_timeout(self):
+        def _service(request, response):
+            time.sleep(10)
```
fujitatomoya, line 243:
```suggestion
            time.sleep(1)
```

fujitatomoya (commented):
overall lgtm, several comments to discuss.

rclpy/test/test_client.py
KKSTB: Is there any method to run flake8 checks locally? I got this exception on my computer:
```
 -- run_test.py: invoking following command in '/home/kk/ros2_rolling_ws/src/ros2/rclpy/rclpy':
53:  - /opt/ros/rolling/bin/ament_flake8 --xunit-file /home/kk/ros2_rolling_ws/build/rclpy/test_results/rclpy/flake8.xunit.xml
53: Traceback (most recent call last):
53:   File "/opt/ros/rolling/bin/ament_flake8", line 33, in <module>
53:     sys.exit(load_entry_point('ament-flake8==0.16.0', 'console_scripts', 'ament_flake8')())
53:   File "/home/kk/ros2_rolling_ws/build/ament_flake8/ament_flake8/main.py", line 31, in main
53:     rc, _ = main_with_errors(argv=argv)
53:   File "/home/kk/ros2_rolling_ws/build/ament_flake8/ament_flake8/main.py", line 88, in main_with_errors
53:     report = generate_flake8_report(
53:   File "/home/kk/ros2_rolling_ws/build/ament_flake8/ament_flake8/main.py", line 231, in generate_flake8_report
53:     opts, _ = parse_config_file(config_file)
53:   File "/home/kk/ros2_rolling_ws/build/ament_flake8/ament_flake8/main.py", line 199, in parse_config_file
53:     opts_manager = manager.OptionManager(
53: TypeError: OptionManager.__init__() missing 1 required keyword-only argument: 'formatter_names'
53: -- run_test.py: return code 1
53: -- run_test.py: generate result file '/home/kk/ros2_rolling_ws/build/rclpy/test_results/rclpy/flake8.xunit.xml' with failed test
53: -- run_test.py: verify result file '/home/kk/ros2_rolling_ws/build/rclpy/test_results/rclpy/flake8.xunit.xml'
53/57 Test #53: flake8 ............................***Failed    0.58 sec
```

fujitatomoya: How about these?

```bash
cd <colcon_workspace>
colcon build --symlink-install --packages-select rclpy
colcon test --event-handlers console_direct+ --packages-select rclpy --ctest-args -R test_client
colcon test-result --verbose | less
```

rclpy/test/test_client.py
KKSTB: Good suggestion thanks.

rclpy/test/test_client.py
KKSTB: `colcon test --event-handlers console_direct+ --packages-select rclpy --ctest-args -R test_client` only do the test on `test_client`. Yet flake8 cannot be done locally.

fujitatomoya:

clalancette:
The Rpr jobs aren't going to succeed here until we do releases of `rcl`, `rclpy`, and `rclcpp`.  We'll just have to rely on the jobs from https://ci.ros2.org for now.

rclpy/test/test_client.py
fujitatomoya: hmm, if i am not mistaken, i do that for my test to check flake8 warnings... anyway you addressed my comments, so i will start the CI.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19802/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14306/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20526/)

KKSTB:
Yet another flake8 and fixed

fujitatomoya:
yeah you fixed the final flake8 error, thanks starting CI 🤞 

Building on the built-in node in workspace /var/lib/jenkins/jobs/ci_launcher/workspace
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19803/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14307/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20527/)

tonynajjar:
Additionally a comment about the PR. It's surprising to me that you went with returning None if the call timed out instead of raising an Exception. Just returning None increases the risk of "silent bugs", especially for services not expecting a response.
@fujitatomoya would you be open if I opened a PR to raise an Exception?

fujitatomoya:
> Additionally a comment about the PR. It's surprising to me that you went with returning None if the call timed out instead of raising an Exception. Just returning None increases the risk of "silent bugs", especially for services not expecting a response. @fujitatomoya would you be open if I opened a PR to raise an Exception?

@tonynajjar @KKSTB can you review https://github.com/ros2/rclpy/pull/1271?

i think it makes more sense to raise `TimeoutError` instead of `None` that goes not give us any information.

fujitatomoya:
> I was just about to implement this and saw that it's already on rolling, can this be backported to humble please?

@tonynajjar i am not inclined to do that. https://github.com/ros2/rclpy/pull/1188#pullrequestreview-1686030933, probably we eventually deprecate this method in the future... @clalancette @sloretz @adityapande-1995 @wjwwood what do you think?

wjwwood:
We can't change a stable function in Humble to suddenly be able to throw an exception. You could consider a rewrite of this feature as a separate function, so it's purely new API. But even then we have to be careful that we don't accidentally change the behavior of existing functions either intentionally or as a new bug. That's why we're generally not inclined to back port features to older versions.

KKSTB:
Thanks @tonynajjar. You are right that returning None increases the chance of silent bug. What I considered at that time was to return whatever `future.result()` has without modifying the function heavily, therefore resulting in this. Maybe its an improvement to raise `TimeoutError`.

As for backporting to older version, I also hope this to happen in ROS Iron which I am using. But I go with the way similar to what @wjwwood suggest so that the fix (or any other customization or convenient wrappers) becomes readily available in my version of ROS Iron:
1. Make a new class inheriting `Node` and `Client`, and customize the `Client.call()`
2. override `node.create_client()` that returns your custom `Client` class, by assigning `__class__` to your new `Client` class, like this: https://stackoverflow.com/questions/8062161/can-i-dynamically-convert-an-instance-of-one-class-to-another
3. You may create a custom `rclpy` class that overrides `rclpy.create_node()` to conveniently create your custom node.

Its just my personal suggestion as a ROS user, and not representing what ROS team is suggesting. Hope that help.

PR #1192, opened 2023-10-28
Adjust python usage of the type_description service API

mjcarroll:
In https://github.com/ros2/rcl/pull/1112, the API is updated to populate the service structure, rather than `init`, `get`, and `fini`.  This updates rclpy to use the new API pattern.

### DIFF
--- a/rclpy/src/rclpy/type_description_service.cpp
+++ b/rclpy/src/rclpy/type_description_service.cpp
@@ -22,25 +22,12 @@ namespace rclpy
 
 TypeDescriptionService::TypeDescriptionService(Node & node)
 {
-  rcl_ret_t ret = rcl_node_type_description_service_init(node.rcl_ptr());
+  auto srv_ptr = std::make_shared<rcl_service_t>();
+  rcl_ret_t ret = rcl_node_type_description_service_init(srv_ptr.get(), node.rcl_ptr());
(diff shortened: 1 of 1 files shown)


rclpy/src/rclpy/type_description_service.cpp
```diff
@@ -22,25 +22,12 @@ namespace rclpy
...
-  if (RCL_RET_OK != ret) {
-    throw RCLError("Failed to initialize type description service");
-  }
-  rcl_service_t * srv_ptr = nullptr;
-  ret = rcl_node_get_type_description_service(node.rcl_ptr(), &srv_ptr);
+  auto srv_ptr = std::make_shared<rcl_service_t>();
+  rcl_ret_t ret = rcl_node_type_description_service_init(srv_ptr.get(), node.rcl_ptr());
   if (RCL_RET_OK != ret) {
     throw RCLError("Failed to retrieve type description service implementation");
   }
```
fujitatomoya, line 29: error message is not correct one.

```suggestion
  if (RCL_RET_OK != ret) {
    throw RCLError("Failed to initialize type description service");
  }
```

fujitatomoya (changes requested):

fujitatomoya (approved):

Issue #1195, opened 2023-11-01
Exception when instantiating ActionServer as part of ROS package tests


sea-bass:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - Binaries
- Version or commit hash:
  - Latest binaries of Humble/Iron/Rolling
- DDS implementation:
  - Fast DDS or Cyclone DDS (doesn't make a difference)
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue
I have a test that is failing as part of this PR (https://github.com/RobotWebTools/rosbridge_suite/pull/886) when I try to instantiate an action server, noting that:

* This only happens on Iron and Rolling, but not on Humble
* This only happens when I run using `colcon test` (which in turn uses `ament_add_pytest_test` CMake capability), but not if I run it directly using Python

Can't seem to figure out why this would be the case, so figured I'd elevate it here? Thank you!

The code:
```python
from example_interfaces.action import Fibonacci

class MyClass:
    def __init__(self, node):
        self.action_server = ActionServer(
            node,  # Just an rclpy.Node
            Fibonacci,
            "/fibonacci_action_1",
            self.execute_callback,
        )
```

The error:
```
          self.action_server = ActionServer(
      /opt/ros/rolling/lib/python3.10/site-packages/rclpy/action/server.py:264: in __init__
          self._logger = self._node.get_logger().get_child('action_server')
      _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
      
      self = <rclpy.impl.rcutils_logger.RcutilsLogger object at 0x7f730bfe9270>
      name = 'action_server'
      
          def get_child(self, name):
              if not name:
                  raise ValueError('Child logger name must not be empty.')
          
              if self.name:
                  # Prepend the name of this logger
      >           _rclpy.rclpy_logging_rosout_add_sublogger(self.name, name)
      E           RuntimeError: Failed to call rcl_logging_rosout_add_sublogger
```

#### Expected behavior
Should not raise an exception.

#### Actual behavior
Raises an exception.

sea-bass:
Seems the issue originates from https://github.com/ros2/rclpy/pull/982

fujitatomoya:
@iuhilnehc-ynos can you take a look at this?

PR #1196, opened 2023-11-03
aligh with rcl that a rosout publisher of a node might not exist

iuhilnehc-ynos:
align with https://github.com/ros2/rcl/pull/1115

### DIFF
--- a/rclpy/src/rclpy/_rclpy_logging.cpp
+++ b/rclpy/src/rclpy/_rclpy_logging.cpp
@@ -217,7 +217,8 @@ rclpy_logging_rosout_add_sublogger(const char * logger_name, const char * sublog
     rcl_reset_error();
     return false;
   } else {
-    throw rclpy::append_rcl_error("Failed to call rcl_logging_rosout_add_sublogger");
+    throw std::runtime_error(
+            rclpy::append_rcl_error("Failed to call rcl_logging_rosout_add_sublogger"));
   }
 }
 
--- a/rclpy/src/rclpy/_rclpy_logging.cpp
+++ b/rclpy/src/rclpy/_rclpy_logging.cpp
@@ -206,12 +206,16 @@ rclpy_logging_get_logging_directory()
 }
 
 /// Add a subordinate logger based on a logger
-void
+bool
 rclpy_logging_rosout_add_sublogger(const char * logger_name, const char * sublogger_name)
 {
   rclpy::LoggingGuard scoped_logging_guard;
   rcl_ret_t rcl_ret = rcl_logging_rosout_add_sublogger(logger_name, sublogger_name);
-  if (RCL_RET_OK != rcl_ret) {
+  if (RCL_RET_OK == rcl_ret) {
+    return true;
+  } else if (RCL_RET_NOT_FOUND == rcl_ret) {
+    return false;
+  } else {
     rcutils_reset_error();
     throw std::runtime_error("Failed to call rcl_logging_rosout_add_sublogger");
   }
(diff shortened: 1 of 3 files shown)


rclpy/src/rclpy/_rclpy_logging.cpp
```diff
@@ -206,14 +206,18 @@ rclpy_logging_get_logging_directory()
...
-  if (RCL_RET_OK != rcl_ret) {
-    rcutils_reset_error();
-    throw std::runtime_error("Failed to call rcl_logging_rosout_add_sublogger");
+  if (RCL_RET_OK == rcl_ret) {
+    return true;
+  } else if (RCL_RET_NOT_FOUND == rcl_ret) {
+    rcl_reset_error();
+    return false;
+  } else {
+    throw rclpy::append_rcl_error("Failed to call rcl_logging_rosout_add_sublogger");
```
fujitatomoya, line 220: having exception and false looks redundant? i think for user perspective, it is not really clear the difference between having exception and false via this call.
how about keeping this `void` and if `RCL_RET_NOT_FOUND` print warning (there will no rosout) w/o exception, and everything else goes to exception?

rclpy/test/test_rosout_subscription.py
```diff
@@ -114,13 +114,6 @@ def call_logger(logger):
         self.executor.spin_until_future_complete(self.fut, 3)
         self.assertTrue(self.fut.done())
 
-    def test_node_logger_not_exist(self):
-        node = rclpy.create_node('test_extra_node', context=self.context)
-        logger = node.get_logger()
-        node = None
-        with self.assertRaises(RuntimeError):
-            logger.get_child('child')
```
fujitatomoya, line 122: then this test needs to make sure there will be no exception but warning?

rclpy/src/rclpy/_rclpy_logging.cpp
iuhilnehc-ynos: Using a boolean is to make sure https://github.com/ros2/rclpy/blob/059bac65707c0e605a08bcf9451738f891ed0d26/rclpy/rclpy/impl/rcutils_logger.py#L238-L239 not to update `logger.logger_sublogger_namepair` if it's false(there is no parent logger existent),
and then it's unnecessary to call `_rclpy.rclpy_logging_rosout_remove_sublogger` https://github.com/ros2/rclpy/blob/059bac65707c0e605a08bcf9451738f891ed0d26/rclpy/rclpy/impl/rcutils_logger.py#L222-L223

If there exist other errors (e.g. `RCUTILS_RET_INVALID_ARGUMENT`, etc) when calling rcutils_hash_map_get, `rclpy_logging_rosout_add_sublogger` just throw the exceptions similar to other APIs in current file.

If only printing warning for `RCL_RET_NOT_FOUND`, there exists some unnecessary actions(`add logger.logger_sublogger_namepair` and `_rclpy.rclpy_logging_rosout_remove_sublogger`).

rclpy/test/test_rosout_subscription.py
```diff
@@ -114,13 +114,6 @@ def call_logger(logger):
         self.executor.spin_until_future_complete(self.fut, 3)
         self.assertTrue(self.fut.done())
 
-    def test_node_logger_not_exist(self):
```
clalancette, line 117: I think we want to add a test to ensure that the situation from https://github.com/ros2/rclpy/issues/1195 doesn't occur again.  The following code (which needs to be adapted into a test) reproduced for me:

```
import rclpy
from rclpy.action import ActionServer

from example_interfaces.action import Fibonacci

class MyClass:
    def __init__(self, node):
        self.action_server = ActionServer(
            node,
            Fibonacci,
            "/fibonacci_action_1",
            self.execute_callback,
        )

    def execute_callback(self):
        print("Execute callback")

rclpy.init()
node = rclpy.create_node("mynode", enable_rosout=False)
x = MyClass(node)
node.destroy_node()
rclpy.try_shutdown()
```

clalancette (changes requested):
I think we need to add a test for this situation, but it otherwise looks good.

rclpy/test/test_rosout_subscription.py
iuhilnehc-ynos: Thanks. 
I used the similar logic to test the case. Updated in https://github.com/ros2/rclpy/pull/1196/commits/fc84c5e18c1655d2a0ed88dc4a04ef3d295d8459.

clalancette (approved):

clalancette:
CI for this is in https://github.com/ros2/rcl/pull/1115#issuecomment-1809415104

PR #1201, opened 2023-11-27
Added more Static typing for Node Class and for RcutilsLogger

InvincibleRMC:
Hello there.

I added some explicit return types to functions in the Node and RCutilsLogger Classes.  I updated the Node one because the other properties seemed to have return types. For RCutilsLogger I simply added return types to the most common functions being the logging functions. If there any requests or problems I should fix please let me know.


InvincibleRMC:
I'll close this for now because I think I'm misunderstanding something about pybind which is the reason why no return types were had in the first place.

PR #1204, opened 2023-12-06
Add doc-string warnings for destroy() methods

scpeters:
The constructors for publisher.py and subscription.py have warnings in their doc-string that they should not be used directly to create Publisher or Subscription objects (see [API documentation](https://docs.ros2.org/latest/api/rclpy/api/topics.html)), but that the `create_publisher` or `create_subscription` methods from a node should be called instead. I believe the same warning applies to the `destroy()` method for these classes, and that `destroy_publisher` / `destroy_subscription` should be called from a node instead of calling `destroy()` on these objects directly, so I've added a warning to the doc-string for these `destroy()` methods.

### DIFF
--- a/rclpy/rclpy/publisher.py
+++ b/rclpy/rclpy/publisher.py
@@ -88,6 +88,12 @@ class Publisher:
         return self.__publisher
 
     def destroy(self):
+        """
+        Destroy a container for a ROS publisher.
+
+        .. warning:: Users should not destroy a publisher with this method, instead they should
+           call :meth:`.Node.destroy_publisher`.
+        """
         for handler in self.event_handlers:
             handler.destroy()
         self.__publisher.destroy_when_not_in_use()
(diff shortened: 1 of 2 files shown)


sloretz (approved):

sloretz:
CI ([repos file](https://gist.githubusercontent.com/sloretz/33eb1c0168cbaa6fa309c206ade5575d/raw/dc9897a9f8c683caeef8f03c2ca42dcee6fd0ea0/ros2.repos) build: `--packages-up-to rclpy` test: `--packages-select rclpy`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19930/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14405/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20637/)

fujitatomoya (approved):

Issue #1018, opened 2022-10-03
Asynchronous client calls in a timer cause the timer to never execute a second time.


molasses11:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binaries
- Version or commit hash:
  - foxy (the 2022-09-28 release that was just pushed through the apt repos)
- DDS implementation:
  - Fast-RTPS for sure
  - Any (I think)
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

Use the minimal service server from the examples
``` python
from example_interfaces.srv import AddTwoInts

import rclpy
from rclpy.node import Node


class MinimalService(Node):

    def __init__(self):
        super().__init__('minimal_service')
        self.srv = self.create_service(AddTwoInts, 'add_two_ints', self.add_two_ints_callback)

    def add_two_ints_callback(self, request, response):
        response.sum = request.a + request.b
        self.get_logger().info('Incoming request\na: %d b: %d' % (request.a, request.b))

        return response


def main(args=None):
    rclpy.init(args=args)

    minimal_service = MinimalService()

    rclpy.spin(minimal_service)

    rclpy.shutdown()


if __name__ == '__main__':
    main()
```
and a slightly modified service client:
``` python
from example_interfaces.srv import AddTwoInts
import rclpy
from rclpy.node import Node

from rclpy.executors import MultiThreadedExecutor
from rclpy.callback_groups import MutuallyExclusiveCallbackGroup


class MinimalClient(Node):
    def __init__(self):
        super().__init__("minimal_client_async")
        self.cb = None
        self.timer_cb = MutuallyExclusiveCallbackGroup()
        self.cli = self.create_client(
            AddTwoInts, "add_two_ints", callback_group=self.cb
        )
        while not self.cli.wait_for_service(timeout_sec=1.0):
            self.get_logger().info("service not available, waiting again...")
        self.req = AddTwoInts.Request()

        # Create a timer for the main loop execution
        self.timer = self.create_timer(
            1.0, self.main_loop, callback_group=self.timer_cb
        )

    def send_request(self, a, b):
        self.get_logger().info("send_request: enter")
        self.req.a = a
        self.req.b = b

        # Only works once, then never executes the timer again
        self.future = self.cli.call_async(self.req)
        rclpy.spin_until_future_complete(self, self.future)
        self.get_logger().info("send_request: exit")
        return self.future.result()

        # Used to work, but blocks now
[...]

PR #1211, opened 2024-01-20
Fix recursive spinning from within a callback?

meyerj:
Resolves https://github.com/ros2/rclpy/issues/1018. Also https://github.com/ros2/rclpy/pull/919 seems related at first glance.

**Suggested changes:**

- rclpy's global spinning functions (`rclpy.spin*()`) use the node's current executor by default, if set. Only in case the passed `node` does not have an executor assigned yet, fall back to the default executor, for example in the main function.
- only remove the node from the executor if it has actually been added in the same spinning call. [`add_node()`](https://github.com/meyerj/rclpy/blob/825d2c896092a0ab62f847af3e19cd6371bba113/rclpy/rclpy/executors.py#L274) may return `False` if the node was already added to the executor before, like if `spin_until_future_complete()` was called recursively from within another callback. In that case it must not be removed in the same callback. Only the outermost call would do so.

**Rationale:**

This is to support the common case where `rclpy.spin_until_future_complete(self, future)` is used within another callback, by a running executor, where the node has already been added before. At the moment, without a patch like this, it seems very complex to call a service in a blocking way in this case, for example to forward a call to another service within a node? The [synchronous `client.call()`](https://github.com/meyerj/rclpy/blob/825d2c896092a0ab62f847af3e19cd6371bba113/rclpy/rclpy/client.py#L72) cannot be used either, which would be the most obvious solution, because it does not spin the node while waiting or handles the response directly, and hence depends on a spinner for that node in another thread to function properly.

The [basic example for calling a service from the official documentation](https://docs.ros.org/en/rolling/Tutorials/Beginner-Client-Libraries/Writing-A-Simple-Py-Service-And-Client.html#write-the-client-node) only works because the service call is initiated from the main thread directly, which is unusual in a more complex, long running node. It would also work like that inside a callback, for example a timer callback like in #1018, but then the node is removed from the executor and no other callbacks will run anymore afterwards.

I am aware that this patch would be a quite significant change to `rclpy`, with the potential to break user code. So it is more meant to start a discussion on other potential solutions...

**Alternatives considered:**

- Update examples and documentation and tell users to call `node.executor.spin_until_future_complete(future)` instead of `rclpy.spin_until_future_complete(node, future)` if that is used within a callback?
- Throw an exception if the same node is added to an executor more than once, [like in `rclcpp`](https://github.com/ros2/rclcpp/blob/6f6b5f2e3608eafeadb908aa4f6eded5b38c3043/rclcpp/src/rclcpp/executor.cpp#L258-L288), to enforce the above?

**References:**

For reference, [my original comment in #1018](https://github.com/ros2/rclpy/issues/1018#issuecomment-1889280840) with some additional thoughts and suggestions:

> We experienced similar issuses in different contexts. I think the problem is deeper, not only related to client calls and timers.
> 
> Essentially all implementations of `rclpy.spin_*()` methods in [`rclpy/__init__.py`](https://github.com/ros2/rclpy/blob/9536f8ec66020a552cf9ad421a11629ad8804a16/rclpy/rclpy/__init__.py) are somehow broken for the following reason, similar to what @iuhilnehc-ynos already [outlined above](https://github.com/ros2/rclpy/issues/1018#issuecomment-1280524111):
> 
> * If they are called recursively from within a callback, which is not explicitly forbidden, then the `node` instance was already added before to that executor. This results in `False` being returned from [`Executor.add_node(node)`](https://github.com/ros2/rclpy/blob/9536f8ec66020a552cf9ad421a11629ad8804a16/rclpy/rclpy/executors.py#L260-L274), but the return value is not used in the spin functions.
> * Even if the `node` has not be added in this spin call, it will be removed from the executor unconditionally in the `finally` block.
> * So if any of those methods is used recursively, the node will not be spinned anymore afterwards.
> 
> I do think that recursive spinning is a desirable feature and apparently many users rely on that for implementing timeouts without blocking all other callbacks. Before recent #1188 calling a service with a timeout that works in any context was surprisingly hard, but the synchronous `call()` with a timeout does currently not allow other callbacks to run in between, like with recursive spinning. Even [the most basic Python service client example in the official documentation](https://docs.ros.org/en/rolling/Tutorials/Beginner-Client-Libraries/Writing-A-Simple-Py-Service-And-Client.html#write-the-client-node) uses `rclpy.spin_until_future_complete(self, self.future)`, but in a method called directly from `main()` and not by an already running executor, like if a service would be called in another callback.
> 
> **Suggestion:**
> 
> So probably the "best" solution would be to try adding the node to the executor for spinning, but keep the return value and only remove it in `finally` if it was successfully added before? Something like
> 
> ```python
>     try:
>         was_added = executor.add_node(node)
>         executor.spin_until_future_complete(future, timeout_sec)
>     finally:
>         if was_added:
>             executor.remove_node(node)
> ```
> 
> of with a context manager that temporarily adds the node to the executor if it was not added before, instead of the repeated `try ... finally` construct.
> 
> Alternatively there must be a way to call [`executor.spin_until_future_complete(future, timeout_sec)`](https://github.com/ros2/rclpy/blob/9536f8ec66020a552cf9ad421a11629ad8804a16/rclpy/rclpy/executors.py#L310) directly without a specific `node` instance if it was already added to the executor before, like inside another callback of that node.
> 
> Something like
> 
> ```python
> node.executor.spin_until_future_complete(future, timeout_sec)
> ```
> 
> instead of
> 
> ```python
> rclpy.spin_until_future_complete(future, timeout_sec)
> ```
> 
> should also solve the problem described in this issue, and similar ones.
> 
> When checking on how to find the current `executor` of a node, I also found that [`Executor.add_node(node)`](https://github.com/ros2/rclpy/blob/9536f8ec66020a552cf9ad421a11629ad8804a16/rclpy/rclpy/executors.py#L260C9-L274) is _always_ called twice recursively, because it invokes [the setter for `node.executor`](https://github.com/ros2/rclpy/blob/9536f8ec66020a552cf9ad421a11629ad8804a16/rclpy/rclpy/node.py#L289-L301) which in turn calls back into [`executor.add_node(node)`](https://github.com/ros2/rclpy/blob/9536f8ec66020a552cf9ad421a11629ad8804a16/rclpy/rclpy/node.py#L300). This only works because [`executor._nodes_lock`](https://github.com/ros2/rclpy/blob/9536f8ec66020a552cf9ad421a11629ad8804a16/rclpy/rclpy/executors.py#L267) is a recursive lock and `add_node(node)` checks whether the `node` was already added to `self._nodes` before.
> 
> Afaik there is currently no way to get the executor of the current thread, if any, without a reference to the node? I think `asyncio.get_event_loop()` uses thread local storage for that. So calls like `rclpy.spin_until_future_complete(node, self.future)` from within a callback would also fail, or spin the wrong executor, if used inside a callback that is executed by another than the global executor. If there was such a way, then using the executor of the current thread as the default executor in global `rclpy.spin_*()` functions would be a reasonable default, and only fall back to the global executor in case the call is from a thread context without any executor, like the main thread before any spin function was called. Of course that would be a quite severe change to the basics of `rclpy`...
> 
> Interestingly, in [rclcpp](https://github.com/ros2/rclcpp) recursively spinning a node results in an exception being thrown in [`Executor::add_node()`](https://github.com/ros2/rclcpp/blob/6f6b5f2e3608eafeadb908aa4f6eded5b38c3043/rclcpp/src/rclcpp/executor.cpp#L258-L288), so it behaves fundamentally different (in version [`6f6b5f2`](https://github.com/ros2/rclcpp/tree/6f6b5f2e3608eafeadb908aa4f6eded5b38c3043)). See also [@wjwwood's related TODO comment in `rclcpp::spin_node_until_future_complete()`](https://github.com/ros2/rclcpp/blob/6f6b5f2e3608eafeadb908aa4f6eded5b38c3043/rclcpp/include/rclcpp/executors.hpp#L82-L96). On the long term it would be nice if the two most widely used client libraries behave the same with this respect.

And just for completeness and because it is very much related, if an uncaught exception is thrown in any callback in `rclcpp`, the executor is left in an unclean state and the node cannot spin anymore, even if exceptions would be handled by the caller of the spinner: https://github.com/ros2/rclcpp/pull/2017.

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -203,23 +203,25 @@ def spin_once(
     One callback will be executed by the provided executor as long as that callback is ready
     before the timeout expires.
 
-    If no executor is provided (ie. ``None``), then the global executor is used.
-    It is possible the work done is for a node other than the one provided if the global executor
+    If no executor is provided (ie. ``None``), then the node's current executor is used. If the
+    node does not have an executor yet, the global executor is used.
+    It is possible the work done is for a node other than the one provided if the used executor
     has a partially completed coroutine.
 
     This method should not be called from multiple threads with the same node or executor
     argument.
 
     :param node: A node to add to the executor to check for work.
-    :param executor: The executor to use, or the global executor if ``None``.
+    :param executor: The executor to use, or the node's current or the global executor if ``None``.
     :param timeout_sec: Seconds to wait. Block forever if ``None`` or negative. Don't wait if 0.
     """
-    executor = get_global_executor() if executor is None else executor
+    executor = executor or node.executor or get_global_executor()
     try:
-        executor.add_node(node)
+        node_added = executor.add_node(node)
         executor.spin_once(timeout_sec=timeout_sec)
     finally:
-        executor.remove_node(node)
+        if node_added:
+            executor.remove_node(node)
 
 
 def spin(node: 'Node', executor: Optional['Executor'] = None) -> None:
@@ -231,15 +233,17 @@ def spin(node: 'Node', executor: Optional['Executor'] = None) -> None:
     This function blocks.
 
     :param node: A node to add to the executor to check for work.
-    :param executor: The executor to use, or the global executor if ``None``.
+    :param executor: The executor to use, or the node's current or the global executor if ``None``.
     """
-    executor = get_global_executor() if executor is None else executor
+    executor = executor or node.executor or get_global_executor()
+    node_added = False
     try:
-        executor.add_node(node)
+        node_added = executor.add_node(node)
         while executor.context.ok():
             executor.spin_once()
     finally:
-        executor.remove_node(node)
+        if node_added:
+            executor.remove_node(node)
 
 
 def spin_until_future_complete(
@@ -256,13 +260,15 @@ def spin_until_future_complete(
 
     :param node: A node to add to the executor to check for work.
     :param future: The future object to wait on.
-    :param executor: The executor to use, or the global executor if ``None``.
+    :param executor: The executor to use, or the node's current or the global executor if ``None``.
     :param timeout_sec: Seconds to wait. Block until the future is complete
         if ``None`` or negative. Don't wait if 0.
     """
-    executor = get_global_executor() if executor is None else executor
+    executor = executor or node.executor or get_global_executor()
+    node_added = False
     try:
-        executor.add_node(node)
+        node_added = executor.add_node(node)
         executor.spin_until_future_complete(future, timeout_sec)
     finally:
-        executor.remove_node(node)
+        if node_added:
+            executor.remove_node(node)


rclpy/rclpy/__init__.py
```diff
@@ -231,15 +233,17 @@ def spin(node: 'Node', executor: Optional['Executor'] = None) -> None:
     This function blocks.
 
     :param node: A node to add to the executor to check for work.
-    :param executor: The executor to use, or the global executor if ``None``.
+    :param executor: The executor to use, or the node's current or the global executor if ``None``.
     """
-    executor = get_global_executor() if executor is None else executor
+    executor = executor or node.executor or get_global_executor()
```
fujitatomoya, line 238: although this makes https://github.com/ros2/rclpy/issues/1018#issue-1395363459 works with `node.executor`,

If `executor` is specified, and `node.executor` is assigned, this will replace the `node.executor` implicitly. i do not think this is expected behavior by user. instead, we should generate the exception to let the user know that this node has been already associated with executor? i agree with https://github.com/ros2/rclpy/issues/1018#issuecomment-1280524111.

fujitatomoya (commented):
I think we need to add the check if the `node` is already associated with `Executor`, https://github.com/ros2/rclpy/blob/070a323a67996bca78d95912efbd5e667ba05b1b/rclpy/rclpy/executors.py#L260-L274

PR #1216, opened 2024-02-07
forbid parameter to be declared statically without initialization.

fujitatomoya:
address part of https://github.com/ros2/ros2/issues/1517

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -499,16 +499,16 @@ class Node:
                 descriptor.type = second_arg.value
             else:
                 value = second_arg
-                if not descriptor.dynamic_typing and value is not None:
+                if not descriptor.dynamic_typing:
                     # infer type from default value
-                    if not isinstance(value, ParameterValue):
-                        descriptor.type = Parameter.Type.from_parameter_value(value).value
-                    else:
-                        if value.type == ParameterType.PARAMETER_NOT_SET:
-                            raise ValueError(
-                                'Cannot declare a statically typed parameter with default value '
-                                'of type PARAMETER_NOT_SET')
+                    if isinstance(value, ParameterValue):
                         descriptor.type = value.type
+                    else:
+                        descriptor.type = Parameter.Type.from_parameter_value(value).value
+                    if descriptor.type == ParameterType.PARAMETER_NOT_SET:
+                        raise ValueError(
+                            'Cannot declare a statically typed parameter with default value '
+                            'of type PARAMETER_NOT_SET')
 
             # Get value from parameter overrides, of from tuple if it doesn't exist.
             if not ignore_override and name in self._parameter_overrides:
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/node.py
```diff
@@ -499,16 +499,16 @@ def declare_parameters(
...
-                        descriptor.type = Parameter.Type.from_parameter_value(value).value
-                    else:
-                        if value.type == ParameterType.PARAMETER_NOT_SET:
-                            raise ValueError(
-                                'Cannot declare a statically typed parameter with default value '
-                                'of type PARAMETER_NOT_SET')
+                    if isinstance(value, ParameterValue):
                         descriptor.type = value.type
+                    else:
+                        descriptor.type = Parameter.Type.from_parameter_value(value).value
```
fujitatomoya, line 507: if `value` is `None`, it can deduced to `ParameterType.PARAMETER_NOT_SET` here.

wjwwood (approved):
lgtm

This makes sense, it shouldn't be possible to have an unset parameter when not dynamically typed, since "not set" is considered a separate type, i.e. you cannot have a statically typed "string" or "int" with no value set, afaik.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20252/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14668/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20976/)

fujitatomoya:
@clalancette @wjwwood even though this fixes as designed, i think we should not backport this to iron and humble because it could break the userspace. what do you think? btw should we put this to release note?

clalancette:
> @clalancette @wjwwood even though this fixes as designed, i think we should not backport this to iron and humble because it could break the userspace. what do you think? 

Yes, agreed that we should not backport.  The change in behavior would be surprising for users.

> btw should we put this to release note?

Yes, please do.

haianos:
Hi all,

@clalancette @wjwwood 
let me share my concern about this choice, also partially related to the backport (see https://github.com/ros2/rclpy/pull/1216#issuecomment-1935962301)

While I understand the principle of being by correct by construction, it feels like enforcing an `init` value defeats the purpose of  _declaring_ a parameter -- which use case is left for declaring instead of `set` directly? :innocent: 

So far, I did make use of declared parameters without init value to actually define which configuration (from param file) is required to start my node, following this logic:

- if an init value for a param is set, then overlaying or passing this parameter is _optional_
- if a param is declared but has no initial value, then it is _required_ to overlay or provide this parameter at start.
This is useful, for example, when it is simply not possible to assign, for the developer of the node, an useful initial value, which is known only at bootstrap time (e.g. think of a device path for a driver).

How can I implement such behavior with this change? The only thing that comes to my mind is enabling `dynamic_typing` anywhere and set to `None` as initial value, which doesn't sound to be a nice workaround, as I cannot "validate" if the type of the value passed by configuration actually match with what declared.

As far as I understand `dynamic_typing` as not much to do with it, as in the declaration the type could be already "statically" defined -- `dynamic_typing` isn't involved at all in my user-case.

My other post: https://github.com/ros2/rclpy/issues/1222#issuecomment-1966036522

PR #979, opened 2022-07-23
Improve static typings

mateusz-lichota:
As per issue #976 , this PR aims to improve the statics typings of rclpy.

Issue #976, opened 2022-07-22
Better typings


mateusz-lichota:
Currently using rclpy in intended ways causes typing errors due to partially unknown types of multiple methods. Even the simplest functionality (creating a publisher or a subscription) does not typecheck (tested using pylance). I've written much better typing stubs for myself, including making rclpy Subscription and Publisher generic over the mesage type, and Future generic over the return type. Would it be a welcome change to this repo to add those improved typings? Those are not evaluated at runtime, so they can't break existing code, but it would greatly improve the experience for anyone using static type checking

apockill:
I would personally quite appreciate Subscription/Publisher/Future generics!

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/python-type-checking-in-ros2/28507/1

clalancette:
We've added lot of types in the last year or so, though it is not 100% done yet.  Given that, I'm going to close this out as "mostly done", and we can follow-up with individual PRs as needed.

PR #1227, opened 2024-02-27
Add types to logging_service.py

InvincibleRMC:
Added static typing to logging_service.py. This change was already done similarly in https://github.com/ros2/rclpy/pull/979 in response to issue https://github.com/ros2/rclpy/issues/976.  Also fixes a spelling error of Invaild -> Invalid.

### DIFF
--- a/rclpy/rclpy/logging_service.py
+++ b/rclpy/rclpy/logging_service.py
@@ -12,17 +12,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+from typing import Union, TYPE_CHECKING
+
 from rcl_interfaces.msg import LoggerLevel, SetLoggerLevelsResult
 from rcl_interfaces.srv import GetLoggerLevels
 from rcl_interfaces.srv import SetLoggerLevels
 import rclpy
+from rclpy.impl.logging_severity import LoggingSeverity
 from rclpy.qos import qos_profile_services_default
 from rclpy.validate_topic_name import TOPIC_SEPARATOR_STRING
+
+if TYPE_CHECKING:
+    from rclpy.node import Node
 
 
 class LoggingService:
 
-    def __init__(self, node):
+    def __init__(self, node: Node):
         node_name = node.get_name()
 
         get_logger_name_service_name = \
@@ -39,19 +46,21 @@ class LoggingService:
             self._set_logger_levels, qos_profile=qos_profile_services_default
         )
 
-    def _get_logger_levels(self, request, response):
+    def _get_logger_levels(self, request: GetLoggerLevels.Request,
+                           response: GetLoggerLevels.Response) -> GetLoggerLevels.Response:
         for name in request.names:
             logger_level = LoggerLevel()
             logger_level.name = name
             try:
-                ret_level = rclpy.logging.get_logger_level(name)
+                ret_level: Union[int, LoggingSeverity] = rclpy.logging.get_logger_level(name)
             except RuntimeError:
                 ret_level = 0
             logger_level.level = ret_level
             response.levels.append(logger_level)
         return response
 
-    def _set_logger_levels(self, request, response):
+    def _set_logger_levels(self, request: SetLoggerLevels.Request,
+                           response: SetLoggerLevels.Response) -> SetLoggerLevels.Response:
         for level in request.levels:
             result = SetLoggerLevelsResult()
             result.successful = False
--- a/rclpy/rclpy/logging_service.py
+++ b/rclpy/rclpy/logging_service.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from __future__ import annotations
-from typing import Union, TYPE_CHECKING
+from typing import TYPE_CHECKING, Union
 
 from rcl_interfaces.msg import LoggerLevel, SetLoggerLevelsResult
 from rcl_interfaces.srv import GetLoggerLevels
--- a/rclpy/rclpy/logging_service.py
+++ b/rclpy/rclpy/logging_service.py
@@ -51,7 +51,7 @@ class LoggingService:
             logger_level = LoggerLevel()
             logger_level.name = name
             try:
-                ret_level: Union[LoggingSeverity] = rclpy.logging.get_logger_level(name)
+                ret_level: Union[int, LoggingSeverity] = rclpy.logging.get_logger_level(name)
             except RuntimeError:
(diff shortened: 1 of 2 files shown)


InvincibleRMC:
Would the maintainers prefer a small PR per file like I have currently been doing? Or  do you all have some other preference?

clalancette:
> Would the maintainers prefer a small PR per file like I have currently been doing? Or do you all have some other preference?

Small PRs are much preferred; they are much easier to review and run CI for.  Thanks!

rclpy/rclpy/logging_service.py
```diff
@@ -12,17 +12,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
```
clalancette, line 16: Is this necessary?

rclpy/rclpy/logging_service.py
```diff
@@ -12,17 +12,25 @@
...
+from typing import TYPE_CHECKING, Union
+
 from rcl_interfaces.msg import LoggerLevel, SetLoggerLevelsResult
 from rcl_interfaces.srv import GetLoggerLevels
 from rcl_interfaces.srv import SetLoggerLevels
 import rclpy
+from rclpy.impl.logging_severity import LoggingSeverity
```
clalancette, line 23: I think this should be in the `TYPE_CHECKING` block, as we don't need it otherwise.

rclpy/rclpy/logging_service.py
```diff
@@ -39,27 +47,29 @@ def __init__(self, node):
...
 
-    def _get_logger_levels(self, request, response):
+    def _get_logger_levels(self, request: GetLoggerLevels.Request,
+                           response: GetLoggerLevels.Response) -> GetLoggerLevels.Response:
         for name in request.names:
             logger_level = LoggerLevel()
             logger_level.name = name
             try:
-                ret_level = rclpy.logging.get_logger_level(name)
+                ret_level: Union[int, LoggingSeverity] = rclpy.logging.get_logger_level(name)
```
clalancette, line 56: Does this really need to be a `Union`?  As far as I can tell, [get_logger_level](https://github.com/ros2/rclpy/blob/5cbb110b8ac301d7cdf27768010367dc27f92f59/rclpy/rclpy/logging.py#L58-L60) always returns a `LoggingSeverity`.

clalancette (changes requested):

rclpy/rclpy/logging_service.py
InvincibleRMC: I believe so. Without it I get the following. 
def __init__(self, node: Node):
NameError: name 'Node' is not defined. Did you mean: 'None'?

I believe this case was discussed [`here`](https://peps.python.org/pep-0563/#runtime-annotation-resolution-and-type-checking) in PEP563.

clalancette: The thing is, we don't import that pretty much anywhere else.  So it seems unnecessary.  If we are getting that warning, then I think we probably need to just remove the `TYPE_CHECKING` block, no?  That is, unconditionally import `rclpy.Node`.

rclpy/rclpy/logging_service.py
InvincibleRMC: While get_logger_level always returns a LoggingSeverity ret_level can also be set to int. Mypy prefers an explicit expansion of available types. Without the explicit expansion it correctly identifies ret_level as LoggingSeverity but errors when that type is changed to int. Slightly more info on the type inference can be found [here](https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html#explicit-types-for-variables).

The error in question
Incompatible types in assignment (expression has type "int", variable has type "LoggingSeverity")

rclpy/rclpy/logging_service.py
InvincibleRMC: If we unconditionally import rclpy.Node we get circular imports I believe.

ImportError: cannot import name 'Node' from partially initialized module 'rclpy.node' (most likely due to a circular import) (/opt/ros/iron/lib/python3.10/site-packages/rclpy/node.py)

rclpy/rclpy/logging_service.py
clalancette: > While get_logger_level always returns a LoggingSeverity ret_level can also be set to int.

Oh, that's a bug.  Down below when we get a `RuntimeError`, ret_level should be set to `LoggingSeverity.UNSET`, not 0, which I think should resolve the issue.  There *might* be a further issue when we do the assignment `logger_level.level = ret_level`, but if so we should type annotate that differently, in my opinion.

rclpy/rclpy/logging_service.py
clalancette: Actually, re-reading this I think we may be talking about different things.  Here I was referring to the `from __future__ import annotations`.  I don't think that is necessary.

clalancette: Ah yeah.  OK.  We've done the latter solution in other places, like in launch.  So yeah, let's go for that, thanks.

InvincibleRMC: > Actually, re-reading this I think we may be talking about different things. Here I was referring to the `from __future__ import annotations`. I don't think that is necessary.

Without the `__future__ import annotations` at run time the constructor fails because Node is undefined. The `__future__ import annotations` converts all type annotations to their string version eg `node: Node` becomes `node: 'Node'`

fujitatomoya (approved):
lgtm with green CI

clalancette (approved):
This looks great, thanks for iterating!  I'll run CI on it next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20380/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14789/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21129/)

InvincibleRMC:
Any idea if my changes caused these regressions or is it an unrelated? @clalancette

fujitatomoya:
it does not like related to the fix, CI restarted.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20381/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14790/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21130/)

PR #1230, opened 2024-02-27
Add Static Typing to Validate files

InvincibleRMC:
Added static typing to validate files. This change was already done similarly in https://github.com/ros2/rclpy/pull/979 in response to issue https://github.com/ros2/rclpy/issues/976.

### DIFF
--- a/rclpy/rclpy/validate_full_topic_name.py
+++ b/rclpy/rclpy/validate_full_topic_name.py
@@ -12,12 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Literal
+
 from rclpy.exceptions import InvalidServiceNameException
 from rclpy.exceptions import InvalidTopicNameException
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
 
-def validate_full_topic_name(name, *, is_service=False):
+def validate_full_topic_name(name: str, *, is_service: bool = False) -> Literal[True]:
     """
     Validate a given topic or service name, and raise an exception if invalid.
 
--- a/rclpy/rclpy/validate_full_topic_name.py
+++ b/rclpy/rclpy/validate_full_topic_name.py
@@ -28,8 +28,8 @@ def validate_full_topic_name(name: str, *, is_service: bool = False) -> Literal[
     If the name is invalid then rclpy.exceptions.InvalidTopicNameException
     will be raised.
 
-    :param name str: topic or service name to be validated
-    :param is_service bool: if true, InvalidServiceNameException is raised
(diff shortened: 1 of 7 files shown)


fujitatomoya (approved):
lgtm with green CI

rclpy/rclpy/validate_full_topic_name.py
```diff
@@ -12,12 +12,14 @@
...
 
+from typing import Literal
+
 from rclpy.exceptions import InvalidServiceNameException
 from rclpy.exceptions import InvalidTopicNameException
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
 
-def validate_full_topic_name(name, *, is_service=False):
+def validate_full_topic_name(name: str, *, is_service: bool = False) -> Literal[True]:
```
christophebedard, line 22: Should we remove the typing info from the param documentation below?

```diff
-:param name str: topic or service name to be validated
+:param name: topic or service name to be validated
```

same for all other changes.

InvincibleRMC:
I will defer to maintainers on the return types in the docstrings since I have no preference.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20440/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14842/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21183/)

rclpy/rclpy/validate_full_topic_name.py
fujitatomoya: @InvincibleRMC i think we want to remove the typing information from the doc part if that is typed. **i believe that we should avoid inconsistent typing information as documentation, so if typing is available, probably we do not need to have types in doc section. typing is the correct one.**

if typing is available, it shows on API arguments.

image

on the other hand, doc section comes up like this as an example.

image

InvincibleRMC: That makes sense to me.

sloretz (approved):

sloretz:
CI ([repos file](https://gist.githubusercontent.com/sloretz/3accc4558ac6c157dfa5c5f62d919bd8/raw/876dc0d3024c1853977ecc0caa4543776d8317d8/ros2.repos) build: `--packages-up-to rclpy` test: ` --packages-select rclpy`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20447/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14849/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21190/)

PR #1233, opened 2024-02-28
Add types to duration.py

InvincibleRMC:
Adds types to duration.py. Also adds an RCLDuration Type/Object Alias so type inference can be on the _duration_handle.

### DIFF
--- a/rclpy/rclpy/duration.py
+++ b/rclpy/rclpy/duration.py
@@ -18,11 +18,10 @@ import builtin_interfaces.msg
 from rclpy.constants import S_TO_NS
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
-RCLDuration = _rclpy.rcl_duration_t
 
 
-class RCLDurationType:
-    """Object Alias of _rclpy.rcl_duration_t."""
+class DurationType:
+    """Type alias of _rclpy.rcl_duration_t."""
 
     nanoseconds: int
 
@@ -43,7 +42,7 @@ class Duration:
             # pybind11 would raise TypeError, but we want OverflowError
             raise OverflowError(
                 'Total nanoseconds value is too large to store in C duration.')
-        self._duration_handle: RCLDurationType = RCLDuration(total_nanoseconds)
+        self._duration_handle: DurationType = _rclpy.rcl_duration_t(total_nanoseconds)
 
     @property
     def nanoseconds(self) -> int:
@@ -108,7 +107,7 @@ class Duration:
             raise TypeError('Must pass a builtin_interfaces.msg.Duration object')
         return cls(seconds=msg.sec, nanoseconds=msg.nanosec)
 
-    def get_c_duration(self) -> RCLDurationType:
+    def get_c_duration(self) -> DurationType:
         return self._duration_handle
 
 
--- a/rclpy/rclpy/duration.py
+++ b/rclpy/rclpy/duration.py
@@ -19,10 +19,6 @@ from rclpy.constants import S_TO_NS
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
 
-<<<<<<< HEAD
-=======
-
->>>>>>> 3538803 (update type alias name)
 class DurationType:
     """Type alias of _rclpy.rcl_duration_t."""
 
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/duration.py
```diff
@@ -12,15 +12,25 @@
...
 from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
 
+RCLDuration = _rclpy.rcl_duration_t
+
+
+class RCLDurationType:
+    """Object Alias of _rclpy.rcl_duration_t."""
+
+    nanoseconds: int
+
```
fujitatomoya, line 28:
```suggestion
DurationType = _rclpy.rcl_duration_t

```

rclpy/rclpy/duration.py
```diff
@@ -33,51 +43,51 @@ def __init__(self, *, seconds=0, nanoseconds=0):
             # pybind11 would raise TypeError, but we want OverflowError
             raise OverflowError(
                 'Total nanoseconds value is too large to store in C duration.')
-        self._duration_handle = _rclpy.rcl_duration_t(total_nanoseconds)
+        self._duration_handle: RCLDurationType = RCLDuration(total_nanoseconds)
```
fujitatomoya, line 46: How about this?

```suggestion
        self._duration_handle: DurationType = _rclpy.rcl_duration_t(total_nanoseconds)
```

rclpy/rclpy/duration.py
```diff
@@ -98,7 +108,7 @@ def from_msg(cls, msg):
             raise TypeError('Must pass a builtin_interfaces.msg.Duration object')
         return cls(seconds=msg.sec, nanoseconds=msg.nanosec)
 
-    def get_c_duration(self):
+    def get_c_duration(self) -> RCLDurationType:
```
fujitatomoya, line 111:
```suggestion
    def get_c_duration(self) -> DurationType:
```

InvincibleRMC:
The problem with that solution is without making some custom type alias mypy, and pylance have to treat _rclpy.rcl_duration_t as type Any which weakens the type inference. For example since the _duration_handle is type Any it cannot reason about the type of the nanoseconds field.

rclpy/rclpy/duration.py:44: error: Returning Any from function declared to return "int"  [no-any-return]

```python
@property
    def nanoseconds(self) -> int:
        return self._duration_handle.nanoseconds
```

rclpy/rclpy/duration.py
fujitatomoya, line 111: > The problem with that solution is without making some custom type alias mypy, and pylance have to treat _rclpy.rcl_duration_t as type Any which weakens the type inference. For example since the _duration_handle is type Any it cannot reason about the type of the nanoseconds field.
> 
> rclpy/rclpy/duration.py:44: error: Returning Any from function declared to return "int" [no-any-return]
> 
> ```python
> @property
>     def nanoseconds(self) -> int:
>         return self._duration_handle.nanoseconds
> ```

I see, thanks for the explanation. this does not break the userspace, right? (just checking) previously it returns `Any` but now `RCLDurationType`.

InvincibleRMC: Yes previously it has the type annotation `Any`. And would now have the type annotation `RCLDurationType`. The type annotation does not have any effective on what  the actual return type is which is  `_rclpy_pybind11.Duration` I believe.  Because type annotations are ignored at run time it has no effect on the actual code behavior. I'm not entirely clear what you mean by userspace.

fujitatomoya (approved):

fujitatomoya:
@InvincibleRMC thanks for working on this. the reason i do not start CI is to wait for 2nd review. once we have 2nd approval, i will start the CI. (So are other PRs)

InvincibleRMC:
No problem. Thanks so much for the prompt reviews.

sloretz (approved):

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20409/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14819/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21159/)

PR #1240, opened 2024-03-05
Add types to context.py

InvincibleRMC:
Adds types to context.py also creates a Destroyable Protocol.

### DIFF
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -147,11 +147,13 @@ class Context(ContextManager['Context']):
         """Add a callback to be called on shutdown."""
         if not callable(callback):
             raise TypeError('callback should be a callable, got {}', type(callback))
+
         if self.__context is None:
-            if ismethod(callback):
-                self._callbacks.append(WeakMethod(callback, self._remove_callback))
-            else:
-                self._callbacks.append(callback)
+            with self._lock:
+                if ismethod(callback):
+                    self._callbacks.append(WeakMethod(callback, self._remove_callback))
+                else:
+                    self._callbacks.append(callback)
             return
 
         with self.__context, self._lock:
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -12,20 +12,49 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from inspect import ismethod
 import sys
 import threading
-from types import TracebackType
+from abc import abstractmethod
+from inspect import ismethod
+from types import MethodType, TracebackType
 from typing import Callable
 from typing import ContextManager
 from typing import List
 from typing import Optional
+from typing import Protocol
 from typing import Type
-import weakref
+from typing import Union
+
+from weakref import WeakMethod
+
+
+class DestroyableType(Protocol):
+    def __enter__(self) -> None:
+        ...
+
+    def __exit__(self, exc_type: Optional[Type[BaseException]],
+                 exc_val: Optional[BaseException], exctb: Optional[TracebackType]) -> None:
+        ...
+
+    def destroy_when_not_in_use(self) -> None:
+        ...
+
+
+class ContextType(DestroyableType):
+    @abstractmethod
+    def ok(self) -> bool:
+        ...
+
+    @abstractmethod
+    def get_domain_id(self) -> int:
+        ...
+
+    def shutdown(self) -> None:
+        ...
 
 
 g_logging_configure_lock = threading.Lock()
-g_logging_ref_count = 0
+g_logging_ref_count: int = 0
 
 
 class Context(ContextManager['Context']):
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/context.py
```diff
@@ -12,20 +12,49 @@
...
+
+    def __exit__(self, exc_type: Optional[Type[BaseException]],
+                 exc_val: Optional[BaseException], exctb: Optional[TracebackType]) -> None:
+        ...
+
+    def destroy_when_not_in_use(self) -> None:
+        ...
+
+
+class ContextType(DestroyableType):
```
sloretz, line 43: I'd recommend `ContextHandle` or `InternalContext`. As a casual reader I might confuse `ContextType` and `type(Context)`

sloretz (changes requested):
Thank you for the PR!

I just had time for a partial review. I'll look again later.

rclpy/rclpy/context.py
```diff
@@ -12,20 +12,37 @@
...
+    @abstractmethod
+    def ok(self) -> bool:
+        ...
+
+    @abstractmethod
+    def get_domain_id(self) -> int:
+        ...
+
+    def shutdown(self) -> None:
+        ...
```
fujitatomoya, line 41: why `shutdown` does not have decorator `@abstractmethod`, i think this needs to be implemented?

```suggestion
    @abstractmethod
    def shutdown(self) -> None:
        ...
```

InvincibleRMC: I had misinterpreted  the `Protocol` [documentation](https://typing.readthedocs.io/en/latest/spec/protocol.html#merging-and-extending-protocols). I thought we need `@abstractmethod` for functions that didn't return None however I actually needed to inherit from `Protocol` again. From my understanding `@abstractmethod` would imply `_rclpy.Context()` is some sort of abstract class itself.

rclpy/rclpy/context.py
```diff
@@ -123,23 +140,25 @@ def try_shutdown(self):
...
+    def _remove_callback(self, weak_method: WeakMethod[MethodType]) -> None:
         self._callbacks.remove(weak_method)
 
-    def on_shutdown(self, callback: Callable[[], None]):
+    def on_shutdown(self, callback: Callable[[], None]) -> None:
         """Add a callback to be called on shutdown."""
         if not callable(callback):
             raise TypeError('callback should be a callable, got {}', type(callback))
+        if self.__context is None:
+            raise RuntimeError('Context must be initialized before it can be shutdown')
```
sloretz, line 151: I think we can skip this check. I think it's fine to add `on_shutdown` callbacks before initializing the context because it would be a way to avoid a race condition where one thread initialized the context, but another thread (or the signal handler) shutdown the context before the `on_shutdown` callbacks could be registered.

sloretz (changes requested):

rclpy/rclpy/context.py
InvincibleRMC: Were you thinking of something like this? We need to check if `__context` is None otherwise it will error on `self.__context.ok()`

```python
 def on_shutdown(self, callback: Callable[[], None]) -> None:
        """Add a callback to be called on shutdown."""
        if not callable(callback):
            raise TypeError('callback should be a callable, got {}', type(callback))
        if self.__context is None:
            if ismethod(callback):
                self._callbacks.append(WeakMethod(callback, self._remove_callback))
            else:
                self._callbacks.append(callback)
            return

            if not self.__context.ok():
                callback()
            else:
                if ismethod(callback):
                    self._callbacks.append(WeakMethod(callback, self._remove_callback))
                else:
                    self._callbacks.append(callback)
```

rclpy/rclpy/context.py
```diff
@@ -123,23 +140,30 @@ def try_shutdown(self):
...
-    def _remove_callback(self, weak_method):
+    def _remove_callback(self, weak_method: WeakMethod[MethodType]) -> None:
         self._callbacks.remove(weak_method)
 
-    def on_shutdown(self, callback: Callable[[], None]):
+    def on_shutdown(self, callback: Callable[[], None]) -> None:
         """Add a callback to be called on shutdown."""
         if not callable(callback):
             raise TypeError('callback should be a callable, got {}', type(callback))
+        if self.__context is None:
```
sloretz, line 150: Oh I see, before we were holding `self.__context`, but if that was None then the `with` statement failed. Would you be willing to wrap this new block in `with self._lock:`?

rclpy/rclpy/context.py
sloretz: Ah yeah I missed that we were holding `with self.__context` when adding the callback.

sloretz (approved):
Thank you for the PR!

sloretz:
CI ([repos file](https://gist.githubusercontent.com/sloretz/b930bae33fd14d9b544b81482225dd3f/raw/4d1182128c10420047cd24f6f3b06cd5a486e1df/ros2.repos) build: `--packages-above-and-dependencies rclpy` test: `--packages-above rclpy`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20466/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14868/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21209/)

fujitatomoya (approved):

fujitatomoya:
CI retry:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20468/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14870/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21211/)

sloretz:
CI retry
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20475/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14877/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21218/)

PR #1232, opened 2024-02-28
Add types to TypeHash and moved away from __slots__ usage

InvincibleRMC:
Adds static typing to type_hash.py. In the constructor I switch away from `**kwargs` to explicit variable names. If there is some reason the `**kwargs` is preferred for now I can use Any. Just let me know so I can add a TODO to use Unpack[] when python 3.12 is supported. Or I suppose I could use typing_extensions package to use Unpack on older versions of python.

PR #1243, opened 2024-03-09
Revert "Add types to TypeHash and moved away from __slots__ usage"

clalancette:
Reverts ros2/rclpy#1232

This is causing problems in CI: https://ci.ros2.org/job/ci_linux/20452/testReport/junit/ros2cli.ros2cli.test/test_daemon/test_get_subscriptions_info_by_topic/

In particular, it isn't clear to me why we removed the `__slots__` (as that is a performance enhancement), and it is also causing downstream packages to fail.

### DIFF
--- a/rclpy/rclpy/type_hash.py
+++ b/rclpy/rclpy/type_hash.py
@@ -18,44 +18,56 @@ class TypeHash:
 
     _TYPE_HASH_SIZE = 32
 
-    def __init__(self, version: int = -1, value: bytes = bytes(_TYPE_HASH_SIZE)):
-        self.version = version
-        self.value = value
+    __slots__ = [
+        '_version',
+        '_value',
+    ]
+
+    def __init__(self, **kwargs):
+        assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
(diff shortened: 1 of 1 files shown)


InvincibleRMC:
Ah my apologies. I misunderstood the purpose of `__slots__`.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20454/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14856/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21197/)

fujitatomoya:
> In particular, it isn't clear to me why we removed the __slots__ (as that is a performance enhancement), and it is also causing downstream packages to fail.

@clalancette thanks for taking care of this. sorry i thought there is no dependency for `TypeHash.__slot__` but there is https://github.com/ros2/ros2cli/blob/64d216cb8fafef83d046b79ee6294afb06b7c595/ros2cli/ros2cli/xmlrpc/marshal/rclpy.py#L90-L91

and as you mentioned, i think we should keep `__slots__` for performance and security.

PR #1239, opened 2024-03-04
Using Generics for messages

InvincibleRMC:
First Draft.

PR #1247, opened 2024-03-11
Add types to Waitable super class.

InvincibleRMC:
Adds type annotations to waitable.py. Builds off of #1240 and #1239.


InvincibleRMC:
DCO got scuffed beyond repair will open new PR to replace.

PR #1254, opened 2024-03-14
Add types to task.py

InvincibleRMC:


rclpy/rclpy/task.py
```diff
@@ -128,13 +126,14 @@ def set_exception(self, exception):
...
 
-    def _schedule_or_invoke_done_callbacks(self):
+    def _schedule_or_invoke_done_callbacks(self) -> None:
         """
         Schedule done callbacks on the executor if possible, else run them directly.
 
         This function assumes self._lock is not held.
         """
         with self._lock:
+            assert self._executor is not None
```
fujitatomoya, line 136: this check is required here? i think this could be `None`.

fujitatomoya (commented):
i have one comment to confirm. i would like to request another review.

rclpy/rclpy/task.py
InvincibleRMC: I believe if `self._executor` is `None` then an error of  `'NoneType' object is not callable` would occur with the line `executor = self._executor()`

clalancette (approved):
This looks good to me.  I'm going to go ahead and rebase it onto the latest, since it will certainly need that in order to pass CI.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21571/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15860/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1069/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22291/)

PR #1268, opened 2024-04-02
Replace spin_until_future_complete with spin_until_complete, add spin…

christophebedard:
Part of https://github.com/ros2/rclcpp/pull/2475

Supersedes #919

Due to changes in `rclcpp`, deprecate `spin_until_future_complete` and replace it with `spin_until_complete`. Add `spin_for`.


rclpy/rclpy/executors.py
```diff
@@ -307,32 +307,53 @@ def spin(self) -> None:
...
+        elif (callable(condition)):
+            def finish_condition(): return condition()
+        else:
+            raise TypeError('Condition has to be of Future or Callable type')
 
         if timeout_sec is None or timeout_sec < 0:
-            while self._context.ok() and not future.done() and not self._is_shutdown:
-                self.spin_once_until_future_complete(future, timeout_sec)
+            while self._context.ok() and not finish_condition() and not self._is_shutdown:
+                self.spin_once_until_complete(finish_condition, timeout_sec)
```
christophebedard, line 333: The original PR had `condition` here, but that didn't work. I think it should be `finish_condition` here. Same thing below.

christophebedard: This does not apply anymore.

christophebedard:
I've updated this PR to reflect the decision in https://github.com/ros2/rclcpp/pull/2475#discussion_r1550080593 to not deprecate `spin_until_future_complete` and not have `spin_until_complete` accept both a future and a condition.

christophebedard:
@fujitatomoya could you review this? This reflects the new changes to https://github.com/ros2/rclcpp/pull/2475, which should be final-ish now.

christophebedard:
Well this may be going in yet another direction (https://github.com/ros2/rclcpp/pull/2475#discussion_r1559540603), so let's wait a bit.

christophebedard:
> Well this may be going in yet another direction ([ros2/rclcpp#2475 (comment)](https://github.com/ros2/rclcpp/pull/2475#discussion_r1559540603)), so let's wait a bit.

All good now, this is ready for review.

rclpy/rclpy/__init__.py
```diff
@@ -242,6 +243,52 @@ def spin(node: 'Node', executor: Optional['Executor'] = None) -> None:
...
+    node: 'Node',
+    condition: Callable[[], bool],
+    executor: Optional['Executor'] = None,
+    timeout_sec: Optional[float] = None,
+) -> None:
+    """
+    Execute work until the condition is complete.
+
+    Callbacks and other work will be executed by the provided executor until ``condition()``
+    returns ``True`` or the context associated with the executor is shutdown.
```
wjwwood, line 275: nitpick: no mention of the timeout here

rclpy/rclpy/__init__.py
```diff
@@ -242,6 +243,52 @@ def spin(node: 'Node', executor: Optional['Executor'] = None) -> None:
...
+
+    :param node: A node to add to the executor to check for work.
+    :param condition: The callable condition to wait on. If this condition is not related to what
+        the executor is waiting on and the timeout is infinite, this could block forever.
+    :param executor: The executor to use, or the global executor if ``None``.
+    :param timeout_sec: Seconds to wait. Block until the condition is complete
+        if ``None`` or negative. Don't wait if 0.
+    """
```
wjwwood, line 283: To go along with the discussion in the rclcpp version of this (https://github.com/ros2/rclcpp/pull/2475#discussion_r1550039913), I think we need to do a much better job of explaining that the executor must be woken up when the condition changes and condition's value must be changed _before_ the signal to wake up is sent, otherwise you might deadlock here (if not using a timeout).

I think we could make an issue to do this as a follow up, as the docs can be improved post API freeze.

rclpy/test/test_executor.py
```diff
@@ -402,6 +418,72 @@ def timer_callback():
...
+        def condition():
+            nonlocal condition_var
+            return condition_var
+
+        # Condition complete timeout_sec > 0
+        self.assertFalse(condition())
+        t = threading.Thread(target=lambda: set_condition())
+        t.start()
+        executor.spin_until_complete(condition, timeout_sec=1.0)
+        self.assertTrue(condition())
```
wjwwood, line 445: This is a fairly weak test in my opinion, as you could comment out line 444 and have it pass most of the time. Also, if `spin_until_complete()` was broken (and did not return before the timeout if the condition was true) this would still pass.

rclpy/rclpy/__init__.py
```diff
@@ -242,6 +243,52 @@ def spin(node: 'Node', executor: Optional['Executor'] = None) -> None:
...
+    finally:
+        executor.remove_node(node)
+
+
+def spin_until_complete(
+    node: 'Node',
+    condition: Callable[[], bool],
+    executor: Optional['Executor'] = None,
+    timeout_sec: Optional[float] = None,
+) -> None:
```
wjwwood, line 270: Should this not return something to indicate if the function returned due to the condition being complete or due to a timeout instead?

wjwwood (commented):
Very quick look from me and it looks pretty good, but I do think the function ought to return something like it does in c++. However, another rclpy maintainer can override me if they think this could be added later or that it isn't necessary.

Also, I feel that the tests are kind of shallow and wouldn't catch some of the ways the function could fail.

rclpy/rclpy/__init__.py
christophebedard: Then that would mean that `spin_until_future_complete` should also something similar?

christophebedard: Just so we agree before I do the changes: do we just want a `bool` (`True` means it returned due to the condition being complete, or `False` otherwise), or do we something like `rclcpp::SpinUntilCompleteReturnCode` to indicate `SUCCESS` or `INTERRUPTED` or `TIMEOUT`?

Issue #1126, opened 2023-05-11
Warn if Timer frequency is not met


tonynajjar:
#### Feature description
How about adding a warning if the timer frequency is not met? That would be useful for the user to either decrease the frequency or optimize the callback.

#### Implementation considerations

In case the timer is implemented with a ReentrantCallback group, another instance of the callback would always fire; we should decide if we want to warn nevertheless if the first instance of the callback takes longer than expected to complete :thinking:

fujitatomoya:
i think that would be helpful.

> How about adding a warning if the timer frequency is not met?

required accuracy is really dependent on user application. for doing this, i think we do need to ask application about acceptable jitter or something like that to the timer?

my concern is there would be many warning messages. i think there are many possibilities that can cause this warning, executor becomes busy, system is busy and so on. to avoid having many warnings, probably it can print just once when the requirement does not match. and then we are not sure how often it happens, which is really good information if we want to tune the timer frequency. we can have time window for warning message, but it gives more complexity for the timer?

sorry, i just left my note above, i really do not have good idea. i would like to ask for comments on this.

apockill:
I'm all for more visibility when it comes to logging, but I wouldn't want to add more complexity or compute to the system. 

What if before a timer callback gets called, a check is performed to see if the (time - last_time_run) > intended_delay, and log there? It would be simple to implement and maintain, and give some transparency to users.

I also agree logging once might be a good idea.

tonynajjar:
cross-linking as this seems related: https://github.com/ros2/rclcpp/pull/2343

fujitatomoya:
@tonynajjar thanks, i was going to do that, but dropped it from my radar. we need to implement it in rclpy as counterpart of https://github.com/ros2/rclcpp/pull/2343

fujitatomoya:
@tonynajjar @apockill if you can review https://github.com/ros2/rclpy/pull/1292, that would be appreciated!

fujitatomoya:
@apockill @tonynajjar https://github.com/ros2/rclpy/pull/1292 has been merged in rclpy, so that we can get the `TimerInfo` in the application timer callback. i will go ahead to close this one.

PR #1292, opened 2024-06-02
Add TimerInfo to timer callback.

fujitatomoya:
address https://github.com/ros2/rclpy/issues/1126

counterparf of rclcpp implementation https://github.com/ros2/rclcpp/pull/2343


rclpy/rclpy/executors.py
```diff
@@ -366,7 +366,19 @@ def spin_once_until_future_complete(
     def _take_timer(self, tmr):
         try:
             with tmr.handle:
-                tmr.handle.call_timer()
+                info = tmr.handle.call_timer_with_info()
```
fujitatomoya, line 369: This still works but,

something we can do for the performance improvement is to get the signature when we create timer, not calling here every single time, and call either `call_timer()` or `call_timer_with_info()` accordingly?

fujitatomoya:
@mjcarroll @ahcorde @clalancette can you take a look?

CC: @jmachowinski @tonynajjar @aprotyas

mjcarroll (approved):

rclpy/rclpy/executors.py
jmachowinski: in the rcl layer, there is none.

fujitatomoya: @mjcarroll sorry i was not clear on this.

i was thinking that calling `inspect.signature` could be optimized to cache the callback type or something, but i guess that approach is bad assumption since user application would change the callback for `Timer` class. besides, this is gonna be really minor optimization. i will keep the current code for now.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21285/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15604/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/782/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22022/)

fujitatomoya:
according to https://ci.ros2.org/job/ci_linux/21290/console, `test_launch_ros` gets stuck because of this fix. i am expecting that something is missing to support `partial(...)` case. anyway, i will stop all CIs for now.

Barry-Xu-2018:
> according to https://ci.ros2.org/job/ci_linux/21290/console, `test_launch_ros` gets stuck because of this fix. i am expecting that something is missing to support `partial(...)` case. anyway, i will stop all CIs for now.

About the stuck, it is related to ` inspect.signature(tmr.callback).bind(object())`
I still can't explain the root cause of being stuck.

Here it's just checking if the callback has one parameter.
Could it be changed to
```python
    def _take_timer(self, tmr):
        try:
            with tmr.handle:
                info = tmr.handle.call_timer_with_info()
                if len(inspect.signature(tmr.callback).parameters) == 1:
                    timer_info = TimerInfo(
                        expected_call_time=info['expected_call_time'],
                        actual_call_time=info['actual_call_time'],
                        clock_type=tmr.clock.clock_type)                        
                    async def _execute():
                        await await_or_execute(tmr.callback, timer_info)
                else:
                    async def _execute():
                        await await_or_execute(tmr.callback)
                return _execute
        except InvalidHandle:
            # Timer is a Destroyable, which means that on __enter__ it can throw an
            # InvalidHandle exception if the entity has already been destroyed.  Handle that here
            # by just returning an empty argument, which means we will skip doing any real work
            # in _execute_timer below
            pass

        return None
```

rclpy/rclpy/executors.py
```diff
@@ -366,11 +367,35 @@ def spin_once_until_future_complete(
...
+                arg_name = check_argument_type(tmr.callback, target_type=TimerInfo)
+                prefilled_arg = {arg_name: timer_info}
+                if arg_name is not None:
+                    async def _execute():
+                        await await_or_execute(partial(tmr.callback, **prefilled_arg))
+                    return _execute
+                else:
+                    async def _execute():
+                        await await_or_execute(tmr.callback)
+                    return _execute
```
fujitatomoya, line 398: at least this does not block `colcon test --event-handlers console_direct+ --packages-select test_launch_ros` and pass all tests without getting stuck.

rclpy/rclpy/executors.py
```diff
@@ -366,11 +367,35 @@ def spin_once_until_future_complete(
...
+                info = tmr.handle.call_timer_with_info()
+                timer_info = TimerInfo(
+                    expected_call_time=info['expected_call_time'],
+                    actual_call_time=info['actual_call_time'],
+                    clock_type=tmr.clock.clock_type)
+
+                # This needs to be done somewhere else since this is certain overhead to call timer
+                # for every iteration. at this moment stays here for review purpose.
+                # Besides, probably we can make this generic utility function
+                def check_argument_type(callback_func, target_type):
```
fujitatomoya, line 379: I believe user can change the callback in the application, so signature check needs to be done here. but we can do that in the 1st place in this scope, so that we can save overhead especially for the application does not need `TimerInfo`.

fujitatomoya:
> > according to https://ci.ros2.org/job/ci_linux/21290/console, `test_launch_ros` gets stuck because of this fix. i am expecting that something is missing to support `partial(...)` case. anyway, i will stop all CIs for now.
> 
> About the stuck, it is related to ` inspect.signature(tmr.callback).bind(object())` I still can't explain the root cause of being stuck.
> 
> Here it's just checking if the callback has one parameter. Could it be changed to
> 
> ```python
>     def _take_timer(self, tmr):
>         try:
>             with tmr.handle:
>                 info = tmr.handle.call_timer_with_info()
>                 if len(inspect.signature(tmr.callback).parameters) == 1:
>                     timer_info = TimerInfo(
>                         expected_call_time=info['expected_call_time'],
>                         actual_call_time=info['actual_call_time'],
>                         clock_type=tmr.clock.clock_type)                        
>                     async def _execute():
>                         await await_or_execute(tmr.callback, timer_info)
>                 else:
>                     async def _execute():
>                         await await_or_execute(tmr.callback)
>                 return _execute
>         except InvalidHandle:
>             # Timer is a Destroyable, which means that on __enter__ it can throw an
>             # InvalidHandle exception if the entity has already been destroyed.  Handle that here
>             # by just returning an empty argument, which means we will skip doing any real work
>             # in _execute_timer below
>             pass
> 
>         return None
> ```

@Barry-Xu-2018 thanks for the suggestion.

this does not work either, because callback would have multiple parameters.

e.g. https://github.com/ros2/launch_ros/blob/rolling/test_launch_ros/test/test_launch_ros/actions/test_ros_timer.py#L173

fujitatomoya:
CI:
Building on the built-in node in workspace /var/lib/jenkins/jobs/ci_launcher/workspace
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21306/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15622/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/804/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22037/)

Barry-Xu-2018:
> this does not work either, because callback would have multiple parameters.
> e.g. https://github.com/ros2/launch_ros/blob/rolling/test_launch_ros/test/test_launch_ros/actions/test_ros_timer.py#L173

Oh, yeah. Finding parameter names through type is the correct way.

rclpy/rclpy/executors.py
```diff
@@ -366,11 +367,35 @@ def spin_once_until_future_complete(
...
+                            return param.name
+                    # We could not find the target type in the signature
+                    return None
 
-                async def _execute():
-                    await await_or_execute(tmr.callback)
-                return _execute
+                arg_name = check_argument_type(tmr.callback, target_type=TimerInfo)
+                prefilled_arg = {arg_name: timer_info}
+                if arg_name is not None:
```
Barry-Xu-2018, line 391: How about move this line under if
```suggestion
                if arg_name is not None:
                    prefilled_arg = {arg_name: timer_info}
```

fujitatomoya:
https://ci.ros2.org/job/ci_linux/21306/#showFailuresLink does not look related to the fix, but just in case i want to retry.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21309/)

rclpy/rclpy/executors.py
fujitatomoya: @Barry-Xu-2018 thanks for the comment. this can be improved more, i would like to wait for the feedback for this approach before adding the test cases and cosmetic changes. but thanks!

fujitatomoya:
@ahcorde @sloretz @mjcarroll @clalancette i could use review for this.

ahcorde (changes requested):
There is a test failure https://build.ros2.org/job/Rpr__rclpy__ubuntu_noble_amd64/161/testReport/junit/rclpy.rclpy.test/test_timer/test_cancel_reset_0_001_/ is this related with this PR ?

fujitatomoya:
> There is a test failure https://build.ros2.org/job/Rpr__rclpy__ubuntu_noble_amd64/161/testReport/junit/rclpy.rclpy.test/test_timer/test_cancel_reset_0_001_/ is this related with this PR ?

i do not think this is related, all tests are pass in my local environment.

i did add a couple of cosmetic fixes including new test case, and rebasing the current rolling.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21574/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15863/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1072/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22294/)

ahcorde (approved):

fujitatomoya:
https://ci.ros2.org/job/ci_linux/21574/ is not related rclpy action.

PR #1293, opened 2024-06-03
Make nodes context-manager aware.

### LINKED ISSUE #1280: Make nodes, publishers, subscriptions, services, clients, action servers, action clients Python context managers
Let's say that you want to create an rclpy node with a publisher and a subscription.  In order to do this entirely correctly today, you would have to do the following:

```
rclpy.init()
try:
  node = rclpy.create_node('minimal_action_server')
  try:
    publisher = node.create_publisher(String, 'chatter', 1)
    try:
      subscription = node.create_subscription(String, 'chatter', lambda msg: print(msg), 1)
      try:
        do_thing()
      finally:
        node.destroy_subscription(subscription)
    finally:
      node.destroy_publisher(publisher)
  finally:
    node.destroy_node()
finally:
  rclpy.try_shutdown()
```

That is, every time we successfully create an entity, we really need to put a `try..finally` block around it to clean up only those parts that we've initialized so far.  This is very unwieldy, so we don't actually recommend it in our examples or documentation.

However, we should be able to do better.  In particular, if all of the initialization, nodes, publishers, subscriptions, services, clients, action servers, and action clients were [Python context managers](https://docs.python.org/3/library/contextlib.html) (i.e. they all implemented `__enter__` and `__exit__`), this could be cleaned up to something like:

```
with rclpy.init() as init:
  with rclpy.create_node('minimal_action_server') as node:
    with node.create_publisher(String, 'chatter', 1) as pub:
      with node.create_subscription(String, 'chatter', lambda msg: print(msg), 1) as subscription:
        do_thing()
```

Or even better, using a [contextlib.ExitStack](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack):

```
with ExitStack() as stack:
    stack.enter_context(rclpy.init())
    node = stack.enter_context(rclpy.create_node('minimal_action_server'))
    publisher = stack.enter_context(node.create_publisher(String, 'chatter', 1))
    subscription = stack.enter_context(node.create_subscription(String, 'chatter', lambda msg: print(msg), 1))
    do_thing()
```

This came out of discussion around https://github.com/ros2/examples/pull/379

clalancette:
This way it is much easier to create examples that properly clean up after themselves.

This partially fixes #1280, but we'll need more PRs to fully complete that.

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -15,6 +15,7 @@
 import math
 import time
 
+from types import TracebackType
 from typing import Any
 from typing import Callable
 from typing import Dict
@@ -2295,3 +2296,14 @@ class Node:
             flag = fully_qualified_node_name in fully_qualified_node_names
             time.sleep(0.1)
         return flag
+
+    def __enter__(self) -> 'Node':
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> None:
+        self.destroy_node()
--- a/rclpy/test/test_node.py
+++ b/rclpy/test/test_node.py
@@ -2528,6 +2528,17 @@ class TestCreateNode(unittest.TestCase):
         finally:
             rclpy.shutdown(context=context)
 
+    def test_node_context_manager(self):
+        context = rclpy.context.Context()
+        rclpy.init(context=context)
+
+        try:
+            with rclpy.create_node('test_node_no_sim', context=context) as node:
+                self.assertTrue(node.has_parameter(USE_SIM_TIME_NAME))
+                self.assertFalse(node.get_parameter(USE_SIM_TIME_NAME).value)
+        finally:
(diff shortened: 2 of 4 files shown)


mjcarroll (approved):
Yes, this is a great improvement.  Do you want to add at least once test here to cover this before doing the follow up work?

fujitatomoya (approved):

clalancette:
> Yes, this is a great improvement. Do you want to add at least once test here to cover this before doing the follow up work?

Yeah, that makes sense.  I've added a test for the node, plus for the subscriptions and publishers (that we've already merged in).

mjcarroll (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21212/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15521/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21933/)

clalancette:
The failing test is a known flake on Windows, so going ahead and merging this one.  Thanks for the reviews.

PR #1298, opened 2024-06-11
Make rclpy initialization context-manager aware.

clalancette:
This PR does two somewhat controversial things:

1.  It adds in a new "rclpy.managed_init" method to rclpy, which would be used with a context manager instead of rclpy.init.  I had to do things this way because there is no way (as far as I know) for the callee to know whether it was called within a context manager or not, and I don't think we can reasonably change the semantics of rclpy.init() at this point.

2.  It changes the context manager implementation of Context so that it does *not* call init() within itself. This definitely changes the semantics, but as it stands that initialization doesn't make sense because it can't take arguments.  I think this change is warranted, though we may have to search through documentation and examples to make sure this doesn't break anything.

@sloretz I particularly am interested in your feedback on these changes.

### DIFF
--- a/rclpy/rclpy/__init__.py
+++ b/rclpy/rclpy/__init__.py
@@ -40,6 +40,8 @@ all ROS nodes associated with the context), the :func:`shutdown` function should
 This will invalidate all entities derived from the context.
 """
 
+import contextlib
+
 from typing import List
 from typing import Optional
 from typing import TYPE_CHECKING
@@ -86,7 +88,35 @@ def init(
         else:
             signal_handler_options = SignalHandlerOptions.NO
     install_signal_handlers(signal_handler_options)
-    return context.init(args, domain_id=domain_id)
+    context.init(args, domain_id=domain_id)
+
+
+@contextlib.contextmanager
+def managed_init(
+    *,
+    args: Optional[List[str]] = None,
+    context: Optional[Context] = None,
+    domain_id: Optional[int] = None,
+    signal_handler_options: Optional[SignalHandlerOptions] = None,
+) -> None:
+    """
+    Initialize ROS communications for a given context, in a Python context-manager aware way.
+
+    :param args: List of command line arguments.
+    :param context: The context to initialize. If ``None``, then the default context is used
+        (see :func:`.get_default_context`).
+    :param domain_id: ROS domain id.
+    :param signal_handler_options: Indicate which signal handlers to install.
+        If `None`, SIGINT and SIGTERM will be installed when initializing the default context.
+    """
+    init(args=args, context=context, domain_id=domain_id,
+         signal_handler_options=signal_handler_options)
+
+    context = get_default_context() if context is None else context
+    try:
+        yield context
+    finally:
+        context.try_shutdown()
 
 
 # The global spin functions need an executor to do the work
--- a/rclpy/rclpy/context.py
+++ b/rclpy/rclpy/context.py
@@ -187,9 +187,11 @@ class Context(ContextManager['Context']):
             return self.__context.get_domain_id()
 
     def __enter__(self) -> 'Context':
-        # We do not accept parameters here. If one wants to customize the init() call,
-        # they would have to call it manually and not use the ContextManager convenience
-        self.init()
+        if self.__context is None:
+            # This object hasn't been initialized yet, so we can't use a context manager
+            raise RuntimeError(
+                'init() must be called on this Context before using it with a context manager')
+
         return self
 
     def __exit__(
--- a/rclpy/test/test_context.py
+++ b/rclpy/test/test_context.py
@@ -61,6 +61,8 @@ def test_context_manager():
 
     assert not context.ok(), 'the context should not be ok() before init() is called'
 
+    context.init()
+
     with context as the_context:
         # Make sure the correct instance is returned
         assert the_context is context
--- a/rclpy/test/test_init_shutdown.py
+++ b/rclpy/test/test_init_shutdown.py
@@ -109,3 +109,9 @@ def test_signal_handlers():
 def test_init_with_invalid_domain_id():
     with pytest.raises(RuntimeError):
         rclpy.init(domain_id=-1)
+
+
+def test_managed_init():
+    with rclpy.managed_init(domain_id=123) as context:
+        assert context.get_domain_id() == 123
+        assert context.ok()


rclpy/rclpy/context.py
```diff
@@ -187,9 +187,11 @@ def get_domain_id(self) -> int:
             return self.__context.get_domain_id()
 
     def __enter__(self) -> 'Context':
-        # We do not accept parameters here. If one wants to customize the init() call,
-        # they would have to call it manually and not use the ContextManager convenience
-        self.init()
+        if self.__context is None:
+            # This object hasn't been initialized yet, so we can't use a context manager
+            raise RuntimeError(
+                'init() must be called on this Context before using it with a context manager')
```
fujitatomoya, line 193: > It changes the context manager implementation of Context so that it does not call init() within itself. This definitely changes the semantics

i may be totally wrong...

this change is required because `managed_init` calls `init` including `context.init`? so if `managed_init` implements inline without calling `init` and `context.init`, this does not need to be changed?

clalancette: > this change is required because `managed_init` calls `init` including `context.init`? so if `managed_init` implements inline without calling `init` and `context.init`, this does not need to be changed?

Partially yes.  But I also think that the current implementation is just wrong.  Calling `init` inside of `__enter__`, without the ability to pass arguments, just seems incorrect to me.  I think it is much better to require that users already call `context.init()` before entering the context-manager.

fujitatomoya: thanks for the explanation, yes that makes more sense to me too.

clalancette:
After discussing this with @sloretz , we don't need the `managed_init` method at all; we can just return the context from `init()`.  I've now done this in 25a26a5 .

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21269/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15588/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/764/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21995/)

rclpy/rclpy/__init__.py
```diff
@@ -86,7 +86,8 @@ def init(
         else:
             signal_handler_options = SignalHandlerOptions.NO
     install_signal_handlers(signal_handler_options)
-    return context.init(args, domain_id=domain_id)
+    context.init(args, domain_id=domain_id)
+    return context
```
christophebedard, line 90: The return type (`None`) and docstring should probably be updated.

rclpy/rclpy/context.py
```diff
@@ -187,9 +187,11 @@ def get_domain_id(self) -> int:
             return self.__context.get_domain_id()
 
     def __enter__(self) -> 'Context':
-        # We do not accept parameters here. If one wants to customize the init() call,
-        # they would have to call it manually and not use the ContextManager convenience
-        self.init()
+        if self.__context is None:
+            # This object hasn't been initialized yet, so we can't use a context manager
+            raise RuntimeError(
```
sloretz, line 192: What do you think about calling `self.init()` here to keep backwards compatibility?

clalancette: I think I'd be OK with doing that, if we also emit a warning that this usage is deprecated and that `init` should be called before entering the context.

rclpy/rclpy/__init__.py
clalancette: Yeah, good call.  I've done that now (although I also ended up doing a much more substantial update).

clalancette:
All right, I've now heavily revamped this based on feedback from @sloretz and @wjwwood .  In particular, we now are doing a couple of different things:

1.  We have a proxy initialization object where we deal with initialization and shutdown.  This allows us to do *all* of the work we need to do when initializing and shutting down, and also allows us to do it with a context manager.
2.  We track all nodes associated with contexts, and properly destroy the nodes when the context is shutdown.  Since nodes also shutdown all pubs, subs, clients, servers, and timers associated with it when it shuts down, this means that it is now enough to do:

```
with rclpy.init():
    node = rclpy.create_node()
    pub = node.create_publisher()
    sub = node.create_subscription()
```

And everything will be properly cleaned up after when the context is exited.

Please take another look when you get a chance.

clalancette:
Here is new CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21273/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15592/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/768/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21999/)

rclpy/rclpy/context.py
clalancette: And this is now done in [b81046d](https://github.com/ros2/rclpy/pull/1298/commits/b81046d6fe806db44f57385f321f60b59a3babdb)

clalancette:
New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21274/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15594/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/769/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22000/)

sloretz (commented):
This is looking great!

rclpy/rclpy/context.py
```diff
@@ -82,6 +88,10 @@ def init(self,
         Initialize ROS communications for a given context.
 
         :param args: List of command line arguments.
+        :param initialize_logging: Whether to initialize logging for this context.
+            The default is to initialize logging.
```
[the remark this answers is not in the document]
clalancette: Yeah, you are right.  Super confusing here, I'll accept this suggestion.

rclpy/rclpy/context.py
```diff
@@ -106,6 +116,35 @@ def init(self,
                         _rclpy.rclpy_logging_configure(self.__context)
                 self._logging_initialized = True
 
+    def track_node(self, node: 'Node') -> None:
+        """
+        Track a Node associated with this Context.
+
+        During destruction, this list of weak Node references will be used to destroy resources
+        associated with this Context.
```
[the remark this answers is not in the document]
clalancette: > Nit, I think this wording is clearer. Feel free to ignore.

Agreed, your wording is much better.

rclpy/rclpy/context.py
```diff
@@ -106,6 +116,35 @@ def init(self,
...
+        self.__node_weak_ref_list.append(weakref.ref(node))
+
+    def untrack_node(self, node: 'Node') -> None:
+        """
+        Stop tracking a Node associated with this Context.
+
+        If a Node is destroyed before the context, we no longer need to track it for destruction of
+        the Context, so remove it here.
+        """
+        for index, weak_node in enumerate(self.__node_weak_ref_list):
```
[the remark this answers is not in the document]
clalancette: Yep, good call.  Fixed in 103073a88aa9b2e2a29fe12480baa3ed207ad6a5.

rclpy/rclpy/context.py
```diff
@@ -187,9 +232,12 @@ def get_domain_id(self) -> int:
             return self.__context.get_domain_id()
 
     def __enter__(self) -> 'Context':
-        # We do not accept parameters here. If one wants to customize the init() call,
-        # they would have to call it manually and not use the ContextManager convenience
-        self.init()
+        if self.__context is None:
+            # init() hasn't been called yet; for backwards compatibility, initialize and warn
+            warnings.warn('init() must be called on a Context before using it in a Python context '
+                          'manager. Calling init() with no arguments, this usage is deprecated')
```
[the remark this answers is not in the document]
clalancette: I kind of think what is currently there explains a bit better.  That is, I want to tell the user that we called init() on their behalf with no arguments, but that using the Context context manager like this is deprecated.  Further, calling `init()` with no arguments is *not* deprecated; it is just deprecated that the `__enter__` method does it for you.

For now I'm going to leave this as-is, though I am open to other wording that explains all of the above in a better way.

clalancette:
All right, after responding to review, here is another set of CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21277/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15597/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/772/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22005/)

sloretz (approved):

PR #1302, opened 2024-06-18
Add top-level try_shutdown method.

clalancette:
I noticed that this was needed when trying to convert some of the examples over to the context manager.  They currently throw an exception because the signal handler does the shutdown, and then leaving the context tries to do it again.

This is also more symmetric with what shutdown() does, in that it also uninstalls signal handlers.

This is a follow-up to #1298 

@sloretz FYI


sloretz (approved):

fujitatomoya (approved):
lgtm with green CI

Issue #1303, opened 2024-06-18
Async tasks are processed in LIFO not FIFO


jpfeltracco:
## Bug report

**Required Info:**

- Operating System:
  - [ros:jazzy docker tag](https://github.com/osrf/docker_images/blob/0038f1c3a11aa0fc573d698b39ab5c204aad5a40/ros/jazzy/ubuntu/noble/ros-base/Dockerfile)
- Installation type:
  - binaries
- Version or commit hash:
  - Jazzy 0.11.0-1
- DDS implementation:
  - Whatever is default
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

```python3
import rclpy
from rclpy.node import Node
from rclpy.executors import SingleThreadedExecutor

class SimpleNode(Node):
    def __init__(self, n=10):
        super().__init__('simple_node')
        self.n = n
        # Create a one-shot timer with a zero duration to start tasks after initialization
        self.timer = self.create_timer(0, self.start_tasks)

    def start_tasks(self):
        self.timer.cancel()
        # Scheduling tasks to print their execution order
        tasks = [self.executor.create_task(self.print_value(i)) for i in range(self.n)]

    async def print_value(self, value):
        print(f"Task executed with value: {value}")

def main(args=None):
    rclpy.init(args=args)
    node = SimpleNode()
    executor = SingleThreadedExecutor()
    executor.add_node(node)
    try:
        executor.spin()  # Spin the node to process callbacks
    except Exception as e:
        node.get_logger().error(f'Unhandled exception in executor: {e}')
    finally:
        executor.shutdown()
        node.destroy_node()

if __name__ == '__main__':
    main()
```

#### Expected behavior

1, 2, 3..., 9

Would expect output to be roughly 0->9. I understand no guarantees without awaiting explicitly, but generally would expect tasks created earlier in time should be scheduled for execution earlier in FIFO fashion. This matches how asyncio in base Python works.

#### Actual behavior

9, 8, 7, 6..., 0

#### Additional information

See this code snippet for how asyncio works (producing 0, 1, 2 ... 9)

```python
import asyncio

async def print_value(value):
    # Simulate some asynchronous operation
    print(f"Task started with value: {value}")

async def main(n=10):
    for i in range(n):
        asyncio.create_task(print_value(i))
    # Give control to executor
    await asyncio.sleep(1)

# Run the main function with the default number of tasks
asyncio.run(main())
```

sloretz:
[...]

PR #1304, opened 2024-06-20
Executor executes the tasks in FIFO order.

fujitatomoya:
address https://github.com/ros2/rclpy/issues/1303

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -208,9 +208,10 @@ class Executor(ContextManager['Executor']):
         """
         Add a callback or coroutine to be executed during :meth:`spin` and return a Future.
 
-        Created task is queued in the executor in FIFO order.
-
         Arguments to this function are passed to the callback.
+
+        .. warning:: Created task is queued in the executor in FIFO order,
+           but users should not rely on the task execution order.
 
         :param callback: A callback to be run in the executor.
         """
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -208,6 +208,8 @@ class Executor(ContextManager['Executor']):
         """
         Add a callback or coroutine to be executed during :meth:`spin` and return a Future.
 
+        Created task is queued in the executor in FIFO order.
+
         Arguments to this function are passed to the callback.
 
         :param callback: A callback to be run in the executor.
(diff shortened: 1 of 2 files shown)


fujitatomoya:
@sloretz @Barry-Xu-2018 can you take a look at this?

i am not sure if we can backport this to jazzy since it changes behavior. i believe user application should not rely on the execution order, but we cannot say there is no application relies on the current behavior with released distro.

rclpy/rclpy/executors.py
```diff
@@ -208,6 +208,8 @@ def create_task(self, callback: Union[Callable, Coroutine], *args, **kwargs) ->
         """
         Add a callback or coroutine to be executed during :meth:`spin` and return a Future.
 
+        Created task is queued in the executor in FIFO order.
+
```
fujitatomoya, line 212: probably we should add more documentation that says execution order is not guaranteed? so user application should not rely on the execution order?

rclpy/test/test_executor.py
```diff
@@ -270,7 +270,7 @@ def func():
         self.assertTrue(future.done())
         self.assertEqual('Sentinel Result', future.result())
 
-    def test_create_task_dependent_coroutines(self):
+    def test_create_task_fifo_order(self):
```
fujitatomoya, line 273: this new test is similar with `test_create_task_dependent_coroutines` but it is the most simple case to check if the tasks are executed in FIFO order without dependent coroutines.

Barry-Xu-2018:
LGTM

sloretz (approved):
LGTM. I agree that this should not be backported

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21327/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15640/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/834/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22057/)

fujitatomoya:
https://ci.ros2.org/job/ci_linux-rhel/834/#showFailuresLink is unrelated.

fujitatomoya:
Note: we will not backport this since it changes the execution order in task queue. even though this fixes the order into FIFO which is expected order, this would be surprising behavior change for some user application.

PR #1306, opened 2024-06-23
Add missing `Optional` types to `node.py` on `humble` branch

alberthli:
Added `Optional` types to function declarations in `node.py` where relevant on the `humble` branch. Missing `Optional` types cause static type checking failures. Example:
```
class ExampleNode(Node):
    def __init__(self, cbg: Optional[CallbackGroup] = None) -> None:
        super().__init__("example_node")
        self.timer = self.create_timer(
            0.001,
            self.timer_callback,
            callback_group=cbg,  # fails static type checking - this field wasn't annotated as Optional[CallbackGroup]
        )
```

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -117,13 +117,13 @@ class Node:
         self,
         node_name: str,
         *,
-        context: Context = None,
-        cli_args: List[str] = None,
-        namespace: str = None,
+        context: Optional[Context] = None,
+        cli_args: Optional[List[str]] = None,
+        namespace: Optional[str] = None,
         use_global_arguments: bool = True,
         enable_rosout: bool = True,
         start_parameter_services: bool = True,
-        parameter_overrides: List[Parameter] = None,
+        parameter_overrides: Optional[List[Parameter]] = None,
         allow_undeclared_parameters: bool = False,
         automatically_declare_parameters_from_overrides: bool = False
     ) -> None:
@@ -1397,7 +1397,7 @@ class Node:
         srv_name: str,
         *,
         qos_profile: QoSProfile = qos_profile_services_default,
-        callback_group: CallbackGroup = None
+        callback_group: Optional[CallbackGroup] = None
     ) -> Client:
         """
         Create a new service client.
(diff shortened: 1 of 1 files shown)


Barry-Xu-2018:
LGTM.
The rolling version already operates in this way.

alberthli:
Great. Are there any next steps before merging?

Barry-Xu-2018:
The next step is to pass the CI since this PR modifies the code.  
Please wait for the CI test results.  

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21315/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15630/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/816/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22046/)

fujitatomoya (commented):
@alberthli thanks for the PR.

although this fixes static type checking, i guess we would want to proceed these changes from the upstream and then backport to downstream to avoid those fixes are scattered over distros. could you address those issues in rolling or pick up merged PR to be backported to released distros?

btw, @InvincibleRMC has been working on these type checking with many PRs, e.g [InvincibleRMC](https://github.com/InvincibleRMC)

CC: @clalancette @sloretz

clalancette:
> although this fixes static type checking, i guess we would want to proceed these changes from the upstream and then backport to downstream to avoid those fixes are scattered over distros. could you address those issues in rolling or pick up merged PR to be backported to released distros?

It looks like these are already fixed on the `rolling` branch, so we are OK there.

In terms of backporting, I'm actually of the opinion that we should take this PR as-is.  While it would be nice to cherry-pick changes from the upstream branches, I think it will be just as much work as doing them one-by-one here.  And since @alberthli already has this one queued up and ready to go, I think we should just take it (though obviously we have to re-run CI here to make sure it is passing).

fujitatomoya:
thanks, and im okay. sometimes backporting is not straight forward either so would require some adjustment.

clalancette:
CI looks good here, so going ahead and merging this one in.  Thanks @alberthli !

PR #1311, opened 2024-07-07
Avoid causing infinite loop when message is empty (#935)

xueying4402:
Same as [ad389c](https://github.com/ros2/rclpy/pull/935), but for `humble`.

* Avoid causing infinite loop when message is empty
* Extend the sleep time to 0.2

### DIFF
--- a/rclpy/test/test_client.py
+++ b/rclpy/test/test_client.py
@@ -102,8 +102,7 @@ class TestClient(unittest.TestCase):
         try:
             self.assertTrue(cli.wait_for_service(timeout_sec=20))
             cli.call_async(GetParameters.Request())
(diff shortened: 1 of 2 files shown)


Barry-Xu-2018:
Does iron have the same issue ?
Perhaps it is better to use Mergify bot to do backport. @fujitatomoya @jacobperron

fujitatomoya (approved):
lgtm with green CI.

@Barry-Xu-2018 iron has been fixed already with https://github.com/ros2/rclpy/pull/935

sloretz:
Thank you for the PR!

Closing for in favor of the mergify bot #1334

Issue #1209, opened 2024-01-05
Lifecycle node `Transition is not registered` exception kills node


mjforan:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - binaries (in Docker)
- Version or commit hash:
  - Humble
- DDS implementation:
  - rmw_cyclonedds_cpp
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

The example below is based on [the official demo code](https://github.com/ros2/demos/blob/rolling/lifecycle_py/lifecycle_py/talker.py).

Launch this node and then attempt to execute an invalid transition from the command line, e.g. run the `cleanup` transition while already in the `unconfigured` state: 
```ros2 service call /lc_demo/change_state lifecycle_msgs/srv/ChangeState "{transition: {id: 2}}"```

```
from typing import Optional
import rclpy
from rclpy.lifecycle import LifecycleNode
from rclpy.lifecycle import LifecyclePublisher
from rclpy.lifecycle import LifecycleState
from rclpy.lifecycle import TransitionCallbackReturn
from rclpy.qos import QoSProfile, ReliabilityPolicy, DurabilityPolicy, HistoryPolicy
from rclpy.timer import Timer

from std_msgs.msg import Bool

class LifecycleSubDemo(LifecycleNode):
  
  def __init__(self, node_name, **kwargs):
    super().__init__(node_name, **kwargs)

    self._count: int = 0
    self._pub: Optional[LifecyclePublisher] = None
    self._timer: Optional[Timer] = None

  def publish(self):
    self._pub.publish(Bool(data=False))
  

  def on_configure(self, state: LifecycleState) -> TransitionCallbackReturn:
    self._pub = self.create_lifecycle_publisher(Bool, "/test", 1)
    self._timer = self.create_timer(1.0, self.publish, autostart=False)
    return TransitionCallbackReturn.SUCCESS

  def on_activate(self, state: LifecycleState) -> TransitionCallbackReturn:
    # publisher automatically activated by parent class transition method
    r = super().on_activate(state)
    self._timer.reset()
    return r
  
  def on_deactivate(self, state: LifecycleState) -> TransitionCallbackReturn:
    self._timer.cancel()
    return super().on_deactivate(state)

  def on_cleanup(self, state: LifecycleState) -> TransitionCallbackReturn:
    self.destroy_timer(self._timer)
    self.destroy_publisher(self._pub)
[...]

PR #1317, opened 2024-07-22
Reject wrong lifecycle transitions without crash (#1209)

chrisbitter:
Fix for #1209 

Instead of attempting to do a lifecycle transition and potentially crashing the node, it is first checked if a transition is valid. If it is not, it is ignored. The service caller gets notified via `success = False`.

### DIFF
--- a/rclpy/rclpy/lifecycle/node.py
+++ b/rclpy/rclpy/lifecycle/node.py
@@ -350,6 +350,12 @@ class LifecycleNodeMixin(ManagedEntity):
     ):
         self.__check_is_initialized()
         transition_id = req.transition.id
+        
+        available_transition_ids = [t[0] for t in self._state_machine.available_transitions]
+        if transition_id not in available_transition_ids:
+              resp.success = False
+              return resp
+        
         if req.transition.label:
             try:
                 transition_id = self._state_machine.get_transition_by_label(req.transition.label)


Barry-Xu-2018:
Please fix flake8 errors https://build.ros2.org/job/Rpr__rclpy__ubuntu_noble_amd64/181/

rclpy/rclpy/lifecycle/node.py
```diff
@@ -350,6 +350,12 @@ def __on_change_state(
     ):
         self.__check_is_initialized()
         transition_id = req.transition.id
+        
+        available_transition_ids = [t[0] for t in self._state_machine.available_transitions]
+        if transition_id not in available_transition_ids:
+              resp.success = False
+              return resp
+
```
fujitatomoya, line 358: according to the stack trace from https://github.com/ros2/rclpy/issues/1209#issuecomment-2244019251, the root cause of this exception is from `__change_state`. i think what we should do here is to catch the exception from the failure from rcl transition change, and print the error for user application.

Barry-Xu-2018: Yes. This PR only addresses the issue of external state changes through the "change_state" service.   
Modifying __change_state is more appropriate.

mjcarroll:
@fujitatomoya has this been replaced by #1319?  If so, can this one be closed?

fujitatomoya:
@mjcarroll yes, i will go ahead to close this.

PR #1345, opened 2024-08-24
Executors types

InvincibleRMC:


rclpy/rclpy/executors.py
```diff
@@ -597,37 +640,40 @@ def _wait_for_ready_callbacks(
...
-                clients.extend(filter(self.can_execute, node.clients))
-                services.extend(filter(self.can_execute, node.services))
-                node_guards = filter(self.can_execute, node.guards)
-                waitables.extend(filter(self.can_execute, node.waitables))
-                # retrigger a guard condition that was triggered but not handled
-                for gc in node_guards:
-                    if gc._executor_triggered:
-                        gc.trigger()
-                    guards.append(gc)
+            if nodes_to_use:
```
sloretz, line 657: Nit, if I follow correctly `nodes_to_use` is always a list, so we don't need to check `if nodes_to_use` because the `for` loop will do nothing in the case of an empty list.

rclpy/rclpy/executors.py
```diff
@@ -727,53 +776,56 @@ def _wait_for_ready_callbacks(
                         gc._executor_triggered = True
 
                 # Check waitables before wait set is destroyed
+                if nodes_to_use:
```
sloretz, line 779: Same nit about `nodes_to_use`

rclpy/rclpy/executors.py
```diff
@@ -727,53 +776,56 @@ def _wait_for_ready_callbacks(
...
+                        for wt in node.waitables:
+                            # Only check waitables that were added to the wait set
+                            if wt in waitables and wt.is_ready(wait_set):
+                                if wt.callback_group.can_execute(wt):
+                                    handler = self._make_handler(wt, node, self._take_waitable)
+                                    yielded_work = True
+                                    yield handler, wt, node
+
+            # Process ready entities one node at a time
+            if nodes_to_use:
```
sloretz, line 790: Same nit about `nodes_to_use`

sloretz (commented):
LGTM with a couple nitpicks addressed!

rclpy/rclpy/executors.py
```diff
@@ -387,15 +416,19 @@ def check_argument_type(callback_func, target_type):
 
                 # User might change the Timer.callback function signature at runtime,
                 # so it needs to check the signature every time.
-                arg_name = check_argument_type(tmr.callback, target_type=TimerInfo)
-                prefilled_arg = {arg_name: timer_info}
+                if tmr.callback:
+                    arg_name = check_argument_type(tmr.callback, target_type=TimerInfo)
```
fujitatomoya, line 420: these code area is not directly related to type checking, but refactoring makes sense to me. thanks!

rclpy/rclpy/executors.py
```diff
@@ -742,10 +790,11 @@ def _wait_for_ready_callbacks(
...
                         if tmr.handle.is_timer_ready():
-                            if tmr.callback_group.can_execute(tmr):
-                                handler = self._make_handler(tmr, node, self._take_timer)
-                                yielded_work = True
-                                yield handler, tmr, node
+                            if tmr.callback_group:
+                                if tmr.callback_group.can_execute(tmr):
+                                    handler = self._make_handler(tmr, node, self._take_timer)
+                                    yielded_work = True
+                                    yield handler, tmr, node
```
fujitatomoya, line 797:
```suggestion
                            if tmr.callback_group and tmr.callback_group.can_execute(tmr):
                                handler = self._make_handler(tmr, node, self._take_timer)
                                yielded_work = True
                                yield handler, tmr, node
```

rclpy/rclpy/executors.py
```diff
@@ -756,7 +805,7 @@ def _wait_for_ready_callbacks(
 
                 for gc in node.guards:
                     if gc._executor_triggered:
-                        if gc.callback_group.can_execute(gc):
+                        if gc.callback_group and gc.callback_group.can_execute(gc):
```
fujitatomoya, line 808: Why do we not need to add the same check for waitables, subscriptions, clients and services?

rclpy/rclpy/executors.py
```diff
@@ -844,8 +895,9 @@ def _spin_once_impl(
             pass
         else:
             handler()
-            if handler.exception() is not None:
-                raise handler.exception()
+            exception = handler.exception()
+            if exception is not None:
+                raise exception
```
fujitatomoya, line 900: I think this was kind of bug to cause the internal state mismatch? when `handler.exception()` is called `_exception_fetched` is already true, that means we should keep that fetched exception like this. just checking.

fujitatomoya (approved):
overall lgtm with green CI, a few comments.

rclpy/rclpy/callback_groups.py
```diff
@@ -24,7 +24,8 @@
     from rclpy.service import Service
     from rclpy.waitable import Waitable
     from rclpy.guard_condition import GuardCondition
-    Entity = Union[Subscription, Timer, Client, Service, Waitable[Any], GuardCondition]
+    Entity = Union[Subscription[Any], Timer, Client[Any, Any, Any], Service[Any, Any, Any],
```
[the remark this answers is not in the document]
fujitatomoya: Ah, you already addressed this one. ignore me, sorry.

rclpy/rclpy/executors.py
InvincibleRMC: Since `GuardCondition` can have Optional `.callback_group`. And the others always have a `CallbackGroup`

rclpy/rclpy/executors.py
InvincibleRMC: Yes I believe it could cause a mismatch in case somehow some other function got called in between so storing the exception should avoid it.

InvincibleRMC:
@fujitatomoya I fixed a typing mistake should be good to re-run.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21822/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16092/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1324/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22522/)

InvincibleRMC:
@fujitatomoya is this good to merge in?

PR #1346, opened 2024-08-25
Add types to Node.py

InvincibleRMC:


rclpy/rclpy/node.py
```diff
@@ -168,57 +169,59 @@ def __init__(
             to get and set logger levels of this node. Otherwise, logger levels are only managed
             locally. That is, logger levels cannot be changed remotely.
         """
-        self.__handle = None
```
fujitatomoya, line 171: good eye, thanks!

rclpy/rclpy/node.py
```diff
@@ -168,57 +169,59 @@ def __init__(
...
-                validate_node_name(node_name)
-                # emulate what rcl_node_init() does to accept '' and relative namespaces
-                if not namespace:
-                    namespace = '/'
-                if not namespace.startswith('/'):
-                    namespace = '/' + namespace
-                validate_namespace(namespace)
-                # Should not get to this point
-                raise RuntimeError('rclpy_create_node failed for unknown reason')
+        if self._context.handle:
```
fujitatomoya, line 195: This changes the behavior if that is NULL? before it catches `ValueError` but after it passes this if statement and proceed to the following code?

rclpy/rclpy/exceptions.py
```diff
@@ -109,7 +109,7 @@ class InvalidParameterTypeException(ParameterException):
 
     from rclpy.parameter import Parameter
 
-    def __init__(self, desired_parameter: Parameter, expected_type: Parameter.Type) -> None:
+    def __init__(self, desired_parameter: Parameter, expected_type: str) -> None:
```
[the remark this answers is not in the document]
InvincibleRMC: Yes when I initially did the type for that exception I mistakenly thought it took the Parameter.Type object rather than the name field of that object.

rclpy/rclpy/node.py
InvincibleRMC: Do you think it would be better to raise an Exception if it is `None` rather than attempt to carry on with the rest of the `__init__` function?

InvincibleRMC: Ok I switched it to raising an exception if it does not exist.

InvincibleRMC: Another solution to potentially avoid all these `context.handle` None checks would be to move the Context's `init` function directly into the `__init__` function. That way the context handle would always exist. However, there might be some reason I am unaware of for this delayed construction.

rclpy/rclpy/node.py
```diff
@@ -168,29 +169,34 @@ def __init__(
...
         self._descriptors: Dict[str, ParameterDescriptor] = {}
 
         namespace = namespace or ''
+
         if not self._context.ok():
             raise NotInitializedException('cannot create node')
+
+        if self._context.handle is None:
+            raise RuntimeError('Context must be initialized before a node can be created')
+
```
fujitatomoya, line 199: i wonder if we need to have this check for the `Context.handle` because above call `self._context.ok()` already verifies that context is valid. IMO we can remove this change for now, and not directly related to the objective for this PR? what do you think?

fujitatomoya:
@InvincibleRMC thanks for iterating. just one comment i want to hear your opinion. after that one is solved, i will start the CI.

rclpy/rclpy/node.py
InvincibleRMC: I agree that `self._context.ok()` does the same check however in Python's type system there is no way to have a function give information about other attributes.


Would the following be cleaner in your opinion?
```python
if self._context.handle is None or not self._context.ok():
    raise NotInitializedException('cannot create node')
 ```

fujitatomoya: makes sense to me! thanks for the suggestion

fujitatomoya:
Pulls: ros2/rclpy#1346
Gist: https://gist.githubusercontent.com/fujitatomoya/191023b5d7c74b2fd4a3c32a3b9a8829/raw/585d226800af1d203306e8f24f5038004da4c161/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14480
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21751/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16025/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1255/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22453/)

InvincibleRMC:
@fujitatomoya  is this good to merge or does this need another review?

InvincibleRMC:
Resolved conflict in .pyi file

sloretz (approved):

sloretz:
Pulls: ros2/rclpy#1346
Gist: https://gist.githubusercontent.com/sloretz/8ca37d9c72412b90694110953fb816ff/raw/585d226800af1d203306e8f24f5038004da4c161/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14495
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21769/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16041/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1273/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22468/)

sloretz:
Thank you for the PR!

Issue #1352, opened 2024-09-02
`rcl_shutdown` is not thread-safe when used in both `signal_handler` and `Context.__exit__`.


YuanYuYuan:
## Bug report

We found this sporadic failure with rclpy (rolling) due to the race condition while calling `rcl_shutdown`.
In this [issue](https://github.com/ros2/rmw_zenoh/issues/275), the conflict happens if `rmw_shutdown` is slow so that `rcutils_atomic_store` on Thread 1 is set after the check `rcl_context_is_valid` on Thread 2. Therefore `rcl_shutdown` would be called twice and cause an error.

## Analysis
- Thread 1
  - [rclpy::shutdown_contexts](https://github.com/ros2/rclpy/blob/8a2e2d3de5a96eaa50b9c9336d97396f8489bda4/rclpy/src/rclpy/signal_handler.cpp#L144)
  - [rcl_shutdown](https://github.com/ros2/rclpy/blob/8a2e2d3de5a96eaa50b9c9336d97396f8489bda4/rclpy/src/rclpy/context.cpp#L49)
  - [check rcl_context_is_valid](https://github.com/ros2/rcl/blob/522f3476c76d911d72922584b18df1fa1761f608/rcl/src/rcl/init.c#L289)
  - [rmw_shutdown](https://github.com/ros2/rcl/blob/522f3476c76d911d72922584b18df1fa1761f608/rcl/src/rcl/init.c#L294)  // Slow
  - [rcutils_atomic_store](https://github.com/ros2/rcl/blob/522f3476c76d911d72922584b18df1fa1761f608/rcl/src/rcl/init.c#L301)

- Thread 2
  - [InitContextManger.\_\_exit\_\_](https://github.com/ros2/rclpy/blob/8a2e2d3de5a96eaa50b9c9336d97396f8489bda4/rclpy/rclpy/__init__.py#L112)
  - [_try_shutdown](https://github.com/ros2/rclpy/blob/8a2e2d3de5a96eaa50b9c9336d97396f8489bda4/rclpy/rclpy/__init__.py#L200)
  - [g_default_context.try_shutdown()](https://github.com/ros2/rclpy/blob/8a2e2d3de5a96eaa50b9c9336d97396f8489bda4/rclpy/rclpy/utilities.py#L100)
  - [if self.__context.ok()](https://github.com/ros2/rclpy/blob/8a2e2d3de5a96eaa50b9c9336d97396f8489bda4/rclpy/rclpy/context.py#L172)
[...]

PR #1353, opened 2024-09-03
Fix the race condition while calling rcl_shutdown

Barry-Xu-2018:
Address #1352

### DIFF
--- a/rclpy/src/rclpy/context.cpp
+++ b/rclpy/src/rclpy/context.cpp
@@ -151,6 +151,9 @@ void
 Context::shutdown()
 {
   std::lock_guard<std::mutex> guard{g_contexts_mutex};
+  if (already_shutdown_) {
+    throw CONTEXT_ALREADY_SHUTDOWN("Context already shutdown.");
+  }
   auto iter = std::find(g_contexts.begin(), g_contexts.end(), rcl_context_.get());
   if (iter != g_contexts.end()) {
     g_contexts.erase(iter);
@@ -158,6 +161,7 @@ Context::shutdown()
     if (RCL_RET_OK != ret) {
       throw RCLError("failed to shutdown");
     }
+    already_shutdown_ = true;
   }
 }
 
--- a/rclpy/src/rclpy/context.cpp
+++ b/rclpy/src/rclpy/context.cpp
@@ -152,7 +152,7 @@ Context::shutdown()
 {
   std::lock_guard<std::mutex> guard{g_contexts_mutex};
   if (already_shutdown_) {
-    throw CONTEXT_ALREADY_SHUTDOWN("Context already shutdown.");
+    throw ContextAlreadyShutdown("Context already shutdown.");
   }
   auto iter = std::find(g_contexts.begin(), g_contexts.end(), rcl_context_.get());
   if (iter != g_contexts.end()) {
--- a/rclpy/src/rclpy/context.cpp
+++ b/rclpy/src/rclpy/context.cpp
@@ -154,11 +154,10 @@ Context::shutdown()
   auto iter = std::find(g_contexts.begin(), g_contexts.end(), rcl_context_.get());
   if (iter != g_contexts.end()) {
     g_contexts.erase(iter);
-  }
-
-  rcl_ret_t ret = rcl_shutdown(rcl_context_.get());
-  if (RCL_RET_OK != ret) {
-    throw RCLError("failed to shutdown");
+    rcl_ret_t ret = rcl_shutdown(rcl_context_.get());
+    if (RCL_RET_OK != ret) {
+      throw RCLError("failed to shutdown");
+    }
   }
 }
 
--- a/rclpy/src/rclpy/context.cpp
+++ b/rclpy/src/rclpy/context.cpp
@@ -150,12 +150,10 @@ Context::ok()
 void
 Context::shutdown()
 {
-  {
-    std::lock_guard<std::mutex> guard{g_contexts_mutex};
-    auto iter = std::find(g_contexts.begin(), g_contexts.end(), rcl_context_.get());
-    if (iter != g_contexts.end()) {
-      g_contexts.erase(iter);
-    }
+  std::lock_guard<std::mutex> guard{g_contexts_mutex};
+  auto iter = std::find(g_contexts.begin(), g_contexts.end(), rcl_context_.get());
+  if (iter != g_contexts.end()) {
+    g_contexts.erase(iter);
   }
 
   rcl_ret_t ret = rcl_shutdown(rcl_context_.get());
--- a/rclpy/src/rclpy/exceptions.hpp
+++ b/rclpy/src/rclpy/exceptions.hpp
@@ -80,7 +80,7 @@ class InvalidHandle : public std::runtime_error
   using std::runtime_error::runtime_error;
 };
 
-class CONTEXT_ALREADY_SHUTDOWN : public std::runtime_error
+class ContextAlreadyShutdown : public std::runtime_error
 {
   using std::runtime_error::runtime_error;
 };
--- a/rclpy/src/rclpy/exceptions.hpp
+++ b/rclpy/src/rclpy/exceptions.hpp
@@ -80,6 +80,11 @@ class InvalidHandle : public std::runtime_error
   using std::runtime_error::runtime_error;
 };
 
+class CONTEXT_ALREADY_SHUTDOWN : public std::runtime_error
+{
+  using std::runtime_error::runtime_error;
+};
+
 }  // namespace rclpy
 
 #endif  // RCLPY__EXCEPTIONS_HPP_
--- a/rclpy/src/rclpy/context.hpp
+++ b/rclpy/src/rclpy/context.hpp
@@ -101,6 +101,7 @@ public:
 
 private:
   std::shared_ptr<rcl_context_t> rcl_context_;
+  bool already_shutdown_{false};
 };
 
 /// Define a pybind11 wrapper for an rclpy::Context


Barry-Xu-2018:
@sloretz 

Regarding the fix for issue #1352, I'd like to hear your thoughts.

About issue #1352, there is an issue

One thread is calling rclpy::shutdown()

https://github.com/ros2/rclpy/blob/8a2e2d3de5a96eaa50b9c9336d97396f8489bda4/rclpy/src/rclpy/context.cpp#L44-L53

Another thread is calling Context::shutdown()

https://github.com/ros2/rclpy/blob/8a2e2d3de5a96eaa50b9c9336d97396f8489bda4/rclpy/src/rclpy/context.cpp#L151-L165

rcl_shutdown() will be called twice. And rcl_shutdown() may be called at the same time (Calling rcl_shutdown() isn't protected by `g_contexts_mutex` in Context::shutdown()).  
I changed the codes in Context::shutdown().   
But this brings a problem: multiple calls to Context::shutdown() will no longer trigger an exception. This will cause the below existing test to fail.
https://github.com/ros2/rclpy/blob/8a2e2d3de5a96eaa50b9c9336d97396f8489bda4/rclpy/test/test_init_shutdown.py#L77-L83

My question is whether we still need to ensure that multiple calls to Context::shutdown() throw an exception ?  
If so, I need to add a member in the Context to indicate whether shutdown() has already been called.

Barry-Xu-2018:
Friendly ping @sloretz ?

rclpy/test/test_init_shutdown.py
```diff
@@ -74,15 +74,6 @@ def test_double_init():
         rclpy.shutdown(context=context)
 
 
-def test_double_shutdown():
-    context = rclpy.context.Context()
-    rclpy.init(context=context)
-    assert context.ok()
-    rclpy.shutdown(context=context)
-    with pytest.raises(RuntimeError):
```
fujitatomoya, line 82: Do we need to remove this test?

I think we can keep this test, but expects `RCLError` exception here to catch from 2nd `rcl_shutdown` (internally `RCL_RET_ALREADY_SHUTDOWN` returns in rcl) ?

fujitatomoya (commented):
IMO this is correct fix.

rclpy/test/test_init_shutdown.py
Barry-Xu-2018: Thanks for your comments.

> I think we can keep this test, but expects RCLError exception here to catch from 2nd rcl_shutdown (internally RCL_RET_ALREADY_SHUTDOWN returns in rcl) ?

With the current fix, if the context is not in g_contexts, rcl_shutdown will not be called.  

According to the issue described in #1352, it is indeed possible for two threads to call rcl_shutdown on the same context. However, one call is from rclpy::shutdown() and the other from Context::shutdown(). In this situation, we do not want to see an exception being thrown.

The test calls Context::shutdown() twice, and it is reasonable to expect an exception on the second call. Therefore, I am considering setting a flag variable in the Context class. If Context::shutdown() has been successfully called, this variable will be set. If Context::shutdown() is called again, it will check if the flag has been set and, if so, will throw an exception CONTEXT_ALREADY_SHUTDOWN (not from RCL). What do you think?

Barry-Xu-2018: > Therefore, I am considering setting a flag variable in the Context class. If Context::shutdown() has been successfully called, this variable will be set. If Context::shutdown() is called again, it will check if the flag has been set and, if so, will throw an exception CONTEXT_ALREADY_SHUTDOWN (not from RCL). What do you think?

Please review https://github.com/ros2/rclpy/pull/1353/commits/0eb7ee57731d0b2dbf44288493290e5049b55fc0

rclpy/src/rclpy/context.cpp
```diff
@@ -150,17 +150,18 @@ Context::ok()
...
-    }
+  std::lock_guard<std::mutex> guard{g_contexts_mutex};
+  if (already_shutdown_) {
+    throw CONTEXT_ALREADY_SHUTDOWN("Context already shutdown.");
   }
-
-  rcl_ret_t ret = rcl_shutdown(rcl_context_.get());
-  if (RCL_RET_OK != ret) {
-    throw RCLError("failed to shutdown");
+  auto iter = std::find(g_contexts.begin(), g_contexts.end(), rcl_context_.get());
```
fujitatomoya, line 157: so the `g_contexts` is the collection of valid context, and this means if it cannot find the context in the `g_contexts` during this shutdown call, that is the  invalid operation? so we can generate the exception without introducing `already_shutdown_`?

Barry-Xu-2018: > so the g_contexts is the collection of valid context, and this means if it cannot find the context in the g_contexts during this shutdown call, that is the invalid operation?

Yes.  
`rc_context_` is added to g_contexts in constructor of Context.  Before calling rcl_shutdown(), `rc_context_` is removed from g_contexts.

> so we can generate the exception without introducing already_shutdown_?

`already_shutdown_` is only used for multiple calls to Context::shutdown() to throw the exception.   

About issue reported by https://github.com/ros2/rclpy/issues/1352, it describes a possible scenario. There are two threads to call rcl_shutdown for the same context. One call is from rclpy::shutdown() and the other from Context::shutdown(). In this situation, we do not want to get an exception in Context::shutdown().

Yadunund:
@fujitatomoya do you mind doing another round of review here whenever you get a chance?

fujitatomoya:
@Barry-Xu-2018 is this still draft? are we waiting for someone or something to make if official?

Barry-Xu-2018:
> @Barry-Xu-2018 is this still draft? are we waiting for someone or something to make if official?

Oh, I forgot to change the status.

rclpy/src/rclpy/context.cpp
fujitatomoya: that is exactly why we have `std::lock_guard<std::mutex> guard{g_contexts_mutex};` lock escalated to the this function scope? so that when [rclpy::shutdown_contexts](https://github.com/ros2/rclpy/blob/8a2e2d3de5a96eaa50b9c9336d97396f8489bda4/rclpy/src/rclpy/context.cpp#L49) is under process, we will not find the iterator with this context, so generate the exception. i really do not understand to have `already_shutdown_` flag internally...

Barry-Xu-2018: @fujitatomoya 

> i really do not understand to have already_shutdown_ flag internally...

Let me first summarize the two scenarios we currently need to address:

1. The scenario mentioned in #1352. There are two threads to call rcl_shutdown for the same context. One call is from rclpy::shutdown() and the other from Context::shutdown(). For this scenario, we don't want to get exception.   
2. For one context instance, Context::shutdown() is being called multiple times. At this scenario, we want to throw an exception. (There is an existed test to check if an exception is thrown.)  The fix for 1 would prevent 2 from throwing an exception. That's why already_shutdown_ ``` was added to indicate that Context::shutdown() has already been called once on the context instance. Do you have any better suggestions?

Barry-Xu-2018: @fujitatomoya Friendly ping.

fujitatomoya: thanks for the explanation, i understand what you mean. if we take `already_shutdown_`, then we need to reinitialize `already_shutdown_` during context initialization? unless, `already_shutdown_` will never be False?

IMO, `Context::shutdown()` should generate the exception only when `rcl_shutdown` fails with valid context. this actually aligns with `rclcpp` behavior. of course, if we change like this, we need to adjust the unit tests accordingly.

maybe we can keep this open for more feedback.

Barry-Xu-2018: > thanks for the explanation, i understand what you mean. if we take already_shutdown_, then we need to reinitialize already_shutdown_ during context initialization? unless, already_shutdown_ will never be False?

Yes. 

> IMO, Context::shutdown() should generate the exception only when rcl_shutdown fails with valid context. this actually aligns with rclcpp behavior. of course, if we change like this, we need to adjust the unit tests accordingly.

Yes. The addition of already_shutdown_ is just to pass the existing tests (Ensure consistency with previous behavior.)

> maybe we can keep this open for more feedback.

Yes.

rclpy/src/rclpy/exceptions.hpp
```diff
@@ -80,6 +80,11 @@ class InvalidHandle : public std::runtime_error
   using std::runtime_error::runtime_error;
 };
 
+class CONTEXT_ALREADY_SHUTDOWN : public std::runtime_error
```
clalancette, line 83:
```suggestion
class ContextAlreadyShutdown : public std::runtime_error
```

clalancette (changes requested):
I'm generally fine with this logic (and it seems to fix some of the errors we've been having in Zenoh).  I've left one naming thing to fix, and this needs to be rebased.  Once those things are done, I'm happy to approve and run CI on this.

rclpy/src/rclpy/exceptions.hpp
Barry-Xu-2018: Thanks. I will update codes.

fujitatomoya:
@Barry-Xu-2018 

> if we take already_shutdown_, then we need to reinitialize already_shutdown_ during context initialization? unless, already_shutdown_ will never be False?

i think this still needs to be addressed. with that, i am okay with this PR!

Barry-Xu-2018:
Run CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22105/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16360/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1624/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22845/)

Barry-Xu-2018:
@fujitatomoya 

> if we take already_shutdown_, then we need to reinitialize already_shutdown_ during context initialization? unless, already_shutdown_ will never be False?

One context has one` already_shutdown_`.  Based on my understanding, once context has been shut down, it should not be used again.  That is, user need to create another context.  
So already_shutdown_ need not change from 'True' to 'False'.

ahcorde:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22847/)

clalancette (approved):

fujitatomoya (approved):
@Barry-Xu-2018 thanks for the explanation, as you say there is no reinitialize way to do for user application.

fujitatomoya:
@Barry-Xu-2018 @clalancette btw, we can backport this to Jazzy? i think we can, in the `cpp` code it breaks the ABI, but this is rclpy python client library, so nobody should rely on the cpp code in here.

Barry-Xu-2018:
>  we can backport this to Jazzy?

I think yes. What do you think ? @clalancette

PR #1387, opened 2024-12-11
Add support for operator overloading of `Duration`

arjo129:
This PR adds support for operator overloading of the `Duration` class in python. This should improve the overall UX of the rclpy library.

### DIFF
--- a/rclpy/rclpy/duration.py
+++ b/rclpy/rclpy/duration.py
@@ -49,6 +49,21 @@ class Duration:
             return 'Infinite'
         return f'{self.nanoseconds} nanoseconds'
 
+    def __add__(self, other: 'Duration') -> 'Duration':
+        if isinstance(other, Duration):
+            return Duration(nanoseconds=other.nanoseconds + self.nanoseconds)
+        return NotImplemented
+
+    def __sub__(self, other: 'Duration') -> 'Duration':
+        if isinstance(other, Duration):
+            return Duration(nanoseconds=self.nanoseconds - other.nanoseconds)
+        return NotImplemented
+
+    def __mul__(self, other: int) -> 'Duration':
+        if isinstance(other, int):
+            return Duration(nanoseconds=self.nanoseconds * other)
+        return NotImplemented
+
     def __eq__(self, other: object) -> bool:
         if isinstance(other, Duration):
             return self.nanoseconds == other.nanoseconds
--- a/rclpy/rclpy/duration.py
+++ b/rclpy/rclpy/duration.py
@@ -71,10 +71,6 @@ class Duration:
             return Duration(nanoseconds=int(self.nanoseconds * other))
         return NotImplemented
 
-    def __floordiv__(self, other: int) -> 'Duration':
-        if isinstance(other, int):
-            return Duration(nanoseconds=self.nanoseconds // other)
-        return NotImplemented
 
     def __eq__(self, other: object) -> bool:
         if isinstance(other, Duration):
--- a/rclpy/rclpy/duration.py
+++ b/rclpy/rclpy/duration.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import math
 from typing import Union
 
 import builtin_interfaces.msg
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):
lgtm with green CI

fujitatomoya:
this is part of https://github.com/ros2/ros2/issues/1625

fujitatomoya:
Pulls: ros2/rclpy#1387
Gist: https://gist.githubusercontent.com/fujitatomoya/cc6f8e4150322bcf9e6a19e5be8008b3/raw/fa8c89f821580b760357b33bb764989451ce353b/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14937
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22249/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16509/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1771/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23023/)

rclpy/rclpy/duration.py
```diff
@@ -49,6 +49,21 @@ def __str__(self) -> str:
...
+        if isinstance(other, Duration):
+            return Duration(nanoseconds=other.nanoseconds + self.nanoseconds)
+        return NotImplemented
+
+    def __sub__(self, other: 'Duration') -> 'Duration':
+        if isinstance(other, Duration):
+            return Duration(nanoseconds=self.nanoseconds - other.nanoseconds)
+        return NotImplemented
+
+    def __mul__(self, other: int) -> 'Duration':
```
fujitatomoya: should we take floating point here? so that user can do `Duration * 0.5`?

arjo129: I was a bit hesitant about the can of worms that floats tend to open up. So I assume we take a float and then recast it to an int?

fujitatomoya: that is what rclcpp does, https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/duration.cpp#L225-L246

and IMO that is one of the use case. @clalancette @sloretz what do you thiunk?

arjo129: [4df59eb](https://github.com/ros2/rclpy/pull/1387/commits/4df59ebe1f210ed99dc588de4779bc1ef41c06f7)

rclpy/rclpy/duration.py
```diff
@@ -49,6 +49,33 @@ def __str__(self) -> str:
...
+                else:
+                    raise ValueError("Can't multiply duration with nan")
+            return Duration(nanoseconds=int(self.nanoseconds * other))
+        return NotImplemented
+
+    def __floordiv__(self, other: int) -> 'Duration':
+        if isinstance(other, int):
+            return Duration(nanoseconds=self.nanoseconds // other)
+        return NotImplemented
+
```
fujitatomoya: I am not sure if we need to support division, instead user can do `Duration * 0.5` as multiplication? (note that rclcpp does not have this operator.)

rclpy/rclpy/duration.py
clalancette: Yeah, agreed, I think we should make this API feel like the rclcpp one as much as possible.  So I'm for adding floats here.

rclpy/rclpy/duration.py
arjo129: I can open a PR in RCLCPP for this if we need. My take is that if someone were to want to do division by 3 its much better to say `//3` than `*0.333`.

fujitatomoya: i think that is okay to have them, thanks. i am happy to review rclcpp too.

probably we need to add,

- `__truediv__` and roundup the result? to support `Duration / 3`?
- catch `ZeroDivisionError` that is missing here.

clalancette: Division always makes me somewhat uneasy.  One reason for that is https://en.cppreference.com/w/cpp/chrono/duration/operator_arith4, particularly API (6).

At the end of the day, I'm fine with implementing division, as long as it is division by a scalar (not by another Duration).  If we are going to do this, then I agree that we should do it both here in rclpy and rclcpp.

arjo129: > __truediv__ and roundup the result? to support Duration / 3?

 I don't think we should implement `__truediv__`, in fact I think that this is exactly the use case for `__floordiv__` is as it guarantees that the output is an integer and makes the programmer aware that rounding is occurring. 
 
 > catch ZeroDivisionError that is missing here
 
 Im curious what I would do with the exception, do we bubble a different exception up? I dont know if silently failing is a good idea.
 
 >  I'm fine with implementing division, as long as it is division by a scalar (not by another Duration)
 
Also curious about the rationale for this. There is a narrow use case for division by duration that I can think of (in fact something that I'm actively pursuing): If we are doing some form of discretized time based planning then division by duration is kind of nice for us to be able to set a size.

I'm open to removing division as well or separating this out into a seperate PR along with the relevant work in RCLCPP.

clalancette: > Also curious about the rationale for this. There is a narrow use case for division by duration that I can think of (in fact something that I'm actively pursuing): If we are doing some form of discretized time based planning then division by duration is kind of nice for us to be able to set a size.

See API (6) in `std::chono::duration` in https://en.cppreference.com/w/cpp/chrono/duration/operator_arith4 .  It doesn't return the same "type" as e.g. API (5), or API (3) and (4) for multiplication.  That makes me think that there are subtleties there that we need to worry about, and unless we have an immediate use case for it, I'd rather not go down that particular rabbit hole.

arjo129: I'm removing division from here. I'll probably open a seperate PR for it later so that we can move this PR along.

Issue #1098, opened 2023-03-28
Tasks that raise an exception crash the entire executor, even when the exception is caught


haudren-woven:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - Binary
- Version or commit hash:
  - Foxy
- DDS implementation:
  - Cyclone
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

```python
import rclpy


async def throwing_task():
    raise FileNotFoundError()


async def test_1():
    try:
        await throwing_task()
    except FileNotFoundError:
        pass
    return True


async def test_2(executor):
    task = executor.create_task(throwing_task)
    try:
        await task
    except FileNotFoundError:
        print("Error but should keep going")
    return True


rclpy.init()

node = rclpy.create_node("test_wait")

executor = rclpy.get_global_executor()
executor.add_node(node)

task = executor.create_task(test_1)

executor.spin_until_future_complete(task)

assert task.result()

task = executor.create_task(test_2, executor)

executor.spin_until_future_complete(task)

assert task.result()

rclpy.shutdown()
```

#### Expected behavior
The above code should complete and pass since the `FileNotFoundError` is caught in both cases

#### Actual behavior
The executor itself raises an exception on `test_2` which is uncaught and brings down the entire test...

#### Additional information
I was trying to run a number of `async` callbacks in the rclpy executor, and I was raising exceptions in some cases. For some reason, those would bring down the entire node. I tracked it down to these lines:
https://github.com/ros2/rclpy/blob/f6d652fa2d5d3a8164cc69fb543ae0693e958262/rclpy/rclpy/executors.py#L724

I am not sure why the executor is raising the exception of the task :thinking: It might be due to the fact that subscribers, timers etc... are tasks that are never explicitly awaited? I don't have much knowledge about the internals of rclpy but this is very surprising coming from `asyncio`... It is also very surprising to see that the exception is only improperly raised when packaged as an asynchronous task, and not when `await`ed directly.

Thank you!

fujitatomoya:
I think `task = executor.create_task(throwing_task)` is called with another `handler` in the executor, which does not have `try&catch` statement at all.

haudren-woven:
The stacktrace is fairly clear:
```
Traceback (most recent call last):
  File "/tmp/test_rclpy.py", line 40, in <module>
    executor.spin_until_future_complete(task)
[...]

PR #1377, opened 2024-11-08
Check if Task(Future) is canceled.

fujitatomoya:
closes https://github.com/ros2/rclpy/issues/1099

Issue #1099, opened 2023-03-28
Cancelling asynchronous tasks has no effect?


haudren-woven:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - Binary
- Version or commit hash:
  - Foxy
- DDS implementation:
  - Cyclone
- Client library (if applicable):
  - rclpy
  - 
#### Steps to reproduce issue

```python
import rclpy


async def test_cancel():
    future = rclpy.task.Future()
    future.cancel()

    print(future, future.cancelled(), future.done())

    return await future


if __name__ == "__main__":
    rclpy.init()

    node = rclpy.create_node("test")
    node.create_timer(0.1, lambda: None)

    executor = rclpy.get_global_executor()
    executor.add_node(node)
    task = executor.create_task(test_cancel)

    executor.spin_until_future_complete(task)
```

#### Expected behavior
This should terminate immediately, raising a `CancelledError` or equivalent

#### Actual behavior
This runs forever

#### Additional information
This is related to #1098 , as I was experimenting with exceptions and asynchronous tasks. I was wondering what happens when you `await` a task that's cancelled. since in `asyncio` it raises a `CancelledError`. It turns out that in `rclpy` it does call the done callbacks, but the task is not marked as done / doesn't raise.

I'm guessing that we need to adjust this logic:
https://github.com/ros2/rclpy/blob/f6d652fa2d5d3a8164cc69fb543ae0693e958262/rclpy/rclpy/task.py#L56

To be something like:
```python
        while not self._done and not self._cancelled:
            yield
        if self._cancelled:
            raise CancelledError()
        return self.result()
```

But I'd like to hear the maintainers' opinion first!

weber-niklas:
Is there any updates here or plans to adjust the code to the suggested version?

fujitatomoya:
@weber-niklas i think noone is working on this at this moment, besides Foxy is already E.O.L, we probably need to check if the same issue can happen with rolling and available distros. i will try to take a look in this week.

tonygoldman:
It's not fixed in rolling...
In `_wait_for_ready_callbacks` only done tasks are filtered, cancelled tasks remain executed.
https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/executors.py#L612
Also awaiting a cancelled task blocks forever, as the `__await__` method returns only when the task is done.
[...]

PR #1392, opened 2025-01-04
Throw CancelledException into cancelled tasks

nadavelkabets:
PR #1377 addresses issue #1099 by filtering out cancelled tasks. The original attempt encountered issues where cancelled coroutines were not closed properly. This is resolved by throwing CancelledException into the coroutine, similar to the behavior of asyncio.

The solution is not ready yet, as the current implementation raises CancelledException when calling task.result(), which crashes the executor due to the way exceptions in tasks are currently treated: https://github.com/ros2/rclpy/blob/4e8b071127228d5dace5aebf61d02260ecb91253/rclpy/rclpy/executors.py#L853-L855

Potential Solutions
1. Change Executor Behavior to Log Exceptions: Modify the executor to log exceptions instead of crashing, aligning with asyncio behavior. This approach would also resolve issue #1098.
2. Ignore Cancelled Tasks in `task.result()`: Prevent `CancelledException` from being set in `self._exception`. Users would need to call `task.cancelled()` to check cancellation status instead.
3. Executor Ignores `CancelledException`: Allow the exception to be raised but modify the executor to prevent it from crashing.
4. Introduce a Flag (`is_awaited`): The flag is set upon entering `__await__`, causing the executor to ignore or log any exception.

Would love to hear from you :)
@weber-niklas @fujitatomoya  @haudren-woven @sloretz

### DIFF
--- a/rclpy/rclpy/exceptions.py
+++ b/rclpy/rclpy/exceptions.py
@@ -155,3 +155,6 @@ class ROSInterruptException(Exception):
 
     def __init__(self) -> None:
         Exception.__init__(self, 'rclpy.shutdown() has been called')
+
+class CancelledException(BaseException):
+    """The Future or Task was cancelled."""
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -620,8 +620,6 @@ class Executor(ContextManager['Executor']):
                 with self._tasks_lock:
                     # Get rid of any tasks that are done
                     self._tasks = list(filter(lambda t_e_n: not t_e_n[0].done(), self._tasks))
-                    # Get rid of any tasks that are cancelled
-                    self._tasks = list(filter(lambda t_e_n: not t_e_n[0].cancelled(), self._tasks))
 
             # Gather entities that can be waited on
             subscriptions: List[Subscription] = []
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -182,6 +182,7 @@ class Executor(ContextManager['Executor']):
         self._nodes_lock = RLock()
         # Tasks to be executed (oldest first) 3-tuple Task, Entity, Node
         self._tasks: List[Tuple[Task, Optional[WaitableEntityType], Optional[Node]]] = []
+        # Prevent creating tasks while filtering old tasks
         self._tasks_lock = Lock()
         # This is triggered when wait_for_ready_callbacks should rebuild the wait list
         self._guard = GuardCondition(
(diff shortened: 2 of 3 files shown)


weber-niklas:
I am by no means an expert, but the changes made to so far seem very logical to me.

For your potential solutions to the executor problem:
1. if this would still allow the catching of the Exceptions when awaiting or calling future.result() etc., following asyncios lead is probably a good idea. Big plus would be the fix of #1098 which has been bothering me too :D
2. Personally, I do not like this approach as you would need to always save the task separately:
```python
task = executor.create_task(some_coro)
try:
    await task
except SomeException:
    ....
if task.cancelled():
    ...
```
instead of

```python
try:
    await executor.create_task(some_coro)
except CancelledException:
    ...
except SomeOtherException:
    ....
```
As I wrote this I realized you need to do so anyways if you want to cancel the task, so not as much of an issue, but I'd prefer the expanded try except over another condition check
3. Can work, but probably prefer option 1 over this since it kills two birds with one stone
4. Would this work when just working with done callbacks? The user never actively awaits the task

```python
task =  executor.create_task(some_coro)
task.add_done_callback(done_cb)
task.cancel()

def done_cb(task):
    try:
        task.result()
    except Exception:
        ....   
```

Please correct me if I am wrong: As far as I understand all exceptions raised during the execution of the task are already set as the tasks exception:
https://github.com/ros2/rclpy/pull/1392/files#diff-e85e03231947306447b58d0673c912753a94400f92ea660b56a5e4ec56b7f078R268-R277

All we need to do is make sure the executor does not crash when the exceptions are raised in the first place?
If that is the case, logging them seems valid.

nadavelkabets:
> 1. if this would still allow the catching of the Exceptions when awaiting or calling future.result() etc., following asyncios lead is probably a good idea. Big plus would be the fix of [Tasks that raise an exception crash the entire executor, even when the exception is caught #1098](https://github.com/ros2/rclpy/issues/1098) which has been bothering me too :D

Let's say we're going with the asyncio logging approach, we have to decide on the logging behavior.
In asyncio, unretrieved exceptions are logged on garbage collection. This is a big inconvenience since garbage collection in python is inconsistent and is usually done on program shutdown. It would be very annoying if your program would not work as expected, only to find out you've had an exception on shutdown.
Also, regular callable functions are treated differently in asyncio. If I recall correctly an exception in a regular callable crashes the asyncio event loop. Here in ROS tasks the exceptions must be handled the same way for coroutines and callables...

As I see it, the two possible ways to retrieve exceptions from tasks is either awaiting the task or fetching the exception in a done_callback.
I updated my code with the following behaviors:
1. _wait_for_ready_callbacks now iterates directly on self._tasks - since the executor is now handling tasks FIFO, scheduled callbacks will have the opportunity to retrieve the exception before filtering the task from the list
2. When filtering the tasks, check if there is an unretrieved exception. If there is one and the task is not awaited, raise the exception.

What do you think? Should we stick with the asyncio approach? 
Also I need to think how this may effect the multi threaded executor

PR #1396, opened 2025-01-11
Fix redundant checks for a canceled future

nadavelkabets:
There is no need to check if futures are canceled since we now consider canceled futures to be done.

### DIFF
--- a/rclpy/rclpy/task.py
+++ b/rclpy/rclpy/task.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from enum import auto, StrEnum
 import inspect
 import sys
 import threading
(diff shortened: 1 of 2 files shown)


fujitatomoya (commented):
@nadavelkabets 

can you take a look at my comment in https://github.com/ros2/rclpy/pull/1377?

> To keep the consistent behavior including backports for jazzy and humble, probably we want to keep this as following?

i am not convinced to change this behavior... `canceled` and `completed(done)` are different semantics for me.

nadavelkabets:
> can you take a look at my comment in #1377?

What comment are you referring to? I can't find the one you quoted in #1377...


> i am not convinced to change this behavior... `canceled` and `completed(done)` are different semantics for me.

I copied the behavior of asyncio where a future is considered done if it's no longer executing, whether canceled or completed successfully. This was part of my original PR which you already merged. 
I believe we should take the asyncio approach in the long term. 
In the short term, considering a backport is planned, we might want to keep the same behavior. 
I'll make the necessary fixes in this branch.

fujitatomoya:
@nadavelkabets 

> What comment are you referring to? I can't find the one you quoted in https://github.com/ros2/rclpy/pull/1377...

Ah so sorry, comments were not submitted... 😓 apologies...

PR #1397, opened 2025-01-13
Cancelled future is not done

nadavelkabets:
Future.done() should return true only if FutureState is FINISHED for backwards compatibility.

### DIFF
--- a/rclpy/rclpy/task.py
+++ b/rclpy/rclpy/task.py
@@ -67,16 +67,20 @@ class Future(Generic[T]):
 
     def __await__(self) -> Generator[None, None, Optional[T]]:
         # Yield if the task is not finished
-        while not self.done():
+        while self._pending():
             yield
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/task.py
```diff
@@ -262,7 +276,6 @@ def __call__(self) -> None:
                     handler.send(None)
                 except StopIteration as e:
                     # The coroutine finished; store the result
-                    handler.close()
```
fujitatomoya, line 265: when a `StopIteration` exception is raised by a coroutine, it indicates that the coroutine has finished execution and there is no need to manually close the coroutine. The `StopIteration` exception signifies that the coroutine has naturally reached its end, and the resources associated with it should be automatically cleaned up by Python's garbage collector.

fujitatomoya (approved):

PR #1425, opened 2025-03-08
Check parameter callback signature during registration.

### LINKED ISSUE #1423: set_parameters callback returns None
The callback returns None, and I get an error: 

```bash
AttributeError: 'NoneType' object has no attribute 'successful'
```
in [this line 832](https://github.com/ros2/rclpy/blob/humble/rclpy/rclpy/node.py#L832)

fujitatomoya:
closes https://github.com/ros2/rclpy/issues/1423

### DIFF
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -999,6 +999,9 @@ class Node:
         elif self._on_set_parameters_callbacks:
             for callback in self._on_set_parameters_callbacks:
                 result = callback(parameter_list)
+                if result is None:
+                    # This should not happen. But if it does, it should not succeed..
+                    result = SetParametersResult(successful=False, reason='Callback returned None')
                 if not result.successful:
                     return result
         result = SetParametersResult(successful=True)
@@ -1194,6 +1197,10 @@ class Node:
         :param callback: The function that is called whenever parameters are being validated
                          for the node.
         """
+        if not callable(callback):
+            raise TypeError('Callback must be callable, got {}', type(callback))
+        if not isinstance(callback([Parameter('dummy', value=0)]), SetParametersResult):
+            raise TypeError('Callback must return a SetParametersResult, got {}', type(callback))
         self._on_set_parameters_callbacks.insert(0, callback)
 
     def add_post_set_parameters_callback(
@@ -1215,6 +1222,8 @@ class Node:
 
         :param callback: The function that is called after parameters are set for the node.
         """
+        if not callable(callback):
+            raise TypeError('Callback must be callable, got {}', type(callback))
         self._post_set_parameters_callbacks.insert(0, callback)
 
     def remove_pre_set_parameters_callback(
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -1000,7 +1000,7 @@ class Node:
             for callback in self._on_set_parameters_callbacks:
                 result = callback(parameter_list)
                 if result is None:
-                    # This should not happen. But if it does, it should not succeed..
+                    warnings.warn('Callback returned None, it should return SetParameterResult.')
                     result = SetParametersResult(successful=False, reason='Callback returned None')
                 if not result.successful:
                     return result
@@ -1199,8 +1199,6 @@ class Node:
         """
         if not callable(callback):
             raise TypeError('Callback must be callable, got {}', type(callback))
-        if not isinstance(callback([Parameter('dummy', value=0)]), SetParametersResult):
-            raise TypeError('Callback must return a SetParametersResult, got {}', type(callback))
         self._on_set_parameters_callbacks.insert(0, callback)
 
     def add_post_set_parameters_callback(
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/node.py
```diff
@@ -999,6 +999,9 @@ def _set_parameters_atomically_common(
         elif self._on_set_parameters_callbacks:
             for callback in self._on_set_parameters_callbacks:
                 result = callback(parameter_list)
+                if result is None:
+                    # This should not happen. But if it does, it should not succeed..
+                    result = SetParametersResult(successful=False, reason='Callback returned None')
```
fujitatomoya: this is not expected to happen. but just in case, return failure with message that user callback does not return expected result.

rclpy/rclpy/node.py
```diff
@@ -1194,6 +1197,10 @@ def add_on_set_parameters_callback(
         :param callback: The function that is called whenever parameters are being validated
                          for the node.
         """
+        if not callable(callback):
+            raise TypeError('Callback must be callable, got {}', type(callback))
+        if not isinstance(callback([Parameter('dummy', value=0)]), SetParametersResult):
```
fujitatomoya: i think this is not good solution, we need to check the signature without calling the callback. (this bothers the application.)

rclpy/rclpy/node.py
```diff
@@ -1194,6 +1197,8 @@ def add_on_set_parameters_callback(
         :param callback: The function that is called whenever parameters are being validated
                          for the node.
         """
+        if not callable(callback):
+            raise TypeError('Callback must be callable, got {}', type(callback))
```
fujitatomoya, line 1203: we can do type hint checking here something like,

```diff
+        if inspect.signature(callback).return_annotation is not SetParametersResult:
+            warnings.warn('Callback does not have the correct return annotation. '
+                          'Callback must return a SetParametersResult.')
```

but this breaks the downstream code if those do not have annotation for return type. (and also breaks the existed tests in `rclpy`.)

so i would suggest we can check the actual return value and then print the warning for the user application, instead of enforcing the type annotation.

fujitatomoya:
@Barry-Xu-2018 @ahcorde either of you, can you review this?

rclpy/rclpy/node.py
```diff
@@ -999,6 +999,9 @@ def _set_parameters_atomically_common(
         elif self._on_set_parameters_callbacks:
             for callback in self._on_set_parameters_callbacks:
                 result = callback(parameter_list)
+                if result is None:
+                    warnings.warn('Callback returned None, it should return SetParameterResult.')
+                    result = SetParametersResult(successful=False, reason='Callback returned None')
```
Barry-Xu-2018: Apart from the cases where None is returned, should we check if the returned result is not of the SetParametersResult type?  Because if it is not of the SetParametersResult type, the subsequent code `result.successful` will also cause an error.

fujitatomoya: that type is `Any`. or are you suggesting we should have `SetParameterResult` as python class object?

fujitatomoya: Ah sorry, i misunderstood. just checking if `SetParametersResult` instance?

fujitatomoya: fixed in https://github.com/ros2/rclpy/pull/1425/commits/ed4aa6df8fc5ce185bee486d8c7868653e58f6e0

fujitatomoya:
Pulls: ros2/rclpy#1425
Gist: https://gist.githubusercontent.com/fujitatomoya/eaabdb404b7b9652d5df9168802ca109/raw/b39d50ff6bfa8968d3797083f553da8ff69e1814/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15316
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22670/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16910/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2184/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23496/)

Barry-Xu-2018 (approved):

rclpy/rclpy/node.py
```diff
@@ -999,6 +999,11 @@ def _set_parameters_atomically_common(
         elif self._on_set_parameters_callbacks:
             for callback in self._on_set_parameters_callbacks:
                 result = callback(parameter_list)
+                if not isinstance(result, SetParametersResult):
```
Timple, line 1002: _Random bystander here_
Python is not a strong typed language, so a callback is free to return anything.

So in the concept of duck typing, why not only check what you really need?

`if not hassatr(result, successful):`

Not feeling strong about it though, so feel free to ignore this :)

fujitatomoya: i think right thing to do here is to check the expected object is returned by callback, instead of checking the attribution.

ahcorde (approved):

fujitatomoya:
@Mergifyio backport jazzy humble

PR #1428, opened 2025-03-11
Fix Duration, Clock, and QoS Docs

### LINKED ISSUE #1419: qos docs not being generated in Sphinx
I was looking at QoS settings and recommended defaults and found this MR where they had made sure the documentation would work for this case: 
* https://github.com/ros2/rclpy/pull/653

But looking at the live docs, the QoS documentation is blank for some reason:
* https://docs.ros.org/en/rolling/p/rclpy/rclpy.qos.html
* https://docs.ros.org/en/rolling/p/rclpy/rclpy.qos_event.html

InvincibleRMC:
Closes #1419.

Adding the following to autodock_mock_imports causes it to be blank on the documentation.

I can build it locally with sphinx and rosdoc2. I believe the addition of `_rclpy_pybind11.pyi` is what resolved previous errors with underlying c objects.

Note: `rclpy.qos_event` is supposed to be empty since it is deprecated and `rclpy.event_handlers` should be used instead.

### DIFF
--- a/rclpy/package.xml
+++ b/rclpy/package.xml
@@ -55,13 +55,6 @@
 
   <doc_depend>python3-sphinx</doc_depend>
   <doc_depend>python3-sphinx-rtd-theme</doc_depend>
-  <!-- We add these modules as doc_depends in order to include them into the
-  autodoc_mock_imports list. This prevents TypeErrors from being thrown
-  when these modules import constants from rclpy C modules.
-  TODO: Remove these depends once a better solution is found.-->
-  <doc_depend>rclpy.duration</doc_depend>
-  <doc_depend>rclpy.qos</doc_depend>
-  <doc_depend>rclpy.clock</doc_depend>
 
   <export>
     <build_type>ament_cmake</build_type>


fujitatomoya (commented):
`_rclpy_pybind11.pyi` might indeed resolve the errors related to underlying C objects, making those specific `doc_depend` entries redundant. So it could be a suitable replacement for the previous workaround. As long as we can verify this change by building and checking the documentation to ensure no other issues arise, i am good to go with this.

kscottz (approved):
I didn't test this but after a cursory look I  think this is probably the correct solution.

christophebedard:
Just to be safe, could you rebase on `rolling`? Then I can run CI

christophebedard:
Pulls: ros2/rclpy#1428
Gist: https://gist.githubusercontent.com/christophebedard/9d9426b03121db2d8284a82a57d2a1b9/raw/2e09b22c42fc7547a93a66ee636ef1ccb4bb719f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15431
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22800/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17030/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2317/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23680/)

christophebedard:
The failures on Windows seem unrelated.

Thanks for the fix and for validating!

Looks like Jazzy uses the same workaround: https://github.com/ros2/rclpy/blob/dc598aba6ffd77e0990940555e98a21901fd948a/rclpy/package.xml#L55-L61. @InvincibleRMC can we make the same change on Jazzy now?

InvincibleRMC:
You could give it a try but, I think the fix was  from the inclusion `_rclpy_pybind11.pyi` and that file does not exist in jazzy. However it could be something else entirely and it might work.

InvincibleRMC:
This should be reverted when I was checking out the `rclpy` docs and I noticed the documentation is gone again. When I was checking out `sphinx` I didn't notice that the ability to use type stubs for generated docs is a [newer](https://github.com/sphinx-doc/sphinx/issues/7630) feature not available in the current version of sphinx on Ubuntu. Since I sourced my installed before building I guess that added the missing documentation but, from what I understand on the build farm no building is done in the documentation generation step.

fujitatomoya:
@InvincibleRMC i created revert PR https://github.com/ros2/rclpy/pull/1447, can you review and add your comment there?

PR #1420, opened 2025-03-01
Add remaining type hints and `ament_mypy`

InvincibleRMC:
With this pull request `rclpy` can finally be type checked with `ament_mypy`.  Please let me know if there are any questions/suggestions about the changes. Close #1257.

PR #1370, opened 2024-10-04
Executors types

InvincibleRMC:
A redo of #1345 . Assuming this doesn't trigger the mysterious ros2cli flake8 seg fault it should be fine to merge.

PR #1431, opened 2025-03-12
Fix Task constructor type bug

InvincibleRMC:
Uses overload of constructor instead of Union since the narrower Coroutine type was getting hidden inside the TypeVar T.

Split up from #1420

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -24,13 +24,13 @@ import time
 from types import TracebackType
 from typing import Any
 from typing import Callable
-from typing import cast
 from typing import ContextManager
 from typing import Coroutine
 from typing import Dict
 from typing import Generator
 from typing import List
 from typing import Optional
+from typing import overload
 from typing import Set
 from typing import Tuple
 from typing import Type
@@ -51,7 +51,6 @@ from rclpy.service import Service
 from rclpy.signals import SignalHandlerGuardCondition
 from rclpy.subscription import MessageInfo
 from rclpy.subscription import Subscription
-from rclpy.task import FunctionOrCoroutineFunction
 from rclpy.task import Future
 from rclpy.task import Task
 from rclpy.timer import Timer, TimerInfo
@@ -118,14 +117,20 @@ class _WorkTracker:
         return True
 
 
-async def await_or_execute(callback: FunctionOrCoroutineFunction[T], *args: Any) -> T:
+@overload
+async def await_or_execute(callback: Callable[..., Coroutine[Any, Any, T]], *args: Any) -> T: ...
+
+
+@overload
+async def await_or_execute(callback: Callable[..., T], *args: Any) -> T: ...
+
+
+async def await_or_execute(callback: Callable[..., Any], *args: Any) -> Any:
     """Await a callback if it is a coroutine, else execute it."""
     if inspect.iscoroutinefunction(callback):
         # Await a coroutine
-        callback = cast(Callable[..., Coroutine[None, None, T]], callback)
         return await callback(*args)
     else:
-        callback = cast(Callable[..., T], callback)
         # Call a normal function
         return callback(*args)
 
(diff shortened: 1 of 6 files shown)


rclpy/rclpy/executors.py
```diff
@@ -118,14 +117,20 @@ def wait(self, timeout_sec: Optional[float] = None) -> bool:
...
-async def await_or_execute(callback: FunctionOrCoroutineFunction[T], *args: Any) -> T:
+@overload
+async def await_or_execute(callback: Callable[..., Coroutine[Any, Any, T]], *args: Any) -> T: ...
+
+
+@overload
+async def await_or_execute(callback: Callable[..., T], *args: Any) -> T: ...
+
+
+async def await_or_execute(callback: Callable[..., Any], *args: Any) -> Any:
```
christophebedard, line 128: Today I learned! https://docs.python.org/3/library/typing.html#typing.overload

I do have a question: the `@overload`s above means that this does not actually take in a `Callable` that returns `Any`, right? They _must_ return either a `Coroutine[Any, Any, T]` or `T`, right?

rclpy/rclpy/task.py
```diff
@@ -16,22 +16,17 @@
...
 
 if TYPE_CHECKING:
-    from typing import TypeAlias
 
     from rclpy.executors import Executor
 
 T = TypeVar('T')
 
-FunctionOrCoroutineFunction: 'TypeAlias' = Union[Callable[..., T],
-                                                 Callable[..., Coroutine[None, None, T]]]
```
christophebedard, line 33: Just a note: I would generally be cautious about simply removing something, but it was added 2 months ago in #1370, so it's fine

rclpy/test/test_task.py
```diff
@@ -203,32 +209,32 @@ def test_executing(self) -> None:
...
+        f: Future[str] = Future()
         f.set_result('Sentinel Result')
         self.assertEqual('Sentinel Result', f.result())
         self.assertTrue(f.done())
 
     def test_set_exception(self) -> None:
-        f = Future()
-        f.set_exception('Sentinel Exception')
+        f: Future[Any] = Future()
+        f.set_exception('Sentinel Exception')   # type: ignore[arg-type]
```
christophebedard, line 230: This is kind of abusing the method/type (using a `str` instead of an `Exception` like some of the other tests do), but for the purposes of this test it's fine.

christophebedard (approved):
Just a question and some minor comments/notes, otherwise looks good to me!

christophebedard:
Pulls: ros2/rclpy#1431
Gist: https://gist.githubusercontent.com/christophebedard/bd6bdbbd2a53c65ca8adf1f4394b6e62/raw/a0c9dde4a9bdd40fca41e38f9e902e05a3d1ba9a/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15361
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22720/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16957/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2237/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23584/)

rclpy/rclpy/executors.py
InvincibleRMC: I might be misunderstanding but, someone could pass in a function which returns Any which would update the return type of await_or_execute to return Any. To stop that the user would have to run a type checker with --disallow-any on their code.

christophebedard: Yeah, sorry, I meant: This works fine at runtime, but mypy would complain about a call to `await_or_execute()` that does not use a `Callable` of return type `Coroutine[Any, Any, T]` or `T`, correct?

InvincibleRMC: Yeah.

christophebedard:
There are test failures on Windows. While I can't find any similar test failures in other recent jobs, this PR doesn't affect anything at runtime, so I think this is good.

Thanks for the PR!

PR #1436, opened 2025-03-25
Update `_rclpy_pybind11.pyi` order and add `EventsExecutor`

InvincibleRMC:
This may look scary but this was mostly just reordering the classes and functions to match the order in `_rclpy_pybind11.cpp`. I did this since it was getting difficult to find things intuitively. The main new changes was the addition of `EventsExecutor` and adding the `configure_introspection` functions to `ActionClient` and `ActionServer`. I also updated some type helper objects to be private (by adding an `_` infront) to prevent a user from trying to instantiate them since they are not exposed in the underlying C++.


christophebedard:
Should we automatically generate the `.pyi` file using mypy's stubgen? https://mypy.readthedocs.io/en/stable/stubgen.html

This is how it's done for `rosbag2_py`: https://github.com/ros2/rosbag2/tree/rolling/rosbag2_py

InvincibleRMC:
I think one day it might be nice to, however I don't think so yet. While pybind11 static type hinting has come a long way especially in the upcoming `3.0.0` release it is still missing support for python generics and typed dictionaries which are used quite extensively in `_rclpy_pybind11.pyi`. The current vendored version of pybind11 is `2.11.2` which is missing the `typing.h` header which really makes using tools like mypy's stubgen and pybind11-stubgen a lot more practical. Orginal when I started writing the types in `_rclpy_pybind11.pyi` I tried the rosbag_py approach but the types were typically too broad to be useful as well it really struggle with imports from `rmw.h`

christophebedard:
Thanks for the information! That makes sense.

christophebedard:
Pulls: ros2/rclpy#1436
Gist: https://gist.githubusercontent.com/christophebedard/2d97dbdb1d86575e6ca23323ff9f2f35/raw/934538cb4c1c75f742617644cc6a7d232ff6689c/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15536
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22962/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17183/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2449/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23837/)

fujitatomoya (approved):
lgtm with green CI. @InvincibleRMC thanks!

christophebedard:
There are weird test failures in the ci_linux and ci_linux-aarch64 jobs. I can't easily tell what's going on.

christophebedard:
@InvincibleRMC would you mind taking a look at the test failures to see if it might be related to these changes?

InvincibleRMC:
The ci_linux failure seems to be happening on [nightly](https://ci.ros2.org/job/nightly_linux-aarch64_release/3056/testReport/)

The ci_linux-aarch64 failure seems to be happening on [nightly repeated](https://ci.ros2.org/job/nightly_linux-aarch64_repeated/3086/testReport/junit/sros2.test.sros2.commands.security.verbs/test_generate_policy/test_generate_policy/). It doesn't seem to happen on regular nightly which to me seems very strange. I'm not sure if there is some subtle configuration differences.

Since both tests are failing with `rmw_zenoh_cpp` middleware, I imagine it something specific to that implementation.  Since all of my changes should be middleware agnostic.

christophebedard:
Alright, thanks for checking.

These CLI tests are known to be kind of flaky, maybe even more so with a newer `rmw` implementation.

PR #1469, opened 2025-06-07
Fix issues with resuming async tasks awaiting a future

### LINKED ISSUE #1405: Using rclpy.spin_once and await in a coroutine causes busy waiting
I noticed that when an executor is executing a coroutine which called await, all calls to `rclpy.spin_once` end immediately, causing busy waiting which results in 100% CPU consumption. Here's a script that demonstrates this issue:
```Python
import rclpy
import rclpy.logging
from rclpy.task import Future
from rclpy.executors import SingleThreadedExecutor
from rclpy.node import Node

async def coroutine():
    print("Coroutine")
    future = Future()
    await future

rclpy.init()
rclpy.logging.set_logger_level("rcl", rclpy.logging.LoggingSeverity.DEBUG)
node = Node("test_node")
node.get_logger().set_level(rclpy.logging.LoggingSeverity.DEBUG)
node.create_timer(1.0, coroutine)

# This causes busy waiting
while rclpy.ok():
    rclpy.spin_once(node)

# # This does not
# rclpy.spin(node)

# # This does not either
# executor = SingleThreadedExecutor()
# executor.add_node(node)
# while rclpy.ok():
#     executor.spin_once()

node.destroy_node()
rclpy.shutdown()
```

Tried it in Humble, Iron, Jazzy and Rolling, getting the same result every time.

I suspect the issue might be with the way the `__async__` method in Future objects is implemented:
https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/task.py#L73-L77

### LINKED ISSUE #1462: `async` service server callback gets stuck on wait set
Hello,

I have been debugging an issue which I encountered when trying to add a `async` service callback. What I want to achieve is a service callback which doesn't block the spinning (I have simplified the problem in the code below). Long story short: in the `async` service callback, I want to `await` an event.

I got it tracked down to that the executor gets stuck on `wait_set.wait(..)` in `executors.py` (line 685 on jazzy release):

```python
                # Wait for something to become ready
                wait_set.wait(timeout_nsec)
```

I'm still lacking more knowledge of the executor's details to pinpoint where to keep digging. Before I dig further,  I thought it might be worth asking if someone has a suggestion what the issue could be. It could make the process lots more time efficient - would greatly appreciate any input!

What's really curious (at least to my ignorant self) is the following: it works fine if I have a timed callback set up. Each time the  timed callback gets called, the wait set gets "unstuck" for exactly two iterations, and my `__await__` method gets called. This unblocks the process and the service call completes successfully. If you comment out the timed callback (`self.create_timer(1, self.timer_callback)`), the service call gets stuck. 

This could be related to #1405 .

I tested this on jazzy (Ubuntu 24.04) and humble (Ubuntu 22.04).

Below is my test code, run with

```
python3 simple_coroutine_test.py
```
and the service call
```
ros2 service call /simple_trigger std_srvs/srv/Trigger "{}"
```

```python
import rclpy
import time
from rclpy.node import Node
from std_srvs.srv import Trigger
from rclpy.callback_groups import ReentrantCallbackGroup, MutuallyExclusiveCallbackGroup


class AwaitCount:
    """Implements the await method which finishes once the method has been called a number of times."""
    def __init__(self, logger):
        self._logger = logger
        self._cnt = 0

    def __await__(self):
        while self._cnt < 10:
            # Yield control so the event loop can do other work (ROS spinning work)
            self._logger.info("Awaiting...")
            self._cnt += 1
            yield
        return  # Done waiting


class SimpleServiceNode(Node):
    """Simple test node for testing async service callback."""

    def __init__(self):
        super().__init__('simple_service_node')
        callback_group = MutuallyExclusiveCallbackGroup() # ReentrantCallbackGroup()

        # Create the service for the trigger request.
        self.create_service(Trigger, 'simple_trigger', self.trigger_callback, callback_group = callback_group)

        # Create a timer that logs a "heartbeat" - shows the ROS loop is still running and somehow
        # unblocks the service callback.
        self.create_timer(1, self.timer_callback)

    async def trigger_callback(self, request, response):
        """Callback for trigger service."""
        self.get_logger().info('Trigger called.')

        # Wait asynchronously (while allowing ROS to spin).
        await AwaitCount(self.get_logger())

        self.get_logger().info("Received result!")

        response.success = True
        response.message = "Finished test successfully."
        return response

    async def timer_callback(self):
        self.get_logger().info('Heartbeat: timer is alive.')

def main():
    rclpy.init()
    node = SimpleServiceNode()

    use_single = True  # set to True to use SingleThreadedExecutor, False for MultiThreadeExecutor
    try:
        if use_single:
            rclpy.spin(node)
        else:
            executor = rclpy.executors.MultiThreadedExecutor(num_threads=2)
            executor.add_node(node)
            executor.spin()
    except KeyboardInterrupt:
        pass

if __name__ == '__main__':
    main()
```

### LINKED ISSUE #1467: Calling future.set_result from another thread does not wake up the executor
### Generated by Generative AI

_No response_

### Operating System:

```Linux blazej-legion 6.12.30 #1-NixOS SMP PREEMPT_DYNAMIC Thu May 22 12:29:54 UTC 2025 x86_64 GNU/Linux```

### ROS version or commit hash:

humble, jazzy (testing), kilted (testing), Rolling (testing)

### RMW implementation (if applicable):

rmw_fastrtps_cpp

### RMW Configuration (if applicable):

_No response_

### Client library (if applicable):

rclpy

### 'ros2 doctor --report' output

<details><summary>ros2 doctor --report</summary>

```console
   NETWORK CONFIGURATION
inet         : 127.0.0.1
inet4        : ['127.0.0.1']
inet6        : ['::1']
netmask      : 255.0.0.0
device       : lo
flags        : UP,LOOPBACK,RUNNING
mtu          : 65536
inet         : 192.168.1.105
inet4        : ['192.168.1.105']
ether        : 98:fc:84:e0:2e:9e
inet6        : ['fe80::e7d3:eef1:498d:f03e%enp8s0f3u1u1']
netmask      : 255.255.255.0
device       : enp8s0f3u1u1
flags        : UP,BROADCAST,RUNNING,MULTICAST
mtu          : 1500
broadcast    : 192.168.1.255
inet         : 192.168.1.138
inet4        : ['192.168.1.138']
ether        : 00:41:0e:7a:71:85
inet6        : ['fe80::a705:5297:dd0e:d28%wlo1']
netmask      : 255.255.255.0
device       : wlo1
flags        : UP,BROADCAST,RUNNING,MULTICAST
mtu          : 1500
broadcast    : 192.168.1.255
ether        : 08:8f:c3:fd:5f:71
device       : enp2s0
flags        : UP,BROADCAST,MULTICAST
mtu          : 1500

   PACKAGE VERSIONS
image_transport_plugins                   : latest=5.0.3, local=5.0.3
zstd_image_transport                      : latest=5.0.3, local=5.0.3
webrtc_image_streamer                     : latest=N/A, local=0.0.0
leo_robot                                 : latest=2.2.0, local=2.3.0
leo_bringup                               : latest=2.2.0, local=2.3.0
web_video_server                          : latest=2.1.0, local=2.1.0
theora_image_transport                    : latest=5.0.3, local=5.0.3
rosbridge_test_msgs                       : latest=2.3.0, local=2.3.0
rosbridge_suite                           : latest=2.3.0, local=2.3.0
rosbridge_server                          : latest=2.3.0, local=2.3.0
rosbridge_msgs                            : latest=2.3.0, local=2.3.0
rosapi                                    : latest=2.3.0, local=2.3.0
rosbridge_library                         : latest=2.3.0, local=2.3.0
rosbag2_to_video                          : latest=1.0.1, local=1.0.1
rosapi_msgs                               : latest=2.3.0, local=2.3.0
libdatachannel_vendor                     : latest=N/A, local=0.0.0
leo_desktop                               : latest=3.0.0, local=3.0.0
leo_viz                                   : latest=3.0.0, local=3.0.0
leo                                       : latest=3.1.0, local=3.1.0
leo_teleop                                : latest=3.1.0, local=3.1.0
leo_simulator                             : latest=2.0.2, local=2.0.2
leo_fw                                    : latest=2.2.0, local=2.3.0
leo_msgs                                  : latest=3.1.0, local=3.1.0
leo_gz_bringup                            : latest=2.0.2, local=2.0.2
leo_gz_worlds                             : latest=2.0.2, local=2.0.2
leo_gz_plugins                            : latest=2.0.2, local=2.0.2
leo_filters                               : latest=2.2.0, local=2.3.0
leo_description                           : latest=3.1.0, local=3.1.0
compressed_image_transport                : latest=5.0.3, local=5.0.3
compressed_depth_image_transport          : latest=5.0.3, local=5.0.3
aruco_opencv                              : latest=6.0.2, local=6.0.2
aruco_opencv_msgs                         : latest=6.0.2, local=6.0.2
ament_package                             : latest=0.18.0, local=0.18.0
canopen_core                              : latest=0.3.0, local=0.3.0
canopen_master_driver                     : latest=0.3.0, local=0.3.0
canopen_base_driver                       : latest=0.3.0, local=0.3.0
canopen_proxy_driver                      : latest=0.3.0, local=0.3.0
v4l2_camera                               : latest=0.7.1, local=0.7.1
ros_workspace                             : latest=1.0.3, local=1.0.3
ament_cmake_core                          : latest=2.8.2, local=2.8.1
ament_cmake_export_definitions            : latest=2.8.2, local=2.8.1
ament_cmake_libraries                     : latest=2.8.2, local=2.8.1
ament_cmake_export_dependencies           : latest=2.8.2, local=2.8.1
ament_cmake_export_include_directories    : latest=2.8.2, local=2.8.1
ament_cmake_export_libraries              : latest=2.8.2, local=2.8.1
ament_cmake_export_link_flags             : latest=2.8.2, local=2.8.1
ament_cmake_export_targets                : latest=2.8.2, local=2.8.1
ament_cmake_gen_version_h                 : latest=2.8.2, local=2.8.1
ament_cmake_python                        : latest=2.8.2, local=2.8.1
ament_cmake_include_directories           : latest=2.8.2, local=2.8.1
ament_cmake_target_dependencies           : latest=2.8.2, local=2.8.1
ament_cmake_test                          : latest=2.8.2, local=2.8.1
ament_cmake_version                       : latest=2.8.2, local=2.8.1
ament_cmake                               : latest=2.8.2, local=2.8.1
ament_cppcheck                            : latest=0.20.0, local=0.20.0
ament_cmake_cppcheck                      : latest=0.20.0, local=0.20.0
ament_cpplint                             : latest=0.20.0, local=0.20.0
ament_cmake_cpplint                       : latest=0.20.0, local=0.20.0
ament_lint                                : latest=0.20.0, local=0.20.0
ament_flake8                              : latest=0.20.0, local=0.20.0
ament_cmake_flake8                        : latest=0.20.0, local=0.20.0
ament_pep257                              : latest=0.20.0, local=0.20.0
ament_cmake_pep257                        : latest=0.20.0, local=0.20.0
uncrustify_vendor                         : latest=3.2.0, local=3.2.0
ament_uncrustify                          : latest=0.20.0, local=0.20.0
ament_cmake_uncrustify                    : latest=0.20.0, local=0.20.0
ament_index_python                        : latest=1.12.0, local=1.12.0
rosidl_typesupport_interface              : latest=4.10.0, local=4.10.0
rosidl_cli                                : latest=4.10.0, local=4.10.0
rosidl_adapter                            : latest=4.10.0, local=4.10.0
rosidl_parser                             : latest=4.10.0, local=4.10.0
rosidl_pycommon                           : latest=4.10.0, local=4.10.0
rosidl_cmake                              : latest=4.10.0, local=4.10.0
rosidl_generator_type_description         : latest=4.10.0, local=4.10.0
rpyutils                                  : latest=0.7.0, local=0.7.0
ament_cmake_ros_core                      : latest=0.15.0, local=0.15.0
ament_index_cpp                           : latest=1.12.0, local=1.12.0
libyaml_vendor                            : latest=1.8.0, local=1.8.0
spdlog_vendor                             : latest=1.8.0, local=1.8.0
rmw_implementation_cmake                  : latest=7.9.0, local=7.9.0
tracetools                                : latest=8.7.0, local=8.7.0
console_bridge_vendor                     : latest=1.9.0, local=1.9.0
ament_black                               : latest=0.2.6, local=0.2.6
ament_clang_format                        : latest=0.20.0, local=0.20.0
ament_cmake_gtest                         : latest=2.8.2, local=2.8.1
ament_cmake_gmock                         : latest=2.8.2, local=2.8.1
ament_cmake_auto                          : latest=2.8.2, local=2.8.1
ament_cmake_black                         : latest=0.2.6, local=0.2.6
ament_cmake_clang_format                  : latest=0.20.0, local=0.20.0
ament_copyright                           : latest=0.20.0, local=0.20.0
ament_cmake_copyright                     : latest=0.20.0, local=0.20.0
ament_lint_cmake                          : latest=0.20.0, local=0.20.0
ament_cmake_lint_cmake                    : latest=0.20.0, local=0.20.0
ament_mypy                                : latest=0.20.0, local=0.20.0
ament_cmake_mypy                          : latest=0.20.0, local=0.20.0
ament_cmake_pytest                        : latest=2.8.2, local=2.8.1
zenoh_cpp_vendor                          : latest=0.7.1, local=0.7.1
ament_cmake_vendor_package                : latest=2.8.2, local=2.8.1
ament_xmllint                             : latest=0.20.0, local=0.20.0
ament_cmake_xmllint                       : latest=0.20.0, local=0.20.0
ament_lint_auto                           : latest=0.20.0, local=0.20.0
ament_lint_common                         : latest=0.20.0, local=0.20.0
angles                                    : latest=1.16.0, local=1.16.0
yaml_cpp_vendor                           : latest=9.2.0, local=9.2.0
osrf_pycommon                             : latest=2.1.4, local=2.1.4
launch                                    : latest=3.9.0, local=3.9.0
tinyxml2_vendor                           : latest=0.11.0, local=0.11.0
launch_xml                                : latest=3.9.0, local=3.9.0
eigen3_cmake_module                       : latest=0.5.0, local=0.5.0
orocos_kdl_vendor                         : latest=0.8.0, local=0.8.0
sdl2_vendor                               : latest=3.3.0, local=3.3.0
rttest                                    : latest=0.19.0, local=0.19.0
tlsf                                      : latest=0.11.0, local=0.11.0
pybind11_vendor                           : latest=3.3.0, local=3.3.0
python_orocos_kdl_vendor                  : latest=0.8.0, local=0.8.0
launch_yaml                               : latest=3.9.0, local=3.9.0
launch_testing                            : latest=3.9.0, local=3.9.0
launch_testing_ament_cmake                : latest=3.9.0, local=3.9.0
ros_environment                           : latest=4.4.0, local=4.4.0
rosidl_runtime_py                         : latest=0.15.0, local=0.15.0
keyboard_handler                          : latest=0.5.0, local=0.5.0
zstd_vendor                               : latest=0.33.0, local=0.33.0
liblz4_vendor                             : latest=0.33.0, local=0.33.0
mcap_vendor                               : latest=0.33.0, local=0.33.0
sqlite3_vendor                            : latest=0.33.0, local=0.33.0
python_qt_binding                         : latest=2.4.0, local=2.4.0
tango_icons_vendor                        : latest=0.5.0, local=0.5.0
qt_gui                                    : latest=2.10.1, local=2.10.1
qt_gui_py_common                          : latest=2.10.1, local=2.10.1
qt_dotgraph                               : latest=2.10.1, local=2.10.1
gz_cmake_vendor                           : latest=0.3.1, local=0.3.1
gz_utils_vendor                           : latest=0.3.0, local=0.3.0
rviz_ogre_vendor                          : latest=15.1.3, local=15.1.3
libcurl_vendor                            : latest=3.8.1, local=3.8.1
rviz_assimp_vendor                        : latest=15.1.3, local=15.1.3
domain_coordinator                        : latest=0.15.0, local=0.15.0
generate_parameter_library_py             : latest=0.4.0, local=0.4.0
tcb_span                                  : latest=1.0.2, local=1.0.2
tl_expected                               : latest=1.0.2, local=1.0.2
gz_dartsim_vendor                         : latest=0.1.2, local=0.1.2
gz_tools_vendor                           : latest=0.1.2, local=0.1.2
gz_ogre_next_vendor                       : latest=0.1.0, local=0.1.0
lely_core_libraries                       : latest=0.3.0, local=0.3.0
xacro                                     : latest=2.0.13, local=2.0.13
gz_math_vendor                            : latest=0.3.1, local=0.3.1
sdformat_vendor                           : latest=0.2.5, local=0.2.5
ros2_control_test_assets                  : latest=5.1.0, local=5.1.0
gz_common_vendor                          : latest=0.2.3, local=0.2.3
gz_msgs_vendor                            : latest=0.2.3, local=0.2.3
gz_fuel_tools_vendor                      : latest=0.2.1, local=0.2.1
gz_plugin_vendor                          : latest=0.2.2, local=0.2.2
gz_rendering_vendor                       : latest=0.2.1, local=0.2.1
gz_transport_vendor                       : latest=0.2.2, local=0.2.2
gz_gui_vendor                             : latest=0.2.1, local=0.2.1
gz_physics_vendor                         : latest=0.2.1, local=0.2.1
gz_sensors_vendor                         : latest=0.2.1, local=0.2.1
gz_sim_vendor                             : latest=0.2.1, local=0.2.1
rcutils                                   : latest=6.10.1, local=6.10.1
rosidl_runtime_c                          : latest=4.10.0, local=4.10.0
rosidl_dynamic_typesupport                : latest=0.4.0, local=0.4.0
rmw                                       : latest=7.9.0, local=7.9.0
rosidl_generator_c                        : latest=4.10.0, local=4.10.0
rcpputils                                 : latest=2.14.0, local=2.14.0
rosidl_typesupport_c                      : latest=3.4.0, local=3.4.0
rosidl_generator_py                       : latest=0.25.0, local=0.25.0
rosidl_runtime_cpp                        : latest=4.10.0, local=4.10.0
rosidl_typesupport_cpp                    : latest=3.4.0, local=3.4.0
rosidl_generator_cpp                      : latest=4.10.0, local=4.10.0
rosidl_typesupport_fastrtps_cpp           : latest=3.9.0, local=3.9.0
rosidl_typesupport_fastrtps_c             : latest=3.9.0, local=3.9.0
rosidl_typesupport_introspection_c        : latest=4.10.0, local=4.10.0
rosidl_typesupport_introspection_cpp      : latest=4.10.0, local=4.10.0
rosidl_core_runtime                       : latest=0.4.0, local=0.4.0
builtin_interfaces                        : latest=2.4.0, local=2.4.0
service_msgs                              : latest=2.4.0, local=2.4.0
unique_identifier_msgs                    : latest=2.8.0, local=2.8.0
action_msgs                               : latest=2.4.0, local=2.4.0
rosidl_default_runtime                    : latest=1.8.0, local=1.8.0
example_interfaces                        : latest=0.14.0, local=0.14.0
rcl_interfaces                            : latest=2.4.0, local=2.4.0
rcl_logging_interface                     : latest=3.3.0, local=3.3.0
rcl_logging_spdlog                        : latest=3.3.0, local=3.3.0
rcl_yaml_param_parser                     : latest=10.2.1, local=10.2.1
rmw_security_common                       : latest=7.9.0, local=7.9.0
rmw_dds_common                            : latest=3.3.0, local=3.3.0
rmw_fastrtps_shared_cpp                   : latest=9.4.0, local=9.4.0
rosidl_dynamic_typesupport_fastrtps       : latest=0.5.0, local=0.5.0
rmw_fastrtps_cpp                          : latest=9.4.0, local=9.4.0
rmw_cyclonedds_cpp                        : latest=4.1.0, local=4.1.0
rmw_implementation                        : latest=3.1.0, local=3.1.0
type_description_interfaces               : latest=2.4.0, local=2.4.0
rcl                                       : latest=10.2.1, local=10.2.1
statistics_msgs                           : latest=2.4.0, local=2.4.0
libstatistics_collector                   : latest=2.1.0, local=2.1.0
rosgraph_msgs                             : latest=2.4.0, local=2.4.0
rclcpp                                    : latest=29.6.0, local=29.6.0
rcl_action                                : latest=10.2.1, local=10.2.1
rclcpp_action                             : latest=29.6.0, local=29.6.0
class_loader                              : latest=2.9.0, local=2.9.0
composition_interfaces                    : latest=2.4.0, local=2.4.0
rclcpp_components                         : latest=29.6.0, local=29.6.0
action_tutorials_cpp                      : latest=0.37.0, local=0.37.0
lifecycle_msgs                            : latest=2.4.0, local=2.4.0
rcl_lifecycle                             : latest=10.2.1, local=10.2.1
rclpy                                     : latest=9.2.0, local=9.2.0
action_tutorials_py                       : latest=0.37.0, local=0.37.0
std_msgs                                  : latest=5.7.0, local=5.7.0
actuator_msgs                             : latest=0.0.1, local=0.0.1
rmw_test_fixture                          : latest=0.15.0, local=0.15.0
rmw_zenoh_cpp                             : latest=0.7.1, local=0.7.1
rmw_test_fixture_implementation           : latest=0.15.0, local=0.15.0
ament_cmake_ros                           : latest=0.15.0, local=0.15.0
async_web_server_cpp                      : latest=2.0.0, local=2.0.0
geometry_msgs                             : latest=5.7.0, local=5.7.0
sensor_msgs                               : latest=5.7.0, local=5.7.0
camera_calibration_parsers                : latest=6.2.0, local=6.2.0
rclcpp_lifecycle                          : latest=29.6.0, local=29.6.0
camera_info_manager                       : latest=6.2.0, local=6.2.0
canopen_interfaces                        : latest=0.3.0, local=0.3.0
diagnostic_msgs                           : latest=5.7.0, local=5.7.0
nav_msgs                                  : latest=5.7.0, local=5.7.0
shape_msgs                                : latest=5.7.0, local=5.7.0
std_srvs                                  : latest=5.7.0, local=5.7.0
stereo_msgs                               : latest=5.7.0, local=5.7.0
trajectory_msgs                           : latest=5.7.0, local=5.7.0
visualization_msgs                        : latest=5.7.0, local=5.7.0
common_interfaces                         : latest=5.7.0, local=5.7.0
launch_ros                                : latest=0.29.0, local=0.29.0
composition                               : latest=0.37.0, local=0.37.0
cv_bridge                                 : latest=4.1.0, local=4.1.0
message_filters                           : latest=7.2.1, local=7.2.1
pluginlib                                 : latest=5.7.0, local=5.7.0
image_transport                           : latest=6.2.0, local=6.2.0
control_msgs                              : latest=6.1.0, local=6.1.0
filters                                   : latest=2.2.1, local=2.2.1
parameter_traits                          : latest=0.4.0, local=0.4.0
generate_parameter_library                : latest=0.4.0, local=0.4.0
realtime_tools                            : latest=4.3.0, local=4.3.0
tf2                                       : latest=0.42.1, local=0.42.1
tf2_msgs                                  : latest=0.42.1, local=0.42.1
tf2_ros                                   : latest=0.42.1, local=0.42.1
tf2_py                                    : latest=0.42.1, local=0.42.1
tf2_ros_py                                : latest=0.42.1, local=0.42.1
tf2_geometry_msgs                         : latest=0.42.1, local=0.42.1
control_toolbox                           : latest=5.4.0, local=5.4.0
urdf_parser_plugin                        : latest=2.13.0, local=2.13.0
urdf                                      : latest=2.13.0, local=2.13.0
joint_limits                              : latest=5.1.0, local=5.1.0
pal_statistics_msgs                       : latest=2.6.2, local=2.6.2
pal_statistics                            : latest=2.6.2, local=2.6.2
sdformat_urdf                             : latest=2.0.1, local=2.0.1
hardware_interface                        : latest=5.1.0, local=5.1.0
controller_interface                      : latest=5.1.0, local=5.1.0
controller_manager_msgs                   : latest=5.1.0, local=5.1.0
diagnostic_updater                        : latest=4.4.6, local=4.4.6
ros2cli                                   : latest=0.39.0, local=0.39.0
ros2node                                  : latest=0.39.0, local=0.39.0
ros2topic                                 : latest=0.39.0, local=0.39.0
ros2service                               : latest=0.39.0, local=0.39.0
ros2param                                 : latest=0.39.0, local=0.39.0
ros2pkg                                   : latest=0.39.0, local=0.39.0
ros2run                                   : latest=0.39.0, local=0.39.0
controller_manager                        : latest=5.1.0, local=5.1.0
demo_nodes_cpp                            : latest=0.37.0, local=0.37.0
demo_nodes_cpp_native                     : latest=0.37.0, local=0.37.0
demo_nodes_py                             : latest=0.37.0, local=0.37.0
image_geometry                            : latest=4.1.0, local=4.1.0
depthimage_to_laserscan                   : latest=2.5.1, local=2.5.1
dummy_map_server                          : latest=0.37.0, local=0.37.0
dummy_sensors                             : latest=0.37.0, local=0.37.0
kdl_parser                                : latest=2.13.0, local=2.13.0
robot_state_publisher                     : latest=3.5.1, local=3.5.1
dummy_robot_bringup                       : latest=0.37.0, local=0.37.0
examples_rclcpp_minimal_action_client     : latest=0.21.0, local=0.21.0
examples_rclcpp_minimal_action_server     : latest=0.21.0, local=0.21.0
examples_rclcpp_minimal_client            : latest=0.21.0, local=0.21.0
examples_rclcpp_minimal_composition       : latest=0.21.0, local=0.21.0
examples_rclcpp_minimal_publisher         : latest=0.21.0, local=0.21.0
examples_rclcpp_minimal_service           : latest=0.21.0, local=0.21.0
examples_rclcpp_minimal_subscriber        : latest=0.21.0, local=0.21.0
examples_rclcpp_minimal_timer             : latest=0.21.0, local=0.21.0
examples_rclcpp_multithreaded_executor    : latest=0.21.0, local=0.21.0
examples_rclpy_executors                  : latest=0.21.0, local=0.21.0
examples_rclpy_minimal_action_client      : latest=0.21.0, local=0.21.0
examples_rclpy_minimal_action_server      : latest=0.21.0, local=0.21.0
examples_rclpy_minimal_client             : latest=0.21.0, local=0.21.0
examples_rclpy_minimal_publisher          : latest=0.21.0, local=0.21.0
examples_rclpy_minimal_service            : latest=0.21.0, local=0.21.0
examples_rclpy_minimal_subscriber         : latest=0.21.0, local=0.21.0
image_tools                               : latest=0.37.0, local=0.37.0
intra_process_demo                        : latest=0.37.0, local=0.37.0
joy                                       : latest=3.3.0, local=3.3.0
lifecycle                                 : latest=0.37.0, local=0.37.0
logging_demo                              : latest=0.37.0, local=0.37.0
pcl_msgs                                  : latest=1.0.0, local=1.0.0
pcl_conversions                           : latest=2.7.2, local=2.7.2
pendulum_msgs                             : latest=0.37.0, local=0.37.0
tlsf_cpp                                  : latest=0.19.0, local=0.19.0
pendulum_control                          : latest=0.37.0, local=0.37.0
quality_of_service_demo_cpp               : latest=0.37.0, local=0.37.0
quality_of_service_demo_py                : latest=0.37.0, local=0.37.0
tf2_bullet                                : latest=0.42.1, local=0.42.1
tf2_eigen                                 : latest=0.42.1, local=0.42.1
tf2_eigen_kdl                             : latest=0.42.1, local=0.42.1
tf2_kdl                                   : latest=0.42.1, local=0.42.1
sensor_msgs_py                            : latest=5.7.0, local=5.7.0
tf2_sensor_msgs                           : latest=0.42.1, local=0.42.1
tf2_tools                                 : latest=0.42.1, local=0.42.1
geometry2                                 : latest=0.42.1, local=0.42.1
launch_testing_ros                        : latest=0.29.0, local=0.29.0
ros2action                                : latest=0.39.0, local=0.39.0
ros2component                             : latest=0.39.0, local=0.39.0
ros2doctor                                : latest=0.39.0, local=0.39.0
ros2interface                             : latest=0.39.0, local=0.39.0
ros2launch                                : latest=0.29.0, local=0.29.0
ros2lifecycle                             : latest=0.39.0, local=0.39.0
ros2multicast                             : latest=0.39.0, local=0.39.0
sros2                                     : latest=0.16.0, local=0.16.0
ros2cli_common_extensions                 : latest=0.5.0, local=0.5.0
rosidl_core_generators                    : latest=0.4.0, local=0.4.0
rosidl_default_generators                 : latest=1.8.0, local=1.8.0
sros2_cmake                               : latest=0.16.0, local=0.16.0
ros_core                                  : latest=0.13.0, local=0.13.0
rosbag2_storage                           : latest=0.33.0, local=0.33.0
rosbag2_cpp                               : latest=0.33.0, local=0.33.0
rosbag2_compression                       : latest=0.33.0, local=0.33.0
rosbag2_interfaces                        : latest=0.33.0, local=0.33.0
rosbag2_transport                         : latest=0.33.0, local=0.33.0
rosbag2_py                                : latest=0.33.0, local=0.33.0
ros2bag                                   : latest=0.33.0, local=0.33.0
rosbag2_compression_zstd                  : latest=0.33.0, local=0.33.0
rosbag2_storage_mcap                      : latest=0.33.0, local=0.33.0
rosbag2_storage_sqlite3                   : latest=0.33.0, local=0.33.0
rosbag2_storage_default_plugins           : latest=0.33.0, local=0.33.0
rosbag2                                   : latest=0.33.0, local=0.33.0
ros_base                                  : latest=0.13.0, local=0.13.0
rqt_gui                                   : latest=1.10.0, local=1.10.0
rqt_gui_py                                : latest=1.10.0, local=1.10.0
rqt_py_common                             : latest=1.10.0, local=1.10.0
rqt_console                               : latest=2.4.0, local=2.4.0
rqt_msg                                   : latest=1.7.0, local=1.7.0
rqt_action                                : latest=2.4.0, local=2.4.0
rqt_bag                                   : latest=2.1.0, local=2.1.0
rqt_plot                                  : latest=1.7.1, local=1.7.1
rqt_bag_plugins                           : latest=2.1.0, local=2.1.0
rqt_graph                                 : latest=1.8.0, local=1.8.0
qt_gui_cpp                                : latest=2.10.1, local=2.10.1
rqt_gui_cpp                               : latest=1.10.0, local=1.10.0
rqt_image_view                            : latest=1.3.0, local=1.3.0
rqt_publisher                             : latest=1.10.0, local=1.10.0
rqt_py_console                            : latest=1.5.0, local=1.5.0
rqt_reconfigure                           : latest=1.8.0, local=1.8.0
rqt_service_caller                        : latest=1.5.0, local=1.5.0
rqt_shell                                 : latest=1.4.0, local=1.4.0
rqt_srv                                   : latest=1.4.0, local=1.4.0
rqt_topic                                 : latest=1.9.0, local=1.9.0
rqt_common_plugins                        : latest=1.2.0, local=1.2.0
interactive_markers                       : latest=2.8.0, local=2.8.0
laser_geometry                            : latest=2.11.0, local=2.11.0
map_msgs                                  : latest=2.6.0, local=2.6.0
point_cloud_transport                     : latest=5.2.0, local=5.2.0
resource_retriever                        : latest=3.8.1, local=3.8.1
rviz_rendering                            : latest=15.1.3, local=15.1.3
rviz_common                               : latest=15.1.3, local=15.1.3
rviz_resource_interfaces                  : latest=15.1.3, local=15.1.3
rviz_default_plugins                      : latest=15.1.3, local=15.1.3
rviz2                                     : latest=15.1.3, local=15.1.3
teleop_twist_joy                          : latest=2.6.3, local=2.6.3
teleop_twist_keyboard                     : latest=2.4.0, local=2.4.0
topic_monitor                             : latest=0.37.0, local=0.37.0
turtlesim_msgs                            : latest=1.10.1, local=1.10.1
turtlesim                                 : latest=1.10.1, local=1.10.1
desktop                                   : latest=0.13.0, local=0.13.0
diff_drive_controller                     : latest=5.0.2, local=5.0.2
forward_command_controller                : latest=5.0.2, local=5.0.2
gps_msgs                                  : latest=2.1.0, local=2.1.0
tracetools_image_pipeline                 : latest=7.0.0, local=7.0.0
image_proc                                : latest=7.0.0, local=7.0.0
image_publisher                           : latest=7.0.0, local=7.0.0
joint_state_broadcaster                   : latest=5.0.2, local=5.0.2
joint_state_publisher                     : latest=2.4.0, local=2.4.0
joint_state_publisher_gui                 : latest=2.4.0, local=2.4.0
joy_linux                                 : latest=3.3.0, local=3.3.0
ros_gz_interfaces                         : latest=3.0.1, local=3.0.1
vision_msgs                               : latest=4.2.0, local=4.2.0
ros_gz_bridge                             : latest=3.0.1, local=3.0.1
ros_gz_image                              : latest=3.0.1, local=3.0.1
ros_gz_sim                                : latest=3.0.1, local=3.0.1
ros2test                                  : latest=0.9.0, local=0.9.0
ros_testing                               : latest=0.9.0, local=0.9.0
test_msgs                                 : latest=2.4.0, local=2.4.0

   PLATFORM INFORMATION
system           : Linux
platform info    : Linux-6.12.30-x86_64-with-glibc2.39
release          : 6.12.30
processor        : x86_64

   QOS COMPATIBILITY LIST
compatibility status    : No publisher/subscriber pairs found

   RMW MIDDLEWARE
middleware name    : rmw_fastrtps_cpp

   ROS 2 INFORMATION
distribution name      : rolling
distribution type      : ros2
distribution status    : rolling
release platforms      : {'debian': ['bookworm'], 'rhel': ['9'], 'ubuntu': ['noble']}

   TOPIC LIST
topic               : none
publisher count     : 0
subscriber count    : 0   
```
</details>

### Steps to reproduce issue

Run this script:
```Python
#!/usr/bin/env python3

import threading

import rclpy
from rclpy.task import Future
from rclpy.executors import SingleThreadedExecutor

rclpy.init()
executor = SingleThreadedExecutor()
future = Future()


async def await_task():
    print("Waiting for future to complete...")
    await future
    print("Future completed with result:", future.result())


def future_thread():
    print("Simulating work in a separate thread...")
    threading.Event().wait(2)  # Simulate some work
    future.set_result("Task completed!")
    print("Future set!")
    # executor.wake()  # Wake the executor to process the future

thread = threading.Thread(target=future_thread)
thread.start()

executor.create_task(await_task())
executor.spin()
```

### Expected behavior

The executor wakes up and completes `await_task` coroutine after another threads runs `future.set_result`.

### Actual behavior

The task gets stuck on `await future` unless I uncomment the `executor.wake()` call

### Additional information

_No response_

bjsowa:
## Description
Fixes #1467 
Fixes #1405
Fixes #1462

- ~The `_pending()` method in Future has been renamed to `pending()`~
- The user no longer has to do any workarounds with waking up the executor when a coroutine gets stuck awaiting a future that is already finished.

Some of the comments and docstrings might be partially written by Github Copilot.

### Additional Information

~For #1467, I propagate a guard condition from a task to the awaited future and add this guard condition to the wait set in the executor. When a future is completed by an entity not controller by the executor (e.g. a separate thread), the guard condition is triggered and wakes up the executor so it can resume the task.~

~For #1405, I just had to make sure the task is not yielded from `_wait_for_ready_callbacks` when it is not ready to be resumed, hence the `Task.ready()` method.~

EDIT: Check [this comment](https://github.com/ros2/rclpy/pull/1469#issuecomment-3052365815)


rclpy/rclpy/task.py
```diff
@@ -68,20 +72,41 @@ def __del__(self) -> None:
...
+        :param guard_condition: The guard condition to use.
+        """
+        with self._lock:
+            self._guard = guard
+
+        # If the future is already done, trigger the guard condition immediately
+        if not self.pending():
+            guard.trigger()
+
+    def pending(self) -> bool:
```
nadavelkabets: Until now, rclpy Future and Task objects aligned with the api of asyncio and Concurrent.futures.
Diverting from the standard Future api will disable our ability to switch to the standard implementation if we ever wanted to do that.

https://peps.python.org/pep-3148/#future-objects
https://docs.python.org/3/library/asyncio-future.html

rclpy/rclpy/executors.py
```diff
@@ -699,6 +696,13 @@ def _wait_for_ready_callbacks(
             if timeout_timer is not None:
                 timers.append(timeout_timer)
 
+            with self._tasks_lock:
+                for task, _, _ in self._tasks:
+                    if task.paused():
+                        # If the task is paused, its guard condition needs to be added to the wait
+                        # set so that it can wake up the executor when it is ready to run.
+                        guards.append(task.guard)
```
nadavelkabets: A “paused” task can either await a bare yield (which should result in revisiting the task again immediately), or some future object, which should yield itself.
As suggested in #1399 and #971, when a future yields itself, the task can be “suspended” from the executor, resuming only when the blocking future is finished or cancelled.
I believe that triggering the global guard condition (`executor.wake`) from the resume method like `call_soon` should be sufficient to solve the issue.

rclpy/rclpy/task.py
bjsowa: My newer solution doesn't change the api of Task and Future

bjsowa:
After looking at #971, I revised my solution to not modify the API of Task or Future.

Changes:
- Added a `call_soon` method which schedules the task to be run in the next spin and wakes up the executor.
- The tasks schedule themselves to be run in the next spin, using the new `call_soon` method, if they are ready to be resumed.
- executor._tasks list now contains only tasks that are ready to run or resumed, not all tasks.

Improvements:
- Fixes #1467, #1405 and now also #1462
- Skips checking if tasks are ready to be resumed in each spin.

Regressions:
- When a coroutine task awaits other task, when the other task finishes, the coroutine task needs 2 spins to be resumed as the first spin will only execute a callback which schedules the coroutine task to be run in next spin. This is reflected in the test_create_task_dependent_coroutines test.
- There are 2 tests now that result in a Segmentation fault when run with the EventsExecutor. If this solution is accepted I might look into fixing it, any help would be appreciated.

bjsowa:
@nadavelkabets Could you take a look?

nadavelkabets:
> @nadavelkabets Could you take a look?

I'll try to get to it this weekend

rclpy/rclpy/executors.py
```diff
@@ -251,12 +251,20 @@ def create_task(self, callback: Callable[..., Any], *args: Any, **kwargs: Any
         :param callback: A callback to be run in the executor.
         """
         task = Task(callback, args, kwargs, executor=self)
+        self.call_soon(task)
+        # Task inherits from Future
+        return task
+
+    def call_soon(self, task: Task) -> None:
```
nadavelkabets: Currently, the sole scheduling method in the executor is create_task, allowing the scheduling of 3 possible object types: standard (non-async) functions, async functions (Callables that return a coroutine), and coroutines (an initialized async function). 
This API contradicts the asyncio event loop, in which create_task only accepts a coroutine, while call_soon is used to execute non-async functions.
This is relevant to us, since asyncio's call_soon is used to resume a task by executing task.__wakeup.

I see two alternatives:

1. Keeping the current create_task API and adding call_soon as a protected, internal method (as there is no reason for users to schedule tasks themselves since it is already handled by the executor).
2. Altering the existing API to match asyncio - deprecating the ability to schedule non-async functions in create_task. This also simplifies the implementation of Executor, allowing us to get rid of "await_or_execute".

I personally prefer the second one, but I think the two alternatives should be discussed in the next client library working group meeting.

On one hand, I would love to see this feature backported, as the way tasks are currently handled is inefficient and buggy, but on the other hand, this feature unfortunately changes the existing behavior of the executor.
In my opinion these changes align better to the API and intentioned behavior but I am not the one to decide if this could be backported or not.

rclpy/rclpy/task.py
```diff
@@ -68,8 +68,8 @@ def __del__(self) -> None:
 
     def __await__(self) -> Generator[None, None, Optional[T]]:
         # Yield if the task is not finished
-        while self._pending():
-            yield
+        if self._pending():
+            yield self
```
nadavelkabets, line 73: Asyncio raises an exception here if the Future is awaited a second time, I think we should do the same.

rclpy/rclpy/task.py
```diff
@@ -299,9 +299,21 @@ def __call__(self) -> None:
 
             if inspect.iscoroutine(self._handler):
                 # Execute a coroutine
-                handler = self._handler
                 try:
-                    handler.send(None)
+                    result = self._handler.send(None)
+                    executor = self._executor()
+                    if executor is None:
```
nadavelkabets: We should either raise here or drop this as it's never supposed to happen - task objects should only be initialized by the executor's create_task method (that always initializes them with the executor parameter).

rclpy/rclpy/task.py
```diff
@@ -299,9 +299,21 @@ def __call__(self) -> None:
...
+                        # If there is no executor, we can't schedule the task to resume
+                        pass
+                    elif isinstance(result, Future):
+                        # Schedule the task to resume when the future is done
+                        result.add_done_callback(lambda _: executor.call_soon(self))
```
nadavelkabets: The change of behavior you noticed - requiring two spins to resume a task originates from this line.
When the awaited future completes, it schedules this done callback as a task. The first spin executes this task, calling call_soon to schedule self for the next executor iteration.
In the AsyncioExecutor, I kept the existing behavior by replacing `result.add_done_callback(lambda _: executor.call_soon(self))` with `result.add_done_callback(self.__wake)`: 
```python
def __wake(self, fut: Future): 
     self._fut_waiter = None 
     if fut.cancelled(): 
         self.cancel()
     elif fut.exception() is not None: 
         self.set_exception(fut.exception()) 
     else: 
         self()
```

rclpy/rclpy/task.py
bjsowa: I we wanna mimic the asyncio behavior, should the `done()` method return true for cancelled futures?

bjsowa: https://github.com/python/cpython/blob/6293d8a1a648a498b7ac899631b74fa25c71c1ac/Lib/asyncio/futures.py#L193

nadavelkabets: Currently rclpy tasks behave differently when canceled. Tasks have 3 states: pending, canceled and finished (when a result or an exception is set). In asyncio a task is done if it is not pending. Currently in rclpy a task is done only if it is finished.
Ideally we would want the behavior to be the same, but it means braking behavior and that would prevent a feature from being back ported.

bjsowa: Got it. I modified the __await__ method to raise an exception when it is awaited a second time. For some reason, it resulted in another test segfaulting on EventsExecutor.

rclpy/rclpy/task.py
bjsowa: Should I make the `executor` parameter "not optional" then? What about the executor in Future?

rclpy/rclpy/task.py
nadavelkabets: EventsExecutor utilizes the python implementation of Task and Future, but does not share implementation with the python Executor class. The changes you made to Executor do not apply to EventExecutor and similar changes must be made there. I do not like that the logic is duplicated in cpp, but that's how it is currently.

rclpy/rclpy/task.py
nadavelkabets: The executor parameter is inherited from Future but should always be there in a task.
Users are not supposed to initialize tasks by themselves, and they should always utilize executor.create_task - as a task should always be bound to a running executor.

bjsowa: So should the parameter stay optional in Future but be mandatory in Task?

rclpy/rclpy/task.py
bjsowa: Understandable. I'll look at the EventsExecutor implementation after this gets accepted.

rclpy/rclpy/task.py
nadavelkabets: Ideally yeah, but I'm not sure we should change that as users are not supposed to interact with this class anyway.
Is it mandatory in asyncio?

bjsowa: In asyncio, if you don't specify the `loop` parameter, it [gets the default event loop](https://github.com/python/cpython/blob/6293d8a1a648a498b7ac899631b74fa25c71c1ac/Lib/asyncio/futures.py#L76-L83)

nadavelkabets: Then we should probably raise.

bjsowa: I now raise an exception but had to skip one test.

rclpy/rclpy/task.py
bjsowa: I have an issue with this approach. I assume the tasks shouldn't be run outside of executor's context. Consider this situation:
1. A Future is created without an executor.
2. The Executor and Task are created.
3. The Task awaits the Future.
4. Another thread calls Future.set_result()

Because the Future was created without an executor, the call to set_result() immediately invokes the callback which wakes up and runs the task, thus the task is run in another thread's context instead of Executor's spin.

rclpy/rclpy/executors.py
bjsowa: If possible, I would prefer the way which allows this to be backported (at least to Jazzy).
Regarding the first option, do we allow Task to call private methods of Executor?

bjsowa: Renamed `call_soon` to `_call_task_in_next_spin`. Let me know what you think.

rclpy/rclpy/task.py
bjsowa: I found a way to make it always work in a single spin and from the executor context but it required checking future's executor and I'm not sure it's allowed. Waiting for another review.

bjsowa:
@nadavelkabets any plans to move this further?

nadavelkabets:
> @nadavelkabets any plans to move this further?

Hi, I'll try to go over it again tomorrow.

bjsowa:
@nadavelkabets friendly ping :)

rclpy/rclpy/executors.py
```diff
@@ -616,8 +624,6 @@ async def handler(entity: 'EntityT', gc: GuardCondition, is_shutdown: bool,
         task: Task[None] = Task(
             handler, (entity, self._guard, self._is_shutdown, self._work_tracker),
             executor=self)
-        with self._tasks_lock:
-            self._tasks.append((task, entity, node))
```
nadavelkabets, line 667: Is this working? I think that the Task instance is created, but is not added to the tasks list.
Is it possible to call create_task here?

rclpy/rclpy/task.py
nadavelkabets: Hi, sorry for taking so long and thanks for your patience!

In asyncio, users are only allowed to create futures from an active event loop using loop.create_future().
As I see it, Futures should not exist without an executor, especially when they are created by the user. 
In rclpy, the executor is optional in Future to enable the use case of calling a service client before attaching the node to an executor - the executor is then set just before execution of the service callback (a strange design choice in my opinion). 
https://github.com/ros2/rclpy/blob/f3a326cdb92f1a7c9ff23e4911df5ca5ab36822d/rclpy/rclpy/executors.py#L496-L515

I think we should also discourage users from initializing a Future object directly, to prevent such case of a missing executor / executor mismatch. 
I added this feature in #1470, but perhaps I should create a separate PR just for it.

I am somewhat conflicted. It would be best to raise an exception if a Task tried to attach to a Future without an executor (this is also covered for the client use case as the task is created only after an executor is attached to the future), but I fear that might break user code since we never encouraged users to add the optional executor parameter and I assume most developers didn’t.
The other option isn’t fantastic either, as it would introduce conflicting behavior (requiring a single spin once if an executor is set, and two spins otherwise).

Given the assumption that multiple executors are never spinning concurrently, I think we should set the executor of the future to be the executor of the task if the executor is not set. We should also probably log here to inform the users that they should provide the executor argument themselves (or preferably use executor.create_future()). 
If it’s a different executor then we should probably raise.

What do you think?

rclpy/rclpy/executors.py
nadavelkabets: Sorry again for not following up. 
We discussed the two alternatives in the last working group meeting and decided on the first alternative (the one you implemented). 
We agreed that the second alternative is superior, but such major change in API requires community feedback and so is not currently viable.

rclpy/test/test_task.py
```diff
@@ -54,6 +56,7 @@ def func() -> str:
         self.assertTrue(t.done())
         self.assertEqual('Sentinel Result', t.result())
 
+    @pytest.mark.skip(reason='Resuming coroutine tasks is not supported without an executor')
```
nadavelkabets: I think you can remove this test entirely from test_task, and add a similar one to test_executor (if there isn’t one already), replacing t() with spin_once().

rclpy/test/test_executor.py
```diff
@@ -315,6 +317,29 @@ async def coroutine():
...
+                    threading.Event().wait(0.1)  # Simulate some work
+                    thread_future.set_result(None)
+
+                t = threading.Thread(target=future_thread)
+                t.start()
+
+                coroutine_future = executor.create_task(coroutine)
+                executor.spin_until_future_complete(coroutine_future, timeout_sec=1.0)
+
+                self.assertTrue(coroutine_future.done())
```
nadavelkabets, line 382: You should probably also assert that the elapsed time is close to 0.1.

rclpy/rclpy/task.py
```diff
@@ -322,6 +337,17 @@ def __call__(self) -> None:
...
+    def _add_resume_callback(self, future: Future[T], executor: 'Executor') -> None:
+        future_executor = future._executor()
+        if future_executor is not None and future_executor is executor:
+            # The future is associated with the same executor, so we can resume the task directly
+            # in the done callback
+            future.add_done_callback(lambda _: self.__call__())
+        else:
+            # If the future is not associated with the same executor, we need to schedule
+            # the task to resume in the next spin of the executor.
+            future.add_done_callback(lambda _: executor._call_task_in_next_spin(self))
```
nadavelkabets: As I explained above, I fear that this might induce unexpected behavior for many people, since setting the executor is currently not an encouraged practice.
I believe we should set the executor of the future if it’s not set (with some informative log), and raise if it’s different.

rclpy/rclpy/executors.py
bjsowa: `_make_handler` is only used in `_wait_for_ready_callbacks` which yields the handler to the executor so it can call it. When a task is ready to be resumed, it will add itself to the `_tasks` list by calling `_call_task_in_next_spin`

rclpy/rclpy/task.py
bjsowa: > Given the assumption that multiple executors are never spinning concurrently, I think we should set the executor of the future to be the executor of the task if the executor is not set. 

I think it's a good idea. I'll get to it after the weekend.

> We should also probably log here to inform the users that they should provide the executor argument themselves (or preferably use executor.create_future()).

Should I log in the constructor or only when calling the task?
Should I add `executor.create_future()`?

nadavelkabets: I think it’s best to add create_future in a separate PR. The code exists in #1740 and I just have to move it out. It is related but not necessary for this PR to work.
I think you should log only when attaching to a future without an executor and setting its future.

rclpy/test/test_task.py
bjsowa: I added `test_create_task_coroutine_yield` to executor tests

rclpy/test/test_executor.py
bjsowa: done

rclpy/rclpy/task.py
nadavelkabets: Hey @bjsowa will you have the time to make these changes?

bjsowa: I'll try to find some, sorry for the lack of response. I still want to finish it.

bjsowa: done, waiting for another review

rclpy/rclpy/executors.py
```diff
@@ -251,12 +251,20 @@ def create_task(self, callback: Callable[..., Any], *args: Any, **kwargs: Any
         :param callback: A callback to be run in the executor.
         """
         task = Task(callback, args, kwargs, executor=self)
+        self._call_task_in_next_spin(task)
+        # Task inherits from Future
```
nadavelkabets: What's the reason for this comment?

bjsowa: It was there before

rclpy/rclpy/task.py
```diff
@@ -299,9 +307,21 @@ def __call__(self) -> None:
...
             if inspect.iscoroutine(self._handler):
                 # Execute a coroutine
-                handler = self._handler
                 try:
-                    handler.send(None)
+                    result = self._handler.send(None)
+                    executor = self._executor()
+                    if executor is None:
+                        raise RuntimeError(
+                            'Task is a coroutine but no executor is available to resume it')
```
nadavelkabets: Perhaps "Task tried to reschedule but no executor was set: tasks should only be initialized through executor.create_task()" would be more indicative. What do you think?

rclpy/rclpy/task.py
```diff
@@ -299,9 +307,21 @@ def __call__(self) -> None:
...
+                    elif isinstance(result, Future):
+                        # Schedule the task to resume when the future is done
+                        self._add_resume_callback(result, executor)
+                    elif result is None:
+                        # The coroutine yielded None, schedule the task to resume
+                        executor._call_task_in_next_spin(self)
+                    else:
+                        raise TypeError(
+                            'Expected coroutine to yield a Future or None, got: {}'.format(result))
```
nadavelkabets: Could you move this logic out of the try block?
I think we should crash the executor directly (Currently the exception is set for the task that is then raised outside but that might not be the case in the future)

rclpy/rclpy/task.py
```diff
@@ -322,6 +342,22 @@ def __call__(self) -> None:
...
             self._task_lock.release()
 
+    def _add_resume_callback(self, future: Future[T], executor: 'Executor') -> None:
+        future_executor = future._executor()
+        if future_executor is None:
+            Task._logger.warning(
+                'A task is awaiting a future that is not associated with an executor. '
+                'This might not be supported in the next ROS distribution. '
+                'Consider using executor.create_future().',
+                once=True)
```
nadavelkabets: Could you try to use python's builtin DeprecationWarning?
It believe it should keep the same behavior of only logging once but that should be validated.

rclpy/rclpy/task.py
```diff
@@ -322,6 +342,22 @@ def __call__(self) -> None:
...
+            future._set_executor(executor)
+        elif future_executor is not executor:
+            raise RuntimeError('A task can only await futures associated with the same executor')
+
+        # The future is associated with the same executor, so we can resume the task directly
+        # in the done callback
+        future.add_done_callback(lambda _: self.__call__())
```
nadavelkabets, line 357: Due to this lambda we will resume the task even if the blocking future is cancelled.
Perhaps we should check if it's cancelled before resuming the task again.

rclpy/test/test_executor.py
```diff
@@ -293,6 +295,41 @@ async def coroutine():
...
+                self.assertFalse(future.done())
+                self.assertFalse(called1)
+                self.assertFalse(called2)
+
+                executor.spin_once(timeout_sec=0)
+                self.assertFalse(future.done())
+                self.assertTrue(called1)
+                self.assertFalse(called2)
+
+                executor.spin_once(timeout_sec=1)
```
nadavelkabets, line 328: Does it not work with timeout=0?

nadavelkabets (changes requested):
Thanks for your fast response.
I think we're almost ready to go forward with this PR.
We can't merge this until the EventsExecutor is updated as well.
Do you feel like this PR is mature enough for you to do that or would you want feedback from someone else first?

rclpy/rclpy/executors.py
nadavelkabets: If you see no reason for it as well, let's remove it. Up to you.

bjsowa: I removed it

rclpy/rclpy/task.py
bjsowa: Sounds better

rclpy/rclpy/task.py
bjsowa: I meant to crash the executor directly, just didn't notice that it would get caught by the except block. I moved the logic to the `else` block and moved the whole coroutine handling to a helper method for more clarity

rclpy/rclpy/task.py
bjsowa: I changed it to use DeprecationWarning. The warning is not shown though as the default warning filters are set so that DeprecationWarning is only shown when triggered from `__main__`

rclpy/rclpy/task.py
bjsowa: I'm not sure I follow. At the time the coroutine yields a future, we know that this future is still pending, so we tell the future to schedule the next call of the task to the executor when the future either finishes or is cancelled. A cancelled future will just return None in the next `send`

bjsowa:
@nadavelkabets I raise a warning when awaiting a future without an executor attached as you suggested. The problem is that now this code will generate a warning:

<details>
    <summary>async_client.py</summary>

    import rclpy
    
    from std_srvs.srv import Empty
    
    rclpy.init()
    
    node = rclpy.create_node("async_client_node")
    
    cli = node.create_client(srv_type=Empty, srv_name="test_service")
    
    async def timer_callback():
        await cli.call_async(Empty.Request())
    
    timer = node.create_timer(1.0, timer_callback)
    rclpy.spin(node)
</details>


because at the time we call `call_async`, the future does not have an executor associated with it yet.

Do you have any suggestions on how to fix it?

rclpy/test/test_executor.py
bjsowa: It does not

nadavelkabets:
> Do you have any suggestions on how to fix it?

I'm not sure that I have a simple solution for it.
The architecture of rclpy was not built for this use case.
Client has no access to node or executor, making it really hard to create a future properly. Furthermore, it is currently possible to call a client without even initializing an executor. A perfect solution for this problem might have to introduce behavior changes that I'm not sure yet we want.
I'll take some time to think this further, but perhaps for now, unfortunately, the best course is to remove the warning.

bjsowa:
> > Do you have any suggestions on how to fix it?
> 
> I'm not sure that I have a simple solution for it. The architecture of rclpy was not built for this use case. Client has no access to node or executor, making it really hard to create a future properly. Furthermore, it is currently possible to call a client without even initializing an executor. A perfect solution for this problem might have to introduce behavior changes that I'm not sure yet we want. I'll take some time to think this further, but perhaps for now, unfortunately, the best course is to remove the warning.

Ok, I removed it for now. If the solution gets accepted I can work on updating the EventsExecutor.

bjsowa:
@nadavelkabets I tried adding the EventsExecutor part and it seems to work for the most fault but segfaults on `test_not_lose_callback`. I tried running the test in gdb but the backtraces were not that helpful and I'm not experienced with debugging C++ code run from Python. Any help would be appreciated.

nadavelkabets:
> @nadavelkabets I tried adding the EventsExecutor part and it seems to work for the most fault but segfaults on `test_not_lose_callback`. I tried running the test in gdb but the backtraces were not that helpful and I'm not experienced with debugging C++ code run from Python. Any help would be appreciated.

I fixed the issue you were experiencing and tests are passing for me.
You should merge the changes to your branch https://github.com/bjsowa/rclpy/pull/1.

I'm really happy with this PR.
I'll look at the code once again after you merge the fixes for final approval.
To merge this, we also need the approval of a project maintainer.
I think it's best to present the changes in the working group meeting this Friday.
Meeting link is https://calendar.app.google/2LNP3u8tdffvyCPXA.
Do you want to present the changes or should I?

bjsowa:
> I fixed the issue you were experiencing and tests are passing for me. You should merge the changes to your branch [bjsowa#1](https://github.com/bjsowa/rclpy/pull/1).
> 
> I'm really happy with this PR. I'll look at the code once again after you merge the fixes for final approval. To merge this, we also need the approval of a project maintainer. I think it's best to present the changes in the working group meeting this Friday. Meeting link is https://calendar.app.google/2LNP3u8tdffvyCPXA. Do you want to present the changes or should I?

I merged your changes. I'll try to attend the meeting if I'll have nothing important to do. I would prefer you present the changes though.

btw. the google calendar link does not seem to work

rclpy/src/rclpy/events_executor/events_executor.cpp
```diff
@@ -75,10 +75,16 @@ pybind11::object EventsExecutor::create_task(
...
+  call_task_in_next_spin(task);
   return task;
 }
 
+void EventsExecutor::call_task_in_next_spin(pybind11::handle task)
+{
+  events_queue_.Enqueue(std::bind(&EventsExecutor::IterateTask, this, task));
+}
+
+
```
mjcarroll: Super tiny nit: unnecessary whitespace

mjcarroll:
Thank you @nadavelkabets for doing a huge amount of work here with the review and thanks to @bjsowa for iterating.  I'm catching up on the context of the conversation, but it appears to be in good shape, has this run through CI yet?

nadavelkabets:
> I'm catching up on the context of the conversation, but it appears to be in good shape, has this run through CI yet?

Hey @mjcarroll, thanks for the kind words, it’s been a real pleasure to work with @bjsowa on this.
Would be great to run CI.

rclpy/src/rclpy/events_executor/events_executor.cpp
bjsowa: Fixed. IMO this should be caught by the formatter but ament_uncrustify is very permissive.

mjcarroll:
Pulls: ros2/rclpy#1469
Gist: https://gist.githubusercontent.com/mjcarroll/a8ebf9a9e1304615ce9ac8681376c7fb/raw/ac292e0fb5a520b80c95e61757fff89b782ebfc0/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16957
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24876/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18910/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4338/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25388/)

rclpy/rclpy/executors.py
```diff
@@ -253,12 +253,19 @@ def create_task(self, callback: Callable[..., Any], *args: Any, **kwargs: Any
...
+        return task
+
+    def _call_task_in_next_spin(self, task: Task) -> None:
+        """
+        Add a task to the executor to be executed in the next spin.
+
+        :param task: A task to be run in the executor.
+        """
         with self._tasks_lock:
             self._tasks.append((task, None, None))
```
nadavelkabets: I now realize that we still have some topics to discuss.
We are changing behavior here - previously, the executor always yielded node and entity, even for blocked tasks. Now, any task that is resumed using _call_task_in_next_spin will yield None for entity and node from the second time onwards.
SingleThreadedExecutor and MultiThreadedExecutor ignore these yielded values, but the node argument is used to filter tasks. I believe that unlike before, the new code will keep running tasks created by a removed node for example.
```
            for task_trio in tasks:
                task, entity, node = task_trio
                if node is None or node in nodes_to_use:
```
We should discuss this behavior change further before proceeding.

rclpy/rclpy/executors.py
```diff
@@ -673,17 +678,17 @@ def _wait_for_ready_callbacks(
...
+                self._tasks = []
+            for task_trio in tasks:
+                task, entity, node = task_trio
+                if node is None or node in nodes_to_use:
+                    yielded_work = True
+                    yield task_trio
+                else:
+                    # Asked not to execute these tasks, so don't do them yet
+                    with self._tasks_lock:
+                        self._tasks.append(task_trio)
```
nadavelkabets: Should we maintain this behavior if the node is not in nodes_to_use?
We get this infinity loop that occurs in the following scenario
- A task created by an entity callback is awaiting something
- The node of the task is removed from the executor

In this case we will keep removing and adding the task forever.

nadavelkabets:
@mjcarroll @fujitatomoya @sloretz 
These aren't dealbreakers but nonetheless I'd appreciate your feedback.

nadavelkabets:
@mjcarroll @fujitatomoya @sloretz @wjwwood 
Hey, would appreciate your feedback.
My intuition is that once an async callback is fired it should be executed to completion even if the node is removed from the executor. 
Could that cause unexpected issues? For example, a service server callback might send a response and I'm not sure that's wanted if the node is removed.
Perhaps a node should cancel its own tasks if it's removed.

rclpy/rclpy/executors.py
bjsowa: Do you see any way this can be fixed?

nadavelkabets: Yes, but that depends on the intended behavior.
If we are okay with the "set and forget" approach, no change is required.
Otherwise, for each node in the executor we can hold a set of running tasks.
When a node is removed from the executor, all tasks related to it should be cancelled.
It's also possible to initialize a task with the node that it originated from as a property instead of this task_trio.

bjsowa: I just realized that holding tuples of `(task, entity, node)` does not really make sense now as the entity and node information is never added after I removed it from `_make_handler`. Maybe we should find some other way to track which node created the task?

rclpy/rclpy/executors.py
bjsowa: I could change it to something like this:
```Python
# Yield tasks in-progress before waiting for new work
with self._tasks_lock:
    # Retrieve a list of tasks that can be executed now
    tasks = [
        (task, entity, node) for (task, entity, node) in self._tasks
        if node is None or node in nodes_to_use
    ]
    # Remove tasks that are going to be executed now from the executor
    # Tasks that need to be executed again will add themselves back to the executor
    self._tasks = [
        (task, entity, node) for (task, entity, node) in self._tasks
        if node is not None and node not in nodes_to_use
    ]

yield from tasks
```
Not much of an optimization but imo looks cleaner and we don't remove and add the same tasks.

rclpy/rclpy/executors.py
bjsowa: @nadavelkabets any ideas?

shaur-k:
is there a timeline to when this gets merged? can i help pick up some work here if needed to get this across the finish line?

nadavelkabets:
> is there a timeline to when this gets merged? can i help pick up some work here if needed to get this across the finish line?

Hey, we have completed most of the work, and only have some minor details to iron out.
I await feedback from other maintainers, we should discuss this at the working group meeting.

rclpy/rclpy/executors.py
nadavelkabets: > I just realized that holding tuples of `(task, entity, node)` does not really make sense now as the entity and node information is never added after I removed it from `_make_handler`. 

Exactly. I would like to discuss this topic with core maintainers to get their opinion.
I'm not sure if we should avoid changing this behavior or if this is a welcomed change.

nadavelkabets:
@bjsowa 
I came up with a solution, and opened a PR in your fork https://github.com/bjsowa/rclpy/pull/2.
Currently, we store the source entity and source node in the same tuple that holds the task itself.
Our refactor changed that, causing 2 problems:
1. The executor no longer held reference to the task (that could lead to unexpected garbage collection)
2. We no longer stored the entity and node data for tasks

My thought process:
- The easiest solution is to add the entity and node as properties of the Task class. I chose to avoid this solution to keep our Task class as close as possible to asyncio, and because I believe that a task should have no interest in what entity or node it originated from.
- We could add a node and entity arguments to the Task's `__call__` function, passing them each iteration of the task, this way the task could reschedule itself with them in the tasks tuple. This seemed irrelevant because it both changes `wait_for_ready_callbacks` api and behavior, and also differs from the asyncio task implementation.
- I then thought that the best course of action is to store the source entity and node information in a dictionary for each pending task. I initially added a done_callback to each task to remove itself from the dictionary after finishing execution, then to discover that tests are failing because they were calling spin_once the exact number of times expected, but the newly added done callback added a new, unexpected event.

The only way to solve this was to keep the same behavior as it is now, filtering the tasks list each iteration of the executor.
Currently this entity and node thing is not covered by unit tests. Perhaps we should add some tests for that.
I have to admit I'm not that fond of my solution as it retains much of the inefficiencies of the current executor, but it fixes our issue nonetheless.
What do you think?

nadavelkabets:
> Should we maintain this behavior if the node is not in nodes_to_use? We get this infinity loop that occurs in the following scenario
> 
> * A task created by an entity callback is awaiting something
> * The node of the task is removed from the executor
> 
> In this case we will keep removing and adding the task forever.

I also tried to solve this issue that annoyed me, by cancelling all tasks related to a node when it is removed from the executor. I think this should be the default behavior, but currently the behavior of rclpy.spin_once adds and removes the node from the executor each call which means that my fix is changing it's behavior because async tasks will get cancelled. For now I left it out of my PR and I'm thinking of another solution, but that should not block this PR from proceeding. For now we maintain the current behavior.

bjsowa:
Thanks @nadavelkabets . I'll try to take a look later today. Just a quick thought. Couldn't we check if the task is done after calling it's handler and remove it from `_pending_tasks` then, instead of waiting for next spin?

nadavelkabets:
> Thanks @nadavelkabets . I'll try to take a look later today. Just a quick thought. Couldn't we check if the task is done after calling it's handler and remove it from `_pending_tasks` then, instead of waiting for next spin?

Good thought, I think this might work. I'll try to implement that.

nadavelkabets:
Follow up:
To implement that, we have to edit Task's `_complete_task` class and add:
```
        executor = self._executor()
        del executor._pending_tasks[self]
```
Unfortunately, _pending_task is currently an internal implementation detail and not API, and it does not exist in EventsExecutor. We could add an empty function def there, but I don't like that.
The other alternative means wrapping each callback in the SingleThreadedExecutor before initializing the Task which I also don't like.

bjsowa:
> Follow up: To implement that, we have to edit Task's `_complete_task` class and add:
> 
> ```
>         executor = self._executor()
>         del executor._pending_tasks[self]
> ```
> 
> Unfortunately, _pending_task is currently an internal implementation detail and not API, and it does not exist in EventsExecutor. We could add an empty function def there, but I don't like that. The other alternative means wrapping each callback in the SingleThreadedExecutor before initializing the Task which I also don't like.

I was thinking just adding a check after calling the task in SingleThreadedExecutor:
```Python
handler()
if handler.done():
    del self._pending_tasks[handler]
```
Of course with added locks for thread safety.

nadavelkabets:
> I was thinking just adding a check after calling the task in SingleThreadedExecutor:
> 
> ```Python
> 
> handler()
> 
> if handler.done():
> 
>     del self._pending_tasks[handler]
> 
> ```
> 
> Of course with added locks for thread safety.

I had this thought it mind. The issue I faced - the handler is yielded to and called from the inherited SingleThreadedExecutor and MultiThreadedExecutor classes, each implementing its own _spin_once_impl. This change will break behavior for wait_for_ready_callbacks users. While it's probably an uncommon use case, it's public api.

bjsowa:
Understood. I think I'm fine with checking the `_pending_tasks` on each spin. As long as we can finally fix the issues with async tasks.

nadavelkabets:
> Understood. I think I'm fine with checking the `_pending_tasks` on each spin. As long as we can finally fix the issues with async tasks.

If you're fine with my solution, let's merge it to this branch and I'll go over this PR again.

bjsowa:
> > Understood. I think I'm fine with checking the `_pending_tasks` on each spin. As long as we can finally fix the issues with async tasks.
> 
> If you're fine with my solution, let's merge it to this branch and I'll go over this PR again.

done

bjsowa:
@nadavelkabets Any way I can help to push this further?

@mjcarroll Could we run the CI again?

nadavelkabets (approved):

nadavelkabets:
@bjsowa
I went over it again and everything looks good to me.
As far as I'm aware, we did not change any behavior apparent to the user.

@mjcarroll @fujitatomoya @sloretz @wjwwood
I believe we are ready to proceed with this PR.
Would be great if you could run CI and go over the changes.

skyegalaxy:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25590/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19619/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/5152/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25827/)

mjcarroll:
@nadavelkabets what do you want to do here? More CI?

nadavelkabets:
> @nadavelkabets what do you want to do here? More CI?

I believe CI failed for other reasons, could we run it again?
Would be great if you could also review the code.

bjsowa:
Could we please rerun the CI? @mjcarroll @skyegalaxy

ahcorde:
Pulls: ros2/rclpy#1469
Gist: https://gist.githubusercontent.com/ahcorde/c5e40bddb7d580655b16c07d5303597f/raw/2680f56c42cbfca556c6ba4fb108ae4b8a25826f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17654
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26357/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20005/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6056/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26097/)

JWhitleyWork:
Looks like CI made it through. Can we please get a review to get this merged? I know everyone is busy but this has been ready to go for a month and we would like to get it back-ported to at least Jazzy and Humble. Thank you! Let me know if I can help in any way.

mjcarroll (approved):
Thanks for all of the iteration here, this is a very high quality refactor.

skyegalaxy (approved):
just finished catching up on all the context and reading through the changes. Looks great!

nadavelkabets:
@mjcarroll @ahcorde @skyegalaxy 
Thank you for taking the time to review this PR.
Are we ready to merge this?

JWhitleyWork:
Thank you all so much for your attention and help with implementing this change! I'll create back-port PRs for at least Jazzy and Humble unless you guys have a bot for that.

skyegalaxy:
https://github.com/Mergifyio backport kilted jazzy humble

PR #1481, opened 2025-07-20
Add new RAW_STEADY_TIME changes

### LINKED ISSUE #488: Use MONOTONIC_RAW instead of MONOTONIC for steady clock
I see that @clalancette recently switched to `MONOTONIC` from `MONOTONIC_RAW` in https://github.com/ros2/rcutils/pull/357 in order to have it similar to `std::chrono::steady_clock` representation. 

I've also seen the discussion in the issue: https://github.com/ros2/rcutils/issues/43. However, I think that we should align more towards the `MONOTONIC_RAW` because if I want to use the steady_clock provided by rcutils for the `ros2_control` applications, I want it to be slew-free rather than being affected by any internal clock change. For instance, Industrial grade motor control boards with EtherCAT protocol might generate some discontinuities in the velocities causing strange noises on the robot. 

It would be great to rethink again on this part to use `MONOTONIC_RAW` instead of `MONOTONIC` (or) if we can add new clock type something like `RCL_STEADY_SLEWFREE_TIME` or something like that to have both implementations?

Thank you

saikishor:
## Description

Needs https://github.com/ros2/rcl/pull/1248 and https://github.com/ros2/rcutils/pull/507 and https://github.com/ros2/rclcpp/pull/2910

Fixes https://github.com/ros2/rcutils/issues/488

NO

NO

@fujitatomoya FYI

### DIFF
--- a/rclpy/rclpy/clock.py
+++ b/rclpy/rclpy/clock.py
@@ -279,6 +279,8 @@ class Clock:
                 event.wait_until_system(self.__clock, until._time_handle)
             elif ClockType.STEADY_TIME == self._clock_type:
                 event.wait_until_steady(self.__clock, until._time_handle)
+            elif ClockType.RAW_STEADY_TIME == self._clock_type:
+                event.wait_until_raw_steady(self.__clock, until._time_handle)
             elif ClockType.ROS_TIME == self._clock_type:
                 event.wait_until_ros(self.__clock, until._time_handle)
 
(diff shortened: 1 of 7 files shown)


fujitatomoya (approved):
the fix itself as rclpy looks good, but we need to discuss on https://github.com/ros2/rclcpp/pull/2910#issuecomment-3118738721

PR #1509, opened 2025-09-27
Update `type_support.py` to use new message abstract base classes 

InvincibleRMC:
## Description

Uses new abstract base classes to allow automatic type inference lowering burden to start using pub/sub/server/client/actionserver/actionclient. This can be seen in how the test cases do not require the manual specification of their generic types and that can now be inferred from the msg/srv/action type automatically.

Fixes one task from this main list
https://github.com/ros2/ros2/issues/1735

The classes inside `type_support.py` are no longer `protocols` but, actual generic classes. Users shouldn't need to change anything for this but, a user could in theory now decide to actual instantiate this class at runtime even if that doesn't really serve a point.


InvincibleRMC:
Converted to draft for the moment since https://github.com/ros2/rosidl_python/pull/241 should come first and might require some changes here.

InvincibleRMC:
@christophebedard Should be ready for review now. Tested it locally with the `BaseImpl` changes to  in `rosidl/rosidl_python`.

InvincibleRMC:
@christophebedard just a reminder whenever you are free I would appreciate a review.

InvincibleRMC:
@christophebedard please let me know if you have any questions/comments on this pr.

rclpy/rclpy/executors.py
```diff
@@ -543,7 +543,8 @@ def _take_subscription(self, sub: Subscription[Any]
                     return None
 
                 if sub._callback_type is Subscription.CallbackType.MessageOnly:
-                    msg_tuple: Union[Tuple[Msg], Tuple[Msg, MessageInfo]] = (msg_info[0], )
+                    msg_tuple: Union[tuple[BaseMessage],
+                                     tuple[BaseMessage, MessageInfo]] = (msg_info[0], )
```
christophebedard: Having to replace `Msg` with `BaseMessage` looks weird. Isn't `Msg` a type alias for `BaseMessage` (from `type_support.py`), so couldn't we use `Msg` here and everywhere else? Same with `BaseAction`/etc.

rclpy/test/test_events_executor.py
```diff
@@ -40,12 +40,14 @@
...
-                                                        test_msgs.action.Fibonacci.Feedback]
+                                                        test_msgs.action.Fibonacci.Feedback,
+                                                        test_msgs.action.Fibonacci.Impl]
 
 
 FibonacciClientGoalHandle: TypeAlias = ClientGoalHandle[test_msgs.action.Fibonacci.Goal,
                                                         test_msgs.action.Fibonacci.Result,
-                                                        test_msgs.action.Fibonacci.Feedback]
+                                                        test_msgs.action.Fibonacci.Feedback,
+                                                        test_msgs.action.Fibonacci.Impl]
```
christophebedard, line 50: Explicitly including "Impl" here (and in many other locations) looks a bit weird, since it's pretty generic (heh), especially since it looks like users will need to use it too. Is it necessary?

christophebedard (commented):
I have some questions

rclpy/test/test_events_executor.py
InvincibleRMC: While it is definitely annoying to have to include it does the allow type checker to catch more errors.

Without it any reference to `Fibonacci.Impl` it currently impossible for type checkers to determine the contents of `Impl` when referenced at a later time. This is side effect of [HKTs](https://github.com/python/typing/issues/548) not existing in python so we have to dance around their non-existence. If you think adding the `Fibonacci.Impl` is to much of a burden we can remove it but, will make it impossible for anyone to type check after the `.Impl`.

For a practical example without it `b = foo(...)` and `c = foo(...)` would both be allowed under type checking rules where only `b = foo(...)` should be.

```python

FibonacciActionClient: TypeAlias = ActionClient[Fibonacci.Goal,
                                                Fibonacci.Result,
                                                Fibonacci.Feedback,
                                                Fibonacci.Impl]

a = FibonacciActionClient()
reveal_type(a)
reveal_type(a._action_type)
reveal_type(a._action_type.Impl)
reveal_type(a._action_type.Impl.SendGoalService)
reveal_type(a._action_type.Impl.SendGoalService.Response)

def foo(bar: Fibonacci.Impl.SendGoalService.Response) -> int:
    return 3

b = foo(a._action_type.Impl.SendGoalService.Response()) # Correct
c = foo(a._action_type.Impl.SendGoalService.Response) # Argument 1 to "foo" has incompatible type "type[Fibonacci_SendGoal_Response]"; expected "Fibonacci_SendGoal_Response"  [arg-type]
```

rclpy/rclpy/type_support.py
```diff
@@ -12,125 +12,84 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Any, ClassVar, Optional, Protocol, Type, TypeVar, Union
+from typing import Any
+from typing import Generic
+from typing import TypeVar
+from typing import Union
+
```
christophebedard, line 19: nitpick

```suggestion
```

rclpy/test/test_events_executor.py
christophebedard: thanks for the explanation! It definitely makes sense to want to catch these type errors. Ideally, we'd do something closer to C++'s `Client<ActionT>`, e.g., `ActionClient[Fibonacci]`, and then it could use `Fibonacci`'s `Goal`/`Result`/`Feedback`/`Impl` without having to specify them manually, but if this is the best way right now, then all good.

christophebedard (approved):

rclpy/test/test_events_executor.py
InvincibleRMC: Yeah it saddens me so. The problem is Python just don't have a good way of doing these sort namespaces of `ActionT`/`SrvT` so the messages do their best to skirt around them but, leads to some jankness like this.

christophebedard (approved):
LGTM with green CI!

Are you able to run CI now?

InvincibleRMC:
> Are you able to run CI now?

Yep! I can run it when I get home from work.

InvincibleRMC:
Pulls: ros2/rclpy#1509
Gist: https://gist.githubusercontent.com/InvincibleRMC/eeb597d0a11b24eb3c96c84a87365890/raw/fbf3c3949e80a0e70e2f976ee0f15089bbefb44b/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rclpy
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17940
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26929/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20444/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6685/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26400/)

InvincibleRMC:
Pulls: ros2/rclpy#1509
Gist: https://gist.githubusercontent.com/InvincibleRMC/c68e67579af57c6f4bb8e03fa2346d75/raw/fbf3c3949e80a0e70e2f976ee0f15089bbefb44b/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rclpy
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17979
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27092/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20487/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6864/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26440/)

InvincibleRMC:
Pulls: ros2/rclpy#1509
Gist: https://gist.githubusercontent.com/InvincibleRMC/4db11e63e8fa3ba902477afdc21c576c/raw/fbf3c3949e80a0e70e2f976ee0f15089bbefb44b/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rclpy
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17988
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27101/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20496/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6879/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26449/)

InvincibleRMC:
@fujitatomoya would you mind doing a second and final review? If I am remembering correctly each PR needs 2 approvals.

christophebedard:
Someone can correct if I'm wrong, but I think it's just "a minimum of 1 approval from a fellow developer who did not author the pull request:" https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Developer-Guide.html#change-control-process

InvincibleRMC:
Ah that seems correct. Feel free to still review if you want @fujitatomoya but, if not will merge.

PR #1531, opened 2025-10-23
ParameterEventHandler support ContentFiltering

Barry-Xu-2018:
## Description

Implement this feature ros2/rclcpp#2957 for rclpy

In `ParameterEventHandler`,  
New API `configure_nodes_filter` affects the behavior of the following two functions.

- add_parameter_event_callback()

    The callback will only be called for parameter events from the specified nodes which are configured in this function.

- add_parameter_callback()

    The callback will only be called for parameter events from the specified nodes which are configured in this function and add_parameter_callback().
    If the nodes specified in this function are different from the nodes specified in add_parameter_callback(), the callback will never be called.

No

### Additional Information

No

### DIFF
--- a/rclpy/rclpy/parameter_event_handler.py
+++ b/rclpy/rclpy/parameter_event_handler.py
@@ -297,6 +297,10 @@ class ParameterEventHandler:
 
         Callbacks are called in FILO manner.
 
+        The configure_nodes_filter() function will affect the behavior of this function.
+        If the node specified in this function isn't included in the nodes specified in
+        configure_nodes_filter(), the callback will never be called.
+
         :param parameter_name: Name of a parameter to tie callback to
         :param node_name: Name of a node, that the parameter should be related to
         :param callback: A callable to be called when the parameter is modified
@@ -359,7 +363,7 @@ class ParameterEventHandler:
         """
         Configure which node parameter events will be received.
 
-        This function depends on middleware support for content filtering.
+        This function depends on rmw implementation support for content filtering.
         If middleware doesn't support contentfilter, return false.
 
         If node_names is empty, the configured node filter will be cleared.
@@ -379,10 +383,13 @@ class ParameterEventHandler:
         :param node_names: Node names to filter parameter events from
 
         :return: True if the filter was successfully applied, False otherwise.
+        :raises: RCLError if internal error occurred when calling the rcl function.
         """
         if node_names is None or len(node_names) == 0:
             # Clear content filter
             self.parameter_event_subscription.set_content_filter('', [])
+            if (self.parameter_event_subscription.is_cft_enabled is True):
+                return False
             return True
 
         filter_expression = ' OR '.join([f'node = %{i}' for i in range(len(node_names))])
--- a/rclpy/rclpy/parameter_event_handler.py
+++ b/rclpy/rclpy/parameter_event_handler.py
@@ -352,6 +352,48 @@ class ParameterEventHandler:
         """
         self._callbacks.remove_parameter_event_callback(handle)
 
+    def configure_nodes_filter(
+        self,
+        node_names: Optional[List[str]] = None,
+    ) -> bool:
+        """
+        Configure which node parameter events will be received.
+
+        This function depends on middleware support for content filtering.
+        If middleware doesn't support contentfilter, return false.
+
+        If node_names is empty, the configured node filter will be cleared.
+
+        If this function return true, only parameter events from the specified node will be
+        received.
+        It affects the behavior of the following two functions.
+        - add_parameter_event_callback()
+          The callback will only be called for parameter events from the specified nodes which are
+          configured in this function.
+        - add_parameter_callback()
+          The callback will only be called for parameter events from the specified nodes which are
+          configured in this function and add_parameter_callback().
+          If the nodes specified in this function is different from the nodes specified in
+          add_parameter_callback(), the callback will never be called.
+
+        :param node_names: Node names to filter parameter events from
+
+        :return: True if the filter was successfully applied, False otherwise.
+        """
+        if node_names is None or len(node_names) == 0:
+            # Clear content filter
+            self.parameter_event_subscription.set_content_filter('', [])
+            return True
+
+        filter_expression = ' OR '.join([f'node = %{i}' for i in range(len(node_names))])
+
+        # Enclose each node name in "'"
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/parameter_event_handler.py
```diff
@@ -352,6 +352,48 @@ def remove_parameter_event_callback(
...
         self._callbacks.remove_parameter_event_callback(handle)
 
+    def configure_nodes_filter(
+        self,
+        node_names: Optional[List[str]] = None,
+    ) -> bool:
+        """
+        Configure which node parameter events will be received.
+
+        This function depends on middleware support for content filtering.
```
fujitatomoya:
```suggestion
        This function depends on rmw implementation support for content filtering.
```

rclpy/rclpy/parameter_event_handler.py
```diff
@@ -352,6 +352,48 @@ def remove_parameter_event_callback(
...
+        received.
+        It affects the behavior of the following two functions.
+        - add_parameter_event_callback()
+          The callback will only be called for parameter events from the specified nodes which are
+          configured in this function.
+        - add_parameter_callback()
+          The callback will only be called for parameter events from the specified nodes which are
+          configured in this function and add_parameter_callback().
+          If the nodes specified in this function is different from the nodes specified in
+          add_parameter_callback(), the callback will never be called.
```
fujitatomoya, line 381: how about adding the same description in the doc section of add_parameter_callback?

rclpy/rclpy/parameter_event_handler.py
```diff
@@ -352,6 +352,48 @@ def remove_parameter_event_callback(
...
+            return True
+
+        filter_expression = ' OR '.join([f'node = %{i}' for i in range(len(node_names))])
+
+        # Enclose each node name in "'"
+        quoted_node_names = [f"'{node_name}'" for node_name in node_names]
+
+        self.parameter_event_subscription.set_content_filter(filter_expression, quoted_node_names)
+
+        return self.parameter_event_subscription.is_cft_enabled
```
fujitatomoya, line 402: i believe that this can be checked in the very 1st place in this function, and if that is false return immediately without any further processing.

rclpy/rclpy/parameter_event_handler.py
```diff
@@ -352,6 +352,48 @@ def remove_parameter_event_callback(
...
+          add_parameter_callback(), the callback will never be called.
+
+        :param node_names: Node names to filter parameter events from
+
+        :return: True if the filter was successfully applied, False otherwise.
+        """
+        if node_names is None or len(node_names) == 0:
+            # Clear content filter
+            self.parameter_event_subscription.set_content_filter('', [])
+            return True
```
fujitatomoya, line 393: this is against the doc section? even if the rmw implementation does not support cft, this still returns true here. that does not seem to be a design here?

rclpy/rclpy/parameter_event_handler.py
```diff
@@ -352,6 +352,48 @@ def remove_parameter_event_callback(
...
+
+        filter_expression = ' OR '.join([f'node = %{i}' for i in range(len(node_names))])
+
+        # Enclose each node name in "'"
+        quoted_node_names = [f"'{node_name}'" for node_name in node_names]
+
+        self.parameter_event_subscription.set_content_filter(filter_expression, quoted_node_names)
```
fujitatomoya, line 400: i think this possibly throws the exception RCLError, if any error happens to set the filter expression and parameters? it should describe the doc section?

fujitatomoya (changes requested):

rclpy/rclpy/parameter_event_handler.py
Barry-Xu-2018: Updated.

rclpy/rclpy/parameter_event_handler.py
Barry-Xu-2018: Yes. It is better.

rclpy/rclpy/parameter_event_handler.py
Barry-Xu-2018: Here, I should call is_cft_enabled() to check whether CFT setting is still enabled.  
Currently, there is no API available to check whether the RMW implementation supports CFT.  
If RMW Implementation does not support CFT, is_cft_enabled() will return false even if CFT is set. That is, without setting CFT, we cannot know whether the RMW implementation supports CFT or not.

rclpy/rclpy/parameter_event_handler.py
Barry-Xu-2018: Yes. I should add it to the function description.

rclpy/rclpy/parameter_event_handler.py
Barry-Xu-2018: The function `is_cft_enabled` is not used to check whether the RMW Implementation supports CFT.  
When CFT is not set, is_cft_enabled() returns false regardless of whether the RMW Implementation supports CFT or not.  
Only when CFT is set and the RMW Implementation supports CFT does is_cft_enabled() return true.

fujitatomoya (approved):
lgtm with green CI.

@Barry-Xu-2018 same here, can you run CI?

Barry-Xu-2018:
Rebase was done, and I start to run CI.

Barry-Xu-2018:
Pulls: https://github.com/ros2/rclpy/pull/1531
Gist: 
https://gist.githubusercontent.com/Barry-Xu-2018/b7bf6ed7353257611b68a0995631452b/raw/3f03fcd2f8ddebb528bdd5bf863683eff83d4a60/ros2_rolling.repos
BUILD args: --packages-above-and-dependencies rclpy
TEST args: --packages-above rclpy
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17628
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26331/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19979/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6030/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26071/)

Barry-Xu-2018:
CI passed.   
Should I wait for another reviewer?   @fujitatomoya

Issue #1484, opened 2025-07-31
`rclpy` crashes on exit in tests


hidmic:
### Generated by Generative AI

N/A

### Operating System:

```Linux 43219d45bc38 6.14.0-24-generic #24~24.04.3-Ubuntu SMP PREEMPT_DYNAMIC Mon Jul  7 16:39:17 UTC 2 x86_64 x86_64 x86_64 GNU/Linux```

### ROS version or commit hash:

`rolling`

### RMW implementation (if applicable):

N/A

### RMW Configuration (if applicable):

N/A

### Client library (if applicable):

`rclpy`

### 'ros2 doctor --report' output

<details><summary>ros2 doctor --report</summary>

```console

   NETWORK CONFIGURATION
inet         : 127.0.0.1
inet4        : ['127.0.0.1']
inet6        : ['::1']
netmask      : 255.0.0.0
device       : lo
flags        : UP,LOOPBACK,RUNNING
mtu          : 65536
inet         : 172.17.0.2
inet4        : ['172.17.0.2']
ether        : 92:2a:51:6b:1a:f1
netmask      : 255.255.0.0
device       : eth0
flags        : UP,BROADCAST,RUNNING,MULTICAST
mtu          : 1500
broadcast    : 172.17.255.255

   PACKAGE VERSIONS
launch_testing_ros                        : latest=0.29.2, local=0.29.2
ament_cmake_ros                           : latest=0.15.2, local=0.15.2
rmw_test_fixture_implementation           : latest=0.15.2, local=0.15.2
launch_ros                                : latest=0.29.2, local=0.29.2
rclpy                                     : latest=10.0.0, local=10.0.0
rcl_lifecycle                             : latest=10.2.3, local=10.2.3
rcl_action                                : latest=10.2.3, local=10.2.3
rcl                                       : latest=10.2.3, local=10.2.3
rmw_implementation                        : latest=3.1.2, local=3.1.2
rmw_zenoh_cpp                             : latest=0.8.2, local=0.8.2
zenoh_cpp_vendor                          : latest=0.8.2, local=0.8.2
test_msgs                                 : latest=2.4.2, local=2.4.2
std_msgs                                  : latest=5.8.2, local=5.8.2
rosgraph_msgs                             : latest=2.4.2, local=2.4.2
rmw_fastrtps_dynamic_cpp                  : latest=9.4.2, local=9.4.2
rmw_fastrtps_cpp                          : latest=9.4.2, local=9.4.2
[...]

PR #1551, opened 2025-12-02
Avoid taking back GIL when the Python runtime is finalizing

MiguelCompany:
## Description

When using `gil_scoped_release`, care must be taken not to take the GIL when the Python runtime is finalizing.

According to [this](https://github.com/pybind/pybind11/blob/55e4bb9135f9ab19cf9af5ae2178eebb9d9ef23e/include/pybind11/gil.h#L175-L180) one can call `_Py_IsFinalizing()` to check the failure condition and then `disarm()` the `gil_scoped_release` object so it does not take the GIL back.

This would fix (in theroy) #1484 but locally it doesn't seem so. Opening in draft nonetheless, in case someone wants to take this as a base for a proper fix.

### DIFF
--- a/rclpy/src/rclpy/gil_utils.hpp
+++ b/rclpy/src/rclpy/gil_utils.hpp
@@ -1,0 +1,57 @@
+// Copyright 2025 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCLPY__GIL_UTILS_HPP_
+#define RCLPY__GIL_UTILS_HPP_
+
+#include <pybind11/pybind11.h>
+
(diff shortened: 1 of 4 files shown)


rclpy/src/rclpy/gil_utils.hpp
```diff
@@ -0,0 +1,44 @@
...
+{
+
+class gil_scoped_release
+{
+public:
+  gil_scoped_release() = default;
+
+  ~gil_scoped_release() {
+    // Avoid taking the GIL if runtime is finalizing
+    if (_Py_IsFinalizing()) {
```
skyegalaxy: Since the function `_Py_IsFinalizing` [was removed](https://github.com/python/cpython/issues/108014) in Python 3.13 (in favor of a [public method](https://github.com/python/pythoncapi-compat/pull/66/files#diff-0ed658094dc1d83cf8c55d47ab108bf09c64dc07071bd1ca27eba1db7ccf468cR855)) this might require some forward compatibility hackiness. In situations like this there's usually a way to check the python version, e.g. something like 
```
#if PY_VERSION_HEX < Py_PACK_FULL_VERSION(3, 13)) 
    #define IsFinalizing _Py_IsFinalizing
#else
    #define IsFinalizing Py_IsFinalizing
#endif
```

MiguelCompany:
@skyegalaxy Thanks for the review. I applied your suggestion.

Note however that this is still a draft because:

> This would fix (in theroy) #1484 but locally it doesn't seem so. Opening in draft nonetheless, in case someone wants to take this as a base for a proper fix.

MiguelCompany:
Closing in favor of https://github.com/ros2/launch_ros/pull/511 since the problem seems to be in the test, not in rclpy.

PR #1583, opened 2026-01-22
Update release version to 10.0.4 in docs

InvincibleRMC:
## Description

Kind of wacky we are 7 major versions behind on the docs.

Yes all ROS users will stop accidentally installing `3.2.1` on accident.

### DIFF
--- a/rclpy/docs/source/conf.py
+++ b/rclpy/docs/source/conf.py
@@ -42,7 +42,7 @@ author = 'Open Source Robotics Foundation, Inc.'
 # The short X.Y version
 version = ''
 # The full version, including alpha/beta/rc tags
-release = '3.2.1'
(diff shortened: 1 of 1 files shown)


rclpy/docs/source/conf.py
```diff
@@ -42,7 +42,7 @@
 # The short X.Y version
 version = ''
 # The full version, including alpha/beta/rc tags
-release = '3.2.1'
+release = '10.0.4'
```
christophebedard, line 45: I think we need a better/more scalable solution to avoid needing to update the version number here every time, but this is good for now.

christophebedard (approved):

christophebedard:
I think we can just rely on the Rpr job for this one, since it doesn't change code.

Issue #1461, opened 2025-05-16
Feature: asyncio executor


nadavelkabets:
## Motivation

Current wait_set based executors are highly inefficient and don’t integrate cleanly with Python’s `asyncio` ecosystem.  
Examples of inefficiencies in the wait_set executor:
- Reoccurring construction of the wait set  
- Revisiting blocked tasks every iteration  
- Busy looping while a callback group is blocking  

Advantages of integrating with asyncio:
- Better support for asynchronous code – take advantage of many existing features like `call_later()`, `sleep()`, and task exception handling  
- Easy integration with large Python libraries – `FastAPI`, `pyserial-asyncio`, `pytransitions`, `aiohttp`, `python-can`, etc.  
- High performance – `asyncio` is already far more optimized than the existing wait_set executor, and libraries like uvloop are available for even higher performance, matching the current cpp implementation of the EventsExecutor
- Reduction in code duplication - lower maintenance costs and library complexity by transitioning to a built in event loop implementation

## Proposed Implementation
 
- **Shared dispatch logic** 
   Re-use the existing Python executor’s “take_entity” functions to pull items from the queue.

- **Timers via asyncio**  
   Replace the DelayedEventsThread with `loop.call_at()`.

- **AbstractExecutor base class**  
   Define an `AbstractExecutor` (similar to `asyncio.AbstractEventLoop`) to unify typing across `Executor`, `EventsExecutor`, and `AsyncioExecutor`.

## Related Issues & PRs

- #1234 – async wait/sleep implementation  
- #1098 – Tasks that raise an exception crash the entire executor  
- #962  – Runtime error when a long async callback is triggered  
- #1395 – Ignore exceptions raised in user-created tasks  
- #971  – Enable limited use of asyncio with executors

wjwwood:
Going to discuss in the Client Library Working Group.

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/next-client-library-wg-meeting-friday-30th-may-2025-8am-pt/44015/1

jiongtuogenchen:
[Audit Feedback from Cheng Logic Framework]
The memoryview crash under Python 3.13 isn't just a version mismatch; it's a regression in how buffer ownership is managed during asynchronous callbacks.

[...]

PR #1620, opened 2026-03-11
Feature: async node

nadavelkabets:
## Summary

`AsyncNode` brings native `asyncio` support for rclpy, enabling `async`/`await` throughout subscription callbacks, service handlers, client calls, timers, and clock sleeps.

## Usage examples

**`run()` — simple reactive node:**

```python
with rclpy.init():
    node = AsyncNode("my_node")
    node.create_subscription(String, "/topic", callback, 10)
    await node.run()
```

**`async with` — composable, user-controlled lifetime:**

```python
with rclpy.init():
    async with AsyncNode("my_node") as node:
        node.create_subscription(String, "/topic", callback, 10)
        await some_async_function()
```

**`client.call()` — async service call, no futures or spinning:**

```python
async with AsyncNode("my_node") as node:
    client = node.create_client(SetBool, "/set_bool")
    response = await client.call(SetBool.Request(data=True))
```

**`clock.sleep()` — sim-time-aware, cancels on node shutdown:**

```python
async with AsyncNode("my_node") as node:
    await node.get_clock().sleep(2.0)  # respects sim time
```

**aiohttp — compose with any asyncio library in a callback:**

```python
async def handle_set(request, response):
    async with aiohttp.request("GET", "http://api.example.com/data") as resp:
        response.message = await resp.text()
    return response

node = AsyncNode("my_node")
node.create_service(SetBool, "/my_service", handle_set)
await node.run()
```

**serial — bridge ROS topics to a serial port:**

```python
reader, writer = await serial_asyncio.open_serial_connection(url="/dev/ttyUSB0")
async with AsyncNode("serial_bridge") as node:
    pub = node.create_publisher(Byte, "/serial/rx", 10)
    while data := await reader.read(1):
        pub.publish(Byte(data=data))
```

### Core mechanism

DDS pushes work onto the asyncio event loop instead of an executor polling a wait set:

```
DDS thread → on_new_message_callback → loop.call_soon_threadsafe(event.set) → asyncio executes
```

One reader task per entity waits on an asyncio Event, takes data, and runs the callback. The DDS callback only sets the event. This gives natural backpressure — the entity won't take another message until the current callback yields.

### Structured concurrency

The design follows structured concurrency (Trio, asyncio TaskGroup). Every task has a clear owner, lifetimes are bounded by `async with` scopes, and no task outlives its parent.

The node's outer TaskGroup owns entity reader tasks. Subscriptions and services can run callbacks concurrently via a nested inner TaskGroup. When an entity is destroyed, the inner group cancels all in-flight callbacks before the entity handle is cleaned up — no orphaned callbacks. This is required for service correctness (a callback needs the live service handle to send its response), and subscriptions use the same pattern for consistency. Clients and timers don't need inner groups — clients only route responses to futures, timers dispatch sequentially.

Resource cleanup is deterministic: when the `async with` block exits, all reader tasks have finished, all DDS callbacks have been cleared, and no orphaned coroutines remain.

### Two entry points

Both entry points share the same lifecycle:
- `async with` is for composable use cases where the user controls the lifetime (bridges, tests, multi-protocol applications).
- `run()` is for simple reactive nodes.

### Class hierarchy

```
BaseNode
├── Node                    (executor-based, existing)
└── AsyncNode               (asyncio-based, new)

BasePublisher
├── Publisher               (executor: callback groups, executor event)
└── AsyncPublisher          (no task — fire-and-forget publish)

BaseSubscription
├── Subscription            (executor: callback groups, executor event)
└── AsyncSubscription       (owns reader task, optional concurrent dispatch)

BaseService
├── Service                 (executor: callback groups, executor event)
└── AsyncService            (owns reader task, optional concurrent dispatch)

BaseClient
├── Client                  (executor: callback groups, call_async → rclpy.Future)
└── AsyncClient             (owns reader task, call → asyncio.Future)

BaseTimer
├── Timer                   (executor: callback groups, executor event)
└── AsyncTimer              (owns timer loop, sequential dispatch)

BaseClock
├── Clock                   (blocking sleep)
└── AsyncClock              (async sleep)
```

### Entity-owned architecture

Each async entity class owns its full DDS bridge: event creation, DDS callback registration, the take loop, callback dispatch, and cleanup. The node is a thin coordinator — it creates C handles via the base class, wraps them in async entity classes, and hands them a reference to its TaskGroup so they can spawn their own reader tasks.

BaseNode holds shared logic (parameters, clock, logger, name resolution, graph discovery) and calls factory methods polymorphically during init, so AsyncNode and Node each produce their own entity types without the base class knowing which subclass it's in.

Entities self-remove from the node's tracking set when destroyed, via a callback passed at construction. This keeps the node's bookkeeping consistent regardless of whether destruction is triggered by the node, by user code, or by the entity's own cleanup.

### Shutdown

Shutdown is synchronous and uses task cancellation as the only mechanism. Destroying the node destroys all entities (cancelling their tasks and DDS handles), cancels pending clock sleeps, then marks the node handle for deferred destruction. Each entity's destroy is idempotent.

### Clock sleep and timers

Clock sleep is the async replacement for blocking sleep. For wall time it schedules a delayed callback on the event loop, for sim time it registers a jump callback that resolves when simulated time advances past the target. If the time source changes during a sleep (ROS time activated or deactivated), the sleep raises an error since the target is no longer meaningful. All pending sleeps are cancelled on node shutdown.

Timers use the same dual-mode wait pattern. They support cancel (parks the loop until reset) and reset (wakes the parked loop). Timer callbacks are always dispatched sequentially — if a callback runs longer than the period, the next tick is delayed.

## Known limitations
Actions and waitables are not yet supported. Waitable support requires a `set_on_ready_callback` API on the waitable interface, matching the approach used by rclcpp’s EventsExecutor, which is not yet available in rclpy.

### CPU usage
I ran the `test_rclpy_performance.py` benchmark from the EventsExecutor PR adapted for the AsyncNode.
AsyncNode nearly matches EventsExecutor performance while significantly outperforming SingleThreadedExecutor.
Screenshot 2026-03-12 at 01 51 31
Running the test with `uvloop.run` instead of `asyncio.run` achieved an even lower cpu usage of 10%.

### Timer latency and jitter
At 50 Hz (20 ms period), AsyncNode's mean jitter is slightly higher than the existing executors, at ~0.16 ms above SingleThreadedExecutor and ~0.38 ms above EventsExecutor. 
I didn't invest much in trying to optimize this, honestly the difference is really small.
Screenshot 2026-03-12 at 17 43 11

## Related work

- **#1399**
- **#1461**


peci1:
This looks quite interesting! I haven't acquainted myself with async patterns too much yet, so I won't be of much value as a reviewer of the whole PR.

On thing has, however, caught my eye: always subscribing to clock.

This is something I wouldn't recommend. I sometimes deliberately don't set sim time on a node if I know it doesn't need it (e.g. a plain relay that just gets a message, computes something and outputs another message with the input timestamp). The reason why I do it is performance: if you have a 1000 Hz sim clock and 50 nodes... You should use ROS 1 :D So I definitely don't want to lose the ability to not subscribe clock.

Another thing: is it possible to specify service call timeouts with the async client?

nadavelkabets:
> 1000 Hz sim clock

The existing SingleThreadedExecutor would probably fry your cpu if you published at 1000hz haha.
I reverted this change.

> Another thing: is it possible to specify service call timeouts with the async client?

Yeah you could.
Notice that there are 2 separate mechanisms:
- An asyncio.timeout() context manager cancels the async call operation.
- Lifespan QoS setting drops the request at the middleware layer.
```python
async with AsyncNode("node") as node:
    qos_profile = QoSProfile(lifespan=Duration(seconds=2))
    client = node.create_client(SetBool, "/set_bool", qos_profile=qos_profile)
    with asyncio.timeout(2):  
        response = await client.call(SetBool.Request(data=True))
```
That's a limitation of the current client rcl/rmw api.

mjcarroll:
As we have discussed before, this is a very high quality contribution.  It brings our interactions in the Python client library in line with what many Python developers would expect.  

I agree with the refactoring into Base classes.  We wouldn't want to inherit from what is already there because it could cause confusion. This is a good move.  

Otherwise the API ergonomics look great, and putting it in the experimental namespace should give us a little bit of latitude on getting improvements in over time.

With that, I would suggest we do this in ~6 PRs.  You mentioned 12 on zulip, but I feel like that would be a lot to manage.

I think (just guessing here):

* Base Entity Classes (should be quick to review)
* Moving shared logic into BaseNode
* Clock and time
* Async infrastructure (utilities/event loop)
* Add the async entities and asyncnode objects
* Tests and documentation

I think those are logical chunks that should each be pretty easy to review.

nadavelkabets (commented):
Reminder - test AsyncNode with python 3.14

christophebedard:
Pulls: ros2/rclpy#1620
Gist: https://gist.githubusercontent.com/christophebedard/c8dc171fed3cf6b2d1b15a5670655752/raw/ef85a1eead83cacf130a0167098300798165f9c8/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19010
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28775/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21593/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8616/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27653/)

christophebedard:
CI for https://github.com/ros2/rclpy/pull/1620 with:

1. Linux dep from https://github.com/ros2/ci/pull/868 (using https://github.com/ros2/ci/tree/feature/add-pytest-asyncio-dep)
2. Windows dep from https://github.com/ros2/ros2/pull/1818
3. RHEL 10 (not the current default, RHEL 9)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28785/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21603/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8629/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27666/)

christophebedard:
I retriggered the RHEL job with RHEL 10 instead of RHEL 9.

christophebedard:
Windows is complaining: https://ci.ros2.org/job/ci_windows/27657/console

```
Error: 
  × failed to solve the conda requirements of 'default' 'win-64'
  ╰─▶ Cannot solve the request because of: pytest-asyncio ==1.3.0 cannot be
      installed because there are no viable options:
      └─ pytest-asyncio 1.3.0 would require
         └─ pytest >=8.2,<10, for which no candidates were found.
```

Also, seems like RHEL 10 is currently broken due to Qt6: https://ci.ros2.org/job/ci_linux-rhel/8627/console

```
Error: Unable to find a match: python3-qt6-devel
```

christophebedard:
The RHEL issue is being addressed in https://github.com/ros2/ci/pull/869

christophebedard:
I re-triggered the Windows job above with the updated `pixi.toml` from https://github.com/ros2/ros2/pull/1818

christophebedard:
aaand I re-triggered the RHEL job above after rebasing the ros2/ci branch on `master` after https://github.com/ros2/ci/pull/869 was merged.

rclpy/rclpy/experimental/asyncio/async_service.py
```diff
@@ -0,0 +1,97 @@
...
+        """DDS bridge read loop for services."""
+        self._loop = asyncio.get_running_loop()
+        try:
+            if self._concurrent:
+                async with asyncio.TaskGroup() as tg:
+                    async for request, header in self._requests():
+                        tg.create_task(await_or_execute(self._handle_request, request, header))
+            else:
+                async for request, header in self._requests():
+                    await await_or_execute(self._handle_request, request, header)
```
nadavelkabets: The await_or_execute calls here are not necessary

christophebedard:
The Windows build failure should hopefully be resolved once https://github.com/ros2/rmw_connextdds/pull/225 gets merged

rclpy/rclpy/experimental/asyncio/async_timer.py
```diff
@@ -0,0 +1,167 @@
...
+        clock: AsyncClock,
+        context: Context,
+        callback: TimerCallbackType,
+        autostart: bool,
+        on_destroy: Callable[['AsyncTimer'], None],
+        tg: Optional[asyncio.TaskGroup] = None,
+    ) -> None:
+        super().__init__(callback, timer_period_ns, clock, context=context,
+                         on_destroy=on_destroy, autostart=autostart)
+        self._pass_info = self._detect_wants_info(callback)
```
nadavelkabets: Might fail if users set timer.callback to a callable with the opposite bool value

mjcarroll (approved):

mjcarroll:
Overall LGTM.  I think we need tutorials and documentation for folks who plan on trying this feature, but I'm not going to block on it here (or for the freeze)

rclpy/rclpy/experimental/async_clock.py
```diff
@@ -0,0 +1,107 @@
...
+                    self._resolve_future(future)
+
+    def _destroy(self) -> None:
+        """Cancel all pending sleeps. Called by AsyncNode.destroy_node()."""
+        if self._destroyed:
+            return
+        self._destroyed = True
+        self._jump_handle.unregister()
+        for future in self._pending_sleeps:
+            future.cancel()
```
mjcarroll, line 73: I think this is technically mutating self._pending_sleeeps while iterating it because the `finally` will pop the future.  You should make a copy of the keys?

rclpy/rclpy/experimental/asyncio/async_timer.py
```diff
@@ -0,0 +1,167 @@
...
+        if self._jump_handle is not None:
+            self._jump_handle.unregister()
+        self.handle.clear_on_reset_callback()
+        super()._destroy()
+
+    def cancel(self) -> None:
+        super().cancel()
+        self._resolve()
+
+    def _on_reset(self, _count: int):
```
mjcarroll: Missing return type

rclpy/rclpy/experimental/asyncio/async_clock.py
```diff
@@ -0,0 +1,107 @@
...
+    async def sleep(self, duration_sec: float) -> None:
+        """
+        Sleep for a duration respecting sim time.
+
+        Cancelled on clock destruction. Raises TimeSourceChangedError if ROS
+        time is activated or deactivated during the sleep.
+        """
+        if self._destroyed:
+            raise RuntimeError('Cannot sleep on a destroyed clock')
+        duration_ns = int(duration_sec * 1e9)
```
mjcarroll: Could use `rclpy.duration.Duration`

rclpy/rclpy/experimental/async_clock.py
nadavelkabets: I believe this should be safe as is, but we could make a defensive copy of the keys.
* asyncio is strictly single threaded
* there are no yield points during the iteration
* canceling a future does not execute any code except for scheduling tasks to run at the next yield point

Any client.call() coroutines should only resume after finishing iteration and exiting destroy()
Do you feel comfortable with that, or would you prefer the defensive action?

rclpy/rclpy/experimental/asyncio/async_node.py
```diff
@@ -0,0 +1,455 @@
...
+            This flag affects the behavior of parameter-related operations.
+        :param automatically_declare_parameters_from_overrides: If True, the "parameter overrides"
+            will be used to implicitly declare parameters on the node during creation.
+        :param enable_logger_service: ``True`` if ROS2 services are created to allow external nodes
+            to get and set logger levels of this node. Otherwise, logger levels are only managed
+            locally. That is, logger levels cannot be changed remotely.
+        """
+        self._clock = AsyncClock(clock_type=ClockType.ROS_TIME)
+        self._tg: Optional[asyncio.TaskGroup] = None
+        self._entities: Set[AsyncEntity] = set()
```
mjcarroll: super minor nit, I think you can use just `set` here and not import the type anymore, right?

rclpy/rclpy/experimental/async_clock.py
mjcarroll: I think we are fine as-is.

christophebedard:
@nadavelkabets once you're done, let me know and I can trigger another full CI run. And I can keep an eye on it and hopefully merge it later once it's green (or green enough)

mjcarroll (approved):

mjcarroll (approved):

christophebedard:
CI for https://github.com/ros2/rclpy/pull/1620 for `rclpy` and above with:

1. Linux dep from https://github.com/ros2/ci/pull/868 (using https://github.com/ros2/ci/tree/feature/add-pytest-asyncio-dep)
2. Windows dep from https://github.com/ros2/ros2/pull/1818
3. (RHEL 9 like the current default, not RHEL 10)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28796/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21614/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8640/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27669/)

sloretz:
Looks like some issues:

* RHEL: 100 test failures seem to be due to pytest-asyncio being extra spammy about it's on config, even in unrelated tests https://github.com/pytest-dev/pytest-asyncio/issues/293
* Linux amd64 and aarch64: 14 test failures, `pytest.PytestUnknownMarkWarning: Unknown pytest.mark.asyncio` (CI issue? I don't see `python3-pytest-asyncio` in the logs, so did it use the wrong CI script branch?)
* Windows: still waiting for results

christophebedard:
Yes, I forgot to add the ros2/ci branch

christophebedard:
Windows passed with a single test failure before.

christophebedard:
And ci_linux passed before the last minor changes

christophebedard:
> Windows passed with a single test failure before.

And it now failed with the same single test failure. Nadav says this might be a Windows clock accuracy thing. I think this should be fixable in a follow-up PR

sloretz:
CI  - just linux and rhel, and just testing rclpy since those test are the only ones that failed, and they appear to have failed for a CI issue that should be fixed by https://github.com/ros2/ci/pull/871 . I assume the above windows results will still be valid and don't need a re-run

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/28811/)
* Linux aarch 64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/21616/)
* Linux RHEL (9) [Build Status](https://ci.ros2.org/job/ci_linux-rhel/8654/)

christophebedard:
@sloretz are you able to keep an eye on CI and merge if it looks good (enough)? It’s 1:30 am here 😴 😅

nadavelkabets:
> CI  - just linux and rhel, and just testing rclpy since those test are the only ones that failed, and they appear to have failed for a CI issue that should be fixed by https://github.com/ros2/ci/pull/871 . I assume the above windows results will still be valid and don't need a re-run
> 
> 
> 
> * Linux [Build Status](https://ci.ros2.org/job/ci_linux/28798/)
> 
> * Linux aarch 64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/21616/)
> 
> * Linux RHEL (9) [Build Status](https://ci.ros2.org/job/ci_linux-rhel/8642/)

Do we need to restart Linux and rhel?

christophebedard:
I updated the comment

christophebedard:
aarch64 looks good, and now RHEL also looks good (no complaints from pytest). ci_linux (amd64) is stuck in the queue, but I think aarch64 should be fairly representative of it, so I’m calling it good and merging

PR #1624, opened 2026-03-14
Refactor: base entity classes

nadavelkabets:
Part of #1620 

## Changes
* New BasePublisher, BaseService, BaseSubscription, BaseClient and BaseTimer classes
* `__init__` parameter changes
    * Subscription, Publisher: `callback_group` and `event_callbacks` are now keyword only
    * Service, Client, Timer: `callback_group` is now keyword only
    * Timer: callback_group defaults to None

## Notes
* Callback groups and event handlers were not moved to base classes
    * Callback groups are exclusive to executors and would not be supported in AsyncNode. More detailed discussion is in #1399.
    * We have no callback mechanism for waitables yet, so AsyncNode won't support event handlers for now.
* Service's send_response() is used publicly only in executor, so it was made protected in the BaseService

### DIFF
--- a/rclpy/rclpy/publisher.py
+++ b/rclpy/rclpy/publisher.py
@@ -26,7 +26,7 @@ from rclpy.type_support import MsgT
 MsgType = TypeVar('MsgType')
 
 
-class Publisher(Generic[MsgT]):
+class BasePublisher(Generic[MsgT]):
 
     def __init__(
         self,
@@ -34,30 +34,11 @@ class Publisher(Generic[MsgT]):
         msg_type: Type[MsgT],
         topic: str,
         qos_profile: QoSProfile,
-        event_callbacks: PublisherEventCallbacks,
-        callback_group: CallbackGroup,
     ) -> None:
-        """
-        Create a container for a ROS publisher.
-
-        .. warning:: Users should not create a publisher with this constructor, instead they should
-           call :meth:`.Node.create_publisher`.
-
-        A publisher is used as a primary means of communication in a ROS system by publishing
-        messages on a ROS topic.
-
-        :param publisher_impl: Publisher wrapping the underlying ``rcl_publisher_t`` object.
-        :param msg_type: The type of ROS messages the publisher will publish.
-        :param topic: The name of the topic the publisher will publish to.
-        :param qos_profile: The quality of service profile to apply to the publisher.
-        """
         self.__publisher = publisher_impl
         self.msg_type = msg_type
         self.topic = topic
         self.qos_profile = qos_profile
-
-        self.event_handlers: List[EventHandler] = event_callbacks.create_event_handlers(
-            callback_group, publisher_impl, topic)
 
     def publish(self, msg: Union[MsgT, bytes]) -> None:
         """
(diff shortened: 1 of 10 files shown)


rclpy/rclpy/subscription.py
```diff
@@ -88,11 +91,9 @@ def __init__(
          subscription_impl: '_rclpy.Subscription[MsgT]',
          msg_type: Type[MsgT],
          topic: str,
-         callback: GenericSubscriptionCallback[MsgT],
-         callback_group: CallbackGroup,
+         callback: GenericSubscriptionCallbackUnion[MsgT],
```
mjcarroll, line 94: Just a note that if someone isn't using `create_subscription`, this may end up being a problem.  In general, people should _not_ be doing this, but we should include it in the changelog.

rclpy/rclpy/timer.py
```diff
@@ -66,54 +66,31 @@ def actual_call_time(self) -> Time:
...
 
 
-class Timer:
+class BaseTimer:
 
     def __init__(
         self,
-        callback: Optional[TimerCallbackType],
-        callback_group: Optional[CallbackGroup],
+        callback: Optional[TimerCallbackUnion],
```
mjcarroll, line 79: Same note at the subscription above.  Directly instancing timers will break.

rclpy/rclpy/publisher.py
```diff
@@ -116,6 +89,53 @@ def assert_liveliness(self) -> None:
...
+
+class Publisher(BasePublisher[MsgT], Generic[MsgT]):
+
+    def __init__(
+        self,
+        publisher_impl: '_rclpy.Publisher[MsgT]',
+        msg_type: Type[MsgT],
+        topic: str,
+        qos_profile: QoSProfile,
+        event_callbacks: PublisherEventCallbacks,
```
mjcarroll, line 113: For this (and each of the other entities) we should make the positional arguments match the base.  Anything beyond we should mark as keyword only arguments to make it explicit.

Mostly this will be so that Subscription and AsyncSubscription (and the others) keep the same indicies between them.

We should do this for Timer, Subscription, Client, etc.

rclpy/rclpy/subscription.py
nadavelkabets: The existing docstring already covers this:
```
.. warning:: Users should not create a subscription with this constructor, instead they
   should call :meth:`.Node.create_subscription`.
```
I added the parameter reorder to the PR description. 
Is there anywhere else you'd like it mentioned?

rclpy/rclpy/publisher.py
nadavelkabets: Done. All entity `__init__` signatures now match their base class, with the remaining parameters as keyword only.

rclpy/rclpy/subscription.py
mjcarroll: Ah, you are correct, I think we are good here, then.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28068/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21038/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7917/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27009/)

PR #1643, opened 2026-04-07
Bugfix: executor doesn't propagate exception from task that awaited a future

### LINKED ISSUE #1642: SingleThreadedExecutor silently swallows exception when coroutine raises after awaiting a Future
This bug was introduced on #1469 

## Environment
ROS2 Rolling (should also affect backports to Kilted and Jazzy)
Ubuntu 24.04 (Noble), aarch64
Python 3.12.3

## Description
When an async Task awaits a Future and raises an exception upon resumption, SingleThreadedExecutor silently swallows the exception. The coroutine's error never propagates to spin_until_future_complete or spin.

## Code to reproduce
The test will fail on rolling.
Test was passing on commit 0ce60eb before #1469.
```python
def test_coroutine_exception_after_await(self) -> None:
    """Exception in a coroutine after awaiting a future must propagate."""
    self.assertIsNotNone(self.node.handle)
    for cls in [SingleThreadedExecutor]:
        with self.subTest(cls=cls):
            executor = cls(context=self.context)
            executor.add_node(self.node)

            first_fut = executor.create_future()
            second_fut = executor.create_future()

            async def coro_that_raises() -> None:
                first_fut.set_result(None)
                await second_fut
                raise RuntimeError('Expected error after await')

            task = executor.create_task(coro_that_raises)

            executor.spin_until_future_complete(first_fut, timeout_sec=5)
            self.assertFalse(task.done())
            # Resolve the inner future — triggers resume
            second_fut.set_result(None)

            # The resumed coroutine should raise, and spin must propagate it
            with self.assertRaises(RuntimeError) as cm:
                executor.spin_until_future_complete(task, timeout_sec=5)
            self.assertIn('Expected error after await', str(cm.exception))
```

## Root Cause
When a Task awaits a Future, it reschedules itself to resume when the future is done
via `Future.add_done_callback`. When scheduling done callbacks, a new task is created for each callback. This causes the original resumed task to be wrapped with a new task that the executor runs.
The wrapper task is checked for raised exceptions, while those do not propagate from the original task to the wrapper done_callback task.

nadavelkabets:
Fixes #1642 

### Changes
A task now registers itself on the awaited future via new `Future._add_waiting_task`.
The executor then dispatches the same `Task` object and sees its exception through `handler.exception()` as intended.

- `Future._callbacks` now holds `Union[Callable, Task]`
-  `_schedule_or_invoke_done_callbacks` handles `Task` entries through `executor._call_task_in_next_spin`.
- `_wait_for_ready_callbacks` skips tasks that were cancelled or completed between being queued and being popped.
- A `warnings.warn` fires if a waiting `Task` is dropped because the executor weakref could not be resolved.

### DIFF
--- a/rclpy/rclpy/task.py
+++ b/rclpy/rclpy/task.py
@@ -172,6 +172,9 @@ class Future(Generic[T]):
             # No executor, call right away
             for callback in callbacks:
                 if isinstance(callback, Task):
+                    warnings.warn(
+                        'Dropping task awaiting future: '
+                        'executor reference could not be resolved')
                     continue
                 try:
                     callback(self)
@@ -222,7 +225,10 @@ class Future(Generic[T]):
                 executor = self._executor()
                 if executor is not None:
                     executor._call_task_in_next_spin(task)
-
+                else:
+                    warnings.warn(
+                        'Dropping task awaiting future: '
+                        'executor reference could not be resolved')
             else:
                 self._callbacks.append(task)
 
@@ -370,6 +376,7 @@ class Task(Future[T]):
         elif future_executor is not executor:
             raise RuntimeError('A task can only await futures associated with the same executor')
 
+        # Register the task to resume when the future is done or cancelled
         future._add_waiting_task(self)
 
     def _complete_task(self) -> None:
--- a/rclpy/rclpy/task.py
+++ b/rclpy/rclpy/task.py
@@ -50,8 +50,8 @@ class Future(Generic[T]):
         # An exception raised by the handler when called
         self._exception: Optional[Exception] = None
         self._exception_fetched = False
-        # callbacks to be scheduled after this task completes
-        self._callbacks: List[Callable[['Future[T]'], None]] = []
+        # callbacks or tasks to be scheduled after this task completes
+        self._callbacks: List[Union[Callable[['Future[T]'], None], 'Task[Any]']] = []
         # Lock for threadsafety
         self._lock = threading.Lock()
         # An executor to use when scheduling done callbacks
@@ -164,10 +164,15 @@ class Future(Generic[T]):
         if executor is not None:
             # Have the executor take care of the callbacks
             for callback in callbacks:
-                executor.create_task(callback, self)
+                if isinstance(callback, Task):
+                    executor._call_task_in_next_spin(callback)
+                else:
+                    executor.create_task(callback, self)
         else:
             # No executor, call right away
             for callback in callbacks:
+                if isinstance(callback, Task):
+                    continue
                 try:
                     callback(self)
                 except Exception as e:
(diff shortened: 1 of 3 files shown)


rclpy/rclpy/task.py
```diff
@@ -209,6 +217,21 @@ def add_done_callback(self, callback: Callable[['Future[T]'], None]) -> None:
...
+        with self._lock:
+            if not self._pending():
+                assert self._executor is not None
+                executor = self._executor()
+                if executor is not None:
+                    executor._call_task_in_next_spin(task)
+                else:
+                    warnings.warn(
+                        'Dropping task awaiting future: '
+                        'executor reference could not be resolved')
```
fujitatomoya, line 231: inconsistency is that the "no executor weakref at all" case gets an assert, while the "executor was GC'd" case gets a graceful warning, yet both represent the same fundamental situation i think...

how about the following? this way a missing executor always degrades to a warning rather than a crash, which is consistent with how the _schedule_or_invoke_done_callbacks method already handles the same situation in its own else branch?

```suggestion
            if not self._pending():
                executor = self._executor() if self._executor is not None else None
                if executor is not None:
                    executor._call_task_in_next_spin(task)
                else:
                    warnings.warn(
                        'Dropping task awaiting future: '
                        'executor reference could not be resolved')
```

fujitatomoya (commented):
i think that the approach of preserving task identity rather than wrapping in a new task is the right call. almost lgtm, with just one minor suggestion.

rclpy/rclpy/task.py
nadavelkabets: I debated that and went for mirroring the existing pattern from `_schedule_or_invoke_done_callbacks`.
```python
    def _schedule_or_invoke_done_callbacks(self) -> None:
        with self._lock:
            assert self._executor is not None
            executor = self._executor()
            callbacks = self._callbacks
```
For some reason `self._executor` is typed Optional while it's never actually None and set at `__init__`. 
I thus assumed this assert was either for type checking only or to strictly make sure that it's never actually None at runtime.

fujitatomoya: i see, thanks for sharing that. let's leave it there, since this is not really related to this specific change.

fujitatomoya (approved):
lgtm with green CI.

fujitatomoya:
Pulls: ros2/rclpy#1643
Gist: https://gist.githubusercontent.com/fujitatomoya/304d03354c28f17e75b36dec7465ed9c/raw/364b136f9e5a7268dd0d7017890a230f6bb89273/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18897
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28638/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21465/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8470/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27500/)

nadavelkabets:
@fujitatomoya could you run windows again?

fujitatomoya:
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21479/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27525/)

mjcarroll (approved):

mjcarroll:
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21571/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27620/)

mjcarroll:
@Mergifyio backport kilted jazzy humble

PR #1616, opened 2026-03-02
Fix incorrect action client/server callback type hints

bjsowa:
## Description
This fixes some incorrect type hints in Action Client and Server callbacks.
1. `feedback_callback` argument changed from `FeedbackT` to `FeedbackMessage[FeedbackT]`
2. `cancel_callback` argument changed from `GoalT` to `ServerGoalHandle`
3. All callback type hints now support passing a coroutine function instead of a normal function.

### Is this user-facing behavior change?

The user no longer has to add `# type: ignore` comments if using a static type checker

PR #1654, opened 2026-04-20
[kilted] Fix incorrect action client feedback callback type hint

laser:
## What Problem Are We Trying to Solve?

- `ActionClient.send_goal_async()` declares `feedback_callback` as `Callable[[FeedbackT], None]`, but the runtime dispatcher in the same file invokes the registered callback with a `FeedbackMessage[FeedbackT]` (the wrapper with `goal_id` and `feedback` fields), not a bare `FeedbackT`.
- Consumers typing their callback correctly according to the declared signature cannot reach the feedback payload. The type hint is simply wrong.
- This was fixed on rolling by #1616 (commit f735d59) but was never backported to kilted. kilted users still see the incorrect hint.

## How Did We Solve the Problem?

- Backport the client-side portion of #1616 to kilted.
- Introduce a `FeedbackCallbackUnion` `TypeAlias` in the `TYPE_CHECKING` block, move `SendGoalKWargs` into the same block as `Generic[FeedbackT]`, and use the new alias in `_feedback_callbacks`, `send_goal`, and `send_goal_async`.
- Update `test_action_client.py` to type `feedback_callback` with `FeedbackMessage[Fibonacci.Feedback]`, documenting the corrected signature.
- Server-side changes from #1616 are not included: they depend on the `ImplT` generic parameter introduced after kilted.
- Pure type-hint correction; no runtime behavior change.

### DIFF
--- a/rclpy/rclpy/action/client.py
+++ b/rclpy/rclpy/action/client.py
@@ -17,6 +17,7 @@ import time
 from types import TracebackType
 from typing import Any
 from typing import Callable
+from typing import Coroutine
 from typing import Dict
 from typing import Generic
 from typing import Optional
@@ -25,6 +26,7 @@ from typing import Type
 from typing import TYPE_CHECKING
 from typing import TypedDict
 from typing import TypeVar
+from typing import Union
 
 import uuid
 import weakref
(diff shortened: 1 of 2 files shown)


bjsowa:
@laser Since I was planning to do this either way, I pushed the complete backport of #1616 to #1655 so I think this can be closed.

laser:
> @laser Since I was planning to do this either way, I pushed the complete backport of #1616 to #1655 so I think this can be closed.

Sounds good. My dreams of landing a commit in rclpy have been dashed, alas, but I'm pleased with the outcome nonetheless :)

bjsowa:
@laser I added you as a co-author since you technically helped me by reminding to do this.

PR #1660, opened 2026-05-03
Fix Windows failures in `test_sleep_multiple_concurrent_waiters`

nadavelkabets:
## Summary
Fixes test from #1620
Asyncio's clock (monotonic) has 15ms accuracy on windows, causing this test to fail.
The failure is due to asyncio's behavior of waking timer calls one clock tick early, failing the assert in the test.

## Test failure
https://ci.ros2.org/view/nightly/job/nightly_win_rel/lastCompletedBuild/testReport/rclpy.rclpy.test/test_async_clock/test_sleep_multiple_concurrent_waiters/
```
+ Exception Group Traceback (most recent call last):
  |   File "C:\pixi_ws\.pixi\envs\default\Lib\site-packages\_pytest\runner.py", line 341, in from_call
  |     result: Optional[TResult] = func()
  |                                 ^^^^^^
  |   File "C:\pixi_ws\.pixi\envs\default\Lib\site-packages\_pytest\runner.py", line 262, in <lambda>
  |     lambda: ihook(item=item, **kwds), when=when, reraise=reraise
  |             ^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "C:\pixi_ws\.pixi\envs\default\Lib\site-packages\pluggy\_hooks.py", line 501, in __call__
  |     return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "C:\pixi_ws\.pixi\envs\default\Lib\site-packages\pluggy\_manager.py", line 119, in _hookexec
  |     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "C:\pixi_ws\.pixi\envs\default\Lib\site-packages\pluggy\_callers.py", line 181, in _multicall
  |     return outcome.get_result()
  |            ^^^^^^^^^^^^^^^^^^^^
  |   File "C:\pixi_ws\.pixi\envs\default\Lib\site-packages\pluggy\_result.py", line 99, in get_result
  |     raise exc.with_traceback(exc.__traceback__)
[...]

PR #1661, opened 2026-05-13
Fix rclpy async executor sleep.

clalancette:
## Description

The python asyncio sleep allows for a bit of clock resolution sloppiness (see https://github.com/python/cpython/blob/50476a7e87926cf6a9ea978943c4a3d5d771c95f/Lib/asyncio/base_events.py#L2036) which means that the asyncio sleep can actually wake up *before* the requested amount of time has passed. This is mostly a problem on Windows, where the tick time is 15ms, and we see some tests in rclpy fail a number of times there before eventually succeeding.

There are 2 ways we can fix this: either by making the tests accept some slop, or by making the rclpy AsyncClock sleep go back to sleep if it wakes up early.  The former is more like Python's asyncio, the latter is more like the rclcpp RCL_TIME sleep.

I chose to go with the latter because this is a robotics framework and waking up too early is as bad as waking up too late.  But everyone has to deal with the latter situation *anyway*, so it seems more predictable to always sleep at least the requested amount of time.  That's what this commit does.

No.

Yes, Claude Opus 4.7

### Additional Information

N/A

### DIFF
--- a/rclpy/rclpy/experimental/async_clock.py
+++ b/rclpy/rclpy/experimental/async_clock.py
@@ -88,20 +88,35 @@ class AsyncClock(BaseClock):
             return
 
         loop = asyncio.get_running_loop()
-        future: asyncio.Future[None] = loop.create_future()
-        timer_handle: Optional[asyncio.TimerHandle] = None
-        target: Optional[Time] = None
 
         if self.ros_time_is_active:
-            target = self.now() + duration
-        else:
-            timer_handle = loop.call_later(
-                duration_sec, AsyncClock._resolve_future, future)
+            # _on_jump resolves the future only when self.now() >= target,
+            # so this branch already provides "at least duration" semantics.
+            future: asyncio.Future[None] = loop.create_future()
+            target: Time = self.now() + duration
+            self._pending_sleeps[future] = target
+            try:
+                await future
+            finally:
+                self._pending_sleeps.pop(future)
+            return
 
-        self._pending_sleeps[future] = target
-        try:
-            await future
-        finally:
-            self._pending_sleeps.pop(future)
-            if timer_handle is not None:
+        # asyncio.call_later is allowed to fire up to loop._clock_resolution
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29067/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21861/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8934/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27981/)

fujitatomoya (approved):
lgtm.

note that https://github.com/ros2/rclpy/pull/1660 tries to mitigate the windows test failures from the same root cause. should we keep https://github.com/ros2/rclpy/pull/1660 with this PR, or this PR should fix them up? IMO, that would be okay to keep the slop in the test. what do you think?

CC: @nadavelkabets

nadavelkabets:
I'll look into this tonight when I get home, that's an interesting question I debated when writing the async timer.

clalancette:
> note that #1660 tries to mitigate the windows test failures from the same root cause. should we keep #1660 with this PR, or this PR should fix them up? IMO, that would be okay to keep the slop in the test. what do you think?

I'm sorry, I should have looked for open PRs before opening this, and I missed #1660 .  My personal opinion is that we should decide whether we want to allow the wakeup sloppiness or not, and based on that decision, take one or the other.  Taking both is not necessarily a bad thing, but it does mean that we aren't testing this particular invariant (either allow early wakeups, or never allow early wakeups).

ahcorde:
@clalancette is this something that we should backport ?

clalancette:
> @clalancette is this something that we should backport ?

Yeah, this would be a good one to get into Lyrical.

clalancette:
@Mergifyio backport lyrical

PR #1662, opened 2026-05-14
Fix a race in the executors.

clalancette:
## Description

This is complicated to explain, but in the MultiThreadedExecutor spin_once_impl hands a task to a worker and adds it to the list of futures.  It then goes into a loop looking to finish/reap futures, but the one we just added is almost never done at this point.

However, if it completes fast enough to beat the outer-loop check in spin_until_future_complete, then the future will be done, we'll exit the loop without calling spin_once_impl again, and we'll never raise.

The fix is simple; after we've broken out of the outer loop in spin_until_future_complete, check one more time to see if the future was marked done with an exception, and if so, raise the exception.  There is no chance of throwing a double exception, since if we earlier threw the exception in spin_once, we never would have made it here.

Sort of.  It means that we'll raise in some cases where we didn't before, but those were cases where we *should* have been raising anyway.  So it is more a bugfix.

Yes, Claude Opus 4.7

### Additional Information

This should make Windows CI slightly less flaky.

### DIFF
--- a/rclpy/test/test_executor.py
+++ b/rclpy/test/test_executor.py
@@ -1032,6 +1032,22 @@ class TestExecutor(unittest.TestCase):
                 finally:
                     executor.shutdown()
 
+    def test_spin_until_future_complete_propagates_already_done_exception(self) -> None:
+        # Test that spin_until_future_complete surfaces a future's exception even if
+        # the future was already done before the spin loop ran.
+        for cls in [SingleThreadedExecutor, MultiThreadedExecutor]:
+            with self.subTest(cls=cls):
+                executor = cls(context=self.context)
+                try:
+                    future: Future[None] = executor.create_future()
+                    future.set_exception(RuntimeError('pre-set exception'))
+
+                    with self.assertRaises(RuntimeError) as cm:
+                        executor.spin_until_future_complete(future, timeout_sec=1)
+                    self.assertIn('pre-set exception', str(cm.exception))
+                finally:
+                    executor.shutdown()
+
 
 if __name__ == '__main__':
     unittest.main()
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29074/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21869/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8941/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27994/)

clalancette:
test_executor.py failed in this Windows CI run, but I'm fairly convinced that this is a different Windows timing error.  I'll probably open up a separate PR to fix that one.

rclpy/test/test_executor.py
```diff
@@ -1032,6 +1032,22 @@ def spin_thread() -> None:
                 finally:
                     executor.shutdown()
 
+    def test_spin_until_future_complete_propagates_already_done_exception(self) -> None:
```
fujitatomoya, line 1035: (just a comment, i guess you already know this but could be hard to simulate.)
this test does cover the code path the fix adds.
but it doesn't reproduce the actual race (handler completing on a worker thread between two iterations of the outer loop). this will be hard to consider how long should be given to the spin loop time to enter wait_for_ready_callbacks... that i do not have good idea to simulate this...

fujitatomoya (approved):
lgtm

rclpy/test/test_executor.py
clalancette: Yeah, I didn't have a good idea on how to reproduce the race, so I settled for this.  I'm going to go ahead and merge this, but if you come up with an idea on how to do this, I'm happy to do a follow-up to add another test.

ahcorde:
@clalancette is this something that we should backport ?

PR #1663, opened 2026-05-15
Switch back to using ROS time for the async_clock.

clalancette:
## Description

This makes it so that we always honor ROS time, and is consistent with what the rest of the file does.

Pointed out by @nadavelkabets .  This is a more correct implementation.

No.

Yes, Claude Opus 4.7

### Additional Information

N/A

### DIFF
--- a/rclpy/rclpy/experimental/async_clock.py
+++ b/rclpy/rclpy/experimental/async_clock.py
@@ -106,14 +106,15 @@ class AsyncClock(BaseClock):
         # Re-arm until the deadline is actually reached so callers get an
         # "at least duration_sec" guarantee, matching the strict semantics
         # of Clock.sleep_for in rclpy/rclcpp.
-        deadline = loop.time() + duration_sec
+        deadline: Time = self.now() + duration
         while True:
-            remaining = deadline - loop.time()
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29092/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21888/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8959/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28021/)

fujitatomoya (approved):

fujitatomoya:
@clalancette rhel and windows warnings seem unrelated, can we go ahead to merge?

clalancette:
Actually, it turns out that this is the wrong thing to do.  The problem that I got twirled around in my head is that self.now() is System time, not monotonic time.  And on Windows that turns out to be the slow 15ms clock.  Further, the loop.time() that *was* there is the correct thing, because at this point we know we are not handling simulation time (that was done above).  We can see that this fails on the Windows test.

I'm going to close this one, sorry for the noise.

PR #1666, opened 2026-05-16
Fix the flaky immediate shutdown.

clalancette:
## Description

The test was trying to ensure that when shutdown is called, it happens "immediately", and not wait until the timeout.  The problem in our current code is that we don't know whether another thread is in "spin"
during the shutdown.  If it is, then there is some possibility that we deleted the shutdown guard condition *before* the spin built the waitset, in which case we would spin until the timeout, and thus fail the test.

The fix is to mark when a thread is in spin, and then have shutdown wait for the spin to complete before destroying resources.  If spin is waiting, then it will be woken by the guard condition immediately, and the shutdown will proceed quickly.  If the spin is stuck in a long-running user callback, then we still need a timeout because we don't want shutdown to take forever.

While I was in here, I also enabled this test for the MultiThreadedExecutor since it should succeed for that now as well.

No.

Yes, Claude Opus 4.7

### Additional Information

N/A

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -20,9 +20,7 @@ from dataclasses import dataclass
 from functools import partial
 import inspect
 import os
-from threading import Condition
-from threading import Lock
-from threading import RLock
+import threading
 import time
 from types import TracebackType
 from typing import Any
@@ -94,7 +92,7 @@ class _WorkTracker:
     def __init__(self) -> None:
         # Number of tasks that are being executed
         self._num_work_executing = 0
-        self._work_condition = Condition()
+        self._work_condition = threading.Condition()
 
     def __enter__(self) -> None:
         """Increment the amount of executing work by 1."""
@@ -212,12 +210,12 @@ class Executor(ContextManager['Executor']):
         super().__init__()
         self._context = get_default_context() if context is None else context
         self._nodes: Set[Node] = set()
-        self._nodes_lock = RLock()
+        self._nodes_lock = threading.RLock()
         # all tasks that are not complete or canceled
         self._pending_tasks: Dict[Task[Any], TaskData] = {}
         # tasks that are ready to execute
         self._ready_tasks: Deque[Task[Any]] = deque()
-        self._tasks_lock = Lock()
+        self._tasks_lock = threading.Lock()
         # This is triggered when wait_for_ready_callbacks should rebuild the wait list
         self._guard: Optional[GuardCondition] = GuardCondition(
             callback=None, callback_group=None, context=self._context)
@@ -225,7 +223,7 @@ class Executor(ContextManager['Executor']):
         self._is_shutdown = False
         self._work_tracker = _WorkTracker()
         # Protect against shutdown() being called in parallel in two threads
-        self._shutdown_lock = Lock()
+        self._shutdown_lock = threading.Lock()
         # State for wait_for_ready_callbacks to reuse generator
         self._cb_iter: Optional[YieldedCallback] = None
         self._last_args: Optional[tuple[object, ...]] = None
@@ -238,24 +236,28 @@ class Executor(ContextManager['Executor']):
         # True when the executor is spinning
         self._is_spinning = False
         # Protects access to _is_spinning
-        self._is_spinning_lock = Lock()
+        self._is_spinning_cond = threading.Condition()
+        self._spinning_thread: Optional[threading.Thread] = None
 
     def _enter_spin(self) -> None:
         """Mark the executor as spinning and prevent concurrent spins."""
-        with self._is_spinning_lock:
+        with self._is_spinning_cond:
             if self._is_spinning:
                 raise RuntimeError('Executor is already spinning')
             self._is_spinning = True
+            self._spinning_thread = threading.current_thread()
 
     def _exit_spin(self) -> None:
         """Clear the spinning flag."""
-        with self._is_spinning_lock:
+        with self._is_spinning_cond:
             self._is_spinning = False
+            self._spinning_thread = None
+            self._is_spinning_cond.notify_all()
 
     @property
     def is_spinning(self) -> bool:
         """Return whether the executor is currently spinning."""
-        with self._is_spinning_lock:
+        with self._is_spinning_cond:
             return self._is_spinning
 
     @property
@@ -327,6 +329,15 @@ class Executor(ContextManager['Executor']):
         # Clean up stuff that won't be used anymore
         with self._nodes_lock:
             self._nodes = set()
+
+        with self._is_spinning_cond:
+            if self._spinning_thread is not threading.current_thread():
+                # Cap the wait to avoid hanging the process if a spinner
+                # is stuck in a callback that never returns.
+                wait_timeout = timeout_sec if (
+                    timeout_sec is not None and timeout_sec > 0) else 5.0
+                self._is_spinning_cond.wait_for(
+                    lambda: not self._is_spinning, timeout=wait_timeout)
 
         with self._shutdown_lock:
             if self._guard:
@@ -1086,7 +1097,7 @@ class MultiThreadedExecutor(Executor):
                 'Use the SingleThreadedExecutor instead.')
         self._futures: List[Future[Any]] = []
         self._executor = ThreadPoolExecutor(num_threads)
-        self._futures_lock = Lock()
+        self._futures_lock = threading.Lock()
 
     def _spin_once_impl(
         self,
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -176,9 +176,10 @@ class _WorkTracker:
                 # A new waiter may have just satisfied an existing
                 # waiter's condition (its in-flight work is now excluded).
                 self._work_condition.notify_all()
+            drained = False
             try:
-                if not self._work_condition.wait_for(other_work_drained, timeout_sec):
-                    return False
+                drained = self._work_condition.wait_for(other_work_drained, timeout_sec)
+                return drained
             finally:
                 # Keep the waiter membership while a callback is still
                 # in flight on this thread -- removing it now would let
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29098/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21894/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8965/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28027/)

fujitatomoya (commented):
this PR breaks the contract in two distinct ways?

1. if a spinner is processing in a long-running callback, shutdown() returns after 5 seconds.  the return value is True, so the caller has no way to know shutdown didn't actually finish?
2. if the caller passes None, they've explicitly said "I'm fine waiting forever." the library shouldn't second-guess that? if timeout_sec does not work in this way, the API implementation is broken?

i may be wrong, but it seems that shutdown could deadlock for 5 seconds when called from inside a callback? callbacks run on ThreadPoolExecutor workers (MTExecutor case), not on _spinning_thread. if a user callback in a worker thread calls executor.shutdown(), it will wait for 5 seconds or so, cz spinning thread won't return until the guard condition fires.

clalancette:
Thank you for the review!

> if the caller passes None, they've explicitly said "I'm fine waiting forever." the library shouldn't second-guess that? if timeout_sec does not work in this way, the API implementation is broken?

Yeah, you are totally right.  I had a separate change that implements this, but it should be included here.  I've pushed it now, and this should fix the contract.

> if a spinner is processing in a long-running callback, shutdown() returns after 5 seconds. the return value is True, so the caller has no way to know shutdown didn't actually finish?

Yes, you are correct.  Again, I had a separate change for this, but it should have all been included together.  The latest push checks to see if the wait_for failed, and if so, it returns False so the caller can try again.

> i may be wrong, but it seems that shutdown could deadlock for 5 seconds when called from inside a callback? callbacks run on ThreadPoolExecutor workers (MTExecutor case), not on _spinning_thread. if a user callback in a worker thread calls executor.shutdown(), it will wait for 5 seconds or so, cz spinning thread won't return until the guard condition fires.

I don't think this deadlock is possible.  If a callback calls shutdown(), then there are 2 possible waits:
* Line 387, `self._work_tracker.wait()` - the implementation of this ignores a threads own works, so it won't deadlock
* Line 401, `self._is_spinning_cond.wait_for` - this thread is in the callback, so is not in the spinner.  It will wait for the spinner to not be spinning anymore.  It is true that the spinner is waiting on the guard condition, but we triggered the guard condition earlier, so it should have woken up.

That said, I did find another deadlock while looking at this, in that *two* callbacks that call shutdown simultaneously would deadlock with each other.  The latest commit fixes that by tracking which threads are waiting, and if another thread is waiting, it wakes it up to check its conditions.

(whew, this escalated kind of quickly, and this PR grew a lot.  But I think we are getting towards the right shape of things here, please take another look when you have time)

clalancette:
Here's another CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29142/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21935/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9015/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28095/)

clalancette:
Another CI after the latest change:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29144/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21937/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9019/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28097/)

clalancette:
Another CI after fixing linters:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29147/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21940/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9031/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28105/)

mjcarroll:
I have been going back and forth with the agent trying to come up with what I think might be a race condition.  I'm going to see if I can create a test case for it as well.

I think we actually need to see if the `self._work_condition.wait_for()` succeeds or not, and make sure to prune the thread from `self._waiting_threads` based on that?

* Step 1: Initial State
  * Two active callbacks are running on different worker threads (WorkerA and WorkerB).
  * `self._executing_thread_counts` is `{WorkerA: 1, WorkerB: 1}`, populated via [track_callback](https://github.com/ros2/rclpy/pull/1666/files#diff-c97ccd9e6abbfc29a72e5485e24cf79d8625c25deb57b3952b1ed114a9905d40R126).
  * `self._waiting_threads` is empty.

* Step 2: WorkerA calls shutdown with a short timeout (0.1s)
  * WorkerA enters `wait()`, adding itself to `self._waiting_threads` at [line 170](https://github.com/ros2/rclpy/pull/1666/files#diff-c97ccd9e6abbfc29a72e5485e24cf79d8625c25deb57b3952b1ed114a9905d40R170).
  * It calls [wait_for](https://github.com/ros2/rclpy/pull/1666/files#diff-c97ccd9e6abbfc29a72e5485e24cf79d8625c25deb57b3952b1ed114a9905d40R175) with the [other_work_drained](https://github.com/ros2/rclpy/pull/1666/files#diff-c97ccd9e6abbfc29a72e5485e24cf79d8625c25deb57b3952b1ed114a9905d40R161) predicate.
  * The predicate check is `{WorkerA, WorkerB} <= {WorkerA}`, which returns `False` because WorkerB is not waiting. WorkerA releases the lock and blocks.

* Step 3: WorkerA's wait times out
  * After 0.1s, the wait times out and returns `False` at [line 175-176](https://github.com/ros2/rclpy/pull/1666/files#diff-c97ccd9e6abbfc29a72e5485e24cf79d8625c25deb57b3952b1ed114a9905d40R175-R176).
  * The [finally block](https://github.com/ros2/rclpy/pull/1666/files#diff-c97ccd9e6abbfc29a72e5485e24cf79d8625c25deb57b3952b1ed114a9905d40R177) runs. Since WorkerA is still in `_executing_thread_counts` (see [diff](https://github.com/ros2/rclpy/pull/1666/files#diff-c97ccd9e6abbfc29a72e5485e24cf79d8625c25deb57b3952b1ed114a9905d40R187)), it is not discarded from `_waiting_threads`.
  * WorkerA returns `False` from `shutdown()` and continues executing the remainder of its callback.
  * *The Leak:* `self._waiting_threads` remains `{WorkerA}`.

* Step 4: WorkerB completes naturally
  * WorkerB finishes Callback B.
  * The [finally block](https://github.com/ros2/rclpy/pull/1666/files#diff-c97ccd9e6abbfc29a72e5485e24cf79d8625c25deb57b3952b1ed114a9905d40R130) of `track_callback` runs and removes WorkerB from `_executing_thread_counts`.
  * Current State: `_executing_thread_counts` is `{WorkerA: 1}`, and `_waiting_threads` is `{WorkerA}`.

* Step 5: MainThread calls shutdown
  * MainThread calls `executor.shutdown(timeout_sec=5.0)`.
  * MainThread adds itself to `_waiting_threads` (`_waiting_threads` becomes `{WorkerA, MainThread}`).
  * MainThread evaluates the [other_work_drained](https://github.com/ros2/rclpy/pull/1666/files#diff-c97ccd9e6abbfc29a72e5485e24cf79d8625c25deb57b3952b1ed114a9905d40R161) predicate.
  * The check `{WorkerA} <= {WorkerA, MainThread}` returns `True`.
  * The `wait_for()` call returns `True` immediately without blocking MainThread.
  * *The Bug:* MainThread proceeds to destroy executor resources (like guard conditions and node references) while WorkerA is still actively executing its callback.

clalancette:
After merging in @mjcarroll 's fix, and doing the suggestion from @fujitatomoya , here is another round of CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29264/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22049/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9145/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28236/)

clalancette:
I've rebased this onto the latest.  Here is another round of CI.  @fujitatomoya @mjcarroll another look at the latest changes would be appreciated.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30044/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22811/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9929/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28953/)

clalancette:
The failing tests on Linux should be fixed by https://github.com/ros2/rmw_cyclonedds/pull/602

Miko997:
I reproduced the handler-taken/pre-submit case from #1724 against `1e47a4339907e66daca68b259e23486d48868674`.

With a finite timeout, base shutdown returned `False`, but `MultiThreadedExecutor.shutdown()` still closed the worker pool. Releasing the spin then raised `RuntimeError: cannot schedule new futures after shutdown`, and the selected task did not execute.

I added a focused regression and a two-line guard that leaves the pool open after a timeout, allowing the pending submission and a later shutdown retry: https://github.com/Miko997/rclpy/commit/264fa4064c2f2e2283a8267de370b759cd671be3

Tests in an isolated Rolling container (Ubuntu 26.04, Python 3.14, Fast DDS):

- `python3 -m pytest -q src/rclpy/rclpy/test/test_executor.py -k test_multi_threaded_shutdown_timeout_before_submit -W error --timeout=90` — 50 consecutive runs passed, covering both `wait_for_threads` values, exactly-once execution, retry, and thread cleanup.
- `python3 -m pytest -q src/rclpy/rclpy/test/test_executor.py -W error --timeout=90` — passed.
- `colcon test --packages-select rclpy --return-code-on-test-failure --event-handlers console_cohesion+ --ctest-args -j1 --output-on-failure` — passed.
- `colcon test-result --all --verbose` — no failures; 78 C++ checks skipped by the installed ament tool due to cppcheck 2.19 performance issues. XML lint used the official schema via a local catalog.

Would you prefer to cherry-pick this commit or have me open a small PR targeting this branch?

clalancette:
> Would you prefer to cherry-pick this commit or have me open a small PR targeting this branch?

Please open up a small PR targeting this branch and I'll take a look.  Thanks!

PR #1668, opened 2026-05-17
Lock the goal-handle / result-future registry in ActionServer

### LINKED ISSUE #1236: KeyError in ActionServer._execute_goal
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - Binaries (.deb package)
- Version or commit hash:
  - 3.3.12-1jammy.20240217.065750
- DDS implementation:
  - Fast-RTPS (2.6.7-1jammy.20240217.034748)
  - rclpy

#### Steps to reproduce issue

* Run an action server node with the `use_sim_time` parameter as `True`
* Have a node emit clock messages faster than wall clock moves (at x60)
* Call the action

#### Expected behavior

Action goal result is set without problem.

#### Actual behavior

The action server fails because `self._result_futures[bytes(goal_uuid)].set_result(result_response)` in the ActionServer _execute_goal method raises a `KeyError` exception:
```
  File "/home/rbailonr/adauv_ws/install/adauv_bluebox/lib/adauv_bluebox/vehicle.py", line 333, in main
    executor.spin()
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 294, in spin
    self.spin_once()
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 794, in spin_once
    self._spin_once_impl(timeout_sec)
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 791, in _spin_once_impl
    future.result()
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/task.py", line 94, in result
    raise self.exception()
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/task.py", line 239, in __call__
    self._handler.send(None)
  File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/action/server.py", line 352, in _execute_goal
    self._result_futures[bytes(goal_uuid)].set_result(result_response)
KeyError: b'EM$\xc6\xb3YA:\xa9\xa1_\xd1\xeb\x88\x06\x88'
```

#### Additional information

This occurrence of this behavior is inconsistent, but only happens in accelerated time. 

It looks like a synchronization issue with `ActionServer._execute_expire_goals` that removes the goal result future from the `_result_futures` dictionary before `_execute_goal` sets its result. [Checking `not self._result_futures[goal_uuid].done()` before removing the expired goal](https://github.com/rafael1193/rclpy/commit/729c31d7994f7cf26f74bce3de2763459ad0c073) seems to mitigate the issue, but the actual source of the problem may be elsewhere.

### LINKED ISSUE #1667: KeyError in ActionServer._execute_goal under goal-timeout abort (race on _result_futures)
### Bug report

**Required Info:**

- Operating System: Ubuntu 22.04 (inside `nvcr.io/nvidia/isaac-sim:5.0.0` container)
- Installation type: binaries (`apt install ros-humble-rclpy`)
- Version or commit hash: `rclpy 3.3.21-1jammy.20260421.025639`
- DDS implementation: `rmw_cyclonedds_cpp`
- Client library: rclpy
- Python: 3.10.12

### Steps to reproduce

`rclpy/action/server.py:357` does `self._result_futures[bytes(goal_uuid)].set_result(result_response)` without checking whether the future is still in the dict. Under real-time load, the future can be removed by `_execute_expire_goals` before `_execute_goal` reaches that line, raising `KeyError` and crashing the executor.

Distinct trigger from #1236 (which was sim-time fast clock): **wall-clock goal-timeout abort under sustained ActionServer load**. Our `NavigateToPose` action server runs a 20 Hz control loop and aborts goals on `elapsed > goal_timeout_s = 1200.0`. Two out of six independent ActionServers crashed with this `KeyError` in a single 64-order multi-robot simulation run.

Minimal repro shape:
1. ActionServer running on `MultiThreadedExecutor`, `ReentrantCallbackGroup`
2. Long-running `execute_callback` that returns after `goal_handle.abort()` on timeout
3. Sustained load — many sequential goals per server, several servers in the process
4. Real time (`use_sim_time=False`)

### Expected behavior

Goal cleanup after abort should not race with the goal-result publish, or the publish should defensively check whether the future still exists before calling `set_result`.

### Actual behavior

```
[pure_pursuit_driver-6] [WARN] [pure_pursuit_driver]: goal aborted: timeout after 1200.0s (>1200.0s)
[pure_pursuit_driver-6] Traceback (most recent call last):
  File "/opt/ros/humble/lib/python3.10/site-packages/rclpy/executors.py", line 323, in spin
    self.spin_once()
  File "/opt/ros/humble/lib/python3.10/site-packages/rclpy/executors.py", line 863, in spin_once
    self._spin_once_impl(timeout_sec)
  File "/opt/ros/humble/lib/python3.10/site-packages/rclpy/executors.py", line 860, in _spin_once_impl
    future.result()
  File "/opt/ros/humble/lib/python3.10/site-packages/rclpy/task.py", line 109, in result
    raise self.exception()
  File "/opt/ros/humble/lib/python3.10/site-packages/rclpy/task.py", line 272, in _execute_coroutine_step
    result = coro.send(None)
  File "/opt/ros/humble/lib/python3.10/site-packages/rclpy/action/server.py", line 357, in _execute_goal
    self._result_futures[bytes(goal_uuid)].set_result(result_response)
KeyError: b'S\xe2\xd7\xc6\xfd\xd7@\x02\x98^E\xc4\xe9\xcd4x'
```

Second concurrent crash in same run, same trace, different uuid (`b'\x95\x86\xafj\x8e\xe9E\xb5\x8e\xb7\x0bd/C\xe5\xa5'`). Full logs: [docs/runs/m7_steady_state_gt/pure_pursuit.log](https://github.com/zeon01/warehouse-digital-twin/blob/main/docs/runs/m7_steady_state_gt/pure_pursuit.log).

### Suggested fix

Defensive lookup at `action/server.py:357`:

```python
# Before
self._result_futures[bytes(goal_uuid)].set_result(result_response)

# After
future = self._result_futures.get(bytes(goal_uuid))
if future is not None and not future.done():
    future.set_result(result_response)
else:
    self._logger.warn(
        f"_execute_goal: result_future for goal "
        f"{bytes(goal_uuid).hex()} unavailable (already expired or completed); "
        f"skipping set_result"
    )
```

Addresses the symptom. A proper fix likely needs `_execute_expire_goals` and `_execute_goal` to coordinate on future ownership (e.g. atomic pop-and-handle rather than independent lookups).

### Prior art / related issues

- **#1236** reports the identical `KeyError` from a different trigger (`use_sim_time=True` with sim time accelerated 60x). It was closed as fixed by **PR #1363**.
- **PR #1363** only escalated the surrounding log line from DEBUG to WARN — it did not fix the race. It was later reverted on the rolling branch (Oct 17, 2024).
- The underlying race that produced #1236 is therefore still present and reproducible under different load shapes.

### Workaround we shipped (client side)

Wrapping `executor.spin()` in a `KeyError` retry loop keeps the node alive but masks the real bug. ([Commit](https://github.com/zeon01/warehouse-digital-twin/commit/9a909ef).)

```python
while rclpy.ok():
    try:
        executor.spin()
        break
    except KeyError as exc:
        node.get_logger().warn(
            f"executor.spin raised KeyError on goal-result cleanup "
            f"(rclpy ActionServer race, goal_uuid={exc!r}); resuming"
        )
        continue
```

Happy to put up a PR for the defensive-lookup change if there's interest.

zeon01:
## Summary

Closes #1236. Closes #1667.

`ActionServer` tracks goals in two dicts — `_goal_handles` and `_result_futures` — that are mutated and read as a pair. With the `MultiThreadedExecutor`, `_execute_expire_goals` can remove a goal from both dicts on one thread while `_execute_goal`, `_execute_get_result_request`, `_execute_cancel_request` or `ServerGoalHandle._set_result` reads them on another. The reader then hits a `KeyError` and the executor crashes.

This adds `_goal_registry_lock` and takes it around every access to the dict pair, so a reader never observes one dict updated without the other.

## Design notes

- **`threading.Lock`, not `asyncio.Lock`.** `MultiThreadedExecutor` steps these coroutines on real OS threads; `asyncio.Lock` is not thread-safe. The race holds even under `MutuallyExclusiveCallbackGroup`, because `_execute_goal` is scheduled as its own executor task via `notify_execute` (`executor.create_task`) and is therefore not serialized against the waitable's `_execute_expire_goals` by the callback group.
- The lock is only ever held across synchronous sections — no `await`, no user-callback invocation. Where a coroutine acts on a `Future`, the reference is fetched under the lock and `set_result()` / `add_done_callback()` is called after release.
- It is a separate lock from the existing `self._lock` and the two are never nested, so it cannot deadlock.
- With the registry consistent, `_execute_get_result_request` reliably detects an expired goal and returns `STATUS_UNKNOWN` instead of crashing.

This replaces the earlier defensive-lookup-only revision of this PR, per review feedback that the root-cause fix is locking the dict pair as a single unit of state.

## Test plan

Two regression tests added to `test_action_server.py`:

- `test_get_result_request_handles_expired_future` (#1236) — removes the result future, sends a get-result request; asserts a `STATUS_UNKNOWN` response instead of a crash.
- `test_execute_goal_handles_expired_future` (#1667) — removes the result future while the execute callback runs; asserts `_execute_goal` does not raise.

Verified against `osrf/ros:rolling-desktop`:

- Without the patch: both new tests fail with `KeyError` (`_execute_goal`, `_execute_get_result_request`).
- With the patch: both pass.
- Full `test_action_server.py` — 28 tests — pass. `ament_flake8` / `ament_pep257` / `ament_copyright` pass.

## Out of scope / follow-ups

- A cleaner long-term shape is to collapse the two dicts into a single `Dict[bytes, GoalEntry]`; kept out of this PR to stay minimal — can follow up if preferred.
- `succeed(response)` / `abort(response)` set the result future and `_execute_goal` then sets it again; rclpy's `Future.set_result()` overwrites + re-fires callbacks rather than raising, so the result can be sent twice. Separate issue from this race.

### DIFF
--- a/rclpy/rclpy/action/server.py
+++ b/rclpy/rclpy/action/server.py
@@ -197,7 +197,11 @@ class ServerGoalHandle(Generic[GoalT, ResultT, FeedbackT, ImplT]):
             result_response.result = response
         else:
             result_response.result = self._action_server._action_type.Result()
-        self._action_server._result_futures[bytes(self.goal_id.uuid)].set_result(result_response)
+        key = bytes(self.goal_id.uuid)
+        with self._action_server._goal_registry_lock:
+            result_future = self._action_server._result_futures.get(key)
+        if result_future is not None:
+            result_future.set_result(result_response)
 
     def execute(
         self,
@@ -376,6 +380,14 @@ class ActionServer(Generic[GoalT, ResultT, FeedbackT, ImplT],
         # key: UUID in bytes, value: Future
         self._result_futures: Dict[bytes, Future[GetResultServiceResponse[ResultT]]] = {}
 
+        # Serializes mutations and reads of the (`_goal_handles`, `_result_futures`)
+        # pair so a reader never observes one dict updated without the other. Needed
+        # because `_execute_goal` runs as its own executor task (see `notify_execute`)
+        # and can race with the waitable's `_execute_expire_goals` under the
+        # MultiThreadedExecutor. Only ever held across synchronous sections — never
+        # across an `await` or a user-callback invocation.
+        self._goal_registry_lock = threading.Lock()
+
         callback_group.add_entity(self)
         self._node.add_waitable(self)
 
@@ -418,9 +430,12 @@ class ActionServer(Generic[GoalT, ResultT, FeedbackT, ImplT],
                     'Failed to accept new goal with ID {0}: {1}'.format(goal_uuid.uuid, e))
                 accepted = False
             else:
-                self._goal_handles[bytes(goal_uuid.uuid)] = goal_handle
-                self._result_futures[bytes(goal_uuid.uuid)] = Future()
-                self.add_future(self._result_futures[bytes(goal_uuid.uuid)])
+                key = bytes(goal_uuid.uuid)
+                result_future: Future[GetResultServiceResponse[ResultT]] = Future()
+                with self._goal_registry_lock:
+                    self._goal_handles[key] = goal_handle
+                    self._result_futures[key] = result_future
+                self.add_future(result_future)
 
         # Send response
         response_msg = self._action_type.Impl.SendGoalService.Response()
@@ -478,7 +493,15 @@ class ActionServer(Generic[GoalT, ResultT, FeedbackT, ImplT],
         result_response = self._action_type.Impl.GetResultService.Response()
         result_response.status = goal_handle.status
         result_response.result = execute_result
-        self._result_futures[bytes(goal_uuid)].set_result(result_response)
+        with self._goal_registry_lock:
+            result_future = self._result_futures.get(bytes(goal_uuid))
+        if result_future is not None:
+            result_future.set_result(result_response)
+        else:
+            # Goal expired before the callback finished; an expected outcome
+            # under load with a short result_timeout, so log at debug level.
+            self._logger.debug(
+                'Goal with ID {0} expired before its result could be set'.format(goal_uuid))
 
     async def _execute_cancel_request(
         self,
@@ -495,12 +518,13 @@ class ActionServer(Generic[GoalT, ResultT, FeedbackT, ImplT],
 
         for goal_info in cancel_response.goals_canceling:
             goal_uuid = bytes(goal_info.goal_id.uuid)
-            if goal_uuid not in self._goal_handles:
+            with self._goal_registry_lock:
+                goal_handle = self._goal_handles.get(goal_uuid)
+            if goal_handle is None:
                 # Possibly the user doesn't care to track the goal handle
                 # Remove from response
                 cancel_response.goals_canceling.remove(goal_info)
                 continue
-            goal_handle = self._goal_handles[goal_uuid]
             response = await await_or_execute(self._cancel_callback, goal_handle)
 
             if CancelResponse.ACCEPT == response:
@@ -536,9 +560,17 @@ class ActionServer(Generic[GoalT, ResultT, FeedbackT, ImplT],
         self._logger.debug(
             'Result request received for goal with ID: {0}'.format(goal_uuid))
 
+        # Atomically check whether the goal is still tracked and grab its result
+        # future. `_execute_expire_goals` removes the goal handle and result future
+        # as a pair, so taking the lock here keeps the membership check and the
+        # future lookup consistent with each other.
+        key = bytes(goal_uuid)
+        with self._goal_registry_lock:
+            result_future = self._result_futures.get(key) if key in self._goal_handles else None
+
         # If no goal with the requested ID exists, then return UNKNOWN status
         # or the goal with the requested ID has been already expired
-        if bytes(goal_uuid) not in self._goal_handles:
+        if result_future is None:
             self._logger.warning(
                 'Sending result response for unknown or expired goal ID: {0}'.format(goal_uuid))
             result_response = self._action_type.Impl.GetResultService.Response()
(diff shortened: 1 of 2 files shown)


fujitatomoya (commented):
@zeon01 thanks for the patch.

i believe this is a defensive patch, this could mitigate the situation a bit, but not a real fix.
to address and get to the root cause, i think that the proper fix is an asyncio.Lock (or threading.Lock depending on which executor path can touch these) that covers the pair of dicts as a single unit of state.

fujitatomoya:
CC: @Barry-Xu-2018

zeon01:
Reworked to a lock-based fix — `_goal_registry_lock` now covers the `_goal_handles` / `_result_futures` pair as a single unit of state.

It's a `threading.Lock`: `MultiThreadedExecutor` steps these coroutines on real OS threads, and `asyncio.Lock` is not thread-safe. The race holds even under `MutuallyExclusiveCallbackGroup` — `_execute_goal` is scheduled as its own executor task via `notify_execute` (`executor.create_task`), so the callback group does not serialize it against the waitable's `_execute_expire_goals`.

The lock is taken around every access to the pair (`_execute_goal_request`, `_execute_goal`, `_execute_get_result_request`, `_execute_cancel_request`, `_execute_expire_goals`, `ServerGoalHandle._set_result`, `destroy()`). It is only ever held across synchronous sections — no `await`, no user-callback invocation: a `Future` reference is fetched under the lock and `set_result()` / `add_done_callback()` called after release, so done-callbacks can't re-enter it. It is disjoint from `self._lock` and never nested.

Verified: regression tests for #1236 and #1667 fail with `KeyError` against the current code and pass with the lock; the full `test_action_server.py` suite (28 tests) and the flake8/pep257/copyright linters pass.

Two out-of-scope notes:
- A cleaner long-term shape would collapse `_goal_handles` and `_result_futures` into a single `Dict[bytes, GoalEntry]` so the pair can't desync structurally. I kept this PR to the lock to stay minimal — happy to follow up if you'd prefer.
- `ServerGoalHandle.succeed(response)` / `abort(response)` call `_set_result`, and `_execute_goal` then sets the same future again, so passing a non-`None` response sets the future twice. `Future.set_result()` doesn't raise on a second call — it overwrites the result and re-fires done callbacks — so the result response can be sent twice. Separate from this race; happy to file an issue if it isn't already tracked.

PR #1674, opened 2026-05-29
Test waiter leak on timeout

mjcarroll:
## Description

Test and fix for a race condition I found in https://github.com/ros2/rclpy/pull/1666

More context specifically here: https://github.com/ros2/rclpy/pull/1666#issuecomment-4578118417

Running with the first commit (test only) and it fails.  Run with fix and it passes.

Assisted-by: Gemini CLI:Gemini 3.5 Flash

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -171,9 +171,10 @@ class _WorkTracker:
                 # A new waiter may have just satisfied an existing
                 # waiter's condition (its in-flight work is now excluded).
                 self._work_condition.notify_all()
+            drained = False
             try:
-                if not self._work_condition.wait_for(other_work_drained, timeout_sec):
-                    return False
+                drained = self._work_condition.wait_for(other_work_drained, timeout_sec)
+                return drained
             finally:
                 # Keep the waiter membership while a callback is still
                 # in flight on this thread -- removing it now would let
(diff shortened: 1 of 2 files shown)


rclpy/rclpy/executors.py
```diff
@@ -184,10 +185,12 @@ def other_work_drained() -> bool:
...
-                if added_self and current not in self._executing_thread_counts:
-                    self._waiting_threads.discard(current)
-                    self._work_condition.notify_all()
-        return True
+                # However, if we timed out (not drained), we are NOT committed
+                # to finishing successfully, so we must discard the membership.
+                if not drained or (added_self and current not in self._executing_thread_counts):
+                    if added_self:
+                        self._waiting_threads.discard(current)
+                        self._work_condition.notify_all()
```
fujitatomoya, line 193: nit; can be simplified?

```suggestion
                    if added_self and (not drained or current not in self._executing_thread_counts):
                        self._waiting_threads.discard(current)
                    self._work_condition.notify_all()
```

fujitatomoya (approved):
lgtm.

a thread that added itself to `_waiting_threads` and then timed out never removed itself. a leaked entry makes a future wait() from another thread see the predicate satisfied prematurely. it returns True (drained) while a callback is still genuinely in flight.

rclpy/rclpy/executors.py
clalancette: I'll do this simplification over on the other PR.

Issue #1672, opened 2026-05-23
parameter_dict_from_yaml_file loads doubles as strings when str() uses exponent notation (e.g. 0.00001, 5e-06, 1e+20)


Lots-ninety-nine:
## Bug report

**Required Info:**

- Operating System: Ubuntu 24.04 (`osrf/ros:rolling-desktop`)
- Installation type: binaries
- Version or commit hash: `rolling` (the offending line is still present on the `rolling` branch, `rclpy/parameter.py:565`)
- DDS implementation: `rmw_fastrtps_cpp`
- Client library (if applicable): rclpy

### Description

`parameter_dict_from_yaml_file` (used by `ros2 param load`) loads floating-point parameters as **strings** whenever the value's Python `str()` representation uses exponent notation, e.g. `0.00001`, `5e-06`, `1.0e+20`. Plain decimals such as `3.14` are unaffected. This also affects double arrays.

### Steps to reproduce issue

`repro.yaml`:

```yaml
/parameter_blackboard:
  ros__parameters:
    a_small: 0.00001
    b_sci: 5e-06
    c_large: 1.0e+20
    d_normal: 3.14
    e_arr: [1.0e-5, 2.0e-5]
```

```bash
ros2 run demo_nodes_cpp parameter_blackboard &
ros2 param load /parameter_blackboard repro.yaml
for p in a_small b_sci c_large d_normal e_arr; do ros2 param get /parameter_blackboard $p; done
```

### Expected behavior

`a_small`, `b_sci`, `c_large` are doubles and `e_arr` is a double array — matching what the C++ params-file loader (`--params-file`) and `ros2 param set` already produce for the same values.

### Actual behavior

```
a_small   -> String value is: 1e-05
b_sci     -> String value is: 5e-06
c_large   -> String value is: 1e+20
d_normal  -> Double value is: 3.14
e_arr     -> String values are: ['1e-05', '2e-05']
```

### Root cause

In `_unpack_parameter_dict` (`rclpy/parameter.py`):

```python
parameter.value = get_parameter_value(str(param_value))
```

[...]

Issue #1606, opened 2026-02-10
`parameter_dict_from_yaml_file` should be replace with `rcl_parse_yaml_file`


fujitatomoya:
### Operating System:

```Linux tomoyafujita-B760M-Pro-RS-D4 6.17.0-14-generic #14~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Jan 15 15:52:10 UTC 2 x86_64 x86_64 x86_64 GNU/Linux```

### ROS version or commit hash:

rolling

### RMW implementation (if applicable):

Any

### Client library (if applicable):

rclpy

### 'ros2 doctor --report' output

<details><summary>ros2 doctor --report</summary>

```console
<COPY OUTPUT HERE>
```
</details>

### Steps to reproduce issue

see https://github.com/ros2/rclpy/blob/e056ea63bcea5b97af84f3395aa790624738d3a4/rclpy/rclpy/parameter.py#L363-L552

it implements the yaml parsing process in rclpy, it does not call on `rcl_parse_yaml_file`. (this is why rclpy has a dependency to python3-yaml)

i believe this breaks the basic design between `rcl` and `rclpy` responsibility. `rcl_yaml_param_parser` package is meant to provide the consistent behavior for all client libraries such as rclcpp and rclpy. actually `rclcpp` does use `rcl_parse_yaml_file` when it loads the yaml file to the node parameters.

### Expected behavior

i believe we should do,

- deprecate/remove client specific yarl parser, saying remove `parameter_map_from_yaml_file`.
- and then, use `rcl_yaml_param_parser` package to call `rcl_parse_yaml_file` when it loads the yaml file to the node parameters.

this provides the same behavior with `rclcpp`.

### Actual behavior

the behavior of load_parameter(s) between rclpy and rclcpp are different.

fujitatomoya:
@Barry-Xu-2018 what do you think about this? i think we have talked about this before.

Barry-Xu-2018:
> what do you think about this? i think we have talked about this before.

Agree with you. This python code is a duplicate implementation, and different implementations may lead to inconsistent behavior.

fujitatomoya:
@Barry-Xu-2018 can you take care of this?

aki1770-del:
@Barry-Xu-2018 — hi, quick ping on this. I know you did the parameter-YAML compatibility work in #1548 / #1549 and @fujitatomoya pointed this direction your way in February.                                                                                                     
                                                            
[...]

PR #1675, opened 2026-06-01
Preserve float type in parameter_dict_from_yaml_file

Lots-ninety-nine:
Towards #1672.

`_unpack_parameter_dict` round-tripped each parsed value through `str()` and `yaml.safe_load`, which lost `float` values whose `str()` representation lacks a decimal point (e.g. `0.00001` → `'1e-05'` reparsed as `str` by PyYAML 1.1). This builds the `ParameterValue` directly from the parsed Python value instead, preserving float type for both scalars and arrays.

Scope is limited to the rclpy-side round-trip. Unquoted `5e-06` / `1e+20` in YAML source are still parsed as strings by PyYAML 1.1 itself and are not addressed here — that is a PyYAML resolver limitation.

Regression test added; existing parameter tests pass.

### DIFF
--- a/rclpy/rclpy/parameter.py
+++ b/rclpy/rclpy/parameter.py
@@ -342,6 +342,46 @@ def parameter_value_to_python(parameter_value: ParameterValue) -> AllowableParam
         raise RuntimeError(f'unexpected parameter type {parameter_value.type}')
 
 
+def _parameter_value_from_python_value(value: Any) -> ParameterValue:
+    """
+    Build a ParameterValue from an already-parsed Python value.
+
+    Mirrors the type-inference of :func:`get_parameter_value` but skips the
+    ``str(value) -> yaml.safe_load`` round-trip. That round-trip lost
+    ``float`` values whose ``str()`` representation lacks a decimal point
+    (e.g. ``0.00001`` -> ``'1e-05'`` -> reparsed as ``str`` by PyYAML 1.1).
+    """
+    pv = ParameterValue()
+    if isinstance(value, bool):
+        pv.type = ParameterType.PARAMETER_BOOL
+        pv.bool_value = value
+    elif isinstance(value, int):
+        pv.type = ParameterType.PARAMETER_INTEGER
+        pv.integer_value = value
+    elif isinstance(value, float):
+        pv.type = ParameterType.PARAMETER_DOUBLE
+        pv.double_value = value
+    elif isinstance(value, list):
+        if all(isinstance(v, bool) for v in value):
+            pv.type = ParameterType.PARAMETER_BOOL_ARRAY
+            pv.bool_array_value = value
+        elif all(isinstance(v, int) for v in value):
+            pv.type = ParameterType.PARAMETER_INTEGER_ARRAY
+            pv.integer_array_value = value
+        elif all(isinstance(v, float) for v in value):
+            pv.type = ParameterType.PARAMETER_DOUBLE_ARRAY
+            pv.double_array_value = value
+        elif all(isinstance(v, str) for v in value):
+            pv.type = ParameterType.PARAMETER_STRING_ARRAY
+            pv.string_array_value = value
+        else:
+            pv.type = ParameterType.PARAMETER_STRING
+            pv.string_value = str(value)
+    else:
+        return get_parameter_value(str(value))
+    return pv
+
+
 def parameter_dict_from_yaml_file(
     parameter_file: str,
     use_wildcard: bool = False,
@@ -562,6 +602,6 @@ def _unpack_parameter_dict(namespace: str,
         else:
             parameter = ParameterMsg()
             parameter.name = full_param_name
-            parameter.value = get_parameter_value(str(param_value))
+            parameter.value = _parameter_value_from_python_value(param_value)
             parameters[full_param_name] = parameter
     return parameters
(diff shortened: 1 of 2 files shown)


fujitatomoya:
@Barry-Xu-2018 can you take a look at this?

rclpy/test/test_parameter.py
```diff
@@ -474,6 +474,32 @@ def test_parameter_dict_from_yaml_file(self) -> None:
...
+            node:
+                ros__parameters:
+                    tiny: 0.00001
+                    arr: [0.00001, 0.00002]
+            """
+        try:
+            with NamedTemporaryFile(mode='w', delete=False) as f:
+                f.write(yaml_string)
+                f.flush()
+                f.close()
```
Barry-Xu-2018: `with` will automatically call close().  
 By using pytest's tmp_path fixture, you can avoid manually creating and deleting temporary files. So try..finally can be removed.

```python
    temp_file = tmp_path / "test_params.yaml"
    temp_file.write_text(yaml_string)
    parameter_dict = parameter_dict_from_yaml_file(str(temp_file))
```

rclpy/test/test_parameter.py
```diff
@@ -474,6 +474,32 @@ def test_parameter_dict_from_yaml_file(self) -> None:
...
+                parameter_dict = parameter_dict_from_yaml_file(f.name)
+                assert parameter_dict['tiny'].value.type == ParameterType.PARAMETER_DOUBLE
+                assert parameter_dict['tiny'].value.double_value == 0.00001
```
Barry-Xu-2018:
```suggestion
                assert parameter_dict['tiny'].value.double_value == pytest.approx(
        0.00001
    )
```

rclpy/test/test_parameter.py
```diff
@@ -474,6 +474,32 @@ def test_parameter_dict_from_yaml_file(self) -> None:
...
+                assert parameter_dict['arr'].value.type == ParameterType.PARAMETER_DOUBLE_ARRAY
+                assert list(parameter_dict['arr'].value.double_array_value) == [
+                    0.00001, 0.00002]
```
Barry-Xu-2018:
```suggestion
                arr_values = list(parameter_dict["arr"].value.double_array_value)
                assert arr_values == pytest.approx([0.00001, 0.00002])
```

Barry-Xu-2018 (commented):
Some comments about test codes

Barry-Xu-2018:
LGTM. Thank you for your patch @Lots-ninety-nine

Barry-Xu-2018:
Pulls: https://github.com/ros2/rclpy/pull/1675
Gist: https://gist.githubusercontent.com/Barry-Xu-2018/b7bf6ed7353257611b68a0995631452b/raw/82b030e924eb81ac97d9fbd1aa053cc42029f1b4/ros2_rolling.repos
BUILD args: --packages-above-and-dependencies rclpy
TEST args: --packages-above rclpy
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19426/
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29253/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22038/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9134/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28196/)

rclpy/rclpy/parameter.py
```diff
@@ -562,6 +602,6 @@ def _unpack_parameter_dict(namespace: str,
         else:
             parameter = ParameterMsg()
             parameter.name = full_param_name
-            parameter.value = get_parameter_value(str(param_value))
+            parameter.value = _parameter_value_from_python_value(param_value)
```
fujitatomoya: is this supposed to be something like this? this internal function looks like that it introduces a second source of truth for the same Python→ParameterValue mapping?


```suggestion
            parameter.value = Parameter(param_name, value=param_value).to_parameter_msg().value
```

fujitatomoya (commented):
personally i do not really like this kind of workaround.

i still believe that https://github.com/ros2/rclpy/issues/1606 is the right way to proceed.

my understanding is that this PR can fix 1.0e+20 → double but 5e-06 → string, purely on whether the user wrote a decimal point. IMO that's a footgun, and it's the part that still diverges from the C++ --params-file loader (rcl_yaml_param_parser does its own number grammar and accepts 5e-06 as a double). IMO this introduce the more confusion for user.

Lots-ninety-nine:
Adopted in d5b7636, thanks. Routing through `Parameter.Type.from_parameter_value` also normalizes a few edge cases previously produced by the `str()` + `yaml.safe_load` round-trip:

- `[]` → `BYTE_ARRAY` (was `BOOL_ARRAY` via vacuous truth)
- `None` → `NOT_SET` (was string `'None'`)
- quoted scalars like `"true"` → preserved as `STRING` (were reinterpreted as `BOOL`)

All three now match the `Parameter` class semantics used elsewhere in rclpy.

fujitatomoya:
@Lots-ninety-nine i still think https://github.com/ros2/rclpy/pull/1675#pullrequestreview-4407122943 stands instead of this PR. i am not inclined to take this PR, can you instead address https://github.com/ros2/rclpy/issues/1606?

Lots-ninety-nine:
Thanks for the review @fujitatomoya and @Barry-Xu-2018. Understood — closing this PR. Happy to revisit if #1606 becomes the focus.

PR #1374, opened 2024-10-18
Fix a bug on unfinished task increasing while repeatedly calling spin_until_future_complete

Barry-Xu-2018:
Address a bug while investigating #1371 

If write action code like below,
```
            while rclpy.ok:
                future = action_client.send_goal(xxx)
                rclpy.spin_until_future_complete(action_client, future)
```

In spin_until_future_complete(), a callback is added to the done callback of future. And next spin_once_until_future_complete() will be called continuously. 

https://github.com/ros2/rclpy/blob/a09a0312cf9e2bbe0f6dc4467c72060a2fb74ce5/rclpy/rclpy/executors.py#L314-L332

Each time spin_once_until_future_complete() is called, the same callback is added to the future's done callback.
https://github.com/ros2/rclpy/blob/a09a0312cf9e2bbe0f6dc4467c72060a2fb74ce5/rclpy/rclpy/executors.py#L855-L861

This way, the future's done callbacks keep accumulating.

When a future is completed (calls set_result()), _schedule_or_invoke_done_callbacks() is called.
In this function, it creates a Task for each done callback, which leads to many unnecessary Tasks being created and consuming more and more memory.

https://github.com/ros2/rclpy/blob/a09a0312cf9e2bbe0f6dc4467c72060a2fb74ce5/rclpy/rclpy/task.py#L150-L153

So this PR is to remove the code that adds unnecessary done callbacks.

PR #1697, opened 2026-07-19
introduce ActionEndpointInfo to support "ros2 action info (-v)".

fujitatomoya:
## Description

Part of https://github.com/ros2/ros2cli/issues/1210

Fixes # (issue)

### Is this user-facing behavior change?

No, new methods and classes added.

### Did you use Generative AI?

Yes, Claude Fable

PR #1702, opened 2026-07-24
Fix future-wait callback accumulation on every spin path

hugogo1998:
`future.add_done_callback()` appends unconditionally, and every wait path here was calling it fresh on every spin. A single `spin_until_future_complete()` that takes N spins to finish was registering N callbacks on the future, and when it completes, `Future._schedule_or_invoke_done_callbacks()` creates one Task per callback. So waiting N times costs O(N) memory and O(N) work at completion, for no reason. Measured on real Humble rclpy at N=5000: 43.8 ms to complete a future carrying 5000 accumulated callbacks, each one becoming its own Task.

#1374 already fixed this for `spin_until_future_complete`'s internal loop, by rerouting it to a private method that doesn't re-add the callback. That's a real fix and it works: 0.31 ms at N=5000 on that path, matching what this change also gets. But `spin_once_until_future_complete` is public API, and a caller who writes their own wait loop around it directly (`while not future.done(): executor.spin_once_until_future_complete(future)`, a completely reasonable and documented way to use it) still hits the original bug, because that method still adds a fresh lambda on every call. Checked this on real Humble rclpy: looping it directly at N=5000 costs 44.7 ms and 5000 callbacks, basically unchanged from before #1374, because the fix never reached this path.

This registers one stable wake-on-done callback per Executor instance and adds it to a future at most once, checking `future._callbacks` before appending, through a single `_add_wake_on_done()` helper used everywhere a future is waited on: `spin_until_future_complete`, and both `SingleThreadedExecutor`'s and `MultiThreadedExecutor`'s `spin_once_until_future_complete`. Every spin path gets the same O(1) behavior instead of just one of them.

Measured on real Humble rclpy, N=5000:

| path | before | after |
|---|---|---|
| `spin_until_future_complete` | 0.31 ms, 1 callback (already fixed) | 0.30 ms, 1 callback |
| direct `spin_once_until_future_complete` loop | 44.7 ms, 5000 callbacks | 0.30 ms, 1 callback |

Ties the existing fix on the path it already covers, and closes the gap on the one it didn't, 40x-149x faster depending on N, since the leak scales with how many times a caller has to spin.

The wake-on-done semantics are unchanged: the callback's whole job is to break the executor out of its wait, and firing it once is exactly as sufficient as firing it N times, the future still completes with the correct result either way. This isn't a new mechanism, just making the existing one idempotent everywhere it's used, so it doesn't matter whether a caller uses `spin_until_future_complete` or writes their own loop around `spin_once_until_future_complete`, either way they get the fix.

Found via an automated perf-optimization pass profiling the future-completion path; the accumulation showed up directly in the profile as O(N) Task creation.

### DIFF
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -405,18 +405,8 @@ class Executor(ContextManager['Executor']):
             self._exit_spin()
 
     def _add_wake_on_done(self, future: Future[Any]) -> None:
-        """
-        Register a stable wake-on-done callback on a future, at most once.
-
-        future.add_done_callback() appends unconditionally, so calling it fresh on every
-        spin attempt (as every caller here used to) accumulates one callback per spin for
-        a future that takes many spins to complete: O(N) callbacks, each becoming its own
-        Task when the future finally completes. Reusing one callback instance and checking
-        it isn't already registered keeps this at O(1) regardless of how many times a
-        caller spins while waiting on the same future.
-        """
-        if self._wake_on_done_cb not in future._callbacks:
-            future.add_done_callback(self._wake_on_done_cb)
+        """Register the wake-on-done callback at most once per future."""
+        future.add_done_callback(self._wake_on_done_cb, unique=True)
 
     def spin_until_future_complete(
         self,
--- a/rclpy/rclpy/executors.py
+++ b/rclpy/rclpy/executors.py
@@ -211,6 +211,9 @@ class Executor(ContextManager['Executor']):
     def __init__(self, *, context: Optional[Context] = None) -> None:
         super().__init__()
         self._context = get_default_context() if context is None else context
+        # A single stable wake-on-done callback, registered at most once per future (see
+        # _add_wake_on_done), so repeated spins on the same future never accumulate callbacks.
+        self._wake_on_done_cb: Callable[[Future[Any]], None] = lambda future: self.wake()
         self._nodes: Set[Node] = set()
         self._nodes_lock = RLock()
         # all tasks that are not complete or canceled
@@ -401,6 +404,20 @@ class Executor(ContextManager['Executor']):
         finally:
             self._exit_spin()
 
+    def _add_wake_on_done(self, future: Future[Any]) -> None:
+        """
+        Register a stable wake-on-done callback on a future, at most once.
+
+        future.add_done_callback() appends unconditionally, so calling it fresh on every
+        spin attempt (as every caller here used to) accumulates one callback per spin for
+        a future that takes many spins to complete: O(N) callbacks, each becoming its own
+        Task when the future finally completes. Reusing one callback instance and checking
+        it isn't already registered keeps this at O(1) regardless of how many times a
+        caller spins while waiting on the same future.
+        """
+        if self._wake_on_done_cb not in future._callbacks:
+            future.add_done_callback(self._wake_on_done_cb)
+
     def spin_until_future_complete(
         self,
         future: Future[Any],
@@ -410,7 +427,7 @@ class Executor(ContextManager['Executor']):
         # Mark executor as spinning to prevent concurrent spins
         self._enter_spin()
         # Make sure the future wakes this executor when it is done
-        future.add_done_callback(lambda x: self.wake())
+        self._add_wake_on_done(future)
         try:
             if timeout_sec is None or timeout_sec < 0:
                 while (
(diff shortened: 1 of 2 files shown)


hugogo1998:
Hi @skyegalaxy , could you please take a look when you have time? I'd appreciate any feedback, and pls let me know if I overlook anything. Thanks

rclpy/rclpy/executors.py
```diff
@@ -401,6 +404,20 @@ def spin(self) -> None:
         finally:
             self._exit_spin()
 
+    def _add_wake_on_done(self, future: Future[Any]) -> None:
+        """
+        Register a stable wake-on-done callback on a future, at most once.
+
+        future.add_done_callback() appends unconditionally, so calling it fresh on every
+        spin attempt (as every caller here used to) accumulates one callback per spin for
```
skyegalaxy: nit: can make this comment less verbose

skyegalaxy (approved):

skyegalaxy:
Pulls: https://github.com/ros2/rclpy/pull/1702
Gist: https://gist.githubusercontent.com/skyegalaxy/bf89ad9cce444d4bb2333c37671b821e/raw/3c0c20d42f2659ff49f32a35d0612e4bd1429b96/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20092
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30119/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22850/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9968/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29005/)

hugogo1998:
Thanks for running CI @skyegalaxy! The 27 failures across the four platforms are all the same collection-time error, and it's unrelated to this change:

```
AttributeError: module 'rclpy.endpoint_info' has no attribute 'ActionEndpointInfo'
```

`ros2cli`'s xmlrpc marshaller imports `rclpy.endpoint_info.ActionEndpointInfo` at module load, so every ros2cli-based test (ros2node / ros2param / ros2service / ros2topic / ros2log, the daemon, sros2, rqt_topic, …) fails to import and gets counted as a failure. That symbol was added to rclpy in #1697, but this branch predated it (it was 3 commits behind `rolling`). rclpy's own executor/future tests all passed, and only the downstream CLI tools that need the newer symbol broke.

I've rebased onto the latest `rolling`, which brings in #1697, so `rclpy.endpoint_info.ActionEndpointInfo` now exists and the import resolves. The rebase cleanly combined this change with the recent `executors.py` edits on `rolling`. I also addressed your docstring nit on `_add_wake_on_done` (and trimmed the matching `__init__` comment).

Could you re-run CI when you get a chance? Thanks!

rclpy/rclpy/executors.py
```diff
@@ -401,6 +403,16 @@ def spin(self) -> None:
...
             self._exit_spin()
 
+    def _add_wake_on_done(self, future: Future[Any]) -> None:
+        """
+        Register the wake-on-done callback on a future, at most once.
+
+        Adding a fresh callback on every spin would accumulate one per spin (each becoming
+        its own Task on completion); reusing a single callback keeps it O(1).
+        """
+        if self._wake_on_done_cb not in future._callbacks:
```
fujitatomoya: rather than checking future._callbacks directly, could Future expose an idempotent add (performed under its lock)? that would make the dedup safe by construction in the multithreaded case, instead of relying on GIL list semantics to keep the unlocked read benign.

hugogo1998:
Good point. I moved the deduplication into `Future.add_done_callback(..., unique=True)`, so the check and append now happen together under the future's lock. The executor no longer reads `future._callbacks` directly.

rclpy/rclpy/executors.py
hugogo1998: That makes sense, thanks for catching this

fujitatomoya:
@Mergifyio rebase

fujitatomoya:
Pulls: ros2/rclpy#1702
Gist: https://gist.githubusercontent.com/fujitatomoya/b66f72057cb60f12bf9f71c2f4acc48c/raw/3c0c20d42f2659ff49f32a35d0612e4bd1429b96/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20165
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30236/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22930/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10046/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29098/)

ahcorde (approved):

fujitatomoya:
@Mergifyio backport lyrical

PR #1703, opened 2026-07-24
Avoid scanning pending requests on service response completion

### LINKED ISSUE #1709: Avoid scanning pending requests on service response completion
### Description

Service response completion currently calls `Client.remove_pending_request(future)`, which scans every pending request until it finds the matching future. Would the maintainers accept changing the internal done callback to bind the request sequence number and remove that entry directly?

### Motivation

The callback runs for every completed service request. A direct lookup keeps this path constant-time when many requests are in flight. In a ROS Rolling source build, the median callback cost was:

| Pending requests | Current scan | Direct removal |
|---:|---:|---:|
| 8 | 343 ns | 192 ns |
| 32 | 799 ns | 192 ns |
| 128 | 2,399 ns | 195 ns |
| 256 | 4,759 ns | 190 ns |
| 512 | 9,674 ns | 185 ns |

### Design / Implementation Considerations

The sequence number can be bound into the done callback with `functools.partial`. This avoids adding client-specific state to the general-purpose `Future` class, as discussed on #1703.

The direct removal should also confirm that the dictionary entry is still the same future before deleting it. That prevents a delayed callback from removing a newer request if a sequence number is ever reused. The public `remove_pending_request(future)` API can remain unchanged for compatibility.

I have a [two-file implementation and regression test](https://github.com/hugogo1998/rclpy/commit/5adc90e8a56f8b4d03c12494154ec71cecf7cc33) prepared against current `rolling`. It builds from source and the full `rclpy` test run reports 1,364 tests, 0 failures, and 0 errors.

### Additional Information

This follows up on the closed #1703. I am opening the issue first to confirm that the optimization and implementation approach are still wanted before asking to reopen or submitting a replacement PR.

The implementation and test were assisted by OpenAI Codex GPT-5. I reviewed the diff and ran the full test suite locally in a ROS Rolling environment.

hugogo1998:
## Description

Service response completion currently removes its pending request by scanning the dictionary for the completed future. Bind the request sequence number into the done callback instead, so normal completion can remove the entry directly.

The public `remove_pending_request(future)` method remains unchanged. The direct removal also checks that the sequence number still maps to the same future, so a delayed callback cannot remove a newer request if the sequence number is reused.

Fixes #1709

Yes. OpenAI Codex (GPT-5) assisted with the implementation and regression test. I reviewed the final diff and verified it in a ROS Rolling source build.

### Additional Information

The regression test covers sequence-number reuse. The existing concurrent service-call test also exercises the new done-callback path.

Validation:

```text
colcon build --packages-up-to rclpy --cmake-args -DBUILD_TESTING=ON
colcon test --packages-select rclpy
colcon test-result --verbose

Summary: 1364 tests, 0 errors, 0 failures, 79 skipped
```

Median callback time in the same Rolling environment:

| Pending requests | Current scan | Direct removal |
|---:|---:|---:|
| 8 | 343 ns | 192 ns |
| 32 | 799 ns | 192 ns |
| 128 | 2,399 ns | 195 ns |
| 256 | 4,759 ns | 190 ns |
| 512 | 9,674 ns | 185 ns |

This revision addresses the review feedback on the original PR by keeping client bookkeeping out of the general-purpose `Future` class. Issue #1709 was opened before requesting another review so the approach can be confirmed first.

### DIFF
--- a/rclpy/rclpy/client.py
+++ b/rclpy/rclpy/client.py
@@ -236,6 +236,9 @@ class Client(BaseClient[SrvRequestT, SrvResponseT]):
 
             future = Future[SrvResponseT]()
             self._pending_requests[sequence_number] = future
+            # Stash the sequence number on the future so remove_pending_request() is O(1)
+            # instead of scanning _pending_requests to find it by identity.
+            future._sequence_number = sequence_number
 
             future.add_done_callback(self.remove_pending_request)
 
@@ -261,10 +264,10 @@ class Client(BaseClient[SrvRequestT, SrvResponseT]):
         :param future: A future returned from :meth:`call_async`
         """
         with self._lock:
-            for seq, req_future in self._pending_requests.items():
-                if future is req_future:
-                    del self._pending_requests[seq]
-                    break
+            # O(1): the sequence number was stashed on the future by call_async, so drop it
+            # directly instead of scanning _pending_requests to find it by identity.
+            if future._sequence_number is not None:
+                self._pending_requests.pop(future._sequence_number, None)
 
     def wait_for_service(self, timeout_sec: Optional[float] = None) -> bool:
         """
--- a/rclpy/rclpy/task.py
+++ b/rclpy/rclpy/task.py
@@ -58,6 +58,9 @@ class Future(Generic[T]):
         self._executor: Optional[Union[weakref.ReferenceType['Executor'],
                                        Callable[[], None]]] = None
         self._set_executor(executor)
+        # Used by Client to remove this future from its pending-requests dict in O(1)
+        # instead of scanning it by identity; unused/None outside that context.
+        self._sequence_number: Optional[int] = None
 
     def __del__(self) -> None:
         if self._exception is not None and not self._exception_fetched:


rclpy/rclpy/task.py
```diff
@@ -58,6 +58,9 @@ def __init__(self, *, executor: Optional['Executor'] = None) -> None:
         self._executor: Optional[Union[weakref.ReferenceType['Executor'],
                                        Callable[[], None]]] = None
         self._set_executor(executor)
+        # Used by Client to remove this future from its pending-requests dict in O(1)
+        # instead of scanning it by identity; unused/None outside that context.
+        self._sequence_number: Optional[int] = None
```
fujitatomoya, line 63: minor concern here is that, `Future` is a general purpose class used by other classes such as executors, timers and actions. adding client-only field to it leaks `Client` bookkeepiong into shared object.

fujitatomoya (approved):
just one minor comment, lgtm with green CI. i would like to have 2nd review on this.

rclpy/rclpy/client.py
```diff
@@ -236,6 +236,9 @@ def call_async(self, request: SrvRequestT) -> Future[SrvResponseT]:
 
             future = Future[SrvResponseT]()
             self._pending_requests[sequence_number] = future
+            # Stash the sequence number on the future so remove_pending_request() is O(1)
+            # instead of scanning _pending_requests to find it by identity.
+            future._sequence_number = sequence_number
 
             future.add_done_callback(self.remove_pending_request)
```
nadavelkabets, line 243: I agree with @fujitatomoya on leaking a client specific attribute to the generic Future class. 
Perhaps this would be better (adding a new method to remove by sequence):

```suggestion
            future.add_done_callback(partial(self._remove_request_by_sequence, sequence_number))
```
Another alternative would be to construct a reversed pending_requests dict where future is the key, but I suspect that get_pending_request is rarely used by users directly, so the above should solve that for most of the cases.

nadavelkabets (changes requested):

fujitatomoya:
see https://github.com/ros2/rclpy/pull/1704#issuecomment-5098619471

PR #1707, opened 2026-08-01
Migrate from pybind11 to nanobind

sea-bass:
## Description

Following a thread on ROS Discourse, where I suggested upgrading from `pybind11` to `nanobind`, I figured I'd give this a try.

Relevant context: https://nanobind.readthedocs.io/en/latest/why.html

Some additional changes besides pure porting:
* Note that nanobind offers a [CMake macro for Python type stub generation](https://nanobind.readthedocs.io/en/latest/typing.html#stub-generation), so we don't have to manually update it anymore.
* There were several places in the binding where it was specified that an argument _could_ be `None`; however, nanobind provides for-free bindings for this if you just make it a `std::optional` on the C++ side.

Note that `nanobind-dev` as a dependency only transitioned from pip to apt on Ubuntu 24.04, but the 24.04 version is too old for stub generation. So I think on Lyrical onwards this PR is good to go. My recommendation would be not to backport this, though, and simply go forward-looking.

It shouldn't be for the end user, unless there's a bug (or an improvement!) that is noticeable.

Yes, Claude Fable 5.

I of course read through all the changes (except for the autogenerated `.pyi` stub files), identified areas of simplification/inefficiency, and iterated.

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -144,7 +144,8 @@ nanobind_add_stub(_rclpy_nanobind_stub
   MODULE _rclpy_nanobind
   OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/rclpy/impl/_rclpy_nanobind.pyi"
   PYTHON_PATH "${CMAKE_CURRENT_BINARY_DIR}/test_rclpy"
-  DEPENDS _rclpy_nanobind
+  PATTERN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/stubgen_pattern.pat"
+  DEPENDS _rclpy_nanobind "${CMAKE_CURRENT_SOURCE_DIR}/stubgen_pattern.pat"
 )
 nanobind_add_stub(_rclpy_nanobind_service_introspection_stub
   MODULE _rclpy_nanobind.service_introspection
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -38,10 +38,16 @@ find_package(rosidl_runtime_c REQUIRED)
 cmake_policy(SET CMP0094 NEW)
 set(Python3_FIND_UNVERSIONED_NAMES FIRST)
 
-# Find python before pybind11
+# Find python before nanobind
 find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
 
-find_package(pybind11 REQUIRED)
+# nanobind's CMake config requires the new-style FindPython module with the
+# Python::Module target defined. Reuse the interpreter found above so both
+# find_package calls agree on the same Python installation.
+set(Python_EXECUTABLE "${Python3_EXECUTABLE}")
+find_package(Python REQUIRED COMPONENTS Interpreter Development.Module)
+
+find_package(nanobind REQUIRED)
 
 # enables using the Python extensions from the build space for testing
 file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/test_rclpy/__init__.py" "")
@@ -69,10 +75,10 @@ function(configure_build_install_location _library_name)
   )
 endfunction()
 
-# Split from main extension and converted to pybind11
-pybind11_add_module(_rclpy_pybind11
+# Split from main extension and converted to nanobind
+nanobind_add_module(_rclpy_nanobind
   src/rclpy/_rclpy_logging.cpp
-  src/rclpy/_rclpy_pybind11.cpp
+  src/rclpy/_rclpy_nanobind.cpp
   src/rclpy/action_client.cpp
   src/rclpy/action_goal_handle.cpp
   src/rclpy/action_server.cpp
@@ -111,13 +117,13 @@ pybind11_add_module(_rclpy_pybind11
 )
 
 if(CMAKE_C_COMPILER_ID MATCHES "Clang" AND NOT APPLE)
-  target_link_libraries(_rclpy_pybind11 PRIVATE atomic)
+  target_link_libraries(_rclpy_nanobind PRIVATE atomic)
 endif()
 
-target_include_directories(_rclpy_pybind11 PRIVATE
+target_include_directories(_rclpy_nanobind PRIVATE
   src/rclpy/
 )
-target_link_libraries(_rclpy_pybind11 PRIVATE
+target_link_libraries(_rclpy_nanobind PRIVATE
   ament_cmake_ros_core::ament_ros_cxx_standard
   lifecycle_msgs::lifecycle_msgs__rosidl_generator_c
   lifecycle_msgs::lifecycle_msgs__rosidl_typesupport_c
@@ -130,7 +136,7 @@ target_link_libraries(_rclpy_pybind11 PRIVATE
   rcutils::rcutils
   rosidl_runtime_c::rosidl_runtime_c
 )
-configure_build_install_location(_rclpy_pybind11)
+configure_build_install_location(_rclpy_nanobind)
 
 if(NOT WIN32)
   ament_environment_hooks(
@@ -175,7 +181,7 @@ if(BUILD_TESTING)
   target_include_directories(test_python_allocator PRIVATE src/rclpy)
   target_link_libraries(test_python_allocator
     ament_cmake_ros_core::ament_ros_cxx_standard
-    pybind11::embed)
+    Python3::Python)
 
   if(NOT _typesupport_impls STREQUAL "")
     # Run each test in its own pytest invocation to isolate any global state in rclpy
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -138,6 +138,21 @@ target_link_libraries(_rclpy_nanobind PRIVATE
 )
 configure_build_install_location(_rclpy_nanobind)
 
+# Generate the type stubs for the extension module into the source tree,
+# so that they stay committed alongside the bindings.
+nanobind_add_stub(_rclpy_nanobind_stub
+  MODULE _rclpy_nanobind
+  OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/rclpy/impl/_rclpy_nanobind.pyi"
+  PYTHON_PATH "${CMAKE_CURRENT_BINARY_DIR}/test_rclpy"
+  DEPENDS _rclpy_nanobind
+)
+nanobind_add_stub(_rclpy_nanobind_service_introspection_stub
+  MODULE _rclpy_nanobind.service_introspection
+  OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/rclpy/impl/service_introspection.pyi"
+  PYTHON_PATH "${CMAKE_CURRENT_BINARY_DIR}/test_rclpy"
+  DEPENDS _rclpy_nanobind
+)
+
 if(NOT WIN32)
   ament_environment_hooks(
     "${ament_cmake_package_templates_ENVIRONMENT_HOOK_LIBRARY_PATH}"
--- a/rclpy/package.xml
+++ b/rclpy/package.xml
@@ -25,7 +25,7 @@
   <build_depend>rcl_lifecycle</build_depend>
   <build_depend>rcl_logging_interface</build_depend>
   <build_depend>rcl_yaml_param_parser</build_depend>
-  <build_depend>pybind11-dev</build_depend>
+  <build_depend>nanobind-dev</build_depend>
   <build_depend>python3-dev</build_depend>
   <build_depend>rcpputils</build_depend>
   <build_depend>rcutils</build_depend>
--- a/rclpy/rclpy/duration.py
+++ b/rclpy/rclpy/duration.py
@@ -32,7 +32,7 @@ class Duration:
         total_nanoseconds = int(seconds * S_TO_NS)
         total_nanoseconds += int(nanoseconds)
         if total_nanoseconds >= 2**63 or total_nanoseconds < -2**63:
-            # pybind11 would raise TypeError, but we want OverflowError
+            # nanobind would raise TypeError, but we want OverflowError
             raise OverflowError(
                 'Total nanoseconds value is too large to store in C duration.')
         self._duration_handle = _rclpy.rcl_duration_t(total_nanoseconds)
--- a/rclpy/rclpy/experimental/events_executor.py
+++ b/rclpy/rclpy/experimental/events_executor.py
@@ -29,15 +29,6 @@ def EventsExecutor(*, context: typing.Optional[rclpy.Context] = None) -> rclpy.e
     # Python backtrace dumped with the crash.
     faulthandler.enable()
 
-    ex = typing.cast(rclpy.executors.Executor, _rclpy.EventsExecutor(context))
-
-    # rclpy.Executor does this too.  Note, the context itself is smart enough to check
-    # for bound methods, and check whether the instances they're bound to still exist at
-    # callback time, so we don't have to worry about tearing down this stale callback at
-    # destruction time.
-    # TODO(bmartin427) This should really be done inside of the EventsExecutor
-    # implementation itself, but I'm unable to figure out a pybind11 incantation that
-    # allows me to pass this bound method call from C++.
-    context.on_shutdown(ex.wake)
-
-    return ex
+    # Note the EventsExecutor implementation takes care of registering a wake-on-shutdown
+    # callback with the context.
+    return typing.cast(rclpy.executors.Executor, _rclpy.EventsExecutor(context))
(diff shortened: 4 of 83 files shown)


mjcarroll:
Interesting stuff. Did you see any performance improvements out of the gate with this?

sea-bass:
> Interesting stuff. Did you see any performance improvements out of the gate with this?

Yes! I kicked off a pretty rough vibe-benchmark while working on the PR and the general takeaway was twofold:

1. Anything that relies on the C API under the hood doing the heavy lifting is pretty much the same.
2. The more bindings-heavy calls like getting the current time saw up to 3-4x speedup due to nanobind usage.

Other less performance related improvements to reiterate are that binary size is smaller and .pyi stub generation is automatic.

sloretz:
I think we would absolutely consider merging this if the benefits are significant.

* How much faster at runtime is it?
* How much smaller are the binaries?

> Note that nanobind-dev as a dependency only transitioned from pip to apt on Ubuntu 24.04, but the 24.04 version is too old for stub generation

For a change this big, I would recommend against backporting, but it could go into Rolling and beyond. Lyrical and Rolling currently share target platforms. Looking at the [Lyrical target platforms](
https://docs.ros.org/en/lyrical/Releases/lyrical/supported-platforms.html), do these platforms all have a new enough nanobind version?

* Ubuntu Resolute (26.04) (Seems like yes)
* Windows 11 (via pixi / conda-forge)
* macOS (via pixi / conda-forge)
* Debian Trixie
* OpenEmbedded

> .pyi stub generation is automatic.

If stub generation is automatic, do we need `rclpy/rclpy/impl/_rclpy_nanobind.pyi`, or can we generate it?

sea-bass:
When I find some more time I will post some numbers on perf and binary size. In the meantime...

> Looking at the [Lyrical target platforms](https://docs.ros.org/en/lyrical/Releases/lyrical/supported-platforms.html), do these platforms all have a new enough nanobind version?
> 
> * Ubuntu Resolute (26.04) (Seems like yes)
> * Windows 11 (via pixi / conda-forge)
> * macOS (via pixi / conda-forge)
> * Debian Trixie
> * OpenEmbedded

Yes on the first 4 -- I believe nanobind 2.0 is the cutoff, so I see 2.5.0 and newer on all the others.

How do I check OpenEmbedded? My cursory search only had `python3-nanobind` show up but not `nanobind-dev`, which is the one we need. But I also don't really know where to look, to be honest.

EDIT: Looking at rosdistro we may just need to copy how pybind11-dev did it, by adding this to the nanobind-dev key, and we should be good.

```
openembedded: [python3-nanobind@meta-python]
```

However, there are still a bunch of other platforms like rhel, etc. that don't yet have a `nanobind-dev` key so we probably want to add those to `rosdistro` pretty extensively so we don't take down `rclpy` support for existing platforms...

> If stub generation is automatic, do we need `rclpy/rclpy/impl/_rclpy_nanobind.pyi`, or can we generate it?

It is generated! But it's also versioned with the code. This way people get IDE support more easily.

Additionally, it lends itself to testing. For example, in my repo that uses nanobind we have a CI check that builds the code and runs the tests (as normal) and then does a `git diff` to check whether the user updated the stubs by building the source code on their machine.

sea-bass:
Update, I left Claude running a benchmark in the background and it produced a nice report: https://claude.ai/code/artifact/f5514863-a0c8-42fe-81a7-e7bb6442eb04

These are the "headline numbers" I got in my CLI:

``` 
  ┌───────────────────────────────────┬──────────────────────────┬────────────────────┬──────────────┐
  │              Metric               │    pybind11 (de802f7)    │ nanobind (21f8a4d) │    Change    │
  ├───────────────────────────────────┼──────────────────────────┼────────────────────┼──────────────┤
  │ Extension .so                     │                1,484 KiB │            738 KiB │       −50.2% │
  ├───────────────────────────────────┼──────────────────────────┼────────────────────┼──────────────┤
  │ .text (generated code)            │                1,071 KiB │            412 KiB │       −61.5% │
  ├───────────────────────────────────┼──────────────────────────┼────────────────────┼──────────────┤
  │ Geomean over 79 method benchmarks │                        — │                  — │ 1.92× faster │
  ├───────────────────────────────────┼──────────────────────────┼────────────────────┼──────────────┤
  │ Executor spin_once (ready timer)  │                  48.8 µs │            33.9 µs │        1.44× │
  ├───────────────────────────────────┼──────────────────────────┼────────────────────┼──────────────┤
  │ Service round-trip (loopback)     │                 121.3 µs │            88.8 µs │        1.37× │
  ├───────────────────────────────────┼──────────────────────────┼────────────────────┼──────────────┤
  │ import rclpy / cold start / RSS   │ ~55 ms / 125 ms / 66 MiB │               same │       ~1.00× │
  └───────────────────────────────────┴──────────────────────────┴────────────────────┴──────────────┘
```

quite encouraging TBH!

skyegalaxy:
@skyegalaxy to put on the client library WG agenda

jmachowinski:
We discussed this and we are all in favor of moving forward with this.
One thing that came up though was that we should merge this together with also switching the rosidl_py bindings to nanobind.
@sea-bass Would you be interested to look into this as well ?

sea-bass:
Awesome! I could take a look. It is this repo?

https://github.com/ros2/rosidl/tree/rolling/rosidl_buffer_py

jmachowinski:
https://github.com/ros2/rosidl_python/tree/rolling/rosidl_generator_py

That is the generator. It generates the bindings. 

https://github.com/ros2/rosidl_python/pull/193
This PR might collide with this work. It has been stale for years but now we got movement again...

I also have some experimental code around that speed up the existing implementation by ~3x if using pybind that is not that intrusive...

jmachowinski:
https://github.com/ros2/rosidl_python/pull/265

I guess the same optimizations might work for nanobind as well.

sea-bass:
Cool! Would it make sense to land those 2 PRs to `rosidl_python` first, and then we can start a nanobind migration there?

I guess we probably don't want to merge this until both repos are in workable states w nanobind?

sea-bass:
Sorry, I just glanced at those repos and I don't think I understand.

Is there anything there that currently uses pybind11 which would need porting to nanobind? I don't see any binding code at all in the repo or the linked PR. Just CPython stuff which is a level of abstraction down.

Is the idea to change the CPython usage there to nanobind instead?

jmachowinski:
Pulls: ros2/rclpy#1707
Gist: https://gist.githubusercontent.com/jmachowinski/9d0c4e3cc66fb17a604ca57948ab8d39/raw/cb405781297f6b2de4383911b0c56b059129d710/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20161
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30232/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22926/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10042/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29094/)

jmachowinski:
You are right. I was under the assumption that the 'other' python bindings would use the same framework, but this obviously is not the case. 
I think then we are good to go here. I started the CI to see if everything works as expected.

jmachowinski (approved):

jmachowinski:
The CI does not find nanobind, are we missing nanobind-dev as a key for rosdep ?

sea-bass:
I added `nanobind-dev` to the `package.xml`, is CI calling rosdep install / is there a cached set of dependencies being used? Reading the jenkins logs it's still installing pybind11 via apt?

Here's the rosdistro entries: https://github.com/ros/rosdistro/blob/82eefa281fd5702f1869408c66606d395cd635ee/rosdep/base.yaml#L8554

Also re: finding it as a key there may be some missing OSes in our non tier 1 platforms. See: https://github.com/ros2/rclpy/pull/1707#issuecomment-5208250411

jmachowinski:
@mjcarroll @claraberendsen any idea what's going on with the nanobind dependency ?

InvincibleRMC:
These new autogenerated stubs have some regressions around generics. I will check out if nanobind has a nice way of handling it.

sea-bass:
> These new autogenerated stubs have some regressions around generics. I will check out if nanobind has a nice way of handling it.

@InvincibleRMC maybe we can use `nb::sig()` / `nb::typed()` as shown here?

https://nanobind.readthedocs.io/en/latest/api_core.html#_CPPv4N8nanobind3sigE
https://nanobind.readthedocs.io/en/latest/api_core.html#_CPPv4I0DpEN8nanobind5typedE

Combined with this if needed: https://nanobind.readthedocs.io/en/latest/typing.html#pattern-files

InvincibleRMC:
> 
> @InvincibleRMC maybe we can use `nb::sig()` / `nb::typed()` as shown here?
> 

Yeah nb::sig will work, but it means a large amount of the .pyi file is going to be hand maintained (just in cpp instead of .pyi)

sea-bass:
A large amount is better than the whole thing, I guess?

I've just pushed up a commit that proves this out, and it's... not bad, and also not that substantial a portion of the CPP binding code. There are some gaps still, compared to the manual stub curation, but I really would like to hold on to the hope that this can be somewhat automated.

Let me know your thoughts?

claraberendsen:
@sea-bass @jmachowinski CI uses a custom set of dependencies defined on ros2/ci and pixi.toml for Windows. It doesn't use rosdep so there is some manual steps to test this:
Here are the changes needed: 
- https://github.com/ros2/ci/pull/890
- https://github.com/ros2/ros2/pull/1856

Two things to note: on conda-forge there isn't available the exact version as on Resolute so I bumped it to the next one (feel free to modify/point me to what version we want). Is this package available for RHEL? Otherwise CI for it will fail. 

And a fresh run/rebuild  of ci with those branches as base: 

- ci_linux: [Build Status](http://ci.ros2.org/job/ci_linux/30303/)
- ci_linux_aarch:  [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23003/)
- ci_windows: [Build Status](http://ci.ros2.org/job/ci_windows/29189/)

claraberendsen:
Rebuild with side-by-side installation of pybind and nanobind. Windows hasn't triggered yet so the previous build should pick up the changes already.


- ci_linux: [Build Status](http://ci.ros2.org/job/ci_linux/30308/)
- ci_linux_aarch:  [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23004/)

sea-bass:
yay, seems like build passed for Linux; the Windows one seemingly was failing on an unrelated `ament_cmake_gen_version_h` package?

claraberendsen:
I did not look closely at the tests but there are new compiler warnings that seem to be caused by this. 
Here is another run of windows to check if the transient error is resolved. 

- ci_windows: [Build Status](http://ci.ros2.org/job/ci_windows/29217/)

sea-bass:
No space left on disk in the latest windows run :(

claraberendsen:
- ci_windows: [Build Status](http://ci.ros2.org/job/ci_windows/29225/)

sea-bass:
Same deal failure-wise:

```
ament_cmake_gen_version_h (stderr)
--- stderr: ament_cmake_gen_version_h
CMake Error at C:/pixi_ws/.pixi/envs/default/Library/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find GTest (missing: GTEST_LIBRARY GTEST_INCLUDE_DIR
  GTEST_MAIN_LIBRARY)
Call Stack (most recent call first):
  C:/pixi_ws/.pixi/envs/default/Library/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  C:/pixi_ws/.pixi/envs/default/Library/share/cmake-3.28/Modules/FindGTest.cmake:270 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  C:/ci/ws/install/share/ament_cmake_gtest/cmake/ament_cmake_gtest-extras.cmake:23 (find_package)
  C:/ci/ws/install/share/ament_cmake_gtest/cmake/ament_add_gtest_executable.cmake:35 (_ament_cmake_gtest_find_gtest)
  C:/ci/ws/install/share/ament_cmake_gtest/cmake/ament_add_gtest.cmake:68 (ament_add_gtest_executable)
  CMakeLists.txt:26 (ament_add_gtest)
```

It appears GTest is not being found when building `ament_cmake_gen_version_h`, which is bizarre given there is definitely a dependency: https://github.com/ament/ament_cmake/blob/eb6039077fc34e16e3e34edf8a8f49dc5e127c16/ament_cmake_gen_version_h/package.xml#L19

sea-bass:
Any updates here? I've had to resolve a few conflicts over the weeks, so wondering if I can do anything to help get this to the finish line.

jmachowinski:
Pulls: ros2/rclpy#1707
Gist: https://gist.githubusercontent.com/jmachowinski/ea7d07588631bc60fc5fd92e214174f0/raw/cb405781297f6b2de4383911b0c56b059129d710/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20397
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30495/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23187/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10292/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29353/)

sea-bass:
bah Windows CI didn't find nanobind. sad.

EDIT: None of them did... are we supposed to run on a different branch of `ros2_ci`?

jmachowinski:
@claraberendsen What is the trick to run the 'correct' ci ?

claraberendsen:
@jmachowinski for linux you need to run with a custom `ros2/ci` branch that has the correct dependencies installation on the Dockerfile. 
For Windows you need a custom pixi.toml url to have the dependency as well.

jmachowinski:
@claraberendsen Is there a way to do that using the ros-ci-for-pr script ?

claraberendsen:
@jmachowinski I don't think those arguments are mapped right now. It might be worth mapping them on `ros-ci-for-pr` to enable these type of scenarios. 
For now what I did was trigger a ci_launcher with the correct parameters (which is what `ros-ci-for-pr` does behind the hood): https://ci.ros2.org/job/ci_launcher/20447


* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30550/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23237/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10344/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29404/)

claraberendsen:
The Windows build failed because the base of the ros2 branch was outdated. 
Rebuilding with branch rebased to rolling: [Build Status](https://ci.ros2.org/job/ci_windows/29406/)

For Linux RHEL I see the job installed correctly `python3-nanobind-2.10.2-1.el10_2.noarch` is this  https://github.com/ros2/ci/commit/2afde91dc911bc405057440c55bdee8b698ffadb the correct dependency for RHEL @sea-bass ?

sea-bass:
> For Linux RHEL I see the job installed correctly `python3-nanobind-2.10.2-1.el10_2.noarch` is this [ros2/ci@2afde91](https://github.com/ros2/ci/commit/2afde91dc911bc405057440c55bdee8b698ffadb) the correct dependency for RHEL @sea-bass ?

As long as nanobind is at/above version 2.0, which has the stubs generation, it should be fine?

I believe rolling on Ubuntu is on version 2.9.2 or something, so it seems compatible! Unsure why the RHEL job didn't pick it up?

sea-bass:
ugh okay @claraberendsen @jmachowinski I think the issue is that while the other OSes install nanobind as a system library, RHEL installs it as a python package. So the "correct" CMake for that environment is:

```cmake
# 1. Find Python
find_package(Python 3.8 REQUIRED COMPONENTS Interpreter Development.Module)

# 2. Get the CMake path directly from nanobind
execute_process(
    COMMAND "${Python_EXECUTABLE}" -m nanobind --cmake_dir
    OUTPUT_STRIP_TRAILING_WHITESPACE
    OUTPUT_VARIABLE nanobind_ROOT
)

# 3. Now find_package will succeed
find_package(nanobind CONFIG REQUIRED)
```

rather than ~basically what we have now

```cmake
find_package(Python 3.8 REQUIRED COMPONENTS Interpreter Development.Module)

find_package(nanobind CONFIG REQUIRED)
```

I don't really know what the best step forward is here... I just pushed a commit we can try in the RHEL case.

clalancette:
> I don't really know what the best step forward is here... I just pushed a commit we can try in the RHEL case.

Looking at RHEL, I think what you've done here is the best we can do for now, so I'd say we move forward with it.

clalancette:
I also checked out Windows pixi nanobind, and the same CMake change should in theory work there as well.

PR #1629, opened 2026-03-18
Feature: streamlined entity destroy

nadavelkabets:
Part of #1620
## Changes
* Unified `destroy()` pattern across all entity base classes (BasePublisher, BaseSubscription, BaseClient, BaseService, BaseTimer)
* Each base class `destroy()` is now idempotent via `_destroyed` flag, calls an `on_destroy` callback to notify the owning node, then calls `_destroy()` for handle cleanup
* `Node.destroy_*()` methods now delegate entirely to `entity.destroy()` — node-list removal and executor wake-up happen via `on_destroy` callbacks instead of being duplicated in the node
* Entities can be destroyed directly (`entity.destroy()`) without going through `Node.destroy_*()`, enabling AsyncNode to use the same entity classes

## Notes
* `on_destroy` is an optional keyword-only argument defaulting to `None`, so all existing third-party code constructing entities directly continues to work
* The `.. warning:: Users should not destroy ...` docstrings are removed — direct `entity.destroy()` is now the supported API
* `Node.destroy_*()` methods no longer catch `InvalidHandle` — double-destroy is handled by the idempotent `_destroyed` flag instead

PR #1711, opened 2026-08-14
fix: node lingering after destroy

### LINKED ISSUE #1710: Destroyed node does not leave the graph until process exit
### Generated by Generative AI

_No response_

### Operating System:

Ubuntu 26.04 Resolute

### ROS version or commit hash:

lyrical, rolling

### RMW implementation (if applicable):

N/A

### RMW Configuration (if applicable):

_No response_

### Client library (if applicable):

rclpy

### 'ros2 doctor --report' output

N/A

### Steps to reproduce issue

```py
import sys
import time

import rclpy
import rclpy.context

TIMEOUT = 10.0


def main() -> int:
    rclpy.init(context=context)

    observer = rclpy.create_node('observer', context=context)
    doomed = rclpy.create_node('doomed_node', context=context)

    def in_graph() -> bool:
        return any(name == 'doomed_node' for name, _ in observer.get_node_names_and_namespaces())

    deadline = time.monotonic() + TIMEOUT
    while not in_graph() and time.monotonic() < deadline:
        time.sleep(0.05)
    if not in_graph():
        print('FAIL: node never appeared in the graph; environment problem, not this bug')
        return 2
    print('node is in the graph, calling destroy_node()')

    doomed.destroy_node()

    deadline = time.monotonic() + TIMEOUT
    while in_graph() and time.monotonic() < deadline:
        time.sleep(0.05)

    if in_graph():
        print(f'BUG: node still in the graph {TIMEOUT:.0f}s after destroy_node()')
        result = 1
    else:
        print('OK: node left the graph after destroy_node()')
        result = 0

    observer.destroy_node()
    return result


if __name__ == '__main__':
    sys.exit(main())
```

### Expected behavior

After `rclpy.node.Node.destroy_node()`, the node resource is actually destroyed (`rcl_node_fini`) and leaves the graph.

### Actual behavior

The destroyed node lingers until process exits.

### Additional information

_No response_

emersonknapp:
## Description

Add back the explicit destruction of the `TypeDescriptionService` in `BaseNode`, which was removed in https://github.com/ros2/rclpy/pull/1629

Adds a regression test to cover the gap.

Fixes #1710 

In that it fixes an edge-case bug. Most users won't ever see it.

This PR is handwritten. 
The implementation was assisted by Claude Opus 5 with a human editing pass.

### DIFF
--- a/rclpy/CMakeLists.txt
+++ b/rclpy/CMakeLists.txt
@@ -221,6 +221,7 @@ if(BUILD_TESTING)
       test/test_timer.py
       test/test_topic_or_service_is_hidden.py
       test/test_topic_endpoint_info.py
+      test/test_type_description_service.py
       test/test_type_support.py
       test/test_type_hash.py
       test/test_utilities.py
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -2150,6 +2150,8 @@ class BaseNode(ABC):
     def destroy_node(self) -> None:
         self._context.untrack_node(self)
         self._parameter_event_publisher = None
+        # Not owned by self._services, so Node.destroy_node() doesn't destroy it
+        self._type_description_service.destroy()
         self.handle.destroy_when_not_in_use()
 
 
--- a/rclpy/rclpy/node.py
+++ b/rclpy/rclpy/node.py
@@ -2150,8 +2150,6 @@ class BaseNode(ABC):
     def destroy_node(self) -> None:
         self._context.untrack_node(self)
         self._parameter_event_publisher = None
-        # Not owned by self._services, so Node.destroy_node() doesn't destroy it
-        self._type_description_service.destroy()
         self.handle.destroy_when_not_in_use()
 
 
--- a/rclpy/rclpy/type_description_service.py
+++ b/rclpy/rclpy/type_description_service.py
@@ -76,13 +76,6 @@ class TypeDescriptionService:
         if self.enabled:
             self._start_service()
 
-    def destroy(self) -> None:
-        # Required manual destruction because this is not managed by rclpy.Service.
-        # The underlying C++ holds a shared node handle, so leaving this alive keeps node alive.
-        if self._type_description_srv is not None:
-            self._type_description_srv.destroy_when_not_in_use()
-            self._type_description_srv = None
-
     def _start_service(self) -> None:
         node = self._get_node()
         self._type_description_srv = _rclpy.TypeDescriptionService(node.handle)
--- a/rclpy/rclpy/type_description_service.py
+++ b/rclpy/rclpy/type_description_service.py
@@ -76,6 +76,13 @@ class TypeDescriptionService:
         if self.enabled:
             self._start_service()
 
+    def destroy(self) -> None:
+        # Required manual destruction because this is not managed by rclpy.Service.
+        # The underlying C++ holds a shared node handle, so leaving this alive keeps node alive.
+        if self._type_description_srv is not None:
+            self._type_description_srv.destroy_when_not_in_use()
+            self._type_description_srv = None
+
     def _start_service(self) -> None:
         node = self._get_node()
         self._type_description_srv = _rclpy.TypeDescriptionService(node.handle)
--- a/rclpy/src/rclpy/type_description_service.cpp
+++ b/rclpy/src/rclpy/type_description_service.cpp
@@ -37,9 +37,9 @@ TypeDescriptionService::TypeDescriptionService(Node & node)
   service_ = std::make_shared<Service>(node, srv_ptr);
 }
 
-Service TypeDescriptionService::get_impl()
+std::shared_ptr<Service> TypeDescriptionService::get_impl()
 {
-  return *service_;
+  return service_;
 }
 
 py::object TypeDescriptionService::handle_request(
--- a/rclpy/src/rclpy/type_description_service.hpp
+++ b/rclpy/src/rclpy/type_description_service.hpp
@@ -43,7 +43,7 @@ public:
   /**
    * \return The capsule containing the Service
    */
-  Service
+  std::shared_ptr<Service>
   get_impl();
 
   /// Handle an incoming request to the service
--- a/rclpy/test/test_node.py
+++ b/rclpy/test/test_node.py
@@ -74,6 +74,16 @@ TEST_NAMESPACE = '/my_ns'
 TEST_RESOURCES_DIR = pathlib.Path(__file__).resolve().parent / 'resources' / 'test_node'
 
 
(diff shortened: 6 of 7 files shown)


emersonknapp:
@nadavelkabets you implemented the destroy cleanup - does this fix look right, or should I approach it a different way?

nadavelkabets:
> @nadavelkabets you implemented the destroy cleanup - does this fix look right, or should I approach it a different way?


Hey!

This is definitely an issue introduced in my refactor. `node.destroy_node()` destroys `TypeDescriptionService.impl`, but `get_impl()` returns `Service` by value - so `.impl` returns a new copy every call. The node destroys the returned copy, but the original instance inside the C++ wrapper keeps its own share of the node handle, and that's what keeps the node alive.

Your fix reverts the regression and the node leaves the graph again. 
One thing it doesn't cover - the type description service itself still stays in the graph.
I believe this originates from a bug in the TypeDescriptionService constructor:

Service.cpp constructor:
https://github.com/ros2/rclpy/blob/fb4701b97e17073e188ccbea6346a2145d849365/rclpy/src/rclpy/service.cpp#L67-L78
But TypeDescriptionService.cpp constructs:
https://github.com/ros2/rclpy/blob/fb4701b97e17073e188ccbea6346a2145d849365/rclpy/src/rclpy/type_description_service.cpp#L32

`Service` installs an `rcl_service_fini` deleter, `TypeDescriptionService` just uses `make_shared`, so the handle is deleted without ever being finalized. 

Regarding the fix itself - I think this is a design dilemma. The fix itself is valid - with that said, I deliberately removed what you are adding back. It felt like a patch to me, as node cleans up all services automatically in a loop except for this (special) service which has to be cleaned manually.
Reading through type_description_service.cpp surprised me, as I expected TypeDescriptionService to inherit from Service, allowing the standard node cleanup mechanism to destroy that as well.

After digging into it I think we can get that without having to do the inheritance refactor now, by handing out the wrapper instead of a copy of it:

```diff
- Service TypeDescriptionService::get_impl() { return *service_; }
+ std::shared_ptr<Service> TypeDescriptionService::get_impl() { return service_; }
```

This change is enough on its own - with one wrapper instead of two, the normal cleanup loop in `destroy_node()` takes the refcount to zero and the node leaves the graph, no manual destroy needed. Adding the fini deleter on top fixes the service as well. I built both and the full rclpy test suite passes.

Also worth mentioning - `test/test_type_description_service.py` is not registered in CMakeLists so it never run in CI.

nadavelkabets (commented):
> @nadavelkabets you implemented the destroy cleanup - does this fix look right, or should I approach it a different way?


Hey!

This is definitely an issue introduced in my refactor. `node.destroy_node()` destroys `TypeDescriptionService.impl`, but `get_impl()` returns `Service` by value - so `.impl` returns a new copy every call. The node destroys the returned copy, but the original instance inside the C++ wrapper keeps its own share of the node handle, and that's what keeps the node alive.

Your fix reverts the regression and the node leaves the graph again. 
One thing it doesn't cover - the type description service itself still stays in the graph.
I believe this originates from a bug in the TypeDescriptionService constructor:

Service.cpp constructor:
https://github.com/ros2/rclpy/blob/fb4701b97e17073e188ccbea6346a2145d849365/rclpy/src/rclpy/service.cpp#L67-L78
But TypeDescriptionService.cpp constructs:
https://github.com/ros2/rclpy/blob/fb4701b97e17073e188ccbea6346a2145d849365/rclpy/src/rclpy/type_description_service.cpp#L32

`Service` installs an `rcl_service_fini` deleter, `TypeDescriptionService` just uses `make_shared`, so the handle is deleted without ever being finalized. 

Regarding the fix itself - I think this is a design dilemma. The fix itself is valid - with that said, I deliberately removed what you are adding back. It felt like a patch to me, as node cleans up all services automatically in a loop except for this (special) service which has to be cleaned manually.
Reading through type_description_service.cpp surprised me, as I expected TypeDescriptionService to inherit from Service, allowing the standard node cleanup mechanism to destroy that as well.

After digging into it I think we can get that without having to do the inheritance refactor now, by handing out the wrapper instead of a copy of it:

```diff
- Service TypeDescriptionService::get_impl() { return *service_; }
+ std::shared_ptr<Service> TypeDescriptionService::get_impl() { return service_; }
```

This change is enough on its own - with one wrapper instead of two, the normal cleanup loop in `destroy_node()` takes the refcount to zero and the node leaves the graph, no manual destroy needed. Adding the fini deleter on top fixes the service as well. I built both and the full rclpy test suite passes.

Also worth mentioning - `test/test_type_description_service.py` is not registered in CMakeLists so it never run in CI.

emersonknapp:
@nadavelkabets thanks - that approach does seem cleaner, I've applied it. Also enabled `test_type_description_service.py` and fixed the rot in there.

Now, how'd you configure your agent for that review? Did you need to tell it to review, or is it hooked up automatically for review requests?

nadavelkabets (approved):

nadavelkabets:
> Now, how'd you configure your agent for that review? Did you need to tell it to review, or is it hooked up automatically for review requests?

Ha! I actually sat on this myself for a couple of hours on Saturday.
Claude initially suggested to merge your original PR as is.
Wish I had a capable automated bot to do my work for me 😄 

I'm wondering if we should fix the TypeDescriptionService constructor in this PR as well to call rcl_service_fini on destruction. What do you think?

fujitatomoya (commented):
lgtm after the following part is addressed,

> Service installs an rcl_service_fini deleter, TypeDescriptionService just uses make_shared, so the handle is deleted without ever being finalized.

ahcorde (approved):

ahcorde:
Pulls: ros2/rclpy#1711
Gist: https://gist.githubusercontent.com/ahcorde/3f9ef08c466524a670e77bdb03704343/raw/12cb358465bb871175ce532c0dc0626044efcb3c/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20129
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30195/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22893/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10005/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29048/)

ahcorde:
@emersonknapp should we backport this one ?

emersonknapp:
@mergifyio backport lyrical

emersonknapp:
@nadavelkabets haha well then great! It read very much like Claude 😅 Thanks for the review. I didn't get a chance to look at the fini thing, seems fine without? But maybe worth a followup look

PR #1717, opened 2026-09-01
Preserve numeric parameter types loaded from YAML files

### LINKED ISSUE #1672: parameter_dict_from_yaml_file loads doubles as strings when str() uses exponent notation (e.g. 0.00001, 5e-06, 1e+20)

yurekami:
## Summary

- recognize unquoted scientific-notation floats in parameter YAML
- preserve already parsed scalar and array types instead of round-tripping through strings
- keep quoted exponent values as strings and retain safe-loader behavior
- add regression coverage for doubles and double arrays

## Why

`parameter_dict_from_yaml_file` could interpret values such as `5e-06` as strings, including inside arrays. This fixes #1672 while keeping the resolver local to parameter files.

## Verification

- regression smoke matrix for exponent doubles, arrays, quoted strings, bools, ints, infinities, and NaN
- unsafe YAML tag rejection
- loader isolation from global `yaml.SafeLoader`
- `python -m py_compile` on touched files

The full ROS-backed pytest suite was not available locally and remains CI-bound.

Fixes #1672

### DIFF
--- a/rclpy/rclpy/parameter.py
+++ b/rclpy/rclpy/parameter.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import array
 from collections.abc import Sequence
 from enum import IntEnum
+import re
 from typing import Any
 from typing import cast
 from typing import Dict
@@ -34,6 +35,25 @@ from rcl_interfaces.msg import ParameterValue
 import yaml
 
 PARAMETER_SEPARATOR_STRING: Final = '.'
+
+
+class _ParameterFileLoader(yaml.SafeLoader):
+    """YAML loader for ROS parameter files."""
+
+
+_ParameterFileLoader.add_implicit_resolver(
+    'tag:yaml.org,2002:float',
+    re.compile(
+        r'''^(?:[-+]?(?:[0-9][0-9_]*)\.[0-9_]*(?:[eE][-+]?[0-9]+)?
+             |\.[0-9][0-9_]*(?:[eE][-+]?[0-9]+)?
+             |[-+]?(?:[0-9][0-9_]*)(?:[eE][-+]?[0-9]+)
+             |[-+]?[0-9][0-9_]*(?::[0-5]?[0-9])+\.[0-9_]*
+             |[-+]?\.(?:inf|Inf|INF)
+             |\.(?:nan|NaN|NAN))$''',
+        re.X,
+    ),
+    list('-+0123456789.'),
+)
 
 AllowableParameterValue = Union[None, bool, int, float, str,
                                 Sequence[bytes],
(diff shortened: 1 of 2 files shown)


fujitatomoya:
i will close the all PRs from @yurekami because this is categorized into AI spam based on our policy.

- the internal reasoning/audit trailer format of an autonomous coding agent that leaked straight into the commit.
- never ran the real tests.
- did not use our original template.

